PR #10, opened 2021-03-28
Relative path

piotrekklis:


data_pipeline/config.json
```diff
@@ -1,6 +1,6 @@
 {
    "_comment": "coinpaprika ids @ https://api.coinpaprika.com/v1/coins",
-   "output":"/home/ubuntu/l2beat/treasury_optimism.json",
+   "output":"/data/output.json",
```
krzkaczor, line 3: nit: to make it obvious that it's relative path don't begin it with `/` but rather `./`

data_pipeline/l2_job.py
```diff
@@ -9,8 +10,9 @@ def run_l2beat(config=None):
     with open(sys.argv[1]) as json_data:
         config = json.load(json_data)
 
+    parent_directory = os.path.dirname(os.getcwd())
     # create object to store data
-    output_object = create_storage(path=config['output'])
+    output_object = create_storage(path=parent_directory + config['output'])
```
krzkaczor, line 15: use `join` to concat paths: https://stackoverflow.com/questions/10918682/platform-independent-path-concatenation-using

data_pipeline/l2_job.py
```diff
@@ -28,7 +30,8 @@ def run_l2beat(config=None):
     
     output = run_l2beat(config=config)
 
-    with open(config['output'], 'w') as output_file:
+    parent_directory = os.path.dirname(os.getcwd())
+    with open(parent_directory + config['output'], 'w') as output_file:
```
krzkaczor, line 34: reduce duplication by introducing `output_path` var?

krzkaczor (changes requested):

PR #11, opened 2021-03-28
Output path

piotrekklis:
The output path in the config has to remain data/output.json because if it starts with ./ then the lib that we're using to join paths takes it as an absolute path (and ignores the rest of path elements)


data_pipeline/l2_job.py
```diff
@@ -5,14 +5,10 @@
...
-def run_l2beat(config=None):
+def run_l2beat(config=None, output_file_path=None):
 
-    with open(sys.argv[1]) as json_data:
-        config = json.load(json_data)
-
-    parent_directory = os.path.dirname(os.getcwd())
     # create object to store data
-    output_object = create_storage(path=parent_directory + config['output'])
+    output_object = create_storage(path=output_file_path)
```
krzkaczor, line 11: does it work like this? shoudn't `output_file_path` be defined first?

krzkaczor (approved):

PR #41, opened 2021-04-14
[refactor] re-order token and protocols by alphabet

gasolin:
To make token and protocols easier to add and avoid conflict for later protocols, re-order token and protocols by alphabet


krzkaczor:
Thanks but I am gonna close it for now as this won't be relevant after we migrate to new data pipeline.

PR #49, opened 2021-04-15
Add Polygon

bbjay:
I got the bridges info from here:
https://github.com/maticnetwork/static/blob/master/network/mainnet/v1/index.json

Not sure if you want to fill in the frontend info yourself or if I should provide it.


krzkaczor:
Thanks, @bbjay for this. For now, we are reluctant to add any non L2s. 

We might reconsider in the future so I will keep this open.

bbjay:
Hi @krzkaczor 
Polygon is L2, it is even listed in the Plasma section in the article "[Confused? Layer 2 primer](https://ethereum.org/en/developers/docs/layer-2-scaling/)" you link to on l2beat.com

If you use a different definition of L2 or have other criteria for listings, then  please communicate them appropriately.

krzkaczor:
In short: Polygon ships many totally different products. PoS related stuff that you included in the PR is definitely NOT Layer 2. 

I am not sure if Plasma is still considered an L2 given its data availability problems.

sz-piotr:
Closing this PR, because:
1. We are not currently considering Polygon as an L2
2. This PR is referencing a long gone past of the configuration and would require 100% rework to be merged

PR #52, opened 2021-04-18
Generate og images script

paszkowskiDamian:


frontend/pages/_app.tsx
```diff
@@ -21,7 +21,7 @@ function MyApp({ Component, pageProps }: { Component: React.ComponentProps<any>;
 
         <meta property="og:title" content={pageProps.title || TITLE} />
         <meta property="og:description" content={pageProps.description || DESCRIPTION} />
-        <meta property="og:image" content={pageProps.thumbnail || THUMBNAIL} />
+        <meta property="og:image" content={`https://l2beat.com/og/${pageProps.project || 'overview'}.png`} />
```
paszkowskiDamian, line 24: I think now if you managed to build it we can use vercel env instead of hardcoded url `VERCEL_URL` instead of `https://l2beat.com`

PR #56, opened 2021-04-20
add hermez-network information

### LINKED ISSUE #55: Add Hermez rollup
Bridge: https://etherscan.io/address/0xa68d85df56e733a06443306a095646317b5fa633 (~200k USD)

krlosMata:
This PR add the following:
- adds tokens `WETH`, `UNI`, `YFI`
- add hermez config file in `bridges`
- add frontend project information

Close #55

>important note: HEZ token is not supported by `https://api.coinpaprika.com/v1/coins`. Therefore, all value locked in `hermez-network` does not include HEZ locked tokens

### DIFF
--- a/frontend/data/projects-data.json
+++ b/frontend/data/projects-data.json
@@ -256,5 +256,51 @@
         "link": "https://www.immutable.com/blog/immutable-x-alpha-trading-launch"
       }
     ]
+  },
+  "Hermez": {
+    "website": "https://hermez.io/",
+    "color": "#ff9933",
+    "technology": "zk-rollup",
+    "technology-details": "zk-SNARK",
+    "more-info": {
+      "Primary use case": { "text": "Payments" },
+      "Hypothetical level of decentralization": { "text": "High", "sentiment": "good" },
+      "Current level of decentralization": { "text": "Low", "tooltip": "Contracts are upgradable", "sentiment": "bad" },
+      "Can funds be stolen by the operator?": {
+        "text": "Yes, through contract upgrade",
+        "tooltip": "Contracts are upgradable BUT on the upside there is a 7 days timelock",
+        "sentiment": "neutral",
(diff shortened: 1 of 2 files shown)


frontend/data/projects-data.json
```diff
@@ -256,5 +256,51 @@
...
+      "Can funds be stolen by the operator?": {
+        "text": "Yes, through contract upgrade",
+        "tooltip": "Contracts are upgradable BUT on the upside there is a 7 days timelock",
+        "sentiment": "neutral",
+        "pointers": ["https://etherscan.io/address/0xA68D85dF56E733A06443306A095646317B5Fa633#code"]
+      },
+      "Permisonless?": {
+        "text": "No",
+        "sentiment": "bad",
+        "tooltip": "After initial bootstraping period anyone could bid for a slot time to forge new batches and update the state root",
```
krzkaczor, line 278: Sooo is it permissionless or not? The description suggests that there is a bidding process to forge new batches so it looks like permissionless system to me 🤔

krlosMata: The design takes into a account that there is a `bootstrapping period` where the minimum bid to perform is [high enough](https://github.com/hermeznetwork/contracts/blob/master/contracts/auction/HermezAuctionProtocol.sol#L47) that no one can bid (this minimum bid could be changed afterwards by the governance). This has been done in order to avoid stop the network in its initial stage by malicious coordinators. 
Once the network has enough funds locked it will become more costly to stop the network and then minimum bid could be decresead and slots could be decentralized. Note that if minimum bid is set to 0 by the contract, [it could not be changed again](https://github.com/hermeznetwork/contracts/blob/master/contracts/auction/HermezAuctionProtocol.sol#L405).

Therefore, design will be permisionless once this bootstrapping stage will be over

krzkaczor: got it - makes sense

krzkaczor:
Done via https://github.com/krzkaczor/l2beat/pull/60 Thanks!

PR #64, opened 2021-05-24
Added OMG Network

jet86:
Added OMG Network including both its ETH vault and ERC20 vault. Also added a number of tokens which are currently locked in the OMG Network ERC20 vault.


jet86:
Added changes from #65 into this PR as requested.

sz-piotr:
@jet86 Thank you for the PR. The format changed a bit so I made another, but the information is the same :)

PR #72, opened 2021-06-04
Rewrite python backend in typescript

sz-piotr:
Things to discuss:
- javascript number math and potential precision loss
- python optimism code
- performance: it takes ~20s to fetch data for a given day, because of the network requests and rate limits
- storing cache.json outside the repository
- price differences between coingecko and coinpaprika
- replacing big query with etherscan


krzkaczor (commented):
Looks good! Nit: no eslint/tests?

So `precomputed.json` is like a checkpoint committed to git and then the updates will be only stored in GHA cache?

sz-piotr:
I don't find huge value in eslint, but I can add add it if you want. As for tests the code is mainly integration, and so only "boring" parts would be tested. If you insist I can spend some time and cover most of the logic.

> So precomputed.json is like a checkpoint committed to git and then the updates will be only stored in GHA cache?
Yes.

krzkaczor:
>I don't find huge value in eslint, but I can add add it if you want. As for tests the code is mainly integration, and so only "boring" parts would be tested. If you insist I can spend some time and cover most of the logic.

Your call. I would test some of the more complicated logic around cache but these tests will require introducing abstraction on FS etc so they would require a little bit of extra work.

So to merge it we still need to:
1. Remove old data pipeline
2. Tweak CI CRON to run the new pipeline.

Maybe we can tweak CI CRON to execute from this branch and test the whole integration first (ie if new data will show tmrw)?

As for the open questions:

>javascript number math and potential precision loss
I see that you use BN here and there. Doesn't it fix this?

>python optimism code
We can work on it in separate PR as the current code doesn't work anyway.

> price differences between coingecko and coinpaprika
Do you know which tokens are priced differently? Is it a big difference? I wouldn't worry about it as coingecko seems to be a better source anyway.

sz-piotr:
> Maybe we can tweak CI CRON to execute from this branch and test the whole integration first (ie if new data will show tmrw)?

We can remove today's data from the cache and test it this way. Or we can of course wait. You can also run the code yourself and compare the data.json diff.

> I see that you use BN here and there. Doesn't it fix this?

The values collected from the integrations are correct, but the way I add them up might not be, especially since the error can be compounding. First I multiply all the token balances by price, then add those up, then add those up for each bridge and then add those up for each L2. I think the error isn't significant, but also that the result isn't mathematically correct.

> Do you know which tokens are priced differently? Is it a big difference? I wouldn't worry about it as coingecko seems to be a better source anyway.

Basically almost all prices are different. On master TVL is $383,515,747, with coingecko $376,723,027 so a difference of 1.8%

.github/workflows/ci.yml
```diff
@@ -36,3 +36,35 @@ jobs:
...
+          key: cache-json-v-1-${{ github.run_id }}
+          restore-keys: |
+            cache-json-v-1-
+      - run: yarn --frozen-lockfile
+        working-directory: ./backend
+      - run: yarn lint
+        working-directory: ./backend
+      - run: yarn typecheck
+        working-directory: ./backend
+      - run: yarn start
```
krzkaczor, line 66: We don't need to refresh data.json on every commit but rather on daily schedule. So I would remove this bit and tweak CI CRON job.

sz-piotr: I actually think we should keep this step as it is crucial for this code to work. The `data.json` file is overwritten but not committed and the cache would be reused from the master branch on any PR that doesn't yet have its own.

I can add a flag like `--fetch-only` that does not overwrite `data.json`

sz-piotr: Maybe once we have eslint & tests it is less needed. Let's revisit this topic then :)

sz-piotr:
Things that have changed:
1. The PR now has tests (only for core stuff, not for integrations)
2. Added eslint
3. Updated `refresh-data.yml` to use new backend. It works https://github.com/krzkaczor/l2beat/runs/2764540256?check_suite_focus=true. I also checked that the frontend can run against the new data.json
4. Removed `data_pipeline`. I needed to move the config file to the frontend. A future PR might share this file between frontend and backend. TBD. I also updated the links in the front-end but they are now slightly misleading. Should this PR introduce proper contributing guidelines or can this be done in the future?

.github/workflows/ci.yml
sz-piotr: Removed this step

krzkaczor (approved):

PR #75, opened 2021-06-08
Add netlify deployment

sz-piotr:
This PR introduces direct deployment to netlify circumventing the need for `data.json` being commited to the repository.
This deployment will happen on every push to master as well as everyday at 2am (previously the time data.json was generated)

For details of deployment see this run:
https://github.com/l2beat/l2beat/runs/2774891900

The target netlify site should have the following settings:
image

The repository also needs the following secrets to be set:
- `NETLIFY_AUTH_TOKEN`
- `NETLIFY_SITE_ID`

Because the `data.json` file is no longer inside the repository a `build:mock` command is added that creates `build.json` filled with fake data.


krzkaczor (changes requested):
I am missing some getting started guide. Also: `.nvmrc` file should be moved to the root (it's in `frontend` dir). 

I managed to run it with `yarn build:mock` and then in `frontend` `yarn start`. Can we make this somehow simpler?

krzkaczor (approved):

PR #82, opened 2021-06-11
Update zkswap to validium and validium to not l2

sz-piotr:


ZKSwap-fe:
@sz-piotr Why do you make this change? ZKSwap is a Layer-2 protocol based on zk-rollup.

sz-piotr:
Hey @ZKSwap-fe thanks for reaching out. According to [this article](https://medium.com/zkswap/zkswap-introduces-practical-zk-rollups-zkspeed-achieving-high-tps-and-low-gas-fees-in-real-6effe4e789e0) ZKSwap is indeed based on zk-rollup but does not store 100% of the data needed to reconstruct the state on-chain. This is why we don't classify it as a zk-rollup.

I understand that the validium label is very broad and assumes almost no data is published on chain, which is not the case with ZKSwap, where some data is published on-chain. We are currently working hard on developing a better framework for classifying layer 2 solutions as well as the risk involved in them.

PR #88, opened 2021-06-12
Add DeversiFi

luke7211:


sz-piotr:
@luke7211 sorry for doing so much additional work here. Turns out the PR had the following problems:
1. Tokens had invalid coingeckoIds (in the near future token config will be automatically verified in tests)
2. `precomputed.json` was missing a lot of data. I added a script to update it that does not require human input
3. I also added retries to the coingecko API as it likes to fail often...

PR #98, opened 2021-07-02
Add Habitat Rollup

pinkiebell:
### Additional notes
~~This rollup uses a single bridge, thus I linked all the token symbols.~~

### DIFF
--- a/config/src/projects/habitat.ts
+++ b/config/src/projects/habitat.ts
@@ -63,9 +63,10 @@ export const habitat: Project = {
       {
         name: 'Can the funds be freely withdrawn?',
         value: 'Yes, but not instantly.',
-        tooltip: 'Withdraw delays are subject to the inspection period. Currently 1 week.',
+        tooltip:
+          'Withdraw delays are subject to the inspection period. Currently 1 week.',
         sentiment: 'good',
-      }
+      },
       {
         name: 'Rollup Governance',
         value: 'HBT tokens are needed to participate.',
--- a/config/src/projects/habitat.ts
+++ b/config/src/projects/habitat.ts
@@ -1,0 +1,79 @@
+import { Project } from './Project'
+import { tokenList } from '../tokens'
+
+export const habitat: Project = {
+  name: 'Habitat',
+  bridges: [
+    {
+      address: '0x96E471B5945373dE238963B4E032D3574be4d195',
+      sinceBlock: 12731872,
+      tokens: tokenList.map((e) => e.symbol),
+    },
+  ],
+  details: {
+    website: 'https://0xhabitat.org/',
+    color: '#29b65f',
+    technology: {
+      name: 'application specific optimistic-rollup',
+      details: 'NutBerry',
+    },
+    parameters: [
+      {
+        name: 'Primary use case',
+        value: 'Governance & Treasury Management',
+      },
+      {
+        name: 'Hypothetical level of decentralization',
+        value: 'High',
+        sentiment: 'good',
+      },
+      {
+        name: 'Current level of decentralization',
+        value: 'Low',
+        tooltip:
+          'Community owned Rollup, currently in the bootstrapping phase.',
+        sentiment: 'neutral',
+      },
+      {
+        name: 'Can funds be stolen by the operator?',
+        value: 'Yes, through a contract upgrade via the multisig.',
+        tooltip:
+          'Ownership will be transferred to community governance in the future.',
+        sentiment: 'neutral',
+        pointers: [
+          'https://github.com/0xHabitat/habitat/blob/master/src/rollup/contracts/HabitatV1.sol#L52',
+        ],
+      },
+      {
+        name: 'Permissionless?',
+        value: 'Yes',
+        sentiment: 'bad',
+      },
+      {
+        name: 'Force TX mechanism?',
+        value: 'Yes',
+        tooltip:
+          'Fully permissionless. Rollup node can be run inside web browsers.',
+        sentiment: 'good',
+      },
+      {
+        name: 'Privacy',
+        value: 'No',
+      },
+      {
+        name: 'Smart contracts',
+        value:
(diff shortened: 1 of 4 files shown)


config/src/projects/habitat.ts
```diff
@@ -0,0 +1,79 @@
...
+      {
+        name: 'Permissionless?',
+        value: 'Yes',
+        sentiment: 'bad',
+      },
+      {
+        name: 'Force TX mechanism?',
+        value: 'Yes',
+        tooltip:
+          'Fully permissionless. Rollup node can be run inside web browsers.',
```
sz-piotr, line 56: I don't think that this comment is relevant here. Maybe it refers to the previous point? Force transactions are operations that allow users to force the rollup operator to include their transaction. Does Habitat support this? And if yes is it only for withdrawals or does it include any operation that the user may wish to perform?

config/src/projects/habitat.ts
```diff
@@ -0,0 +1,79 @@
...
+        sentiment: 'good',
+      },
+      {
+        name: 'Privacy',
+        value: 'No',
+      },
+      {
+        name: 'Smart contracts',
+        value:
+          'Limited. Only static contracts that comply to the Habitat Module interface.',
```
sz-piotr, line 66: That is very interesting. I tried to find developer documentation for Habitat, but I was out of luck. Can you point me to some resources?

config/src/tokens.ts
```diff
@@ -216,6 +216,14 @@ export const tokenList: TokenInfo[] = [
     coingeckoId: 'golem',
     sinceBlock: 11281056,
   },
+  {
+    name: 'Habitat Token',
+    symbol: 'HBT',
+    address: '0x0aCe32f6E87Ac1457A5385f8eb0208F37263B415',
+    decimals: 10,
+    coingeckoId: 'habitat',
```
sz-piotr, line 224: The new code on the master branch no longer uses coingecko so this is now not needed.

config/src/projects/habitat.ts
```diff
@@ -0,0 +1,79 @@
+import { Project } from './Project'
+import { tokenList } from '../tokens'
+
+export const habitat: Project = {
+  name: 'Habitat',
+  bridges: [
+    {
+      address: '0x96E471B5945373dE238963B4E032D3574be4d195',
+      sinceBlock: 12731872,
+      tokens: tokenList.map((e) => e.symbol),
```
sz-piotr, line 10: Please limit the tokens to the once the rollup actually holds. I this case those are: HBT, WETH, USDC and GNO. The more tokens there are the more data we need to process, so it is best to keep this list lean :)

sz-piotr (changes requested):
Hey @pinkiebell. Thanks for the PR. I have added some comments for you to look into. If you could also rebase your PR to the lastest master it would be fantastic, since the token configuration changed slightly (the `coingeckoId` field is not needed).

config/src/projects/habitat.ts
```diff
@@ -0,0 +1,79 @@
...
+      {
+        name: 'Hypothetical level of decentralization',
+        value: 'High',
+        sentiment: 'good',
+      },
+      {
+        name: 'Current level of decentralization',
+        value: 'Low',
+        tooltip:
+          'Community owned Rollup, currently in the bootstrapping phase.',
```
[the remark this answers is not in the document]
pinkiebell: Thanks for the scrutiny. 🙂 
Is it ok to answer all those points in different parameters?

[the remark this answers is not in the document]
sz-piotr: Yes. Feel free to place your answers wherever is most appropriate: tooltips, links or even new parameters. The current set of parameters is only a guideline.

config/src/projects/habitat.ts
pinkiebell: We are currently working on that and only have more user-centric documentation @  https://docs.0xhabitat.org/

pinkiebell: Currently deployed modules serving as example: https://github.com/0xHabitat/habitat/blob/master/src/rollup/contracts/modules/

pinkiebell: Btw, this interface is going to include more `callbacks` in the future 🙂 . Right now it's only focused on governance proposals

sz-piotr: Interesting. So is my understanding correct here that the contracts are used as a configuration mechanism for voting? They need to be "pure" like pure functions, right? Also, can anyone deploy new ones?

pinkiebell: That's right. The actual validation / state storage is done by the rollup and are handled down to these contracts.

pinkiebell: > Also, can anyone deploy new ones
Yep. Anyone can register them. But this needs to happen on Layer 1 as data availability and verification source-of-truth. That's also why we do not need archive nodes for Ethereum mainnet. Ethereum light-nodes are sufficient for the NutBerry L2 node.

pinkiebell:
Rebased and edited @sz-piotr . Let me know what you think 👌

sz-piotr:
@pinkiebell The values look good, but you have some syntax errors in your code.
https://github.com/l2beat/l2beat/pull/98/checks?check_run_id=2973640504

pinkiebell:
> @pinkiebell The values look good, but you have some syntax errors in your code.
> https://github.com/l2beat/l2beat/pull/98/checks?check_run_id=2973640504

😄 Friday is good-try day.

sz-piotr:
Looks good. And you even fixed the error. (I need to devise a more permanent solution though :P)

sz-piotr:
@pinkiebell And it's live https://l2beat.com/project/habitat

pinkiebell:
Yay ! But a bit too fast. I just pushed 2 nitpicks  😁

sz-piotr:
Ooops! Can you open another PR? We'll get it merged.

PR #109, opened 2021-07-13
Add a chart without log scale and hover

sz-piotr:


sz-piotr:
https://l2beat-test-0001.surge.sh/

PR #112, opened 2021-07-13
Remove italic text and save 160kb

sz-piotr:


sz-piotr:
https://l2beat-test-0002.surge.sh/

PR #114, opened 2021-07-13
Add new optimism bridges

### LINKED ISSUE #113: Add missing Optimism bridges
Here's a token list with bridges extensions: https://github.com/ethereum-optimism/frontends/blob/master/packages/gateway.optimism.io/src/tokenLists/optimism.tokenlist.json 

For most tokens, it should be enough to track the balance of the L1 bridge (`optimismBridgeAddress ` on chainId = 1). In some cases like DAI, funds are escrowed in a separate contract: (`0x467194771dAe2967Aef3ECbEDD3Bf9a310C76C65` in DAI's case).

sz-piotr:
Closes #113

### DIFF
--- a/config/src/projects/optimism.ts
+++ b/config/src/projects/optimism.ts
@@ -5,14 +5,31 @@ export const optimism: Project = {
   slug: 'optimism',
   bridges: [
     {
+      // TODO: this one seems unused but there are tokens locked
       address: '0x045e507925d2e05D114534D0810a1abD94aca8d6',
       sinceBlock: 11656238,
       tokens: ['SNX'],
     },
     {
+      // TODO: this one seems unused but there are tokens locked
       address: '0x5Fd79D46EBA7F351fe49BFF9E87cdeA6c821eF9f',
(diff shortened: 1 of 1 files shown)


config/src/projects/optimism.ts
```diff
@@ -5,15 +5,32 @@ export const optimism: Project = {
...
       sinceBlock: 12409015,
       tokens: ['SNX'],
     },
+    {
+      address: '0xCd9D4988C0AE61887B075bA77f08cbFAd2b65068',
+      sinceBlock: 12409013,
+      tokens: ['SNX'],
+    },
+    {
+      address: '0x10E6593CDda8c58a1d0f14C5164B376352a55f2F',
```
krzkaczor, line 25: This should be: `0x467194771dAe2967Aef3ECbEDD3Bf9a310C76C65` (funds are escrowed outside of a bridge contract).

sz-piotr: Solved via #115

PR #116, opened 2021-07-13
Make website responsive

sz-piotr:


sz-piotr:
https://l2beat-test-0003.surge.sh/

PR #117, opened 2021-07-13
Fix canvas resolution on high DPI devices

sz-piotr:


sz-piotr:
https://l2beat-test-0004.surge.sh/

PR #155, opened 2021-07-29
Add ETH and LINK to L1 Bridge Address

domolord:
The Optimism L1 Bridge Contract 0x99C9fc46f92E8a1c0deC1b1747d010903E884bE1 also holds ETH and LINK deposits. I suspect that this is also the contract that will hold EURT but I have not confirmed this yet.


sz-piotr:
Looks good. Thank you very much for noticing this and contributing a PR ❤️

PR #211, opened 2021-08-31
DeversiFi: Add DVF token

plutoegg:


config/src/tokens.ts
```diff
@@ -424,6 +424,13 @@ export const tokenList: TokenInfo[] = [
     decimals: 18,
     sinceBlock: 11305469,
   },
+  {
+    name: 'DVF',
```
sz-piotr, line 428: Rename to "DeversiFi Token" because that is the name used in the smart contract

sz-piotr (changes requested):

config/src/projects/deversifi.ts
```diff
@@ -44,6 +44,7 @@ export const deversifi: Project = {
         'LRC',
         'OMG',
         'BAT',
+        'DVF',
```
sz-piotr, line 47: Please also add the `associatedToken: 'DVF',` property to the project definition. Place it between `bridges` and `details`

sz-piotr (approved):

sz-piotr:
@plutoegg Thanks for the PR!

plutoegg:
> @plutoegg Thanks for the PR!

NP, love what you guys are doing!

PR #269, opened 2021-09-24
Update Proxy__L1LiquidityPool

jet86:
This contract was updated as per https://github.com/omgnetwork/Boba-Mainnet-Addresses/blob/main/addresses.json and is also verified.
(this is why the previously listed contract had virtually no ETH in it, and was not verified)


packages/config/src/projects/bobanetwork.ts
```diff
@@ -19,7 +19,7 @@ export const bobanetwork: Project = {
     },
     {
       // Proxy__L1LiquidityPool
-      address: '0x6b4eab8D55A4f701C3FE58bDb8b3e1f181eA7087',
+      address: '0x1A26ef6575B7BBB864d984D9255C069F6c361a14',
       sinceBlock: 13013879,
       tokens: ['ETH'],
```
sz-piotr, line 24: Add OMG, USDT, USDC, SUSHI, DAI

sz-piotr (changes requested):

sz-piotr (approved):

PR #313, opened 2021-11-17
Add Immutable X token

igx-C:


packages/config/src/tokens.ts
```diff
@@ -255,6 +255,13 @@ export const tokenList: TokenInfo[] = [
     decimals: 18,
     sinceBlock: 3557596,
   },
+  {
+    name: 'Gods Unchained Token',
```
sz-piotr, line 259: I think this token should be skipped as it doesn't satisfy the requirement of Top 300 Coingecko. For some reason Etherscan doesn't count both GODS and IMX into the TVL.

sz-piotr (changes requested):

packages/config/src/tokens.ts
igx-C: Hm, what do we need to do to have both GODS and IMX count into the TVL?

sz-piotr: Maybe I should make an exception for GODS, but I added some rules as to which tokens should count into the TVL as including all low market cap tokens would decrease the reliability of the metric. Ideally we'd only consider GODS if it ever gets to >#300 on coingecko (the threshold is arbitrary but some threshold needs to exist).

igx-C: Understood, if you could make that exception for us it'd be greatly appreciated. Let me know if you need anything else from me.

PR #323, opened 2021-11-25
Add FXS token, update Boba Network

jet86:
Add FXS ~~and UST~~ tokens to tokens.ts, and update the tokens tracked for Boba Network.


packages/config/src/tokens.ts
```diff
@@ -514,6 +521,13 @@ export const tokenList: TokenInfo[] = [
     decimals: 18,
     sinceBlock: 4719568,
   },
+  {
+    name: 'Wrapped UST Token',
+    symbol: 'UST',
+    address: '0xC02aaA39b223FE8D0A0e5C4F27eAD9083C756Cc2',
```
sz-piotr, line 527: This is the address of WETH. The correct address is: `0xa47c8bf37f92abed4a126bda807a7b7498661acd`.
Also the activity seems really minimal and this token only has some weird pairs on Uniswap, so getting the price would be problematic. I propose to skip this one for now.
https://etherscan.io/token/0xa47c8bf37f92abed4a126bda807a7b7498661acd?a=0xdc1664458d2f0B6090bEa60A8793A4E66c2F1c00

sz-piotr (changes requested):

packages/config/src/projects/bobanetwork.ts
```diff
@@ -31,13 +31,29 @@ export const bobanetwork: Project = {
         'BAT',
         'ZRX',
         'BOBA',
+        'BUSD',
+        'LINK',
+        'FXS',
+        'UST',
```
sz-piotr, line 37:
```suggestion
```

sz-piotr (changes requested):

PR #348, opened 2021-12-07
Update metis FP description

krzkaczor:


packages/config/src/projects/metis.ts
```diff
@@ -40,9 +40,9 @@ export const metis: Project = {
     provider: 'Optimism',
     riskView: {
       stateValidation: {
-        value: 'In development',
+        value: 'No',
```
sz-piotr, line 43: No is too short IMO.
```suggestion
        value: 'No validation',
```

sz-piotr (changes requested):

PR #351, opened 2021-12-09
Add StarkNet

maciejka:


packages/config/src/projects/common/exits.ts
```diff
@@ -106,6 +106,33 @@ const STARKEX_EMERGENCY: ProjectTechnologyChoice = {
...
+  description:
+    REGULAR('zk', 'no proof').description +
+    ' Note that the withdrawal request can be censored by the Sequencer.',
+  references: [
+    {
+      text: ' Withdrawing is based on l2 to l1 messages - StarkNet documentation',
+      href: 'https://www.cairo-lang.org/docs/hello_starknet/l1l2.html',
+    },
+  ],
+  //  risks: [RISK_CENTRALIZED_VALIDATOR],
```
sz-piotr, line 120: Why is this commented out? Either remove or uncomment

packages/config/src/projects/common/exits.ts
```diff
@@ -2,7 +2,7 @@ import { ProjectRisk, ProjectTechnologyChoice } from '../types'
 
 function REGULAR(
   type: 'zk' | 'optimistic',
-  proof: 'no proof' | 'merkle proof'
+  proof: 'no proof' | 'merkle proof' | 'zk proof'
```
sz-piotr, line 5: Why change the type if it is never called in this way

packages/config/src/projects/common/exits.ts
```diff
@@ -106,6 +106,33 @@ const STARKEX_EMERGENCY: ProjectTechnologyChoice = {
...
+    {
+      text: ' Withdrawing is based on l2 to l1 messages - StarkNet documentation',
+      href: 'https://www.cairo-lang.org/docs/hello_starknet/l1l2.html',
+    },
+  ],
+  //  risks: [RISK_CENTRALIZED_VALIDATOR],
+}
+
+const STARKNET_EMERGENCY: ProjectTechnologyChoice = {
+  ...EMERGENCY('', 'merkle proof'),
```
sz-piotr, line 124: The first argument shouldn't be an empty string. `EMERGENCY` is only a helper, you don't have to use it :)

sz-piotr (changes requested):
The image also needs to be <10kb and 256x256. I can help with that

sz-piotr (approved):

PR #354, opened 2021-12-16
Add deversifi tokens missing from TVL

plutoegg:
Several of the more recently listed tokens on deversifi are not included in the TVL


packages/config/src/tokens.ts
```diff
@@ -249,6 +273,14 @@ export const tokenList: TokenInfo[] = [
     sinceBlock: 5123691,
     category: 'other',
   },
+  {
+    name: 'Entropy',
+    symbol: 'ERP',
```
sz-piotr, line 278: Doesn't meet the requirements

packages/config/src/tokens.ts
```diff
@@ -400,6 +440,14 @@ export const tokenList: TokenInfo[] = [
     sinceBlock: 4620855,
     category: 'other',
   },
+  {
+    name: 'Maple Token',
+    symbol: 'MPL',
```
sz-piotr, line 445: Doesn't meet the requirements

packages/config/src/tokens.ts
```diff
@@ -480,6 +528,14 @@ export const tokenList: TokenInfo[] = [
     sinceBlock: 5257012,
     category: 'other',
   },
+  {
+    name: 'Rai Reflex Index',
+    symbol: 'RAI',
```
sz-piotr, line 533: Doesn't meet the requirements

sz-piotr (changes requested):

PR #364, opened 2022-01-03
projects/habitat: add shutdown notice

pinkiebell:


sz-piotr:
Hey @pinkiebell. Thanks for the PR. I think that removing it outright might not be a good idea. Instead I think we should add a warning that says that the rollup will shut down and users are encouraged to exit. Do you have a link to any official announcement?

pinkiebell:
> Hey @pinkiebell. Thanks for the PR. I think that removing it outright might not be a good idea. Instead I think we should add a warning that says that the rollup will shut down and users are encouraged to exit. Do you have a link to any official announcement?

We prepare an additional blogpost soon. I will ping you once ready. What would be the best way to integrate this warning in the project definition?

sz-piotr:
There is an option to add a warning to the `details` field:
https://github.com/l2beat/l2beat/blob/e9ee816ded0967d42b49235f8cd943316039bbc6/packages/config/src/projects/types/ProjectDetails.ts#L8

pinkiebell:
Ready 👍

sz-piotr (approved):

PR #367, opened 2022-01-12
Add Aventus Layer-2

dr-ngubo:


krzkaczor:
Hey @dr-ngubo!

It seems that `Aventus` is a side-chain and they are using `layer2` purely as a marketing term.

1. According to the source code, the state root of the network can change if the quorum is met: https://etherscan.io/address/0x46eaD2891ecf9c7444cB9197D4FAacC7F54C593C#code#L451
2. There is no dispute period - withdrawals are not delayed.
3. There is a slashing mechanism but it's based on quorom - there is no real fraud-proof system. 
4. No tx data are stored on-chain.
5. In their whitepaper they already hint that consensus relies on 2/3 of validators: 
image



Given all of that, it's hard to claim that Aventus shares the security of L1 - it's fully independent PoS network connected with ethereum via multisig bridge.

PR #378, opened 2022-01-21
Add stricter eslint rules

antooni:


antooni:
EtherscanClient and CoingeckoClient have typing error in httpClient.fetch(...), timeout is not a property of RequestInit

packages/frontend/yarn-error.log
```diff
@@ -0,0 +1,86 @@
+Arguments:
```
sz-piotr, line 1: Please add yarn-error.log to .gitignore

sz-piotr (changes requested):

sz-piotr (approved):

PR #386, opened 2022-02-02
Add CoingeckoQueryService

antooni:


packages/backend/src/peripherals/coingecko/CoingeckoQueryService.ts
```diff
@@ -0,0 +1,33 @@
...
+    constructor(
+       private coingeckoClient: CoingeckoClient
+    ){}
+
+}
+
+export function generateTimestampsList(
+    from: UnixTime,
+    to: UnixTime,
+    granularity: Granularity
```
sz-piotr, line 18: Personally I prefer string literals to enums. There a a few reasons:
1. Less imports. You just write `'hourly'` instead of `import Granularity; Granularity.Hourly`
2. Easier to debug. If you console.log and enum you'll find that `Hourly = 1` and `Daily = 2`. With string literals what you see is what you get.

One small drawback is that because you don't have a separate type (you could but it mostly isn't worth it) you need to write `granularity: 'daily' | 'hourly'` on every function that uses it, but if it turns out too frustrating you could have a `type Granularity = 'daily' | 'hourly'` and use that one. It still wouldn't need to be imported by the user.

packages/backend/test/peripherals/coingecko/CoingeckoQueryService.test.ts
```diff
@@ -0,0 +1,49 @@
...
+        const TO = new UnixTime(1643814000) //Wed Feb 02 2022 15:00:00 GMT+0000
+
+        const RESULT = [new Date(1643806800000), new Date(1643810400000), new Date(1643814000000)]
+
+        it("empty if FROM greater than TO", () => {
+            const result = generateTimestampsList(TO,FROM,GRANULARITY)
+            expect(result).toEqual([])
+        })
+
+        it("empty if FROM is not a full hour", () => {
```
sz-piotr, line 22: In this case shouldn't we start from the next full hour after `FROM`. Similarity for `TO` if it isn't a full hour, shouldn't we use the previous full hour before it? I think it makes more sense then requiring precision. If we require precision then throwing is definitely the answer, because an empty array is an unexpected, undetected, silent failure.

packages/backend/src/peripherals/coingecko/CoingeckoQueryService.ts
```diff
@@ -0,0 +1,33 @@
...
+): Date[] {
```
sz-piotr, line 19: I think it makes sense for the function to return `UnixTime[]`. Let the user of this function worry about the conversion to milliseconds. If `UnixTime` is our preferred time structure in the system then I think we should hide the fact that coingecko returns milliseconds as an implementation detail of `CoingeckoQueryService`

packages/backend/test/peripherals/coingecko/CoingeckoQueryService.test.ts
antooni: I think I will make it both ways inclusive, 
e.g. from = 14:15, to = 15:20, granularity = 'hourly'
returns [14:00,15:00,16:00]

@sz-piotr let me know what u think

sz-piotr (approved):

PR #392, opened 2022-02-16
Integrate Coingecko with PriceService

antooni:


packages/backend/test/old/services/prices/PriceService.test.ts
```diff
@@ -0,0 +1,158 @@
+import { CoingeckoId, mock, SimpleDate, UnixTime } from '@l2beat/common'
+import { TokenInfo } from '@l2beat/config'
+import { expect, mockFn } from 'earljs'
+import { utils } from 'ethers'
+import { Logger } from 'ethers/lib/utils'
```
sz-piotr, line 5: Logger needs to be imported from common

packages/backend/test/old/services/prices/PriceService.test.ts
```diff
@@ -0,0 +1,158 @@
...
+            deltaMs: 0,
+          }))
+        ),
+      })
+      const logger = mock<Logger>({
+        info: mockFn().returns([]),
+      })
+      const priceService = new PriceService(
+        coingeckoQueryService,
+        logger as any
```
sz-piotr, line 33: Instead of mocking the logger you could use `Logger.SILENT`

packages/backend/test/old/services/prices/PriceService.test.ts
```diff
@@ -0,0 +1,158 @@
...
+    it('three dates and two coins and ether', async () => {
+      const dates = [
+        SimpleDate.fromString('2022-02-14'),
+        SimpleDate.fromString('2022-02-15'),
+        SimpleDate.fromString('2022-02-16'),
+      ]
+
+      const DAI_PRICE = 200
+      const ETH_PRICE = 1000
+      const UNI_PRICE = 100
```
sz-piotr, line 70: Replace UNI with a token with less decimals, so that this calculation is covered

packages/backend/test/old/services/prices/PriceService.test.ts
```diff
@@ -0,0 +1,158 @@
...
+    })
+    it('three dates and two coins and ether', async () => {
+      const dates = [
+        SimpleDate.fromString('2022-02-14'),
+        SimpleDate.fromString('2022-02-15'),
+        SimpleDate.fromString('2022-02-16'),
+      ]
+
+      const DAI_PRICE = 200
+      const ETH_PRICE = 1000
```
sz-piotr, line 69: Make the ETH_PRICE = 1234.5678, so that we also check that floating point is handled correctly

packages/backend/src/peripherals/coingecko/CoingeckoQueryService.ts
```diff
@@ -89,6 +89,7 @@ export function pickPrices(
   prices: { price: number; date: Date }[],
   timestamps: UnixTime[]
 ): PriceHistoryPoint[] {
+  if (prices.length === 0) return []
```
sz-piotr, line 92: This seems like a temporary workaround. It's fine to keep it, but I'd add a comment: // TODO: Handle this case properly

sz-piotr (changes requested):

PR #393, opened 2022-02-17
Update token list

sz-piotr:


sz-piotr:
Unfortunately this can't be merged yet, as running this code causes a "JavaScript heap out of memory" error.

PR #394, opened 2022-02-17
Add caching

antooni:


packages/backend/src/old/services/prices/PriceService.ts
```diff
@@ -63,4 +60,80 @@ export class PriceService {
...
+      const coingeckoPrices =
+        await this.coingeckoQueryService.getUsdPriceHistory(
+          CoingeckoId(token.coingeckoId),
+          new UnixTime(earliestUnknownDate.toUnixTimestamp()),
+          new UnixTime(end.toUnixTimestamp()),
+          'daily'
+        )
+      prices.push(...coingeckoPrices)
+
+      if (coingeckoPrices)
```
sz-piotr, line 97: How can this be undefined?

packages/backend/src/old/services/prices/PriceService.ts
```diff
@@ -63,4 +60,80 @@ export class PriceService {
...
+        this.logger.info('Coingecko API call', {
+          token: token.coingeckoId,
+          amountOfPricesFetched: coingeckoPrices.length,
+        })
+
+      coingeckoPrices?.map((price) => {
```
sz-piotr, line 103: map implies that we care about the result. You could use `.forEach` but ideally just `for...of`

sz-piotr (changes requested):

PR #395, opened 2022-02-18
Update ZKSpace branding

antooni:


sz-piotr (changes requested):
I'd make the icon transparent without a background. That's what they do on their blog: https://medium.com/zkswap. I'd also add a redirect from zkswapv2 to zkspace in the _redirects file.

sz-piotr (approved):

PR #419, opened 2022-03-11
Balance-updater

antooni:


packages/backend/test/core/BalanceUpdater.test.ts
```diff
@@ -0,0 +1,619 @@
...
+          holderAddress: HOLDER_C,
+          assetId: ASSET_C,
+          balance: BigInt(100),
+          blockNumber: START_BLOCK_NUMBER,
+        },
+      ])
+    })
+  })
+
+  //       it('E2E', async () => {
```
sz-piotr, line 593: what about the commented out code?

sz-piotr (changes requested):

PR #441, opened 2022-04-30
Add new risk profile for Metis after april 2022 upgrade

luke7211:


sz-piotr (approved):
Do not merge before twitter thread

PR #452, opened 2022-05-24
Optimise report query

antooni:


packages/backend/src/peripherals/database/types.ts
```diff
@@ -26,6 +26,7 @@ declare module 'knex/types/tables' {
     asset_id: string
     usd_tvl: string
     eth_tvl: string
+    isDaily: boolean
```
sz-piotr, line 29: not necessary

packages/backend/src/peripherals/database/ReportRepository.ts
```diff
@@ -31,7 +31,7 @@ export class ReportRepository {
           .andOn('asset_balances.holder_address', '=', 'reports.bridge_address')
       })
 
-    return rows.map((r) => toRecordWithBalance(r, r.balance))
+    return rows.map((r) => toRecordWithBalance(r, '0'))
```
sz-piotr, line 34: revert.

1. getDaily should use isDaily
2. addOrUpdate should recalculate isDaily

sz-piotr (changes requested):

antooni:
stop the server
run sql script to:
 update isDaily
 update balance

PR #458, opened 2022-06-01
Track Optimism's token

sz-piotr:


sz-piotr:
@antooni in a separate PR a test for `addOptimismToken` should be added

Issue #219, opened 2021-09-01
Make the table sortable


NewsGuyTor:
Hey, it would be cool if the table colums could be sortable.

uschtwill:
Are there any constraints? Like "no jQuery"?

torztomasz:
Resolved in #2326 :)

PR #460, opened 2022-06-04
Make main table sortable (#219)

uschtwill:
Minimally intrusive solution (codebase-wise), does include jQuery/client-side JS though. 

I put the commented out stylesheet in there so you can toggle it and check out the styling that comes with Datatables, also see: 
- https://datatables.net/examples/styling/
- https://datatables.net/manual/styling/

Without the stylesheet, the headers are not styled at all. Will probably need a little icon or at least the normal mouse-over cursor change. But I thought you guys may have some thoughts on that, so I didn't implement it yet.

### DIFF
--- a/packages/frontend/src/content/common/head/Preload.tsx
+++ b/packages/frontend/src/content/common/head/Preload.tsx
@@ -24,6 +24,19 @@ export function Preload(props: Props) {
         type="font/woff2"
         crossOrigin="anonymous"
       />
+      {/* <link
+        rel="stylesheet"
+        type="text/css"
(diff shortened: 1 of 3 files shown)


sz-piotr:
The solution is straightforward but has some small problems:
1. We probably don't want jquery as reordering a table on click isn't really hard in vanilla js and the site aims to be really lightweight
2. The plugin leaves some text on the bottom, does it automatically introduce pagination?
image
3. The risk table is automatically affected as well and there the sorting doesn't make much sense.

PR #462, opened 2022-06-08
Update dependencies

sz-piotr:


sin6pi7 (approved):

sin6pi7:
Node version needs to updated in github actions scripts to keep them running - currently it specifically requires node v14.

PR #467, opened 2022-06-09
L2Data parsing

sin6pi7:


packages/frontend/src/content/L2Data.ts
```diff
@@ -1,14 +1,40 @@
...
+  return JSON.parse(data)
 }
 
-export interface ProjectData {
-  aggregate: ChartData
-  byToken: Record<string, ChartData>
+async function getDataFromApi() {
+  const http = new HttpClient()
+  const response = await http.fetch('https://api.l2beat.com/api/data')
+  return response.json()
```
sz-piotr, line 33: You want to check `response.ok`, because if it's 400 but still valid JSON you won't get an error at this state.

sz-piotr (changes requested):

sz-piotr (approved):

PR #469, opened 2022-06-09
Clean up type definitions in common

sz-piotr:


sin6pi7:
`backend` module cannot be built, because `common` module does not export everything that's necessary (check out failed gh action)

PR #481, opened 2022-06-14
Update schema

michalsidzej:


packages/backend/src/peripherals/database/types.ts
```diff
@@ -10,12 +10,14 @@ declare module 'knex/types/tables' {
 
   interface PriceRow {
     coingecko_id: string
+    asset_id?: string
```
sin6pi7, line 13: Why is this optional?

michalsidzej: I just wanted to stay without TS errors everywhere

packages/backend/src/peripherals/database/types.ts
```diff
@@ -10,12 +10,14 @@ declare module 'knex/types/tables' {
...
     price_usd: number
     unix_timestamp: string
   }
 
   interface BalanceRow {
     block_number: number
+    unix_timestamp?: string
```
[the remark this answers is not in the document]
michalsidzej: jw

packages/backend/src/api/controllers/report/filter/filterReportsByProjects.ts
```diff
@@ -1,36 +1,31 @@
...
       return false
     }
     return true
   })
 }
 
-function getBridges(projects: ProjectInfo[]) {
+// change
+// now tokens are filtered based on project not bridge
+// todo discuss it
```
sin6pi7, line 23: Do we have a conclusion here?

packages/backend/src/core/ReportUpdater.ts
```diff
@@ -58,20 +80,64 @@ export class ReportUpdater {
...
+    )
+
+    console.log('balancesPerProject', balancesPerProject)
+
     const reports: ReportRecord[] = []
-    for (const balance of balances) {
+    for (const balance of balancesPerProject) {
       const token = this.tokenByAssetId.get(balance.assetId)
       if (!token) {
+        console.log('dupa', balance.assetId, this.tokenByAssetId)
```
sin6pi7, line 94: I guess these `console.log`s are not supposed to be here 😆

packages/backend/src/peripherals/database/migrations/013_update_balances.ts
```diff
@@ -0,0 +1,61 @@
...
+export async function up(knex: Knex) {
+  await knex.schema.alterTable('asset_balances', (table) => {
+    table.bigInteger('unix_timestamp')
+  })
+  await knex.raw(`
+    UPDATE asset_balances
+    SET unix_timestamp = (
+      SELECT unix_timestamp
+      FROM block_numbers
+      WHERE asset_balances.block_number = block_numbers.block_number
```
sin6pi7, line 25: Does this work as expected?

packages/backend/src/api/controllers/report/filter/filterReportsByProjects.ts
michalsidzej: Let's raise it here: We are now filtering the reports based on projects, not bridge. In an event that our config changes in a way, that one of the bridges (b1) stops tracking a particular asset AND other bridge does track it, this filter will not discard reports from b1. I guess this is an unlikely event, so it shouldn't be a problem to keep it as it is for now.

packages/backend/src/peripherals/database/migrations/013_update_balances.ts
michalsidzej: Yes, updates unix_timestamp column with a on block_number equality

packages/backend/src/api/controllers/report/filter/filterReportsByProjects.ts
sz-piotr, line 23: I think that ultimately there should be no or minimal filtering done. The filtering should occur in ReportUpdater where non-conforming records are deleted from the database

packages/backend/src/api/controllers/report/filter/getSufficientlySynced.ts
```diff
@@ -1,16 +1,16 @@
 import { ReportRecord } from '../../../../peripherals/database/ReportRepository'
-import { getMaxAssetInBridge } from './getMaxAssetsInBridge'
+import { getNewestAssetsInProject } from './getNewestAssetsInProject'
 import { getSyncedTimestamp } from './getSyncedTimestamp'
 
 export function getSufficientlySynced(reports: ReportRecord[]) {
```
sz-piotr, line 5: Again with the filtering I think that in the future this logic should just be removed. As long as ReportUpdater leaves only relevant records it's fine.

packages/backend/src/core/ReportUpdater.ts
```diff
@@ -14,37 +20,53 @@ import {
...
   ) {
     this.logger = this.logger.for(this)
     for (const token of this.tokens) {
       this.tokenByAssetId.set(token.id, token)
     }
+
+    for (const { projectId, bridges } of projects) {
+      this.projectDetailsById.set(projectId, {
+        bridges: bridges.map((b) => EthereumAddress.unsafe(b.address)),
+        assetIds: bridges.flatMap((b) => b.tokens.map((t) => t.id)),
```
sz-piotr, line 51: This should probably be deduplicated

packages/backend/src/peripherals/database/BlockNumberRepository.ts
```diff
@@ -35,6 +35,16 @@ export class BlockNumberRepository extends BaseRepository {
     return rows.map(toRecord)
   }
 
+  async findByBlockNumber(blockNumber: bigint): Promise<BlockNumberRecord> {
+    const knex = await this.knex()
+    const [row] = await knex('block_numbers').where(
```
sz-piotr, line 40: You should use .first() and hadle undefined

sz-piotr (changes requested):

sz-piotr (approved):

PR #483, opened 2022-06-14
Align report dates

sz-piotr:


packages/backend/src/peripherals/database/BalanceRepository.ts
```diff
@@ -51,6 +51,7 @@ export class BalanceRepository extends BaseRepository {
       .insert(rows)
       .onConflict(['block_number', 'holder_address', 'asset_id'])
       .merge()
+      .returning('block_number')
```
sin6pi7, line 54: Why is this necessary?

sz-piotr: Reverted this changed and fixed it properly in #484

michalsidzej (approved):

PR #494, opened 2022-06-17
Update prices table

michalsidzej:


packages/backend/src/peripherals/coingecko/CoingeckoQueryService.ts
```diff
@@ -38,11 +40,15 @@ export class CoingeckoQueryService {
   }
 
   async queryPrices(
-    coinId: CoingeckoId,
+    assetId: AssetId,
     from: UnixTime,
     to: UnixTime,
     granularity: Granularity,
   ): Promise<Price[]> {
+    const coinId = tokenList.find((token) => token.id === assetId)?.coingeckoId
```
antooni, line 48: const coinId = getTokenByAssetId(token)?.coingeckId

antooni (changes requested):
good job!

packages/backend/src/peripherals/coingecko/CoingeckoQueryService.ts
sz-piotr, line 48: This shouldn't work this way, because it introduces a hidden dependency on the token list. One way to resolve it would be to pass the token list in constructor, but I think that CoingeckoQueryService should only care about coingeckoIds and PriceUpdater should receive a list of all tokens and do the conversion there.

packages/backend/src/peripherals/database/migrations/012_update_prices.ts
```diff
@@ -0,0 +1,57 @@
...
+    tokenList.map(({ id, coingeckoId }) => {
+      return knex('coingecko_prices')
+        .update({ asset_id: id.toString() })
+        .whereRaw('coingecko_id = ?', [coingeckoId.toString()])
+    }),
+  )
+  await knex.schema.alterTable('coingecko_prices', (table) => {
+    table.dropPrimary()
+    table.string('asset_id').notNullable().alter()
+    table.setNullable('coingecko_id')
```
sz-piotr, line 32: Why not drop it completely?

sz-piotr (changes requested):

packages/backend/src/peripherals/database/migrations/012_update_prices.ts
michalsidzej: Yes, that's better.

packages/backend/src/peripherals/coingecko/CoingeckoQueryService.ts
michalsidzej: It doesn't solve the problem of a hidden dependency. https://github.com/l2beat/l2beat/pull/494#discussion_r900192399

sz-piotr (approved):

PR #497, opened 2022-06-17
Add reports status

antooni:


antooni:
it is basically the same as in balances

sin6pi7 (approved):

PR #502, opened 2022-06-20
Add env variable to disable syncing

antooni:


sin6pi7 (changes requested):

packages/backend/README.md
```diff
@@ -98,6 +98,7 @@ You can configure the behavior of the app with the following environment variabl
 - `TEST_DB_URL` - Database url used in `yarn test`
 - `LOG_LEVEL` - Integer specifying the log level. See `src/services/Logger.ts`
 - `PORT` - The port on which the application exposes the api
+- `DISABLE_SYNC` (default `false`) - When set to `true` it will prevent app from syncing, useful for local debugging
```
sin6pi7, line 101: It's easier to think about a variable if the default value is true. Therefore I'd rather make it just `SYNC`, which defaults to true in all configs. Then in `Application` you just check `if (config.sync)`.

sin6pi7 (approved):

packages/backend/README.md
sz-piotr: I personally believe that defaulting to false is more intuitive, because then it's consistent with what happens in JS. undefined and false are falsy and true is truthy

sz-piotr (approved):

PR #504, opened 2022-06-20
HTML for prices status

antooni:
- add getLatestByToken to PriceRepository
- add html to prices status


packages/backend/.gitignore
```diff
@@ -1,3 +1,3 @@
 node_modules
 build
-.env
+.env
```
sin6pi7, line 3: Why? The default is to keep a newline in other files afaik.

packages/backend/src/Application.ts
```diff
@@ -115,7 +115,7 @@ export class Application {
     )
 
     const statusController = new StatusController(
-      priceRepository,
+      priceUpdater,
```
sin6pi7, line 118: Why? `PriceUpdater.getStatus` just proxies the call to repository, it does not bring any additional logic.

antooni: It is more of a contextual choice. We treat PriceUpdater as a worker, so to ask him about his state of work seems more intuitive for me. I am open to rethink it if u have other perspective on it

antooni: The end game is to put only updaters in StatusController, so person looking at the code will see that this is a status of updaters work

sin6pi7: I see your point. I am OK with this, but let's stay consistent - we can add this interface for all statuses in a separate PR, for now let's keep repositories.

packages/backend/src/api/controllers/status/StatusController.ts
```diff
@@ -89,28 +75,45 @@ export class StatusController {
...
+  return !!timestamp && now.equals(timestamp)
+}
+
 const unixTimeToString = (date: UnixTime) => {
   return date.toDate().toString().slice(4, 21)
 }
 
 const SECONDS_PER_HOUR = 60 * 60
-const getSyncStatus = (date: UnixTime): string => {
+const getSyncStatus = (
```
sin6pi7, line 88: I presume this function will get removed once we have html pages for every kind of status?

antooni: yes

packages/backend/src/api/routers/StatusRouter.ts
```diff
@@ -1,19 +1,19 @@
 import Router from '@koa/router'
 
-import { StatusController } from '../controllers/StatusController'
+import { StatusController } from '../controllers/status/StatusController'
 
 export function createStatusRouter(statusController: StatusController) {
   const router = new Router()
 
-  router.get('/api/status/prices', async (ctx) => {
+  router.get('/status/prices', async (ctx) => {
```
sin6pi7, line 8: Not urgent, but I guess we'd like to have basicauth for all these endpoints on prod? cc @sz-piotr

packages/backend/src/api/controllers/status/view/renderHtml.ts
```diff
@@ -0,0 +1,6 @@
+import { ReactElement } from 'react'
+import { renderToStaticMarkup } from 'react-dom/server'
+
+export function renderHtml(page: ReactElement) {
```
sin6pi7, line 4: `reactToHtml(element: ReactElement)` - this signature is more explicit

packages/backend/src/api/controllers/status/StatusController.ts
```diff
@@ -23,31 +18,22 @@ export class StatusController {
...
+        return {
+          name: token.coingeckoId.toString(),
+          timestamp: latest?.timestamp.toNumber(),
+          value: latest?.priceUsd.toString(),
+          isSynced: isSynced(latest?.timestamp),
+        }
       })
-    }
-    return result
+      .sort((a, b) => Number(a.isSynced) - Number(b.isSynced))
```
sin6pi7, line 34: Wouldn't it be nicer to sort by `timestamp` itself? That way it's guaranteed that two rows with the same value of `isSynced` are actually ordered by what was the list time they were synced.

antooni: this feature was intended to move unsynced values to the top, by sorting with timestamp we will loose the alphabetical order. I am not sure whether this is intended behaviour.

packages/backend/src/api/controllers/status/Status.ts
```diff
@@ -0,0 +1,6 @@
+export interface Status {
+  name: string
+  timestamp: number | undefined
```
sin6pi7, line 3: Why not `UnixDate`?

packages/backend/src/api/controllers/status/Status.ts
```diff
@@ -0,0 +1,6 @@
...
+  value: string | undefined
```
sin6pi7, line 4: Why `undefined` instead of optional values?

antooni: `timestamp: latest?.timestamp.toNumber()` during object creation we are using .? operator so it will set undefined. Having optional parameter is more clean?

packages/backend/src/api/controllers/status/view/StatusPage.tsx
```diff
@@ -0,0 +1,78 @@
...
+            <th>Latest value</th>
+            <th>Status</th>
+          </tr>
+        </thead>
+        <tbody>
+          {statuses.map((status, i) => (
+            <tr key={i} style={{ color: status.isSynced ? undefined : 'red' }}>
+              <td>{status.name}</td>
+              <td>{status.timestamp ?? '-'}</td>
+              <td>{timestampToIsoString(status.timestamp) ?? '-'}</td>
```
sin6pi7, line 36: If `timestamp` was `UnixTime`:
```suggestion
<td>{status.timestamp?.toDate().toISOString() ?? '-'}</td>
```

packages/backend/src/api/controllers/status/Status.ts
sin6pi7: "clean" is not very specific, but for sure it is less verbose. It has a different meaning though, because optional parameter does not have to be explicitly set as `undefined` whereas with `| undefined`  it does have to. This property is neither good nor bad - it just depends on what we want to achieve. As far as my understanding goes, here we don't gain anything from explicit union with `undefined`, but the code becomes more verbose.

sin6pi7 (approved):

PR #505, opened 2022-06-20
Remove old code

sz-piotr:


packages/config/package.json
```diff
@@ -5,12 +5,11 @@
   "main": "build/src/index.js",
   "types": "build/src/index.d.ts",
   "scripts": {
+    "build": "tsc",
     "format": "prettier --check .",
     "format:fix": "prettier --write .",
     "lint": "eslint --ext .ts --max-warnings 0 src test",
     "lint:fix": "yarn lint --fix",
-    "typecheck": "tsc --noEmit",
```
sin6pi7, line 12: Why are we removing typecheck commands?

sz-piotr: Because it's redundant. Build already covers that use case, but only for some packages. Removing that command skips running it for those packages on CI

sin6pi7 (approved):

PR #508, opened 2022-06-22
Fix BalanceRepository tests

sin6pi7:


packages/backend/test/peripherals/database/BalanceRepository.test.ts
```diff
@@ -215,11 +215,13 @@ describe(BalanceRepository.name, () => {
 
     repository.addOrUpdateMany(additionalRecords)
```
sz-piotr, line 216: add await here too

sz-piotr (changes requested):

sz-piotr (approved):

PR #510, opened 2022-06-23
Statuses html

sin6pi7:


packages/backend/src/api/controllers/status/StatusController.ts
```diff
@@ -1,119 +1,120 @@
...
+    return tokenDecimals
   }
 
-  if (diff > 0 && diff < 24 * SECONDS_PER_HOUR) {
-    return {
-      isSynced: false,
-      message: `out of sync for ${diff / SECONDS_PER_HOUR} hour(s)`,
-    }
+  private getDecimals(assetId: AssetId): number {
+    return this.tokenDecimals.get(assetId) ?? 0
```
michalsidzej, line 108: We sometimes use the pattern that if map.get() returnes undefined, we throw `Programmer error`. Do you think it makes sense here?

packages/backend/src/api/controllers/status/view/BalancesPage.tsx
```diff
@@ -0,0 +1,51 @@
...
+  balances: Balance[]
+}
+
+export function BalancesPage({ balances }: BalancesPageProps) {
+  return (
+    <Page title="Balances">
+      {balances.map((balance) => (
+        <div>
+          <h2>
+            {balance.holderAddress} - {balance.holderAddress}
```
michalsidzej, line 31:
```suggestion
            {balance.projectName} - {balance.holderAddress}
```

packages/backend/src/api/controllers/status/StatusController.ts
antooni: hmm, interesting idea, I am just not sure whether throwing is needed here, because if there is no decimal then it will just output 123456789 instead of e.g. 123.46

packages/backend/src/api/controllers/status/view/ReportsPage.tsx
```diff
@@ -0,0 +1,57 @@
...
+  reports: Report[]
+}
+
+export function ReportsPage({ reports }: ReportsPageProps) {
+  return (
+    <Page title="Reports">
+      {reports.map((balance) => (
+        <div>
+          <h2>
+            {balance.holderAddress} - {balance.holderAddress}
```
michalsidzej, line 33:
```suggestion
            {balance.projectName} - {balance.holderAddress}
```

packages/backend/src/api/controllers/status/StatusController.ts
sin6pi7: In this case throwing is not suitable, because it is an expected case (something might not have synced yet).

packages/backend/src/api/controllers/status/view/StatusTable.tsx
```diff
@@ -0,0 +1,36 @@
...
+  rows: Row[]
+}
+
+export function StatusTable(props: StatusTableProps) {
+  return (
+    <table>
+      <thead>
+        <tr>
+          {props.columns.map((column) => (
+            <th>{column}</th>
```
sz-piotr, line 19: Missing react key

packages/backend/src/api/routers/StatusRouter.ts
```diff
@@ -1,21 +1,37 @@
...
-    ctx.body = await statusController.getReportsStatus()
+    const { from, to } = getFromTo(ctx.query)
+
+    ctx.body = await statusController.getReportsStatus(from, to)
   })
 
   return router
 }
+
+function getFromTo(query: ParsedUrlQuery): { from: UnixTime; to: UnixTime } {
```
sz-piotr, line 31: I would have from and to be possibly undefined and then StatusController could set this to Clock.getLastHour() and Clock.getFirstHour()

packages/backend/src/peripherals/database/BalanceStatusRepository.ts
```diff
@@ -48,4 +50,20 @@ export class BalanceStatusRepository extends BaseRepository {
     const knex = await this.knex()
     return await knex('balance_status').delete()
   }
+
+  async getFromTo(
```
sz-piotr, line 54: Rename FromTo to Between everywhere

sz-piotr: Also note that this method should be wrapped in the constructor

packages/backend/src/peripherals/database/PriceRepository.ts
```diff
@@ -95,11 +95,16 @@ export class PriceRepository extends BaseRepository {
     )
   }
 
-  async getLatestByToken(): Promise<Map<AssetId, PriceRecord>> {
+  async getLatestByTokenFromTo(
```
sz-piotr, line 98: wrap in the constructor

packages/backend/src/Application.ts
```diff
@@ -129,10 +129,11 @@ export class Application {
 
     const statusController = new StatusController(
       priceRepository,
-      balanceRepository,
-      reportRepository,
+      balanceStatusRepository,
+      reportStatusRepository,
       config.tokens,
       config.projects,
+      clock,
```
sz-piotr, line 136: Pass clock before tokens

sz-piotr (approved):

PR #512, opened 2022-06-27
Fix eslint setup

sz-piotr:


packages/frontend/src/content/pages/Project/props/getChart.ts
```diff
@@ -26,6 +26,6 @@ function getTokens(project: Project, l2Data: L2Data) {
 }
 
 function getTVL(project: Project, l2Data: L2Data, token: string) {
-  const data = l2Data.byProject[project.name].byToken[token].data
-  return data[data.length - 1][2]
+  const data = l2Data.byProject[project.name]?.byToken[token]?.data
```
antooni, line 29: much needed :D

packages/backend/src/peripherals/database/shared/migrations.d.ts
```diff
@@ -0,0 +1,13 @@
+declare module 'knex/lib/migrations/migrate/sources/fs-migrations' {
```
antooni, line 1: should .d.ts file be tracked with git?

antooni (approved):
great job, much needed changes

packages/backend/src/peripherals/database/shared/migrations.d.ts
sz-piotr: Yes, because it is a custom .d.ts file hand written by me. The @types/knex pacakage doesn't include typings for this because it's only used as a workaround to support .js/.ts migrations

PR #513, opened 2022-06-28
feature - add too data endpoint tvl data filtered by project technology

ImmanuelSegol:


sz-piotr:
The idea makes sense, but the implementation less so. If you wanted to add this to the frontend then you don't really need to add any more data to the API, because it's already there. I think it would be better to sum calculate the sum by technology during the generation of the frontend.

Also, in the coming week we'll be moving away from the one big json with everything in it anyway

ImmanuelSegol:
> The idea makes sense, but the implementation less so. If you wanted to add this to the frontend then you don't really need to add any more data to the API, because it's already there. I think it would be better to sum calculate the sum by technology during the generation of the frontend.
> 
> Also, in the coming week we'll be moving away from the one big json with everything in it anyway

Hey,
I didn't implement it in the frontend just because I wanted to hide the complexity from the `createApi` function.
That being the case should I close this PR or should I implement this in a different way?

ImmanuelSegol:
> > That being the case should I close this PR or should I implement this in a different way?
> 
> I think that the first thing to do is decide on how this would look on the frontend. Would this only be available as an endpoint? If you want to display the chart how would the user view the chart?

The frontend design would need to be defined. If you are interested I can help with that as well.
In the meanwhile, I wanted to add this to the endpoint.

sz-piotr:
I'm against adding this data to the endpoint at this point. The endpoint already returns something like 20mb of data, there is no need to make it larger especially with data that already is in the endpoint. I guess I'm open to having a separate endpoint for this specifically.

ImmanuelSegol:
@sz-piotr Sorry for not responding. This feature seems a bit redundant from your comments. If you need any help with other tasks feel free to write me.

PR #515, opened 2022-06-28
Change sinceBlock to sinceTimestamp

michalsidzej:


packages/config/src/projects/types/ProjectBridge.ts
```diff
@@ -1,8 +1,10 @@
+import { UnixTime } from '@l2beat/common'
+
 export interface ProjectBridge {
   /** Address of the bridge. Use etherscan to verify its correctness. */
   address: string
   /** Block number of the deployment transaction of the bridge contract. */
```
sz-piotr, line 6: Update the comment

since.txt
```diff
@@ -0,0 +1,291 @@
+packages/backend/test/api/controllers/report/aggregateReportsDaily.test.ts:
```
sz-piotr, line 1: This file shouldn't be committed

update.js
```diff
@@ -0,0 +1,284 @@
+const fs = require("fs");
```
sz-piotr, line 1: This file shouldn't be committed

packages/config/src/tokens.ts
```diff
@@ -41,7 +41,7 @@ export interface TokenInfo {
   /** Token decimals as dictated by the token contract */
   decimals: number
   /** Block number of the token contract deployment transaction */
```
sz-piotr, line 43: Update comment

sz-piotr (changes requested):

PR #520, opened 2022-06-30
Check sinceTimestamp before creating a report

michalsidzej:


packages/backend/test/core/reports/fakes.ts
```diff
@@ -90,7 +90,7 @@ export function fakeTokenInfo(token: Partial<TokenInfo>): TokenInfo {
     symbol: 'FKT',
     decimals: 18,
     address: EthereumAddress.random(),
-    sinceTimestamp: new UnixTime(0),
+    sinceTimestamp: new UnixTime(0).add(2, 'days'),
```
sz-piotr, line 93: I think this one needs a comment

sz-piotr (changes requested):

packages/backend/test/core/reports/createReports.test.ts
```diff
@@ -35,3 +38,87 @@ describe(createReports.name, () => {
     ])
   })
 })
+
+describe(aggregateBalancesPerProject.name, () => {
+  it('discards balances before bridge sinceTimestamp', () => {
```
antooni, line 43:
```suggestion
  it('skips balances before bridge sinceTimestamp', () => {
```

antooni: Do I understand it correctly that there is no discarding, just filtering ?

antooni (changes requested):

packages/backend/test/core/reports/createReports.test.ts
michalsidzej: Yes, that's right.

sz-piotr (approved):

PR #522, opened 2022-07-01
zkSync monitor

sin6pi7:


krzkaczor (approved):
eth-sdk integration looks good

packages/zksync-monitor/src/run.ts
```diff
@@ -0,0 +1,96 @@
...
+    Contracts.mainProxy,
+    additionalAddressPosition,
+  )
+  return '0x' + cellValue.slice(-40)
+}
+
+export async function run() {
+  dotenv.config()
+  const alchemyApiKey = process.env.ALCHEMY_API_KEY
+  const provider = new providers.AlchemyProvider('mainnet', alchemyApiKey)
```
krzkaczor, line 60: it's the very first time i have seen integration with RPC node done this way. 

If you use just a regular ethers JSON provider + put node URL in env you avoid alchemy lock-in.

PR #531, opened 2022-07-08
Define /api/main format

sin6pi7:
How to get derived data during website build:
 - total tvl -> main.charts.hourly[last][1]
 - project total tvl -> main.projects[index].charts.hourly[last][1]
 - breakdown -> the same calculation as we do now, but get category from config.tokens[assetId].category and tvl
 - token display name -> config.tokens[assetId].name
 - tvl warning -> use breakdown and compare with some value
 - market share -> project total tvl / total tvl
 - 7 day change -> [main || projects[index]].charts.daily[last][1] - [main || projects[index]].charts.daily[last - 7][1]


packages/common/src/types/ApiMain.ts
```diff
@@ -0,0 +1,43 @@
...
+const Chart = z.object({
+  types: z.tuple([z.literal('timestamp'), z.string(), z.string()]),
+  data: z.array(ChartPoint),
+})
+export type Chart = z.infer<typeof Chart>
+
+export const Charts = z.object({
+  hourly: Chart,
+  qh6: Chart,
+  daily: Chart,
```
sz-piotr, line 23: Maybe `everyHour`, `every6Hours` and `everyDay`?

sz-piotr (changes requested):

packages/common/src/types/ApiMain.ts
sz-piotr: Or even `sixHourly` which is maybe not fully grammatically correct but self explanatory

michalsidzej: `sixHourly` sounds good to me

PR #537, opened 2022-07-11
Update permissioned addresses

sz-piotr:


packages/config/src/projects/bobanetwork.ts
```diff
@@ -143,6 +143,39 @@ export const bobanetwork: Project = {
...
+        {
+          name: 'Owner',
+          accounts: [
+            {
+              address: '0x1f2414D0af8741Bc822dBc2f88069c2b2907a840',
+              type: 'EOA',
+            },
+          ],
+          description:
+            'This address is the owner of the following contracts: OVM_L1CrossDomainMessenger, L1StandardBridge, LibAddressManager. This allows it to censor messages or pause message bridge altogether, upgrade bridge implementation potentially gaining access to all funds stored in a bridge and change the sequencer, block proposer or any other system component (unlimited upgrade power).',
```
antooni, line 156: maybe we should link to the contract's address ?

sz-piotr: I agree, that would be cool, but it would be quite a challenge to implement this right now into the code, as we'd have to add inline markdown support.

michalsidzej (approved):

PR #548, opened 2022-07-13
[/api/main] Adjust frontend

michalsidzej:


packages/frontend/src/static/_redirects
```diff
@@ -4,4 +4,5 @@
 /project/*                  /projects/:splat
 /projects/leverj            /projects/gluon
 /projects/fuel              /projects/fuelv1
-/projects/zkswapv2          /projects/zkspace
\ No newline at end of file
+/projects/zkswapv2          /projects/zkspace
+/api/projects/:project/tvl/assets/:asset          https://api.l2beat.com/api/projects/:project/tvl/assets/:asset
```
sz-piotr, line 8: What about serving locally?

sin6pi7: I guess we could have a `API_BASE_URL` env var, which would prepended (if set in the environment) to paths in the frontend that should essentially point to api e.g. here https://github.com/l2beat/l2beat/pull/548/files#diff-e19d3950687c41524d803f6f4da536d6ff3a3604efb7c54a168004cca148502aR19

packages/common/src/types/UnixTime.ts
```diff
@@ -105,4 +105,8 @@ export class UnixTime {
   toJSON() {
     return this.timestamp
   }
+
+  toDateDays() {
```
sz-piotr, line 109: Maybe `toYYYYMMDD()` or `toFormat('YYYY-MM-DD')` where `"YYYY-MM-DD"` would be the only accepted parameter.

packages/frontend/src/content/pages/MetaImages/MetaImage.tsx
```diff
@@ -23,7 +23,7 @@ export function MetaImage(props: MetaImageProps) {
           tvl={props.tvl}
           sevenDayChange={props.sevenDayChange}
         />
-        <Chart endpoint={props.apiEndpoint} days={30} />
+        <Chart endpoint={props.apiEndpoint} days={7} />
```
sz-piotr, line 26: Why do we want this change? Why not keep 30 days?

packages/frontend/src/content/pages/Home/props/getProps.ts
```diff
@@ -1,27 +1,29 @@
...
 import { getFinancialView } from './getFinancialView'
 import { getPageMetadata } from './getPageMetadata'
 import { getRiskView } from './getRiskView'
 
-export function getProps(projects: Project[], l2Data: L2Data): HomePageProps {
-  const tvl = getFromEnd(l2Data.aggregate.data, 0)?.[1] ?? 0
-  const tvlSevenDaysAgo = getFromEnd(l2Data.aggregate.data, 7)?.[1] ?? 0
+export function getProps(projects: Project[], apiMain: ApiMain): HomePageProps {
+  const tvl = getFromEnd(apiMain.charts.daily.data, 0)?.[1] ?? 0
+  const tvlSevenDaysAgo = getFromEnd(apiMain.charts.daily.data, 7)?.[1] ?? 0
```
sz-piotr, line 12: If we're using hourly data why would we display a daily tvl in the list?

packages/frontend/src/content/pages/Project/props/getHeader.ts
```diff
@@ -1,13 +1,13 @@
...
 import { formatUSD, getFromEnd, getPercentageChange } from '../../../utils'
 
-export function getHeader(project: Project, l2Data: L2Data): HeaderProps {
-  const aggregate = l2Data.byProject[project.name]?.aggregate.data ?? []
-  const tvl = getFromEnd(aggregate, 0)?.[1] ?? 0
-  const tvlSevenDaysAgo = getFromEnd(aggregate, 7)?.[1] ?? 0
+export function getHeader(project: Project, apiMain: ApiMain): HeaderProps {
+  const daily = apiMain.projects[project.name]?.charts.daily.data ?? []
+  const tvl = getFromEnd(daily, 0)?.[1] ?? 0
+  const tvlSevenDaysAgo = getFromEnd(daily, 7)?.[1] ?? 0
```
sz-piotr, line 10: Same daily/hourly concern here

packages/frontend/src/scripts/chart/state/makeChartState.ts
```diff
@@ -27,7 +27,11 @@ export function makeChartState(chart: HTMLElement, onChange: () => void) {
 
   onRadioChange(controls.range, (control) => {
     state.days = toDays(control.value)
-    onChange()
+    if (state.endpoint) {
+      updateInput(state.endpoint)
+    } else {
+      onChange()
+    }
```
sz-piotr, line 34: For some reason selecting 30D and 90D shows the exact same data.

sz-piotr: Also, could someone explain this change?

sz-piotr (changes requested):
Left comments :)

packages/frontend/src/content/pages/MetaImages/MetaImage.tsx
sin6pi7: Refactor mistake, thanks.

packages/frontend/src/scripts/chart/ui/toUiState.ts
```diff
@@ -16,9 +23,11 @@ export function toUiState(state: ChartStateWithInput): UiState {
       points: [],
     }
   }
-  const dateRange = toDateRange(
+  const withTime = state.days === 7
```
sz-piotr, line 26: state.days <= 90, because we still want to display an hour then.

sz-piotr (changes requested):

sz-piotr:
Also, change the `on.schedule.cron` in `.github/workflows/deploy-site.yml` to every hour.

sz-piotr (approved):

PR #549, opened 2022-07-13
Add invert command

sz-piotr:


sin6pi7 (approved):
Big ❤️  for the handcrafted usage!

PR #557, opened 2022-07-14
Aggregate reports

michalsidzej:


sin6pi7 (approved):

sin6pi7:
Currently OP token is added to `/api/main` based on `(aggregated_)reports` on the fly during http request handling. The upsides are savings on storage space (we don't store denormalised data) and instant reflection of changes in calculations on the api. However, this complicates the code a lot and increases the response time and memory footprint during http request handling. Maybe it would be a good idea to add OP token to `(aggregated_)reports` in `ReportUpdater` This way we keep special-case logic in one place at the very beginning of processing pipeline at the small expense of a bit more data being stored. Wdyt @michalsidzej @sz-piotr ?

sz-piotr:
> Currently OP token is added to `/api/main` based on `(aggregated_)reports` on the fly during http request handling. The upsides are savings on storage space (we don't store denormalised data) and instant reflection of changes in calculations on the api. However, this complicates the code a lot and increases the response time and memory footprint during http request handling. Maybe it would be a good idea to add OP token to `(aggregated_)reports` in `ReportUpdater` This way we keep special-case logic in one place at the very beginning of processing pipeline at the small expense of a bit more data being stored. Wdyt @michalsidzej @sz-piotr ?

I think this makes sense. We also have a relatively easy to use mechanism to recalculate this, so it should be fine. And I like simplifying request handlers

packages/backend/test/peripherals/database/fakes.ts
```diff
@@ -0,0 +1,5 @@
+import { Hash256 } from '@l2beat/common'
+
+export function fakeHash(hash?: Hash256, forcedResyncId = 0): string {
```
sz-piotr, line 3: Maybe `fakeConfigHash`

packages/backend/test/core/reports/optimism.test.ts
```diff
@@ -0,0 +1,89 @@
...
+
+import {
+  addOpTokenReport,
+  OP_TOKEN_ID,
+  OP_TOKEN_SINCE_TIMESTAMP,
+  OPTIMISM_PROJECT_ID,
+} from '../../../src/core/reports/optimism'
+import { ReportRecord } from '../../../src/peripherals/database/ReportRepository'
+
+describe(addOpTokenReport.name, () => {
```
sz-piotr, line 12: It needs a test with multiple reports so that only the relevant one gets updated

packages/backend/src/core/reports/charts.ts
```diff
@@ -0,0 +1,26 @@
...
+  return timestamps
+}
+
+export function addMissingDailyTimestamps(points: ChartPoint[]): ChartPoint[] {
+  if (points.length === 0) return []
+  const [min] = points[0]
+  const [max] = points[points.length - 1]
+  const daily = getDailyTimestamps(min, max)
+
+  return daily.reduce((acc, timestamp, i) => {
```
sz-piotr, line 17: This algorithm would be more efficient and cleaner if the reduce was just replaced with the loop from getDailyTimestamps and getDailyTimestamps would go away.

packages/backend/src/peripherals/database/migrations/018_aggregate_reports.ts
```diff
@@ -0,0 +1,29 @@
...
+
+import { Knex } from 'knex'
+
+export async function up(knex: Knex) {
+  await knex.schema.createTable('aggregate_reports', function (table) {
+    table.bigInteger('unix_timestamp').notNullable()
+    table.string('project_id').notNullable()
+    table.decimal('tvl_usd', 80, 0).notNullable()
+    table.decimal('tvl_eth', 80, 0).notNullable()
+    table.boolean('is_daily').notNullable()
```
sz-piotr, line 22: Add index

packages/backend/src/peripherals/database/migrations/018_aggregate_reports.ts
```diff
@@ -0,0 +1,29 @@
...
+should create a new migration file that fixes the issue.
+
+*/
+
+import { Knex } from 'knex'
+
+export async function up(knex: Knex) {
+  await knex.schema.createTable('aggregate_reports', function (table) {
+    table.bigInteger('unix_timestamp').notNullable()
+    table.string('project_id').notNullable()
```
sz-piotr, line 19: Add index

packages/backend/src/peripherals/database/migrations/018_aggregate_reports.ts
```diff
@@ -0,0 +1,29 @@
...
+If you find that something was incorrectly set up in the `up` function you
+should create a new migration file that fixes the issue.
+
+*/
+
+import { Knex } from 'knex'
+
+export async function up(knex: Knex) {
+  await knex.schema.createTable('aggregate_reports', function (table) {
+    table.bigInteger('unix_timestamp').notNullable()
```
sz-piotr, line 18: Add index (unless Postgesql creates indexes for the fields separately when the public key is a composite)

packages/backend/src/peripherals/database/ReportRepository.ts
```diff
@@ -48,10 +56,15 @@ export class ReportRepository extends BaseRepository {
   async addOrUpdateMany(reports: ReportRecord[]) {
```
sz-piotr, line 56: Maybe we could add a simple check to this function that all reports have the same timestamp?

packages/backend/src/core/reports/createReport.ts
```diff
@@ -51,6 +51,16 @@ export function convertBalance(
   return { balanceUsd, balanceEth }
 }
 
+export function toBalanceUsd(
```
sz-piotr, line 54: This function seems unused

packages/backend/src/core/reports/ReportUpdater.ts
```diff
@@ -49,11 +53,20 @@ export class ReportUpdater {
...
+    const aggregatedReports = aggregateReports(
+      reports,
+      this.projects,
       timestamp,
-    })
+    )
+    await Promise.all([
+      this.reportRepository.addOrUpdateMany(reports),
+      this.aggregateReportsRepository.addOrUpdateMany(aggregatedReports),
+      this.reportStatusRepository.add({
```
sz-piotr, line 65: This shouldn't happen in parallel. Only save the report status after the other operations have completed

sz-piotr (changes requested):
I've left the comments above

packages/backend/src/peripherals/database/migrations/018_aggregate_reports.ts
sin6pi7: Looks like adding an index to just `project_id` cover all the cases (https://devcenter.heroku.com/articles/postgresql-indexes#multi-column-indexes)

packages/backend/src/peripherals/database/ReportRepository.ts
```diff
@@ -60,16 +80,59 @@ export class ReportRepository extends BaseRepository {
...
+  ): Promise<ReportRecord[]> {
+    const knex = await this.knex()
+    const rows = await this.getByProjectAndAssetQuery(
+      knex,
+      projectId,
+      assetId,
+    ).andWhere(
+      'unix_timestamp',
+      '>=',
+      UnixTime.now().add(-7, 'days').toString(),
```
sz-piotr, line 119: This is part of business logic and arguably makes this query less testable. Could we instead pass this as a function argument? Same goes for sixHourly and the other repository.

sz-piotr (changes requested):

packages/backend/src/api/controllers/report/ReportController.ts
```diff
@@ -64,8 +68,8 @@ export class ReportController {
     const dailyTimestamp = timestamp.toStartOf('day')
     const [hourlyReports, sixHourlyReports, dailyReports, latestReports] =
       await Promise.all([
-        this.aggregateReportRepository.getHourly(),
-        this.aggregateReportRepository.getSixHourly(),
+        this.aggregateReportRepository.getHourly(getHourlyMinTimestamp()),
+        this.aggregateReportRepository.getSixHourly(getSixHourlyMinTimestamp()),
```
sz-piotr, line 72: Shouldn't we pass `timestamp` here?

packages/backend/src/api/controllers/report/ReportController.ts
```diff
@@ -122,8 +126,16 @@ export class ReportController {
...
-      this.reportRepository.getSixHourlyByProjectAndAsset(projectId, assetId),
+      this.reportRepository.getHourlyByProjectAndAsset(
+        projectId,
+        assetId,
+        getHourlyMinTimestamp(),
+      ),
+      this.reportRepository.getSixHourlyByProjectAndAsset(
+        projectId,
+        assetId,
+        getSixHourlyMinTimestamp(),
```
sz-piotr, line 137: Shouldn't we pass `this.reportStatusRepository.findLatestTimestamp()` here?

sz-piotr (changes requested):

sz-piotr (approved):

PR #568, opened 2022-08-08
Add EventRepository

antooni:


packages/backend/src/peripherals/database/migrations/020_events.ts
```diff
@@ -0,0 +1,32 @@
...
+
+If you find that something was incorrectly set up in the `up` function you
+should create a new migration file that fixes the issue.
+
+*/
+
+import { Knex } from 'knex'
+
+export async function up(knex: Knex) {
+  await knex.schema.createTable('events', function (table) {
```
sz-piotr, line 17: Add indexes for `unix_timestamp`, `project_id` and `time_span`

packages/backend/src/peripherals/database/EventRepository.ts
```diff
@@ -0,0 +1,80 @@
...
+    this.getAll = this.wrapGet(this.getAll)
+    this.deleteAll = this.wrapDelete(this.deleteAll)
+    this.getByProjectAndName = this.wrapGet(this.getByProjectAndName)
+
+    /* eslint-enable @typescript-eslint/unbound-method */
+  }
+
+  async getByProjectAndName(
+    projectId: ProjectId,
+    name: string,
```
sz-piotr, line 31: We probably would like to allow specifying multiple names at once

sz-piotr (changes requested):

packages/backend/src/peripherals/database/migrations/020_events.ts
antooni: The index created in private key works for single columns, I tested it by sql analyze

packages/backend/src/peripherals/database/EventRepository.ts
antooni: Do you mean event names, or both event names and projectId?

sz-piotr: event names

PR #573, opened 2022-08-10
Add UTC to timestamps on charts

sin6pi7:


packages/frontend/src/shared/dates.ts
```diff
@@ -26,7 +26,7 @@ function parseTimestamp(timestamp: number) {
 }
 
 function formatTimeAndDate(date: string, time?: string) {
-  return time === undefined ? date : `${time} ${date}`
+  return time === undefined ? date : `${time} UTC ${date}`
```
sz-piotr, line 29: what about `${time} ${date} (UTC)`?

sz-piotr (approved):

PR #579, opened 2022-08-11
Speed up token responses

sin6pi7:
- Remove unused code
- Add indices

### DIFF
--- a/packages/backend/src/peripherals/database/ReportRepository.ts
+++ b/packages/backend/src/peripherals/database/ReportRepository.ts
@@ -14,6 +14,9 @@ export interface ReportRecord {
   balance: bigint
 }
 
+export const DAY = 86400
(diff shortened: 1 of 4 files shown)


packages/backend/src/peripherals/database/ReportRepository.ts
```diff
@@ -14,13 +14,15 @@ export interface ReportRecord {
   balance: bigint
 }
 
+export const DAY = 86400
```
sz-piotr, line 17: I think I would prefer it as `UnixTimestamp.DAY` and `UnixTimestamp.SIX_HOURS` or even `UnixTimestamp.HOUR * 6`

sz-piotr (changes requested):

sz-piotr (approved):

PR #583, opened 2022-08-16
Remove useless reports db fields

sin6pi7:


packages/backend/src/peripherals/database/migrations/022_clean_up_reports.ts
```diff
@@ -0,0 +1,30 @@
...
+should create a new migration file that fixes the issue.
+
+*/
+
+import { Knex } from 'knex'
+
+export async function up(knex: Knex) {
+  await knex.schema.alterTable('reports', function (table) {
+    table.dropColumn('is_daily')
+    table.dropColumn('is_six_hourly')
```
sz-piotr, line 19: Does it automatically drop indexes?

sz-piotr (approved):

packages/backend/src/peripherals/database/migrations/022_clean_up_reports.ts
sin6pi7: Yes

PR #605, opened 2022-08-26
Update gluon.ts

Julius080:


sz-piotr:
Same question as with #606

PR #607, opened 2022-08-29
Uptime Monitor

michalsidzej:


packages/backend/src/core/UptimeUpdater.ts
```diff
@@ -0,0 +1,95 @@
...
+    private clock: Clock,
+    private projects: ProjectInfo[],
+    private logger: Logger,
+  ) {
+    this.logger = this.logger.for(this)
+  }
+
+  start() {
+    this.logger.info('Started')
+    return this.clock.onNewMinute(() => {
```
sz-piotr, line 25: I think that it's better not to add new functionality to the Clock because the idea of a delay makes no sense for this real time scenario. I'd just use setInterval here directly.

packages/backend/src/core/UptimeUpdater.ts
```diff
@@ -0,0 +1,95 @@
...
+      this.taskQueue.addToFront()
```
sz-piotr, line 26: Maybe only add if there isn't anything running? We don't want the tasks to pile up if something stalls

packages/backend/src/core/UptimeUpdater.ts
```diff
@@ -0,0 +1,95 @@
...
+    })
+  }
+
+  async update() {
+    this.logger.info('Update started')
+
+    await Promise.allSettled(
```
sz-piotr, line 33: There needs to be a guarantee that this actually settles in under a minute otherwise the system breaks

packages/backend/src/peripherals/uptime/ApiMonitor.ts
```diff
@@ -0,0 +1,94 @@
...
+  starknetBlockSchema,
+  zkspaceTradesSchema,
+  zksyncBlockSchema,
+} from './schemas'
+
+export class ApiMonitor {
+  constructor(private http: HttpClient) {}
+
+  async aztecCheckBlock(url: string, body: string, maxSinceLastUpdate: number) {
+    const { response, time } = await this.http.timedFetch(url, {
```
sz-piotr, line 19: Add a 10s timeout

packages/backend/src/peripherals/uptime/RpcMonitor.ts
```diff
@@ -0,0 +1,53 @@
...
+      params: [{ to, data }, 'latest'],
+      id: 145,
+    }
+
+    return await this.fetchAndParseRpc(url, JSON.stringify(body))
+  }
+
+  private async fetchAndParseRpc(url: string, body: string) {
+    try {
+      const { response, time } = await this.http.timedFetch(url, {
```
sz-piotr, line 33: Add a 10s timeout

packages/backend/src/core/UptimeUpdater.ts
```diff
@@ -0,0 +1,95 @@
...
+      this.projects.flatMap(({ urls }) =>
+        urls?.map((url) => this.checkUptime(url)),
+      ),
+    )
+
+    this.logger.info('Update completed')
+  }
+
+  async checkUptime(uptime: ProjectUptime) {
+    if (uptime.action === 'rpc_estimateGas') {
```
sz-piotr, line 43: I am not a fan of this list of ifs. Ideally somehow the Updater would receive a mapping of action type to handlers and just pick from the mapping

sz-piotr (changes requested):

michalsidzej (commented):
Feature deprioritized

PR #611, opened 2022-08-30
Add /api/events endpoint

antooni:
Resolves L2B-49

image


packages/types/src/UnixTime.ts
```diff
@@ -115,4 +118,15 @@ export class UnixTime {
   toYYYYMMDD() {
     return this.toDate().toISOString().slice(0, 10)
   }
+
+  static isNumberUnixTimestamp(timestamp: number): boolean {
```
sz-piotr, line 122: I find this method quite weird. Its purpose is to guard against constructor throwing, but it uses different logic than the constructor.
I would much prefer it to have the exact same logic as the constructor and be renamed to `isSafeToCast`. That also means that the `YEAR_2019_TIMESTAMP` constant should be removed.

packages/backend/src/api/controllers/events/EventsController.ts
```diff
@@ -0,0 +1,110 @@
...
+      { granularity: 'daily', timestamp: undefined },
+    ]
+
+    await Promise.all(
+      this.projects.map(async ({ projectId, events }) => {
+        const eventNames = events.map((e) => e.name)
+
+        const eventCharts = await Promise.all(
+          config.map((c) =>
+            this.getEventChart(
```
sz-piotr, line 50: I would probably call this.eventsRepository.getByProject directly and transform the result in a pure function that's easy to test.

sz-piotr (changes requested):

packages/backend/src/api/controllers/events/EventsController.ts
```diff
@@ -0,0 +1,110 @@
...
+    minTimestamp: UnixTime | undefined,
+    eventNames: string[],
+  ): Promise<EventChart> {
+    const records = await this.eventsRepository.getByProject(
+      projectId,
+      granularity,
+      minTimestamp,
+    )
+
+    //Record<timestamp, Record<eventName, count>>
```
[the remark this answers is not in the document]
antooni: I know that a good code does not need a comment, but here it is very useful in my opinion. Other way around there would be needed more types and logic which is not needed here if it can be solved with a really barebone type and a comment. But I am open to discussing it.

sz-piotr (approved):

PR #616, opened 2022-09-02
Update Arbitrum One description to Nitro

michalsidzej:
Resolves L2B-48
Closes #618


michalsidzej:
Theres also a new ProxyAdmin for a new bridge with address 0x554723262467F125Ac9e1cDFa9Ce15cc53822dbD

sz-piotr (changes requested):
Instead of doing arbitrum-check add this config to update-monitor

PR #620, opened 2022-09-06
Rename bridges to escrows

sz-piotr:
Resolves L2B-43


packages/config/src/projects/types/ProjectBridge.ts
```diff
@@ -1,9 +1,9 @@
 import { UnixTime } from '@l2beat/types'
```
michalsidzej, line 1: We should change the filename as well

antooni (commented):
LGTM, please change filename of ProjectBridge.ts to ProjectEscrow.ts

PR #622, opened 2022-09-07
Rename projects to layer2s

sz-piotr:
Resolves L2B-44

I set out to rename every occurrence of project to layer2 but realized that actually it makes sense to keep the name project in the context of the backend and the frontend. The reasoning is that project is going to be a superclass of layer2 and bridge


packages/backend/src/config/config.production.ts
```diff
@@ -32,7 +32,7 @@ export function getProductionConfig(): Config {
       ...token,
       priceStrategy: { type: 'market' },
     })),
-    projects: projects.map(projectToInfo),
+    projects: layer2s.map(projectToInfo),
```
michalsidzej, line 35:
```suggestion
    projects: layer2s.map(layer2ToProject),
```

michalsidzej (approved):
Otherwise, LGTM

PR #624, opened 2022-09-08
Add a risk page for bridges

sz-piotr:


packages/config/src/bridges/index.ts
```diff
@@ -19,6 +19,8 @@ import { wormholeBridge } from '../bridges/wormholeBridge'
 import { xDaiBridge } from '../bridges/xDaiBridge'
 import { BridgeDescription } from './types/bridge'
 
+export * from './types/bridge'
```
antooni, line 22: maybe create index.ts inside /types ? I know it is not needed now but can be useful in the future

antooni (approved):
only a small suggestion about exporting (in the comment above), but it does not have to be part of this PR

michalsidzej (approved):

PR #625, opened 2022-09-08
Add events config

antooni:
BEFORE: There is an already merged EventUpdater but configs of projects are empty

AFTER: Project have filled config and EventUpdater is ready to sync behind feature flag

Resolves L2B-28


antooni:
Resolves L2B-28

krzkaczor (changes requested):
Once merged, it will kick off backedn sync of events. I wouldn't merge it as we are redoing the whole thing.

sz-piotr:
> Once merged, it will kick off backend sync of events. I wouldn't merge it as we are redoing the whole thing.

Maybe if we just add a flag to prevent syncing it would be fine to have the configuration on master

krzkaczor:
actually good idea, @antooni can you add here some flag that would prevent the whole thing from running? Then we can merge this PR.

krzkaczor (approved):

PR #632, opened 2022-09-12
TaskQueue async support

sin6pi7:
- Add number of workers to TaskQueue
- Add tests

Resolves L2B-74

### DIFF
--- a/packages/common/src/tools/TaskQueue.ts
+++ b/packages/common/src/tools/TaskQueue.ts
@@ -4,6 +4,7 @@ import { Logger } from './Logger'
 
 export class TaskQueue<T> {
   private queue: T[] = []
+  private busyWorkers = 0
 
   constructor(
     private executeTask: (task: T) => Promise<void>,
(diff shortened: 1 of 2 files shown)


packages/common/src/tools/TaskQueue.ts
```diff
@@ -1,16 +1,27 @@
...
+    )
+  }
+
+  private get allWorkersBusy() {
+    return this.workers === 0
+  }
 
   addIfEmpty(task: T) {
-    if (this.queue.length === 0 && !this.isExecuting) {
+    if (this.queue.length === 0 && !this.allWorkersBusy) {
```
sz-piotr, line 24: This has now changed, where if there are multiple workers, but only some are busy the queue is considered empty. The correct logic should be that the queue is only empty if there are no queued items and there are no items being executed right now.

sz-piotr (changes requested):

packages/common/src/tools/TaskQueue.ts
sin6pi7: Good catch, changed.

PR #634, opened 2022-09-13
Rpc transaction counting

michalsidzej:
Resolves L2B-75


packages/backend/src/Application.ts
```diff
@@ -59,15 +62,21 @@ export class Application {
...
+    const txCountRepository = new TxCountRepository(database, logger)
 
     const http = new HttpClient()
 
-    const provider = new providers.AlchemyProvider(
+    const ethereumProvider = new providers.AlchemyProvider(
       'mainnet',
       config.alchemyApiKey,
     )
+    const optimismProvider = new providers.AlchemyProvider(
```
antooni, line 73: I am a bit afraid that adding more and more providers will bloat the Application.ts code.
Maybe we should create one aggregate Provider which methods would take parameter {network: ethereum | optimism | etc.}

antooni: or to not break to much of current logic just add MultiNetwork provider with network parametrization

packages/backend/src/core/tx-count/RpcBlockDownloader.ts
```diff
@@ -0,0 +1,67 @@
...
+import { TxCountRepository } from '../../peripherals/database/TxCountRepository'
+import { EthereumClient } from '../../peripherals/ethereum/EthereumClient'
+import { Clock } from '../Clock'
+
+export class RpcBlockDownloader {
+  private taskQueue = new TaskQueue<void>(() => this.update(), this.logger)
+  private blockQueue = new TaskQueue(this.getBlock.bind(this), this.logger, 100)
+
+  constructor(
+    private etheteumClient: EthereumClient,
```
antooni, line 13: typo

packages/backend/src/core/tx-count/RpcBlockDownloader.ts
```diff
@@ -0,0 +1,67 @@
...
+    private projectId: ProjectId,
+  ) {
+    this.logger = logger.for(this)
+  }
+
+  async getBlock(number: number) {
+    const block = await this.etheteumClient.getBlock(number)
+    const timestamp = new UnixTime(block.timestamp)
+
+    if (timestamp.gt(this.clock.getLastHour())) {
```
antooni, line 26: I do not quite get this check, how are we able to fetch block greater than .getLastHour() ? Or is it just sanity check?

packages/backend/src/core/tx-count/RpcBlockDownloader.ts
```diff
@@ -0,0 +1,67 @@
...
+    this.logger.info('Started')
+    return this.clock.onEveryHour(() => {
+      this.taskQueue.addIfEmpty()
+    })
+  }
+
+  async update() {
+    this.logger.info('Update started')
+
+    const missingBlocks = await this.txCountRepository.getMissingByProject(
```
antooni, line 48: I assume we can have missing blocks because RPC can return an error?

packages/backend/src/peripherals/database/TxCountRepository.ts
```diff
@@ -0,0 +1,116 @@
...
+    const knex = await this.knex()
+    const row = await knex('tx_count')
+      .where('project_id', projectId.toString())
+      .orderBy('block_number', 'desc')
+      .first()
+
+    return row ? toRecord(row) : undefined
+  }
+
+  async getMissingByProject(projectId: ProjectId) {
```
antooni, line 56: this method looks pretty complex, have you tested it with big dataset?

packages/backend/src/Application.ts
michalsidzej: This is more a POC of this implementation and optimism is just an example here. So no, we will not create 10 more providers in the Application.ts code

packages/backend/src/core/tx-count/RpcBlockDownloader.ts
michalsidzej: This is basically why this feature is tricky. We do not hold timestamp -> blockNumber mapping in the database for all the networks. Current idea is to download all the blocks (until eth_blockNumber) and then discard all that happened after clock.getLastHour(), to avoid dealing with reorgs.

packages/backend/src/core/tx-count/RpcBlockDownloader.ts
michalsidzej: Yes, also the app can crash/be restarted while downloading blocks.

packages/backend/src/peripherals/database/TxCountRepository.ts
michalsidzej: Onnit

packages/backend/src/core/tx-count/RpcBlockDownloader.ts
```diff
@@ -0,0 +1,67 @@
+import { Logger, TaskQueue } from '@l2beat/common'
+import { ProjectId, UnixTime } from '@l2beat/types'
+
+import { TxCountRepository } from '../../peripherals/database/TxCountRepository'
+import { EthereumClient } from '../../peripherals/ethereum/EthereumClient'
+import { Clock } from '../Clock'
+
+export class RpcBlockDownloader {
```
krzkaczor, line 8: This name is confusing. Primary focus of this tx is to count tx in block and insert into database

packages/backend/src/core/tx-count/RpcBlockDownloader.ts
```diff
@@ -0,0 +1,67 @@
...
+  private taskQueue = new TaskQueue<void>(() => this.update(), this.logger)
```
krzkaczor, line 9: perhpaps this could be renamed to `refreshQueue` or `updateQueue` to make it more obvious what it is

packages/backend/src/core/tx-count/RpcBlockDownloader.ts
krzkaczor: This deserves in code comment

packages/backend/src/peripherals/database/migrations/023_tx_count.ts
```diff
@@ -0,0 +1,27 @@
...
+should create a new migration file that fixes the issue.
+
+*/
+
+import { Knex } from 'knex'
+
+export async function up(knex: Knex) {
+  await knex.schema.createTable('tx_count', function (table) {
+    table.string('project_id').notNullable()
+    table.bigInteger('unix_timestamp').notNullable()
```
krzkaczor, line 19: please let's switch to storing dateTime for dates. This makes working with dates in SQL easier. I did it here: https://github.com/l2beat/l2beat/pull/633/files#diff-7a11975ed53236d7e5a99aa0d987cf7f7fd2bba17c978b2a55e9cc710b0e578eR20

krzkaczor (commented):
Generally speaking I like it. We ended up with a simple implementation. Tests are missing for some important things.

packages/backend/src/core/tx-count/RpcBlockDownloader.ts
michalsidzej: BlockTxCountUpdater?

packages/backend/src/core/tx-count/BlockTxCountUpdater.ts
```diff
@@ -0,0 +1,72 @@
...
+      blockNumber: block.number,
+      count: block.transactions.length,
+    })
+  }
+
+  start() {
+    this.logger.info('Started')
+    return this.clock.onEveryHour(() => {
+      this.updateQueue.addIfEmpty()
+    })
```
sz-piotr, line 46:
```suggestion
    this.updateQueue.addIfEmpty()
    return this.clock.onNewHour(() => {
      this.updateQueue.addIfEmpty()
    })
```

packages/backend/src/core/tx-count/BlockTxCountUpdater.ts
```diff
@@ -0,0 +1,72 @@
...
+
+  async update() {
+    this.logger.info('Update started')
+
+    const missingBlocks = await this.txCountRepository.getMissingByProject(
+      this.projectId,
+    )
+    for (const block of missingBlocks) {
+      this.blockQueue.addToBack(block)
+    }
```
sz-piotr, line 57: Because we call `update()` every hour thanks to the clock & queue, every hour we would add the same blocks to the queue, since they are in the queue but haven't yet been saved to the database. There should be a variable that checks that blocks up to a certain point have been queued up

packages/backend/src/core/tx-count/BlockTxCountUpdater.ts
```diff
@@ -0,0 +1,72 @@
...
+    this.logger.info('Started')
+    return this.clock.onEveryHour(() => {
+      this.updateQueue.addIfEmpty()
+    })
+  }
+
+  async update() {
+    this.logger.info('Update started')
+
+    const missingBlocks = await this.txCountRepository.getMissingByProject(
```
sz-piotr, line 52: This approach seems to me like it would be problematic, potentially returning millions of blocks here. Maybe this could return block ranges?
Block ranges could be returned using two queries:
- find blocks for which the next block is not present
- find blocks for which the previous block is not present

sz-piotr (changes requested):

packages/backend/src/core/tx-count/BlockTxCountUpdater.ts
```diff
@@ -0,0 +1,85 @@
...
+import { TxCountRepository } from '../../peripherals/database/TxCountRepository'
+import { EthereumClient } from '../../peripherals/ethereum/EthereumClient'
+import { Clock } from '../Clock'
+
+export class BlockTxCountUpdater {
+  private updateQueue = new TaskQueue<void>(() => this.update(), this.logger)
+  private blockQueue = new TaskQueue(this.getBlock.bind(this), this.logger, {
+    workers: 100,
+  })
+  private queuedBlocks = new Set<number>()
```
krzkaczor, line 13: i think this implementation is acceptable for now but WDYT about adding this feature directly to TaskQueue? IMO it can be useful in other places too.

packages/backend/src/peripherals/database/TxCountRepository.ts
```diff
@@ -0,0 +1,139 @@
...
+
+  async getMissingRangeByProject(projectId: ProjectId) {
+    const knex = await this.knex()
+    const maxBlock = await this.findLatestByProject(projectId)
+    if (!maxBlock) {
+      return []
+    }
+
+    const noNextBlockQuery = knex
+      .select('t1.block_number')
```
krzkaczor, line 61: I really dont think we should rely on knex query builder for such complicated queries. Plain SQL would be much more readable here IMO. @sz-piotr thoughts?

packages/backend/test/peripherals/database/TxCountRepository.test.ts
```diff
@@ -0,0 +1,111 @@
...
+        PROJECT_A_LATEST,
+      )
+      expect(await repository.findLatestByProject(PROJECT_B)).toEqual(
+        PROJECT_B_LATEST,
+      )
+    })
+  })
+
+  describe(TxCountRepository.prototype.getMissingRangeByProject.name, () => {
+    it('finds holes', async () => {
```
krzkaczor, line 73: this needs more tests

packages/backend/src/peripherals/database/TxCountRepository.ts
```diff
@@ -0,0 +1,139 @@
...
+    const knex = await this.knex()
+    const row = await knex('tx_count')
+      .where('project_id', projectId.toString())
+      .orderBy('block_number', 'desc')
+      .first()
+
+    return row ? toRecord(row) : undefined
+  }
+
+  async getMissingRangeByProject(projectId: ProjectId) {
```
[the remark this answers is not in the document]
krzkaczor: also, please specify what a "range" is exactly here. In this case as far as i can tell it's exclusive.

packages/backend/src/peripherals/database/TxCountRepository.ts
sz-piotr: I agree, especially since we lose type safety anyway

packages/backend/test/peripherals/database/TxCountRepository.test.ts
michalsidzej: My bad, missed the case when block 0 is missing

packages/backend/src/peripherals/database/migrations/023_tx_count.ts
```diff
@@ -0,0 +1,27 @@
...
+
+If you find that something was incorrectly set up in the `up` function you
+should create a new migration file that fixes the issue.
+
+*/
+
+import { Knex } from 'knex'
+
+export async function up(knex: Knex) {
+  await knex.schema.createTable('tx_count', function (table) {
```
sz-piotr, line 17: This PR introduces `tx` as a shorthand for `tx` but in general in we use `transaction` everywhere. Because we dont use other shorthands too I think it's worthwile to maintain the longform. This applies to all files in this PR.

packages/backend/src/peripherals/database/migrations/023_tx_count.ts
```diff
@@ -0,0 +1,27 @@
...
+    table.string('project_id').notNullable()
+    table.dateTime('unix_timestamp').notNullable()
+    table.integer('count').notNullable()
+    table.integer('block_number')
```
sz-piotr, line 21: This table has no primary key and has no indices. I think that it would be good to have a shared primary key of `project_id` and `block_number`. Also, why is `block_number` not `notNullable()`?

packages/backend/src/core/tx-count/BlockTxCountUpdater.ts
```diff
@@ -0,0 +1,85 @@
+import { Logger, TaskQueue } from '@l2beat/common'
+import { ProjectId, UnixTime } from '@l2beat/types'
+
+import { TxCountRepository } from '../../peripherals/database/TxCountRepository'
+import { EthereumClient } from '../../peripherals/ethereum/EthereumClient'
+import { Clock } from '../Clock'
+
+export class BlockTxCountUpdater {
```
sz-piotr, line 8: I think that instead of this being called `BlockTxCountUpdater` it should be called `RpcTransactionUpdater` as most networks have blocks but not all of them use the standard ethereum rpc.

packages/backend/test/peripherals/database/TxCountRepository.test.ts
```diff
@@ -0,0 +1,111 @@
...
+      await repository.addMany([
+        fakeTxCount({ blockNumber: 0, projectId: PROJECT_A }),
+        fakeTxCount({ blockNumber: 1, projectId: PROJECT_A }),
+        fakeTxCount({ blockNumber: 3, projectId: PROJECT_A }),
+      ])
+
+      expect(await repository.getMissingRangeByProject(PROJECT_A)).toEqual([
+        [1, 3],
```
sz-piotr, line 81: This is super unintuitive to me. If there are no holes we get no output. But if there is no data we also get no output. Why not have this return the following instead:
`[[-Infinity, -1], [2, 2], [4, Infinity]]`

packages/backend/src/core/tx-count/BlockTxCountUpdater.ts
```diff
@@ -0,0 +1,85 @@
...
+      }
+    }
+
+    const lastBlock = await this.txCountRepository.findLatestByProject(
+      this.projectId,
+    )
+    const latestBlock = await this.ethereumClient.getBlockNumber()
+    let lastBlockNumber = lastBlock ? lastBlock.blockNumber : 0
+
+    while (lastBlockNumber < Number(latestBlock)) {
```
sz-piotr, line 71: If my earlier suggestion of returning ranges from -Infinity to +Infinity is to be taken into account then the logic in this file collapses to a single nested loop over ranges with Max(start, 0) and Min(end, lastBlock + 1)

sz-piotr (changes requested):

packages/backend/src/peripherals/database/migrations/023_tx_count.ts
michalsidzej: So the idea was primarily that some of the projects don't give us block_number (ex. all the StarkEx implementations)

packages/backend/src/core/tx-count/BlockTxCountUpdater.ts
```diff
@@ -0,0 +1,85 @@
...
+      for (let i = range[0] + 1; i < range[1]; i++) {
+        this.queueBlock(i)
+      }
+    }
+
+    const lastBlock = await this.txCountRepository.findLatestByProject(
+      this.projectId,
+    )
+    const latestBlock = await this.ethereumClient.getBlockNumber()
+    let lastBlockNumber = lastBlock ? lastBlock.blockNumber : 0
```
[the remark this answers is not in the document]
michalsidzej: That's a good question, WYT @krzkaczor? Block 0 will be full of artificial transactions that bring no useful info for us.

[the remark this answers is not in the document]
krzkaczor: I'd say, gather everything in the database. If it's gonna be really odd data, we will filter it out with a simple SQL query.

packages/backend/src/peripherals/database/TxCountRepository.ts
```diff
@@ -50,59 +55,51 @@ export class TxCountRepository extends BaseRepository {
...
+        WHERE b2.block_number IS NULL) AS no_next
+      ORDER BY block_number ASC
+    `)) as unknown as RawBlockNumberQueryResult
 
     // eslint-disable-next-line @typescript-eslint/no-unsafe-assignment
-    const ranges: [TxCountRow, TxCountRow][] = _.zip(noNext, noPrev)
+    const noPrev = (await knex.raw(`
+      WITH 
+          project_blocks AS (
+            SELECT * FROM tx_count WHERE project_id='${projectId.toString()}'
```
sz-piotr: Potential sql injection vulnerability. Use bindings instead of string manipulation.
https://knexjs.org/guide/raw.html#raw-parameter-binding

packages/backend/src/peripherals/database/TxCountRepository.ts
```diff
@@ -50,59 +55,51 @@ export class TxCountRepository extends BaseRepository {
...
-      .with(
-        'no_prev',
-        knex.raw('? EXCEPT ?', [noPrevBlockQuery, noNextBlockQuery]),
-      )
-      .from('no_prev')
+    // eslint-disable-next-line @typescript-eslint/no-unsafe-assignment
+    const noNext = (await knex.raw(`
+      WITH 
+        project_blocks AS (
+          SELECT * FROM tx_count WHERE project_id='${projectId.toString()}'
```
sz-piotr: bindings

packages/backend/src/peripherals/database/TxCountRepository.ts
```diff
@@ -50,59 +55,51 @@ export class TxCountRepository extends BaseRepository {
...
-      .from('no_next')
-
-    const noPrev = await knex
-      .select()
-      .with(
-        'no_prev',
-        knex.raw('? EXCEPT ?', [noPrevBlockQuery, noNextBlockQuery]),
-      )
-      .from('no_prev')
+    // eslint-disable-next-line @typescript-eslint/no-unsafe-assignment
```
sz-piotr: Is this still needed if you cast the expression?

packages/backend/test/core/tx-count/BlockTxCountUpdater.test.ts
```diff
@@ -20,7 +20,7 @@ describe(BlockTxCountUpdater.name, () => {
       })
       const txCountRepository = mock<TxCountRepository>({
         findLatestByProject: async () => fakeTxCount({ blockNumber: 4 }),
-        getMissingRangeByProject: async () => [[1, 3]],
+        getMissingRangesByProject: async () => [[1, 3]],
```
sz-piotr: This is a value that this function will never return

packages/backend/test/core/tx-count/BlockTxCountUpdater.test.ts
```diff
@@ -53,7 +53,7 @@ describe(BlockTxCountUpdater.name, () => {
...
```
sz-piotr: This is a value that this function will never return

packages/backend/test/core/tx-count/BlockTxCountUpdater.test.ts
michalsidzej: This is outdated, fixed in next commit

packages/backend/test/core/tx-count/BlockTxCountUpdater.test.ts
michalsidzej: Outdated

packages/backend/src/peripherals/database/migrations/023_tx_count.ts
michalsidzej: Thinking of this rn, we should probably have 2 tables, where 1 is only for raw block data and the second would hold aggregated data, ie for StarkEx

packages/backend/test/peripherals/database/TransactionCountRepository.test.ts
```diff
@@ -0,0 +1,117 @@
...
+  const PROJECT_B = ProjectId('project-b')
+
+  beforeEach(async () => {
+    await repository.deleteAll()
+  })
+
+  describe(
+    TransactionCountRepository.prototype.getMissingRangesByProject.name,
+    () => {
+      it('finds holes', async () => {
```
sz-piotr, line 25: There is still an edge case test missing for a completely empty repository. We want it to return [-Infinity, Infinity]

sz-piotr (approved):

packages/backend/src/core/tx-count/BlockTxCountUpdater.ts
michalsidzej: @krzkaczor  Should I add this as a task to backlog?

antooni (approved):

krzkaczor (approved):

PR #636, opened 2022-09-13
Add bridge tvl to /api/main

sz-piotr:
Resolves L2B-46

Additionally it changes the api to return unmodified timestamps and the frontend chart to always show the hour on hover.


packages/backend/src/core/reports/getReportConfigHash.ts
```diff
@@ -5,7 +5,7 @@ import { ReportProject } from './ReportProject'
 
 // Increment this value to change the hash which in turn causes the system to
 // recalculate reports
-const REPORT_LOGIC_VERSION = 0
+const REPORT_LOGIC_VERSION = 1
```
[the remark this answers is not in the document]
sz-piotr: I think that makes sense, but only to track the last reason for an update

packages/backend/src/config/config.production.ts
```diff
@@ -32,7 +32,7 @@ export function getProductionConfig(): Config {
       ...token,
       priceStrategy: { type: 'market' },
     })),
-    projects: layer2s.map(layer2ToProject),
+    projects: layer2s.map(layer2ToProject).concat(bridges.map(bridgeToProject)),
```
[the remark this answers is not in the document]
sz-piotr: I am quite unsure of how we'd be able to test the config. It's quite declarative. What would be some example tests you'd like to see?

packages/backend/src/api/controllers/report/generateMain.ts
```diff
@@ -11,23 +10,17 @@ export function generateMain(
...
   return {
-    charts: getProjectCharts(
-      hourlyReports,
-      sixHourlyReports,
-      dailyReports,
-      ProjectId.ALL,
-    ),
-    projects: projects.reduce<ApiMain['projects']>((acc, { projectId }) => {
+    charts: getProjectCharts(...reports, ProjectId.LAYER2S),
+    layers2s: getProjectCharts(...reports, ProjectId.LAYER2S),
```
sin6pi7, line 18: Are `charts` and `layer2s` keys supposed to have the same values?

packages/backend/src/api/controllers/report/generateMain.ts
```diff
@@ -11,23 +10,17 @@ export function generateMain(
   sixHourlyReports: AggregateReportRecord[],
   dailyReports: AggregateReportRecord[],
   latestReports: ReportRecord[],
-  projects: ReportProject[],
+  projectIds: ProjectId[],
 ): ApiMain {
+  const reports = [hourlyReports, sixHourlyReports, dailyReports] as const
```
sin6pi7, line 15: Maybe instead of destructuring this for every call of `getProjectCharts` we'd make `getProjectCharts` accept an object `reports = { hourly, sixHourly, daily }`?

packages/backend/src/api/controllers/report/generateMain.ts
sz-piotr: That's exactly right. In the near future the `charts` property is going to get removed, but it needs to stay here as long as the frontend depends on it.

PR #637, opened 2022-09-13
TaskQueue retries

sin6pi7:
Resolves L2B-79


packages/common/src/tools/queue/TaskQueue.ts
```diff
@@ -0,0 +1,97 @@
...
+  }
+
+  private handleFailure(job: Job<T>) {
+    const result = this.shouldRetry(job)
+    if (!result.retry) {
+      return
+    }
+    job.executeAfter = result.executeAfter
+    job.attempts++
+    this.queue.unshift(job)
```
michalsidzej, line 73: Maybe log retrying job?

michalsidzej: We're logging the error, so it would make sense to acknowledge the fact that it was taken care of imo

packages/common/test/tools/TaskQueue.test.ts
```diff
@@ -3,7 +3,7 @@ import { setTimeout } from 'timers/promises'
 import waitForExpect from 'wait-for-expect'
 
 import { Logger } from '../../src/tools/Logger'
-import { TaskQueue } from '../../src/tools/TaskQueue'
+import { TaskQueue } from '../../src/tools/queue/TaskQueue'
```
sz-piotr, line 6: Also move the test file into a subfolder

packages/common/src/tools/queue/TaskQueue.ts
```diff
@@ -0,0 +1,97 @@
...
+    private executeTask: (task: T) => Promise<void>,
+    private logger: Logger,
+    opts?: TaskQueueOpts<T>,
+  ) {
+    this.workers = opts?.workers ?? 1
+    assert(
+      this.workers > 0 && Number.isInteger(this.workers),
+      'workers needs to be a positive integer',
+    )
+    this.shouldRetry = opts?.shouldRetry ?? Retries.always
```
sz-piotr, line 28: I think having exponential backoff as a default retry strategy would be good. Alternatively specifying this parameter shouldn't be optional so that the choice has to be made every time

packages/common/src/tools/queue/Retries.ts
```diff
@@ -0,0 +1,29 @@
+import { ShouldRetry } from './types'
+
+function always() {
+  return { retry: true }
+}
+
+function exponentialBackOff<T>(stepMs: number): ShouldRetry<T> {
```
sz-piotr, line 7: Maybe an optional maximum would be nice here.

packages/common/src/tools/queue/TaskQueue.ts
```diff
@@ -0,0 +1,97 @@
...
+    this.queue.unshift(job)
+  }
+
+  private async executeUnchecked() {
+    if (this.allWorkersBusy || this.queue.length === 0) {
+      return
+    }
+    const jobIndex = this.queue.findIndex((job) => this.shouldExecute(job))
+    if (jobIndex === -1) {
+      setTimeout(() => this.execute())
```
sz-piotr, line 82: This is spin waiting. It should probably find the job with the smallest value and timeout to that value.

packages/common/src/tools/queue/Retries.ts
sin6pi7: Do you mean maximum waiting time?

packages/common/src/tools/queue/TaskQueue.ts
```diff
@@ -0,0 +1,113 @@
...
+    if (!result.retry) {
+      this.logger.debug('Task should not be retried - removing from queue', {
+        task: String(job.task),
+        attempts: job.attempts,
+      })
+      return
+    }
+    job.executeAfter = result.executeAfter ?? Date.now()
+    this.logger.debug('Retrying task', {
+      task: String(job.task),
```
sz-piotr, line 80: I would prefer if the tasks had an explicit textual description. This solution will most likely print `[Object object]` or some other nonsense

packages/common/src/tools/queue/TaskQueue.ts
```diff
@@ -0,0 +1,113 @@
...
+    opts?: TaskQueueOpts<T>,
+  ) {
+    this.workers = opts?.workers ?? 1
+    assert(
+      this.workers > 0 && Number.isInteger(this.workers),
+      'workers needs to be a positive integer',
+    )
+    this.shouldRetry =
+      opts?.shouldRetry ??
+      Retries.exponentialBackOff(100, { maxAttempts: 10, maxDistance: 10_000 })
```
sz-piotr, line 30: If we include maxAttempts as a default the current system would break because there is an assumption that tasks will be finished eventually. Also 10s is quite a short time, maybe we could have 10 minutes instead?

sin6pi7: That's why I set `always` as default in the first place. I think for the purpose of delivery I'd keep `always` as default and have it thought through as a separate PR. Wdyt?

sz-piotr: I agree. Let's have always for now

packages/common/src/tools/queue/TaskQueue.ts
sz-piotr: Unfortunately there is no guarantee that `JSON.stringify` doesn't throw. For example it will throw if the task includes a bignumber. This is why we have the `json` type. If you'd like to have a default description you could have the queue keep track of an incrementing id and give a default description of `Task #${++id}`. And the signature would be `addToBack(task: T, description?: string)`

packages/common/src/tools/queue/Retries.ts
```diff
@@ -0,0 +1,47 @@
...
+  return ({ attempts }: { attempts: number }) => {
+    if (attempts === maxAttempts) {
+      return {
+        retry: false,
+      }
+    }
+    const distance = Math.pow(2, attempts) * stepMs
+    return {
+      retry: true,
+      executeAfter: now() + Math.min(distance, maxDistance),
```
sz-piotr, line 30: Maybe the code would be simpler to test if we returned only the `Math.min(distance, maxDistance)` value and added `now()` inside the TaskQueue? Then exponentialBackOff becomes pure.

sz-piotr (approved):

packages/common/src/tools/queue/Retries.ts
sin6pi7: I like it, will do that 👍

sz-piotr (approved):

PR #639, opened 2022-09-14
Redo events

krzkaczor:
* stores raw events in db
* associates events with blocks from the top of the hour
* removes code that needs to be tweaked (for example controllers)

For the record: the whole thing would be much simpler if we would store all block info in the database. Also it would be much more flexible in the future.

A plan is to: 
1. Merge this
2. Sync everything on the backend
3. Test various queries on the remote database and find a way to make everything performant. 
4. Redo extracting aggregated data from the database.


packages/backend/src/peripherals/ethereum/EthereumClient.ts
```diff
@@ -36,7 +36,15 @@ export class EthereumClient {
...
   }
 
-  async getLogsUsingBisection(
+  /**
+   * Handles large block ranges by split them to smaller ones when necessary
+   * @param address
+   * @param topic
+   * @param fromBlock
+   * @param toBlock
+   * @returns
```
sz-piotr, line 45: If you're not filling those out then it's pointless to include them

packages/backend/src/peripherals/database/migrations/023_redo_events.ts
```diff
@@ -0,0 +1,28 @@
...
+  await knex.schema.createTable('events', function (table) {
+    table.increments('id')
+    table.dateTime('unix_timestamp').notNullable()
+    table.string('event_name').notNullable()
+    table.string('project_id').notNullable()
+  })
+}
+
+export async function down(knex: Knex) {
+  await knex.schema.dropTable('events')
```
sz-piotr, line 27: This is incorrect, because it doesn't recreate the old events table

packages/backend/src/peripherals/database/EventRepository.ts
```diff
@@ -69,35 +63,32 @@ export class EventRepository extends BaseRepository {
...
+
+  //   const [{ sum }] = await knex('events')
+  //     .where('project_id', projectId.toString())
+  //     .where('event_name', name)
+  //     .where('unix_timestamp', '>=', from.toString())
+  //     .where('unix_timestamp', '<=', to.toString())
+  //     .sum('count')
+  //   //todo rethink if we want to sum in db
+  //   return Number(sum)
+  // }
```
sz-piotr, line 91: Commented code should not make it onto the master branch

packages/backend/src/core/events/findCorrespondingBlocks.ts
```diff
@@ -0,0 +1,48 @@
...
+import { sortBy } from 'lodash'
+
+interface BlockInfo {
+  timestamp: UnixTime
+  blockNumber: bigint
+}
+
+export function findCorrespondingBlocks<T extends { blockNumber: number }>(
+  _blocks: BlockInfo[],
+  _logs: T[],
```
sz-piotr, line 11: Why are argument variables prefixed with `_`? We don't use this convention anywhere else

antooni (approved):

packages/backend/src/core/events/EventUpdater.ts
```diff
@@ -81,73 +79,37 @@ export class EventUpdater {
...
+      from,
+      to,
+    )
+    const logsWithCorrespondingBlocks = findCorrespondingBlocks(blocks, logs)
+
+    const records: EventRecord[] = logsWithCorrespondingBlocks.map((l) => ({
+      name: event.name,
+      projectId: event.projectId,
+      timestamp: l.block.timestamp,
+    }))
```
sz-piotr, line 111: If we are only using the timestamps here then `findCorrespondingBlocks` should be renamed to `findTimestamps(blockNumbers: bigint[]): UnixTime[]`

packages/backend/src/peripherals/ethereum/EthereumClient.ts
```diff
@@ -36,7 +36,10 @@ export class EthereumClient {
     return Bytes.fromHex(bytes)
   }
 
-  async getLogsUsingBisection(
+  /**
+   * Handles large block ranges by split them to smaller ones when necessary
```
sz-piotr, line 40:
```suggestion
   * Handles large block ranges by splitting them into smaller ones when necessary
```

packages/backend/test/peripherals/database/EventRepository.test.ts
```diff
@@ -117,51 +92,46 @@ describe(EventRepository.name, () => {
...
+  //     PROJECT_A,
+  //     EVENT_A,
+  //     START,
+  //     START.add(3, 'hours'),
+  //   )
 
-    expect(result).toEqual(4)
-  })
+  //   expect(result).toEqual(4)
+  // })
```
sz-piotr, line 127: Again, commented out code shouldn't be included

sz-piotr (changes requested):

packages/backend/src/core/events/EventUpdater.ts
krzkaczor: I can't agree. I actually spent some time refining the interface of this function.

In its current form, it's generic. It can be used not only with events (logs) but with anything really: contract deployments, or some other happening. Furthermore, it can be used to extract other things, not only timestamps from blocks. I think this way interface of the function is much clearer.

sz-piotr (approved):

PR #640, opened 2022-09-14
Add bridge tvl page

sz-piotr:
Resolves L2B-35


packages/frontend/src/utils/orderByTvl.ts
```diff
@@ -0,0 +1,10 @@
+import { ApiMain, ProjectId } from '@l2beat/types'
+
+export function orderByTvl<T extends { id: ProjectId }>(
```
krzkaczor, line 3: tests for this?

krzkaczor (approved):

PR #641, opened 2022-09-14
Timer abstraction

sin6pi7:


packages/common/src/tools/timer/types.ts
```diff
@@ -0,0 +1,5 @@
+export interface Timer {
+  now: () => number
+  setTimeout: (callback: () => void, timeMs: number) => number | NodeJS.Timeout
+  clearTimeout: (timeout: number) => void
```
sz-piotr, line 4:
```suggestion
  setTimeout: (callback: () => void, timeMs: number) => ReturnType<typeof setTimeout>
  clearTimeout: (timeout: ReturnType<typeof setTimeout>) => void
```

packages/common/src/tools/timer/NativeTimer.ts
```diff
@@ -0,0 +1,7 @@
+import { Timer } from './types'
+
+export const NativeTimer: Timer = {
```
sz-piotr, line 3: Maybe this should be camelCase since this isn't a class?
```suggestion
export const nativeTimer: Timer = {
```

sz-piotr (changes requested):

packages/common/src/tools/timer/types.ts
sin6pi7: Won't this require us to return NodeJS.Timeout from InMemoryTimer?

antooni:
Maybe when adding such "infra" update it would be a good idea to provide example where it will be used, ideally inside the test. Then the need for the update could be seen by a potential reviewer

sin6pi7:
Closing, because we are already using https://sinonjs.org/releases/latest/fake-timers/ in tests - no point in adding our own abstraction then.

PR #645, opened 2022-09-15
Transaction counting for all L2Rpcs

michalsidzej:
Resolves L2B-76


packages/backend/src/createL2Clients.ts
```diff
@@ -0,0 +1,38 @@
+import { Layer2Rpc } from '@l2beat/config'
```
michalsidzej, line 1: Idea needed where should we put this file

packages/config/src/layer2s/types/Layer2.ts
```diff
@@ -19,4 +20,6 @@ export interface Layer2 {
   details: Layer2Details
   /** Metadata about events emitted by the system */
   events: Layer2Event[]
+  /** RPC or API url used to get transaction count */
+  url?: Layer2Url
```
krzkaczor, line 24: nit: maybe apiUrl instead?

krzkaczor (approved):

packages/config/src/layer2s/types/Layer2Url.ts
```diff
@@ -0,0 +1,16 @@
...
+}
+
+interface JsonRpcProvider {
+  provider: 'jsonRpc'
+  url: string
+}
+
+export type Layer2Rpc = (AlchemyProvider | JsonRpcProvider) & {
+  type: 'rpc'
+  callsPerMinute?: number
```
sz-piotr, line 13: I would prefer those to be repeated inside AlchemyProvider and JsonRpcProvider. Then the type Layer2Rpc could be dissolved. I think it would be clearer if the Layer2Url was a union of interfaces without any type algebra

packages/config/src/layer2s/types/Layer2Url.ts
```diff
@@ -0,0 +1,16 @@
+interface AlchemyProvider {
+  provider: 'alchemy'
+  slug: string
```
sz-piotr, line 3:
```suggestion
  networkName: string
```

packages/config/src/layer2s/types/Layer2Url.ts
```diff
@@ -0,0 +1,16 @@
...
+  provider: 'jsonRpc'
+  url: string
+}
+
+export type Layer2Rpc = (AlchemyProvider | JsonRpcProvider) & {
+  type: 'rpc'
+  callsPerMinute?: number
+}
+
+export type Layer2Url = Layer2Rpc
```
sz-piotr, line 16: How about a different name, more suggestive of the actual use, like: `Layer2TransactionApi`

packages/config/src/layer2s/types/Layer2.ts
sz-piotr:
```suggestion
  transactionApi?: Layer2TransactionApi
```

packages/backend/src/core/tx-count/RpcTransactionUpdater.ts
```diff
@@ -0,0 +1,78 @@
+import { Logger, TaskQueue } from '@l2beat/common'
```
sz-piotr, line 1: This class alongside its test counterpart is already on the master branch but in a different directory. Remove them from the PR

packages/backend/src/createL2Clients.ts
```diff
@@ -0,0 +1,38 @@
...
+import { providers } from 'ethers'
+
+import { Config } from './config'
+import { EthereumClient } from './peripherals/ethereum/EthereumClient'
+
+export function createL2Clients(config: Config) {
```
sz-piotr, line 7: I think I would structure this a bit differently. We create the clients specifically so that we can get their transactions by creating and starting RpcTransactionUpdater. If we are creating the clients in a separate step then when we introduce new types of transaction updaters we'd need to somehow also differentiate the clients making the code quite a bit more complicated.

I suggest the following:
1. Create `/src/setup` that will be the go to directory for extracting logic from `Application` and move this file there
2. Instead of returning EthereumClients create the clients and then immediately create RpcTransactionUpdater. That way when new types of configuration show up you just handle them differently in the if statement
3. Instead of mapping and filtering just create the updaters array and conditionally push to it (this is just a style preference, but I think the code would be more readable)
4. Then Application just receives an array of updates instead of having to create it by itself

sz-piotr (changes requested):

sz-piotr (approved):

PR #649, opened 2022-09-16
Restore EventsController

antooni:
For developement purposes EventsController and EventsRouter was temporary deleted.
This PR will restore them with updated usage of methods from EventsRepository

Resolves L2B-96


packages/backend/src/peripherals/database/EventRepository.ts
```diff
@@ -72,6 +76,38 @@ export class EventRepository extends BaseRepository {
     return rows.map(toRecord)
   }
 
+  async getAggregatedByProjectAndGranularity(
+    projectId: ProjectId,
+    granularity: 'hour' | 'day',
+    from: UnixTime = new UnixTime(0),
+  ): Promise<EventRecordAggregated[]> {
+    const knex = await this.knex()
+    const rows = (await knex('events')
```
krzkaczor, line 85: @sz-piotr @michalsidzej how do we want to approach such, more complex, queries in knex? (i know we have similar queries in tx counting feature). 

I would suggest going with raw SQLs and just prevent SQL injections with some kind of knex helper. I don't see any value in dealing with the complexities of knex API.

sz-piotr: I agree. If the PR is complex enough so that we lose static typing from knex then it probably should be just raw SQL

packages/backend/src/peripherals/database/EventRepository.ts
```diff
@@ -72,6 +76,38 @@ export class EventRepository extends BaseRepository {
...
+      .where('project_id', projectId.toString())
+      .where('unix_timestamp', '>=', from.toDate())
+      .select(
+        knex.raw(
+          `date_trunc('${granularity}', unix_timestamp) AS unix_timestamp`,
```
sz-piotr, line 67: Strings should never be concatenated into SQL because this leaves a possibility for SQL injection vulnerabilities. Use bindings instead

krzkaczor: Yes. Also, feel free to migrate to simply using .raw as @michalsidzej did: https://github.com/l2beat/l2beat/pull/634/files#diff-4fe439c9479ba1dbaf66a5452361b2095a713399847cf7f72e0457f3da6127d9R62

packages/backend/src/peripherals/database/EventRepository.ts
```diff
@@ -59,30 +68,23 @@ export class EventRepository extends BaseRepository {
...
-    }[]
-    return rows.map(toRecordAggregated)
+
+    // eslint-disable-next-line @typescript-eslint/no-unnecessary-condition
+    if (granularity !== 'hour' && granularity !== 'day') {
+      throw new Error('Programmer error or SQL injection')
+    }
+
+    const query =
+      granularity === 'hour'
```
krzkaczor, line 78: why not just pass granuality directly?

antooni: @krzkaczor there was an error with encoding, I fixed it by changing the column name, you can take a look now

packages/backend/test/api/controllers/events/getEventChart.test.ts
```diff
@@ -0,0 +1,77 @@
...
+    expect(result).toEqual({
+      types: ['timestamp', EVENT_A, EVENT_B],
+      data: [
+        [START, 100, 200],
+        [START.add(1, 'hours'), 100, 200],
+        [START.add(2, 'hours'), 100, 200],
+      ],
+    })
+  })
+  it('skips events not provided in eventNames', () => {
```
sz-piotr, line 55: newline :)

packages/backend/test/api/controllers/events/EventsController.test.ts
```diff
@@ -0,0 +1,126 @@
...
+      const result = await eventController.getEvents()
+
+      expect(result).toEqual({
+        projects: {
+          [PROJECT_A.toString()]: {
+            hourly: {
+              types: ['timestamp', EVENT_A],
+              data: [[START, 3]],
+            },
+            sixHourly: { types: ['timestamp', EVENT_A], data: [] },
```
sz-piotr, line 69: Maybe add some values that would show up in here too

packages/backend/src/peripherals/database/EventRepository.ts
```diff
@@ -11,17 +11,32 @@ export interface EventRecord {
   projectId: ProjectId
 }
 
+export interface EventRecordAggregated extends EventRecord {
```
sz-piotr, line 14: AggregatedEventsRecord?

sz-piotr (changes requested):

sz-piotr (approved):

PR #652, opened 2022-09-20
Add test for migration numbering

sz-piotr:


michalsidzej (approved):
I'm honestly amazed how many stupid mistakes can be avoided using simple unit tests

antooni (approved):

packages/backend/test/peripherals/database/migrations.test.ts
```diff
@@ -0,0 +1,19 @@
+import { expect } from 'earljs'
+import { readdirSync } from 'fs'
+import path from 'path'
+
+describe('migrations', () => {
+  it('migrations have consecutive numbering except for 20', () => {
```
krzkaczor, line 6: don't we have already some other tests for migrations?

krzkaczor: Yes we do: `packages/backend/test/peripherals/database/shared/Database.test.ts`. Maybe it makes sense to merge these test files?

krzkaczor (approved):

PR #653, opened 2022-09-20
Remove unused RequestTracker

sz-piotr:


krzkaczor (commented):
What was the idea behind this feature? And why do we remove it now?

sz-piotr:
> What was the idea behind this feature? And why do we remove it now?

The original idea was to have a /status endpoint on the backend. It would show all running information about all the system components in one place. Each service was supposed to have a `getStatus` method. The RequestTracker was supposed to collect information about how much time is spent on each request to include in the status update.

This PR is not removing this feature. The `getStatus` method is never called, and stopped being called months ago with some refactoring. This PR is effectively just removing the dead code.

krzkaczor (approved):

PR #657, opened 2022-09-21
Unique task queue

michalsidzej:
Adds a wrapper for TaskQueue class that only adds a unique task to the queue.

### DIFF
--- a/packages/common/src/tools/queue/UniqueTaskQueue.ts
+++ b/packages/common/src/tools/queue/UniqueTaskQueue.ts
@@ -1,5 +1,6 @@
 import { Logger } from '../Logger'
-import { TaskQueue, TaskQueueOpts } from './TaskQueue'
+import { TaskQueue } from './TaskQueue'
+import { TaskQueueOpts } from './types'
 
 export class UniqueTaskQueue<T extends string | number> extends TaskQueue<T> {
   private taskSet: Set<T> = new Set()
--- a/packages/common/src/tools/queue/UniqueTaskQueue.ts
+++ b/packages/common/src/tools/queue/UniqueTaskQueue.ts
@@ -18,15 +18,15 @@ export class UniqueTaskQueue<T extends string | number> extends TaskQueue<T> {
     super(wrappedExecute, logger, opts)
   }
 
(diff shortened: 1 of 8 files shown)


packages/common/src/tools/queue/UniqueTaskQueue.ts
```diff
@@ -0,0 +1,39 @@
...
+  ) {
+    const wrappedExecute = async (task: T) => {
+      await executeTask(task)
+      this.taskSet.delete(task)
+    }
+
+    super(wrappedExecute, logger, opts)
+  }
+
+  addIfEmpty(task: T): void {
```
krzkaczor, line 21: Shouldnt this use `override`? 

also maybe we can turn on the compiler flag to enforce its use: https://www.typescriptlang.org/docs/handbook/release-notes/typescript-4-3.html#override-and-the---noimplicitoverride-flag

krzkaczor (changes requested):

packages/common/src/tools/queue/UniqueTaskQueue.ts
michalsidzej: Good point, updated TS rule

packages/common/src/tools/queue/UniqueTaskQueue.ts
```diff
@@ -0,0 +1,39 @@
...
+  override addToFront(task: T): void {
+    this.addUnique(task, super.addToFront.bind(this))
+  }
+
+  private addUnique(task: T, method: (task: T) => void) {
+    if (!this.taskSet.has(task)) {
+      this.taskSet.add(task)
+      method(task)
+    }
+  }
```
sz-piotr, line 38: Since this is already merged I am only leaving this comment as a consideration for the future. Instead of rebinding the method every time one could just make the if impure

```ts
  override addToFront(task: T): void {
    if (this.shouldAdd(task)) {
      super.addToFront(task)
    }
  }

  private shouldAdd(task: T) {
    if (!this.taskSet.has(task)) {
      this.taskSet.add(task)
      return true
    }
    return false
  }
```

PR #659, opened 2022-09-21
Add scaling prefix to directories

sz-piotr:
The pages for bridges are already prefixed with `bridges-` so to keep consistency the scaling pages are now prefixed with `scaling-`


antooni (approved):

antooni:
maybe in renaming PRs we should provide some context why it was renamed? in the description

sz-piotr:
> maybe in renaming PRs we should provide some context why it was renamed? in the description

Good point. Added a short explanation

PR #660, opened 2022-09-21
Add staging config support

krzkaczor:
The only change between this and production config is that I turned on `eventsSyncEnabled`. 

Once merged I will tweak env variables on staging environment so this config is used.

### DIFF
--- a/packages/backend/src/config/config.staging.ts
+++ b/packages/backend/src/config/config.staging.ts
@@ -1,0 +1,41 @@
+import { LogLevel } from '@l2beat/common'
+import { bridges, layer2s, tokenList } from '@l2beat/config'
+import { UnixTime } from '@l2beat/types'
+
+import { bridgeToProject, layer2ToProject } from '../model'
+import { Config } from './Config'
+import { getEnv } from './getEnv'
+
+export function getStagingConfig(): Config {
+  const name = 'Backend/Staging'
+  return {
+    name,
+    logger: {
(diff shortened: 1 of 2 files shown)


packages/backend/src/config/config.staging.ts
```diff
@@ -0,0 +1,41 @@
...
+import { UnixTime } from '@l2beat/types'
+
+import { bridgeToProject, layer2ToProject } from '../model'
+import { Config } from './Config'
+import { getEnv } from './getEnv'
+
+export function getStagingConfig(): Config {
+  const name = 'Backend/Staging'
+  return {
+    name,
```
sz-piotr, line 12: what do you think about spreading the `...getProductionConfig()` here instead of all the configuration being repeated?

sz-piotr (approved):

sz-piotr (approved):

packages/backend/src/config/config.staging.ts
```diff
@@ -0,0 +1,17 @@
...
+
+export function getStagingConfig(): Config {
+  const name = 'Backend/Staging'
+  const prodConfig = getProductionConfig()
+  return {
+    ...prodConfig,
+
+    name,
+    databaseConnection: {
+      ...prodConfig,
```
sz-piotr, line 12:
```suggestion
      ...prodConfig.databaseConnection,
```

sz-piotr (approved):

PR #661, opened 2022-09-21
Starkex Transaction Updater

michalsidzej:
Resolves L2B-107


packages/backend/src/peripherals/starkex/StarkexClient.ts
```diff
@@ -0,0 +1,76 @@
...
+    private starkexApiUrl: string,
+    private starkexApiKey: string,
+    private httpClient: HttpClient,
+    private logger: Logger,
+  ) {
+    this.logger = this.logger.for(this)
+  }
+  // Rate limit?
+
+  // TODO: create BaseClient for this and EtherscanClient to avoid duplication
```
sz-piotr, line 18: Do they have this much in common?

packages/backend/test/peripherals/starkex/StarkexClient.test.ts
```diff
@@ -0,0 +1,101 @@
...
+        API_URL,
+        API_KEY,
+        httpClient,
+        Logger.SILENT,
+      )
+
+      await starkexClient.getDailyCount(day, product)
+    })
+
+    it('returns the block number', async () => {
```
sz-piotr, line 84: Is this the correct test name?

sz-piotr (changes requested):

packages/backend/src/peripherals/starkex/StarkexClient.ts
michalsidzej: I was mostly thinking of ErrorHandling, maybe it could be done in HttpClient?

packages/backend/test/peripherals/starkex/StarkexClient.test.ts
michalsidzej: Nope, fixed

sz-piotr (approved):

PR #662, opened 2022-09-21
Share project page components

sz-piotr:
Resolves L2B-116

This PR continues efforts to build foundations for the bridges project page. Moving the components into a shared location allows them to be reused.

I have also added storybook stories for each of the components so that they can be easily iterated on independently in the future.


michalsidzej (approved):
LGTM

PR #664, opened 2022-09-22
Add pages for bridge projects

sz-piotr:
Resolves L2B-94

This PR adds stub pages for all bridge projects. They generally lack content and have only name, description some links and a chart. Future PRs will expand on this.


packages/frontend/src/pages/config.ts
```diff
@@ -9,5 +9,5 @@ export const config = {
   showBanner: Date.now() < new Date('2022-09-23').getTime(),
   showGitcoinOption: true,
   showPermissionedAccounts: true,
-  showBridgePages: false,
+  showBridgePages: true,
```
krzkaczor, line 12: does it mean that we show in on prod?

krzkaczor (approved):

PR #673, opened 2022-09-26
Set transactionCountSyncEnabled from env variable

michalsidzej:


packages/backend/src/config/config.production.ts
```diff
@@ -38,6 +38,9 @@ export function getProductionConfig(): Config {
     syncEnabled: !getEnv.boolean('SYNC_DISABLED', false),
     freshStart: false,
     eventsSyncEnabled: false,
-    transactionCountSyncEnabled: false,
+    transactionCountSyncEnabled: getEnv.boolean(
```
krzkaczor, line 41: We have a stage config now too. What if we just hardcode it to true on stage?

krzkaczor (changes requested):

packages/backend/src/config/config.production.ts
```diff
@@ -38,6 +38,9 @@ export function getProductionConfig(): Config {
...
+      'TRANSACTION_COUNT_ENABLED',
+      false,
+    ),
```
sz-piotr, line 44: I don't think we should use environment variables to trigger feature flags on production. It is fine for local/staging and others, but having production be as explicit as possible would be nice.

krzkaczor (approved):

PR #674, opened 2022-09-26
Add ApiActivity

sin6pi7:
The idea is to prepare a contract between frontend and api for activity feature.

Resolves L2B-137

### DIFF
--- a/packages/types/src/ApiActivity.ts
+++ b/packages/types/src/ApiActivity.ts
@@ -15,15 +15,8 @@ const Chart = z.object({
 })
 export type Chart = z.infer<typeof Chart>
 
-export const Charts = z.object({
-  hourly: Chart,
-  sixHourly: Chart,
-  daily: Chart,
-})
-export type Charts = z.infer<typeof Charts>
-
 export const ApiActivity = z.object({
-  combined: Charts,
-  projects: z.record(z.string(), Charts),
+  combined: Chart,
+  projects: z.record(z.string(), Chart),
 })
 export type ApiActivity = z.infer<typeof ApiActivity>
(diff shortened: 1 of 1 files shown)


packages/types/src/ApiActivity.ts
```diff
@@ -0,0 +1,29 @@
...
+
+export const Charts = z.object({
+  hourly: Chart,
+  sixHourly: Chart,
+  daily: Chart,
+})
+export type Charts = z.infer<typeof Charts>
+
+export const ApiActivity = z.object({
+  combined: Charts,
```
sin6pi7, line 26: Do we actually need combined data here? If I understand it right it's easily derived from project charts. If that's the case, then by removing combined data we could speed up the api http response and lower its memory footprint. Wdyt? cc @sz-piotr @krzkaczor

sz-piotr: I would keep the combined data here, because the fact that we're calculating it in the app layer is an implementation detail, we might have as well had it calculated in a query, it just isn't what we decided on

sin6pi7: That's a good point. At the end of the day it's perf vs dx trade-off - if we find it too slow/memory-intensive nothing stops us from moving calculation to the frontend - for now let's keep it here.

packages/types/src/ApiActivity.ts
```diff
@@ -0,0 +1,29 @@
...
+export type ChartPoint = z.infer<typeof ChartPoint>
+
+const Chart = z.object({
+  types: z.tuple([z.literal('timestamp'), z.string()]),
+  data: z.array(ChartPoint),
+})
+export type Chart = z.infer<typeof Chart>
+
+export const Charts = z.object({
+  hourly: Chart,
```
krzkaczor, line 19: i think we dont support H and 6H granualities for this feature

sz-piotr (approved):

PR #677, opened 2022-09-26
Improve transaction count logging

michalsidzej:


packages/config/src/layer2s/optimism.ts
```diff
@@ -84,6 +84,11 @@ export const optimism: Layer2 = {
         sinceTimestamp: new UnixTime(1636654763),
       },
     ],
+    transactionApi: {
+      type: 'rpc',
+      provider: 'alchemy',
+      networkName: 'optimism',
+    },
```
sz-piotr, line 91: I would prefer the name of this PR be after this change. The logging is just a small improvement. Or this could be moved to another PR.

sz-piotr (approved):

packages/config/src/layer2s/optimism.ts
michalsidzej: Moving this to a separate PR

PR #678, opened 2022-09-26
Add more content to bridge tvl page

sz-piotr:
This PR adds more columns to the previously bare bridge tvl page

Screenshot 2022-09-26 at 14 22 08


packages/frontend/src/pages/bridges-tvl/index.tsx
```diff
@@ -4,7 +4,7 @@ import React from 'react'
 
 import { PageWrapper } from '../../components'
 import { BridgesTvlPage } from './BridgesTvlPage'
-import { getProps } from './getProps'
+import { getProps } from './props'
 
 export function getBridgesTvlPage(bridges: Bridge[], apiMain: ApiMain) {
   const { props, wrapper } = getProps(bridges, apiMain)
```
antooni, line 10: in `line 12` there is a typo, should be `tvl`, I do not know hot to add comment to it ;p

antooni (changes requested):
one typo should be fixed because otherwise the navbar tabs dont work

sin6pi7 (approved):

PR #679, opened 2022-09-26
Limit blockQueue

michalsidzej:
Up to discussion: a simple limiter of blockQueue in RpcTransactionUpdater. 200_000 requests per project per hour will be enough not to cause JS heap out of memory. Proposing this to start syncing rn, working on a better solution already.


sz-piotr (approved):
Approving this but only as a temporary solution

packages/backend/src/core/transaction-count/RpcTransactionUpdater.ts
```diff
@@ -58,12 +62,15 @@ export class RpcTransactionUpdater {
...
     const latestBlock = await this.ethereumClient.getBlockNumber()
 
-    for (const [start, end] of missingRanges) {
+    enqueueBlockLoop: for (const [start, end] of missingRanges) {
       for (
         let i = Math.max(start, 0);
         i < Math.min(end, Number(latestBlock) + 1);
         i++
       ) {
+        if (this.blockQueue.length >= BLOCK_QUEUE_LIMIT) {
```
antooni, line 71: what is the strategy for handling this error, will it be retried later? it is not clear to me at first sight

michalsidzej: This just limits the number of tasks in the queue. When it's more, it can be reasonable to think that the queue will not be completed in an hour, and the next hour the `update` function will be called anyway, populating the queue with new tasks.

PR #684, opened 2022-09-26
Vercel builds

krzkaczor:


vercel.json
```diff
@@ -0,0 +1,9 @@
+{
+  "redirects": [
+    {
+      "source": "/",
+      "destination": "scaling/tvl",
+      "permanent": true
+    }
```
sin6pi7, line 7: What about other redirects?

krzkaczor: I think all other redirects were configured for legacy reasons.

sz-piotr: Not the api one

sin6pi7 (approved):

sz-piotr (approved):

PR #685, opened 2022-09-26
Do not use a global frontend config

sz-piotr:
The current way frontend is written is using a lot of global configuration. One example would be feature flags, another might be the implicit use of tokenList when reading token properties. The way the configuration is being used is also quite convoluted.

**Current config use**
```
        @l2beat/config*       _______ config.ts _______
               |             /            |            \
               V             V            V            V
/api/main ---> . ---> renderPages ---> getProps --> <Component />
                \---> createApi

* - also used futher down the line
````

This PR addresses the first concern. It replaces a global config with one that is passed through function arguments. This gives us flexibility to test and experiment more easily, especially in the context of unit tests or storybook.

It also makes the data flow much more streamlined

**New config use**
```
@l2beat/config* ---> getConfig() ---> /api/main --->

---> . ---> renderPages ---> getProps --> <Component />
      \---> createApi

* - also used futher down the line (but less)
```


sin6pi7 (approved):
👍

PR #686, opened 2022-09-26
Zksync transaction count

michalsidzej:
Resolves L2B-61


packages/backend/src/core/transaction-count/ZksyncTransactionUpdater.ts
```diff
@@ -0,0 +1,78 @@
...
+    for (const [start, end] of missingRanges) {
+      for (
+        let i = Math.max(start, 1);
+        i < Math.min(end, Number(latestBlock) + 1);
+        i++
+      ) {
+        // Skip block 427, as it has a duplicated blockIndex
+        if (i === 427) {
+          continue
+        }
```
sz-piotr, line 71: I dislike this solution. It results with the updater always returning 427 in the query. I propose special handling in updateBlock.

packages/backend/src/core/transaction-count/ZksyncTransactionUpdater.ts
```diff
@@ -0,0 +1,78 @@
...
+  async updateBlock(number: number) {
+    const transactions = await this.zksyncClient.getTransactionsInBlock(number)
+
+    const records: ZksyncTransactionRecord[] = transactions.map(
+      (transaction) => ({
+        blockIndex: transaction.blockIndex,
+        blockNumber: number,
+        timestamp: transaction.createdAt,
+      }),
+    )
```
sz-piotr, line 46: If a block had 0 transactions it would be missing from the database and therefore always re-downloaded. Are there any such blocks?

packages/backend/src/peripherals/zksync/ZksyncClient.ts
```diff
@@ -0,0 +1,112 @@
...
+      const lastTx = transactions.at(-1)
+      if (!lastTx) {
+        throw new Error('Programmer error: Transactions list empty!')
+      }
+
+      const nextPage = await this.call(`blocks/${blockNumber}/transactions`, {
+        from: lastTx.txHash,
+        limit: '100',
+        direction: 'older',
+      })
```
sz-piotr, line 47: Should we add a check that it is indeed that transaction that got downloaded?

packages/backend/src/peripherals/zksync/ZksyncClient.ts
```diff
@@ -0,0 +1,112 @@
...
+
+    const parsed = ZksyncTransactionResultSchema.parse(result)
+
+    if (parsed.list.length === 0) {
+      throw new Error('Transactions list empty!')
+    }
+    let transactions = parsed.list
+
+    let count = parsed.pagination.count
+    while (count > 100) {
```
sz-piotr, line 37: I would actually prefer subtracting the number of txs fetched and comparing against 0

sz-piotr (changes requested):

packages/backend/src/peripherals/zksync/ZksyncClient.ts
```diff
@@ -0,0 +1,112 @@
...
+    if (!httpResponse) {
+      const message = getErrorMessage(error)
+      this.recordError(path, timeMs, message)
+      throw error
+    }
+
+    const text = await httpResponse.text()
+    const zksyncResponse = tryParseZksyncResponse(text)
+
+    if (!httpResponse.ok) {
```
sin6pi7, line 80: This can be done before parsing response body.

packages/backend/src/peripherals/zksync/parseZksyncResponse.ts
```diff
@@ -0,0 +1,10 @@
+import { ZksyncResponse } from './schemas'
+
+export function parseZksyncResponse(value: string): ZksyncResponse {
+  try {
+    const json: unknown = JSON.parse(value)
+    return ZksyncResponse.parse(json)
+  } catch {
+    throw new TypeError('Invalid Zksync response.')
```
sin6pi7, line 8: Afaik `zod` should include in the error what did not fit the schema - would be good to log that info somehow for simpler debugging.

packages/backend/src/peripherals/zksync/ZksyncClient.ts
```diff
@@ -0,0 +1,112 @@
...
+  }
+
+  private recordError(path: string, timeMs: number, message: string) {
+    this.logger.debug({ type: 'error', message, timeMs, path })
+  }
+}
+
+function tryParseZksyncResponse(text: string) {
+  try {
+    return parseZksyncResponse(text)
```
sin6pi7, line 108: `zod` has a `safeParse` method which never throws - https://github.com/colinhacks/zod#safeparse

sin6pi7 (changes requested):
Left few syntax comments. It would also be great to see in PR description why for zkSync we are storing all transactions instead of count per block like for other l2s.

packages/backend/src/peripherals/zksync/ZksyncClient.ts
michalsidzej: Do you mean the lastTx?

packages/backend/src/core/transaction-count/ZksyncTransactionUpdater.ts
michalsidzej: I've found not such blocks up to 100k (now at 107k)

packages/backend/src/peripherals/zksync/parseZksyncResponse.ts
michalsidzej: We use zod in a few places and that would probably be a smart thing to do, but maybe not in the scope of this PR

sz-piotr (approved):

sin6pi7 (approved):

PR #698, opened 2022-09-28
Add separate alchemy keys for Optimism and Alchemy

michalsidzej:


packages/backend/src/setup/createRpcTransactionUpdaters.ts
```diff
@@ -43,18 +45,28 @@ export function createRpcTransactionUpdaters(
...
+      url: rpc.url,
+      timeout: 10000,
+    })
+  }
 
-    default:
-      throw new Error('Unknown provider')
+  let apiKey = ''
+  if (projectId === ProjectId('arbitrum')) {
+    apiKey = config.optimismAlchemyApiKey
```
sin6pi7, line 62: Shouldn't this be `arbitrumAlchemyKey`?

sin6pi7, line 62:
```suggestion
    apiKey = config.arbitrumAlchemyApiKey
```

packages/backend/src/setup/createRpcTransactionUpdaters.ts
```diff
@@ -43,18 +45,28 @@ export function createRpcTransactionUpdaters(
...
+  }
+  if (projectId === ProjectId('optimism')) {
+    apiKey = config.optimismAlchemyApiKey
   }
+  if (!apiKey) {
```
sin6pi7, line 67: Shouldn't this fall back to `config.alchemyApiKey`?

michalsidzej: Maybe we don't want to flood this key with ton of other requests. I think it would make sense to create a key for each project in that case

sin6pi7 (approved):

PR #701, opened 2022-09-28
Simplify Layer2Category

sz-piotr:


krzkaczor (commented):
Are we losing info from "details" or is this duplicated somewhere?

sz-piotr:
> Are we losing info from "details" or is this duplicated somewhere?

This information wasn't displayed anywhere

sin6pi7 (approved):

PR #703, opened 2022-09-28
Unify api responses naming

sin6pi7:
Following suggestion from @sz-piotr. We have 3 features: `tvl`, `activity` and `event`. Each feature has its own endpoint and therefore response types. Naming convention is the following:
- `FeatureApiResponse` for `Feature` endpoint response type,
- `FeatureApiSpecificType` for `Feature` endpoint specific type e.g. `ChartPoint`,
-  ~`featureResponse`~ `featureApiResponse` for variable name of type `FeatureApiResponse`


packages/frontend/src/build/buildPages.ts
```diff
@@ -15,7 +15,7 @@ main().catch((e) => {
 async function main() {
   const env = process.env.DEPLOYMENT_ENV ?? 'production'
   const config = getConfig(env)
-  const apiMain = await fetchApiMain(
+  const apiMain = await fetchTvlApi(
```
sz-piotr, line 18: `apiMain` -> `tvlApiResponse`. This rename should propagate in the entire codebase

packages/backend/src/api/controllers/report/generateTvlResponse.ts
```diff
@@ -1,37 +1,40 @@
...
+  TvlApiResponse,
 } from '@l2beat/types'
 
 import { AggregateReportRecord } from '../../../peripherals/database/AggregateReportRepository'
 import { ReportRecord } from '../../../peripherals/database/ReportRepository'
 import { asNumber } from './asNumber'
 import { getChartPoints } from './charts'
 
-export function generateMain(
+export function generateTvlResponse(
```
sz-piotr, line 14: `generateTvlApiResponse`. This applies to many more things. I believe that we shouldn't make an exception for variable names and include the full type name everywhere. It might be three letters longer but it avoids confusion as to which form should be used.

sz-piotr (changes requested):

packages/types/src/api/EventsApi.ts
```diff
@@ -0,0 +1,32 @@
...
+const EventsApiCharts = z.object({
+  hourly: EventsApiChart,
+  daily: EventsApiChart,
+})
+export type EventsApiCharts = z.infer<typeof EventsApiCharts>
+
+const EventsApiResponse = z.object({
+  projects: z.record(z.string(), EventsApiCharts.optional()),
+})
+export type EventsApiResponse = z.infer<typeof EventsApiResponse>
```
sz-piotr, line 32: Following previous suggestion by @antooni we should rename `Events*` to `Event*` so in this case `EventsApiResponse` -> `EventApiResponse` as we are using singular form for almost everything else in the code

sz-piotr (changes requested):

packages/backend/src/api/controllers/report/generateTvlResponse.ts
sin6pi7: I like it 👍

packages/types/src/api/EventsApi.ts
sin6pi7: Fair enough. What about `EventsRouter`? We are serving data on `/events` path.

sin6pi7: I made everything apart from routes singular, because url path seems like an exception.

packages/frontend/src/build/buildPages.ts
```diff
@@ -15,33 +15,33 @@ main().catch((e) => {
 async function main() {
   const env = process.env.DEPLOYMENT_ENV ?? 'production'
   const config = getConfig(env)
-  const apiMain = await fetchApiMain(
+  const tvlApiResponse = await fetchTvlApi(
     config.backend.apiUrl,
     config.backend.skipCache,
   )
-  const apiActivity = await fetchApiActivity(config.backend.apiUrl)
+  const activityApiResponse = await fetchApiActivity(config.backend.apiUrl)
```
sz-piotr, line 22:
```suggestion
  const activityApiResponse = await fetchActivityApi(config.backend.apiUrl)
```

sz-piotr (approved):

PR #704, opened 2022-09-28
Unify api controller renaming

sin6pi7:
A cherry on top of https://github.com/l2beat/l2beat/pull/703 in http layer.


michalsidzej (approved):
Finally

PR #708, opened 2022-09-28
Add zksync to activity controller

sin6pi7:
Exposes zksync transaction counting to controller. Refactor the controller to make it more generic.

### DIFF
--- a/packages/backend/src/core/transaction-count/ZksyncTransactionUpdater.ts
+++ b/packages/backend/src/core/transaction-count/ZksyncTransactionUpdater.ts
@@ -1,4 +1,5 @@
 import { Logger, TaskQueue, UniqueTaskQueue } from '@l2beat/common'
+import { ProjectId } from '@l2beat/types'
 
 import {
   ZksyncTransactionRecord,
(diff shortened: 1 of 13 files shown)


packages/backend/src/core/transaction-count/ZksyncTransactionUpdater.ts
```diff
@@ -79,4 +81,13 @@ export class ZksyncTransactionUpdater {
 
     this.logger.info('Update enqueued')
   }
+
+  async getDailyTransactionCounts() {
+    const counts =
+      await this.zksyncTransactionRepository.getDailyTransactionCount()
+    return {
+      projectId: ProjectId('zksync'), // TODO: take from config
```
sin6pi7, line 89: @sz-piotr how do you suggest we solve this? Ideally we take it from a constant from config, but I'm open for suggestions.

sz-piotr (approved):

packages/backend/src/core/transaction-count/ZksyncTransactionUpdater.ts
sz-piotr: We could introduce this as a constant like this: `ProjectId.ZKSYNC`

PR #709, opened 2022-09-28
Do not require configuration for turned off features

krzkaczor:
Resolves: L2B-145

Problem: 
* Backend requires configuration for features that are turned off
* It's not easy to conditionally create dependencies in `Application.ts`

I struggled a little bit to find a solution that doesn't introduce too much boilerplate code. Finally, I decided to rely on usage of binary operators (`&&`,` ||`) and force unwrap `undefined` (with `!`) were really needed.

EDIT:
Introduced modules.


packages/backend/src/setup/createRpcTransactionUpdaters.ts
```diff
@@ -59,10 +59,10 @@ function createL2Provider(
 
   let apiKey = ''
   if (projectId === ProjectId('arbitrum')) {
-    apiKey = config.arbitrumAlchemyApiKey
+    apiKey = config.transactionCountSync!.arbitrumAlchemyApiKey
```
sz-piotr, line 62: Not a fan of this. Maybe there is a way for this function to take ConfigWithTransactionCount instead of Config?

.eslintrc.json
```diff
@@ -19,6 +19,7 @@
   },
   "plugins": ["simple-import-sort", "no-only-tests", "unused-imports"],
   "rules": {
+    "@typescript-eslint/no-non-null-assertion": "off",
```
sz-piotr, line 22: It's so easy to abuse this. I prefer the ugly comments required in each instance

packages/backend/src/setup/modules.ts
```diff
@@ -0,0 +1,98 @@
...
+import { ZksyncTransactionRepository } from '../peripherals/database/ZksyncTransactionRepository'
+import { StarkexClient } from '../peripherals/starkex'
+import { ZksyncClient } from '../peripherals/zksync'
+import { createRpcTransactionUpdaters } from './createRpcTransactionUpdaters'
+import { createStarkexTransactionUpdaters } from './createStarkexTransactionUpdaters'
+
+interface AppModule {
+  start?: () => void
+  router?: Router
+}
```
sz-piotr, line 21: This type is unnecessary and potentially does more harm then good. IMO `getActivityModule` should just return its own object or undefined without trying to generalize.

Also, I think that instead of this file being named `modules.ts` it should be named `ActivityModule.ts`

packages/backend/src/tools/assert.ts
```diff
@@ -0,0 +1,6 @@
+// eslint-disable-next-line
+export function assert(condition: any, msg = ''): asserts condition {
+  if (!condition) {
+    throw new Error('Assertion Error: ' + msg)
+  }
+}
```
sz-piotr, line 6:
```suggestion
export function assert(condition: any, message?: string): asserts condition {
  if (!condition) {
    throw new Error(message ? 'Assertion Error: ${message}' : 'Assertion Error')
  }
}
```

sz-piotr (changes requested):

packages/backend/src/config/Config.ts
```diff
@@ -26,7 +25,12 @@ export interface Config {
...
-  arbitrumAlchemyApiKey: string
-  optimismAlchemyApiKey: string
+  transactionCountSync?:
+    | {
+        starkexApiUrl: string
+        starkexApiKey: string
+        arbitrumAlchemyApiKey: string
+        optimismAlchemyApiKey: string
+      }
+    | undefined
```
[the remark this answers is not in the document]
krzkaczor: oh, maybe i can drop this now actually

packages/backend/src/setup/modules.ts
sin6pi7: I agree with @sz-piotr with one addition - imo instead of returning `undefined` when `transactionCountSync` is missing in `config`, `getActivityModule` should accept just `transactionCountSync` instead of full `config`, because it does not make sense to create this module otherwise - caller should be handling this.

krzkaczor: The problem with this is that its dependencies require full configs.

sin6pi7: Ah, you're right. IMO that's an issue with its dependencies (because they in fact do not require full config), but that's a bigger refactor and can be done in a separate PR.

krzkaczor: I think sometimes they actually do require parts of the config outside of counting related stuff. I think generally is simpler to treat config as a blob.

packages/backend/src/config/Config.ts
```diff
@@ -26,7 +25,12 @@ export interface Config {
   syncEnabled: boolean
   freshStart: boolean
   eventsSyncEnabled: boolean
-  transactionCountSyncEnabled: boolean
-  arbitrumAlchemyApiKey: string
-  optimismAlchemyApiKey: string
+  transactionCountSync:
```
sin6pi7, line 28: Maybe rename to `transactionCounting`? Always one letter less ;)

packages/backend/src/config/getEnv.ts
```diff
@@ -49,3 +49,7 @@ getEnv.boolean = function getBooleanFromEnv(
   }
   throw new Error(`Missing environment variable ${name}!`)
 }
+
+getEnv.featureFlag = function getFeatureFlagFromEnv(name: string): boolean {
+  return getEnv.boolean(name, false)
+}
```
sz-piotr, line 55: With how simple the function is is it really necessary?

packages/backend/src/tools/assert.ts
```diff
@@ -0,0 +1,6 @@
+// eslint-disable-next-line
+export function assert(condition: any, message?: string): asserts condition {
```
sz-piotr, line 2: If you used `condition: unknown` you wouldn't have to disable eslint.

packages/backend/src/setup/ActivityModule.ts
```diff
@@ -0,0 +1,94 @@
...
+
+export function getActivityModule(
+  config: Config,
+  logger: Logger,
+  http: HttpClient,
+  database: Database,
+  clock: Clock,
+) {
+  if (!config.transactionCountSync) {
+    return {}
```
sz-piotr, line 25: returning undefined here makes more sense IMO

sz-piotr (changes requested):

packages/backend/src/config/getEnv.ts
krzkaczor: WITH 80 CHARS PER LINE LIMIT I AM FIGHTING FOR EACH CHARACTER TO MAKE THE CONFIG READABLE xD but seriously i also thought about dropping it... but then there are new lines everywhere.

packages/backend/src/tools/assert.ts
krzkaczor: Yeah i thought about this but I am not sure if this won't break type inference somewhere.. I will check it out.

sz-piotr (approved):

PR #716, opened 2022-09-29
Optionally show layer2s on bridge tvl view

sz-piotr:
Resolves L2B-146

I needed make `Layer2 | Bridge` a discriminated union so that's why all of them now have a `type` member.

This PR uses `data-bridges-only` and `data-combined-only` to toggle visibility of elements depending on the checkbox state. The one thing this PR doesn't update is the chart, which will probably be the subject of the next PR.


sin6pi7 (approved):

michalsidzej (approved):
gicik

PR #717, opened 2022-09-29
Subtract Nitro system transactions

michalsidzej:
Resolves L2B-126

### DIFF
--- a/packages/backend/src/core/transaction-count/RpcTransactionUpdater.ts
+++ b/packages/backend/src/core/transaction-count/RpcTransactionUpdater.ts
@@ -1,4 +1,5 @@
 import { Logger, TaskQueue, UniqueTaskQueue } from '@l2beat/common'
+import { AssessCount } from '@l2beat/config'
 import { ProjectId, UnixTime } from '@l2beat/types'
 
 import { RpcTransactionCountRepository } from '../../peripherals/database/RpcTransactionCountRepository'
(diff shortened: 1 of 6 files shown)


packages/backend/test/core/transaction-count/RpcTransactionUpdater.test.ts
```diff
@@ -142,6 +142,70 @@ describe(RpcTransactionUpdater.name, () => {
         ],
       ])
     })
+
+    it('subtracts Nitro system transactions', async () => {
+      const TIME_0 = new UnixTime(0)
+      const ethereumClient = mock<EthereumClient>({
+        getBlock: async () =>
+          fakeBlock({
+            number: 1,
```
krzkaczor, line 151: to make this mock correct here we should return actual block number that was asked.

krzkaczor (commented):
This makes RpcTxUpdater aware of the specifics of an exact network, but I think that is acceptable and it's much more readable than inheritance or anything like that.

packages/backend/src/core/transaction-count/RpcTransactionUpdater.ts
```diff
@@ -61,6 +61,19 @@ export class RpcTransactionUpdater implements TransactionCounter {
     })
   }
 
+  // We need to subtract the one system transaction that appear in each Nitro block
+  private calcTransactionCount(blockNumber: number, transactions: string[]) {
```
sin6pi7, line 65: It would be cleaner if this function was configurable through constructor instead of hardcoded project ids.

sin6pi7 (changes requested):

packages/backend/src/core/transaction-count/RpcTransactionUpdater.ts
michalsidzej: Wanted to do it like this from the beginning, thanks for help!

sin6pi7 (approved):

PR #718, opened 2022-09-29
Loopring Transaction Count

michalsidzej:


packages/backend/src/tools/types.ts
```diff
@@ -32,3 +32,17 @@ export function stringAs<T>(Brand: (s: string) => T) {
...
+    .refine((s) => {
+      try {
+        Brand(s)
+        return true
+      } catch {
+        return false
+      }
+    })
+    .transform(Brand)
+}
```
sin6pi7, line 48: This could be abstracted away the following way:
```

function branded<T extends z.ZodTypeAny, B>(
  schema: T,
  Brand: (t: z.TypeOf<T>) => B,
) {
  return schema
    .refine((s: T) => {
      try {
        Brand(s)
        return true
      } catch {
        return false
      }
    })
    .transform(Brand)
}

export function stringAs<T>(Brand: (t: string) => T) {
  return branded(z.string(), Brand)
}

export function numberAs<T>(Brand: (t: number) => T) {
  return branded(z.number(), Brand)
}
```

sin6pi7 (changes requested):

packages/backend/src/peripherals/database/shared/types.ts
```diff
@@ -82,7 +82,7 @@ declare module 'knex/types/tables' {
     report_status: ReportStatusRow
     events: EventRow
     transactions: {
-      rpc: RpcTransactionCountRow
+      block: BlockTransactionRow
```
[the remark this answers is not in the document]
michalsidzej: Which one do you actually prefer? RN we have a bit inconsistency in repo, so that's probably a good point to bring it back. (we have `ZksyncTransactionRepository` and `StarkexTransactionCountRepository`). Personally I like the former as it's shorter. Or maybe it'd be best to say `BlockActivityRow`. That would be consistent with the module naming and it's most explanatory and the shortest. WYT?

packages/backend/src/core/transaction-count/RpcTransactionUpdater.ts
```diff
@@ -104,6 +104,6 @@ export class RpcTransactionUpdater implements TransactionCounter {
   async getDailyTransactionCounts() {
     return this.rpcTransactionCountRepository.getDailyTransactionCount(
       this.projectId,
-    )
+    )sssssssssssssssssssssssssssssssssssssss
```
sin6pi7, line 107:
```suggestion
    )
```

packages/backend/src/tools/types.ts
```diff
@@ -19,11 +19,14 @@ export function stringAsBigInt(fallback?: bigint) {
...
-  return z
-    .string()
+function branded<T extends z.ZodTypeAny, B>(
+  schema: T,
+  Brand: (t: z.TypeOf<T>) => B,
+) {
+  return schema
     .refine((s) => {
       try {
+        // eslint-disable-next-line @typescript-eslint/no-unsafe-argument
```
sin6pi7, line 29:
```suggestion
    .refine((s: T) => {
      try {
```

sin6pi7 (changes requested):

packages/backend/src/core/transaction-count/RpcTransactionUpdater.ts
michalsidzej: Hmmmmmm

PR #719, opened 2022-09-30
Show layer2s on bridge risk view

sz-piotr:
Resolves L2B-127 and L2B-158


packages/frontend/src/build/fetchActivityApi.ts
```diff
@@ -4,6 +4,7 @@ import { ActivityApiChart, ActivityApiResponse, UnixTime } from '@l2beat/types'
 export async function fetchActivityApi(
   apiUrl: string,
 ): Promise<ActivityApiResponse> {
+  return getMockData()
```
sin6pi7, line 7: Should it still be here?

sin6pi7 (approved):

PR #720, opened 2022-09-30
Limit ethereum block counting

sin6pi7:
There is no point to count transactions in blocks prior to first layer 2 block. Not ideal solution, but lets us continue spiking this feature.


packages/backend/src/config/config.staging.ts
```diff
@@ -21,6 +21,7 @@ export function getStagingConfig(): Config {
       starkexApiKey: getEnv('STARKEX_API_KEY'),
       arbitrumAlchemyApiKey: getEnv('ARBITRUM_ALCHEMY_API_KEY'),
       optimismAlchemyApiKey: getEnv('OPTIMISM_ALCHEMY_API_KEY'),
+      ethereumAlchemyApiKey: getEnv('ALCHEMY_API_KEY'),
```
sin6pi7, line 24: That's for now - we should probably have a separate key for ethereum, so that tvl and activity features do not affect each others performance.

sz-piotr (approved):

PR #721, opened 2022-09-30
Add cache to fetchActivityApiResponse

antooni:


sz-piotr (approved):

packages/frontend/src/build/caching/CachedHttpClient.ts
```diff
@@ -0,0 +1,61 @@
...
+    if (!response.ok) {
+      throw new Error(
+        `Could not get data from api (received status ${response.status})`,
+      )
+    }
+    const json: unknown = await response.json()
+
+    await write(url, JSON.stringify(json))
+
+    return JSON.stringify(json)
```
sz-piotr, line 28: Why stringify the json? I think its fine to return `unknown`

sz-piotr (changes requested):

sz-piotr (approved):

PR #722, opened 2022-09-30
Rollbar integration

krzkaczor:


packages/backend/src/tools/ErrorReporter.ts
```diff
@@ -0,0 +1,25 @@
+import { Context } from 'koa'
+import Rollbar from 'rollbar'
+
+const accessToken = process.env.ROLLBAR_ACCESS_TOKEN
```
sz-piotr, line 4: This could probably go to config and we could make error reporter a class WDYT?

sz-piotr (changes requested):
One thing that this integration misses is errors during system startup. Maybe we can also try to capture those?

krzkaczor:
> One thing that this integration misses is errors during system startup. Maybe we can also try to capture those?

Hmm, as quickly as error reporter file will be required it should notify on any crashes (`captureUncaught: true,`). This is exactly the reason why I wouldn't refactor it as a class or anything like this. Just load it as soon as it's possible and it should be fine. 

But you are right that it's missing reporting at index.ts -- I just added it.

sz-piotr (approved):

PR #724, opened 2022-09-30
Add status to activity endpoints

sin6pi7:
A bit more insight into what is going on - just a start.

### DIFF
--- a/packages/backend/src/core/transaction-count/ZksyncTransactionUpdater.ts
+++ b/packages/backend/src/core/transaction-count/ZksyncTransactionUpdater.ts
@@ -92,6 +92,7 @@ export class ZksyncTransactionUpdater implements TransactionCounter {
     return {
       queuedJobsCount: this.blockQueue.length,
       missingRanges: await this.zksyncTransactionRepository.getMissingRanges(),
+      busyWorkers: this.blockQueue.getBusyWorkers(),
     }
   }
 }
--- a/packages/backend/src/core/transaction-count/ZksyncTransactionUpdater.ts
+++ b/packages/backend/src/core/transaction-count/ZksyncTransactionUpdater.ts
@@ -87,4 +87,11 @@ export class ZksyncTransactionUpdater implements TransactionCounter {
   async getDailyTransactionCounts() {
     return this.zksyncTransactionRepository.getDailyTransactionCount()
   }
+
+  async getStatus() {
+    return {
+      queuedJobsCount: this.blockQueue.length,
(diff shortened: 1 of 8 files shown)


packages/backend/src/peripherals/database/RpcTransactionCountRepository.ts
```diff
@@ -55,14 +55,15 @@ export class RpcTransactionCountRepository extends BaseRepository {
       `
       WITH 
         project_blocks AS (
-          SELECT * FROM transactions.rpc WHERE project_id=?
+          SELECT * FROM transactions.rpc WHERE project_id = ?
```
krzkaczor, line 58: do we need these changes in this PR?

sz-piotr (changes requested):

krzkaczor (approved):

packages/common/src/tools/queue/TaskQueue.ts
```diff
@@ -35,6 +35,10 @@ export class TaskQueue<T> {
     return this.queue.length
   }
 
+  getBusyWorkers() {
+    return this.busyWorkers
+  }
```
michalsidzej, line 40: Why not set `readonly busyWorkers`?

packages/common/src/tools/queue/TaskQueue.ts
```diff
@@ -7,7 +7,7 @@ import { Job, ShouldRetry, TaskQueueOpts } from './types'
 export class TaskQueue<T> {
   private queue: Job<T>[] = []
   private busyWorkers = 0
```
michalsidzej, line 9:
```suggestion
  readonly busyWorkers = 0
```

packages/common/src/tools/queue/TaskQueue.ts
sin6pi7: This will block updating this value from the inside of the class too.

packages/backend/src/core/transaction-count/ZksyncTransactionUpdater.ts
```diff
@@ -87,4 +87,12 @@ export class ZksyncTransactionUpdater implements TransactionCounter {
   async getDailyTransactionCounts() {
     return this.zksyncTransactionRepository.getDailyTransactionCount()
   }
+
+  async getStatus() {
+    return {
+      queuedJobsCount: this.blockQueue.length,
```
sz-piotr, line 93:
```suggestion
      queuedJobCount: this.blockQueue.length,
```

sz-piotr (approved):

PR #726, opened 2022-10-03
Fix tx-count duplicate key

michalsidzej:
Resolves L2B-197
Because we have a lot of ongoing tasks while querying missing blocks ranges, so probably querying those two in one query should solve the problem

### DIFF
--- a/packages/backend/src/peripherals/database/RpcTransactionCountRepository.ts
+++ b/packages/backend/src/peripherals/database/RpcTransactionCountRepository.ts
@@ -1,5 +1,6 @@
 import { Logger } from '@l2beat/common'
 import { ProjectId, UnixTime } from '@l2beat/types'
+import { Knex } from 'knex'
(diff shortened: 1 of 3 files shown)


packages/backend/src/peripherals/database/RpcTransactionCountRepository.ts
```diff
@@ -51,42 +52,55 @@ export class RpcTransactionCountRepository extends BaseRepository {
...
-      `
-      WITH 
-          project_blocks AS (
-            SELECT * FROM transactions.rpc WHERE project_id=?
-          )
-        SELECT * 
-        FROM (
-          SELECT project_blocks.block_number 
+        ),
+        no_next AS (
```
sin6pi7, line 61: I think it **might be possible** that `no_next` and `no_prev` being part of `WITH` statement would slow things down - the alternative would be to keep only `project_blocks` inside `WITH`.

sin6pi7 (approved):

PR #727, opened 2022-10-03
Activity sync

sin6pi7:
(DO NOT MERGE) Solely for duplicating frontend deployment


packages/backend/src/Application.ts
```diff
@@ -200,14 +200,16 @@ export class Application {
       await database.migrateToLatest()
 
       if (config.syncEnabled) {
```
sin6pi7, line 202: Do we actually want/need this? Feature-specific flags seem to be sufficient @krzkaczor @sz-piotr

packages/frontend/src/build/config/config.staging.ts
```diff
@@ -9,7 +9,7 @@ export function getStagingConfig() {
       activity: true,
     },
     backend: {
-      apiUrl: 'https://staging.l2beat.com',
+      apiUrl: 'https://l2beat-activity.herokuapp.com',
```
sin6pi7, line 12:
```suggestion
      apiUrl: 'https://staging.l2beat.com',
```

packages/backend/src/Application.ts
sin6pi7: IMO we should remove it, but this can be done in a separate PR.

packages/common/src/tools/queue/Retries.ts
```diff
@@ -42,4 +42,13 @@ export const Retries = {
   always,
   exponentialBackOff,
   maxAttempts,
+  // never drops failing tasks
+  defaultExponentialBackOff: exponentialBackOff(100, {
```
sz-piotr, line 46: I think that simply `default` would be better here

packages/common/src/tools/queue/Retries.ts
```diff
@@ -42,4 +42,13 @@ export const Retries = {
...
+    maxDistance: 3_000,
+  }),
+  // drops failing tasks after few retries
+  defaultExponentialBackOffAndDrop: exponentialBackOff(100, {
```
sz-piotr, line 50: I am not sure if there can be two defaults. Maybe just `backOffAndDrop`? Or maybe we should differentiate and have the functions be `camelCase` and values be `CONST_CASE`

sz-piotr: After some more thoughts:

in TaskQueue `const DEFAULT_RETRY = Retires.expo...`
in `/core/transaction-count/utils.ts` `export const BACK_OFF_AND_DROP = Retires...`

packages/backend/src/peripherals/database/migrations/029_update_transactions.ts
```diff
@@ -0,0 +1,22 @@
...
+
+If you find that something was incorrectly set up in the `up` function you
+should create a new migration file that fixes the issue.
+
+*/
+
+import { Knex } from 'knex'
+
+export async function up(knex: Knex) {
+  await knex.schema.withSchema('transactions').renameTable('rpc', 'block')
```
sz-piotr, line 17: Please fix primary key constraint for this table and starkex

packages/backend/src/core/balances/BalanceUpdater.ts
```diff
@@ -39,6 +36,10 @@ export class BalanceUpdater {
   ) {
     this.logger = this.logger.for(this)
     this.configHash = getBalanceConfigHash(projects)
+    this.taskQueue = new TaskQueue(
+      this.update.bind(this),
```
krzkaczor, line 40: can we normalize to () => this.update()

packages/backend/src/core/reports/ReportUpdater.ts
```diff
@@ -32,6 +29,10 @@ export class ReportUpdater {
   ) {
     this.logger = this.logger.for(this)
     this.configHash = getReportConfigHash(projects)
+    this.taskQueue = new TaskQueue(
+      this.update.bind(this),
```
krzkaczor, line 33: same here

packages/common/src/tools/queue/TaskQueue.ts
```diff
@@ -93,9 +119,13 @@ export class TaskQueue<T> {
     this.busyWorkers++
     const job = this.queue.splice(jobIndex, 1)[0]
     try {
+      this.logger.info('Executing job', { job: JSON.stringify(job) })
```
krzkaczor, line 122: this should be debug

packages/backend/src/core/balances/BalanceUpdater.ts
sin6pi7: I will do this, but in a separate PR - will rename updaters to counters - lets RIP this PR.

sz-piotr (approved):

PR #732, opened 2022-10-03
Improve app configurability 

krzkaczor:
* add the ability to turn off reporting-related APIs
* add the ability to configure # of workers for transaction sync

Resolves: L2B-199

### DIFF
--- a/packages/backend/src/config/config.staging.ts
+++ b/packages/backend/src/config/config.staging.ts
@@ -16,12 +16,17 @@ export function getStagingConfig(): Config {
       ...(productionConfig.databaseConnection as Knex.PgConnectionConfig),
       application_name: name,
     },
+    tvlReportSync: false,
     transactionCountSync: {
       starkexApiUrl: getEnv('STARKEX_API_URL'),
       starkexApiKey: getEnv('STARKEX_API_KEY'),
       arbitrumAlchemyApiKey: getEnv('ARBITRUM_ALCHEMY_API_KEY'),
       optimismAlchemyApiKey: getEnv('OPTIMISM_ALCHEMY_API_KEY'),
       ethereumAlchemyApiKey: getEnv('ALCHEMY_API_KEY'),
+      rpcWorkQueueLimit: 200_000,
+      rpcWorkQueueWorkers: 1,
+      zkSyncWorkQueueWorkers: 1,
+      starkexWorkQueueWorkers: 1,
     },
   }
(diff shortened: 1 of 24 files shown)


packages/backend/src/config/config.staging.ts
```diff
@@ -16,12 +16,17 @@ export function getStagingConfig(): Config {
       ...(productionConfig.databaseConnection as Knex.PgConnectionConfig),
       application_name: name,
     },
+    tvlReportSync: false,
```
sin6pi7, line 19:
```suggestion
    tvlReportSync: true,
```

Is there a reason why we should not sync tvl on staging?

packages/backend/src/core/transaction-count/RpcTransactionUpdater.ts
```diff
@@ -26,18 +23,19 @@ export class RpcTransactionUpdater implements TransactionCounter {
...
-  private startBlock: number
+  private readonly assessCount: AssessCount
+  private readonly startBlock: number
 
   constructor(
-    private ethereumClient: EthereumClient,
-    private rpcTransactionCountRepository: RpcTransactionCountRepository,
-    private clock: Clock,
-    private logger: Logger,
+    private readonly config: TransactionCountSyncConfig,
```
sin6pi7, line 34: Why passing a separate object instead of adding new parameters to `opts`? Changing the signature requires us to update more files to add this change. What is more, many keys from `TransactionCountSyncConfig` are not used in this class.
This applies to other updaters as well.

packages/backend/src/config/config.staging.ts
sin6pi7: Nvm, staging now === activity app.

packages/backend/src/core/transaction-count/RpcTransactionUpdater.ts
```diff
@@ -26,24 +24,26 @@ export class RpcTransactionUpdater implements TransactionCounter {
     this.updateBlock.bind(this),
     this.logger,
     {
-      workers: 100,
+      workers: this.opts?.workQueueSizeLimit,
```
sin6pi7, line 27:
```suggestion
      workers: this.opts?.workQueueWorkers,
```

sin6pi7 (changes requested):

sin6pi7 (approved):

PR #745, opened 2022-10-04
Re-style navbar with support for bridges

sz-piotr:
Resolves L2B-174

The actual implementation varies from the designs mainly due to the values in design not being aligned with tailwind's scale.

The mobile navbar is also mostly my invention based loosely on designs.

Some z values had to be adjusted in the chart because they were showing over the sidebar.

The desktop and mobile navbars as well as the sidebar are slightly different when bridges and activity are disabled.

Screen Shot 2022-10-04 at 15 46 30
Screen Shot 2022-10-04 at 15 46 10
Screen Shot 2022-10-04 at 15 45 33
Screen Shot 2022-10-04 at 15 46 03

<table>
<tr>
<td>

</td>
<td>

</td>
</tr>
<tr>
<td>

</td>
<td>

</td>
</tr>
</table>


krzkaczor (commented):
image

I find it odd that we provide so much space for "light/night mode" switch on this screen. Wouldn't it be better to just add it as another small icon next to the other ones?

sz-piotr:
> I find it odd that we provide so much space for "light/night mode" switch on this screen. Wouldn't it be better to just add it as another small icon next to the other ones?

In this case I just roughly implemented the design. I also find that it potentially takes to much space but IMO we should just work with the designer on a better version rather than holding back on this PR

PR #746, opened 2022-10-04
Improve error logging for TaskQueue

krzkaczor:


packages/common/src/tools/queue/TaskQueue.ts
```diff
@@ -23,7 +26,12 @@ export class TaskQueue<T> {
       this.workers > 0 && Number.isInteger(this.workers),
       'workers needs to be a positive integer',
     )
-    this.shouldRetry = opts?.shouldRetry ?? Retries.always
+    this.shouldRetry =
```
sin6pi7, line 29: Afaik some queues depend on the fact that a task will always be retried e.g. updaters `updateQueue`s - correct me if I'm wrong @sz-piotr @krzkaczor

krzkaczor: The more I thing about it, i think the default strategy should be `always` but with a delay (do not drop tasks).

sin6pi7: Then you can just skip `maxAttempts` and it will default to that - exponential back-off with maximum waiting time, but running forever.

sin6pi7 (approved):

PR #747, opened 2022-10-04
Wait until queue is empty

sin6pi7:


sz-piotr (approved):

packages/backend/src/core/transaction-count/RpcTransactionUpdater.ts
```diff
@@ -90,6 +90,8 @@ export class RpcTransactionUpdater implements TransactionCounter {
   async update() {
     this.logger.info('Update started', { project: this.projectId.toString() })
 
+    await this.blockQueue.waitTilEmpty()
```
krzkaczor, line 93: this deserves a long comment on why it's so important to wait until the queue is empty :)

krzkaczor (changes requested):

krzkaczor (approved):

PR #748, opened 2022-10-04
Pass Coingecko API key to CI

sz-piotr:


krzkaczor (approved):

krzkaczor (commented):
Shouldn't CoingeckoClient support retries too?

sz-piotr:
> Shouldn't CoingeckoClient support retries too?

The problem is that it was mostly test timeouts not an explicit API error, so retries wouldn't have helped

PR #749, opened 2022-10-04
Add near rainbow bridge info

luke7211:


packages/config/src/bridges/near.ts
```diff
@@ -48,5 +53,179 @@ export const near: Bridge = {
...
+        { address: '0xC3E34337B6505d503DcA3e99303Ad8BE6DB05984', type: 'EOA' },
+        { address: '0xD127323856C3703f94Dc1de18985cdCa2a8bA2b5', type: 'EOA' },
+      ],
+      name: 'MultiSig Participants',
+      description: 'Participants of the 3/6 Aurora MultiSig.',
+    },
+  ],
+}
+
+/*
```
sz-piotr, line 163: Please remove the code comment

sz-piotr (changes requested):

packages/config/src/bridges/near.ts
luke7211: Removed

luke7211: Comment removed

sz-piotr (approved):

PR #751, opened 2022-10-04
Add stargate details

luke7211:


sz-piotr (approved):

sz-piotr:
It would be nice to run the descriptions through grammarly at some point. There are definitely a lot of missing articles like `a` or `the` in them and I am sure that more suggestions will show up.

PR #752, opened 2022-10-04
[Prototype] Project activity chart

michalsidzej:
Resolves L2B-139
Screenshot from 2022-10-04 19-33-17


packages/frontend/src/components/chart/configure/state/getEndpoint.ts
```diff
@@ -1,7 +1,30 @@
...
+  }
+
+  if (selectedToken) {
+    return getTvlEndpoint(selectedToken)
+  }
+  return getTvlEndpoint(chart)
+}
+
+export function getTvlEndpoint(element: HTMLElement) {
+  const endpoint = element.dataset.tvlendpoint
```
sz-piotr, line 17: if you use `data-tvl-endpoint` then you can access it as `element.dataset.tvlEndpoint`. Same for activity

packages/frontend/src/scripts/configureTvlActivity.ts
```diff
@@ -0,0 +1,10 @@
+export function configureTvlActivity() {
```
sz-piotr, line 1: This could probably be moved to the same folder as the actual header

sz-piotr (changes requested):

packages/frontend/src/components/chart/configure/state/getEndpoint.ts
michalsidzej: Oh, that's definitely useful

sz-piotr (approved):

PR #754, opened 2022-10-05
Introduce medium font weight

sz-piotr:


packages/frontend/src/components/navbar/PageLink.tsx
```diff
@@ -9,15 +9,16 @@ export interface PageLinkProps {
...
-    props.large ? 'text-base md:text-lg px-2 md:px-4' : 'px-2',
-    'flex items-center h-full',
-    props.selected &&
-      'pt-0.5 text-pink-900 dark:text-pink-100 border-current border-b-2',
-  )
-
   return (
-    <a className={className} href={props.href}>
+    <a
+      className={classNames(
```
michalsidzej, line 14: Nit: we sometimes import `classNames` as `cx`. Would probably make sense to unify this in the future.

michalsidzej (approved):

PR #755, opened 2022-10-05
Sanity check only relevant projects

sz-piotr:


packages/frontend/src/build/buildPages.ts
```diff
@@ -80,11 +80,15 @@ function sanityCheck(tvlApiResponse: TvlApiResponse) {
...
 
+  const ids = [...bridgesInApi, ...layer2sInApi].map((x) => x.id.toString())
+
   const emptyChartsExist = [
     tvlApiResponse.bridges,
     tvlApiResponse.layers2s,
     tvlApiResponse.combined,
-    ...Object.values(tvlApiResponse.projects).map((x) => x?.charts),
+    ...Object.entries(tvlApiResponse.projects)
+      .filter(([k]) => ids.includes(k))
```
sin6pi7, line 90:
```suggestion
      .filter(([id]) => ids.includes(id))
```

sin6pi7 (approved):
LGTM 👍 just two small suggestions - if these land on a new line why not make it more readable - wdyt?

PR #756, opened 2022-10-05
Revert stage config changes

krzkaczor:


packages/backend/src/config/config.staging.ts
```diff
@@ -16,16 +16,16 @@ export function getStagingConfig(): Config {
...
     transactionCountSync: {
       starkexApiUrl: getEnv('STARKEX_API_URL'),
       starkexApiKey: getEnv('STARKEX_API_KEY'),
       arbitrumAlchemyApiKey: getEnv('ARBITRUM_ALCHEMY_API_KEY'),
       optimismAlchemyApiKey: getEnv('OPTIMISM_ALCHEMY_API_KEY'),
       ethereumAlchemyApiKey: getEnv('ALCHEMY_API_KEY'),
       rpcWorkQueueLimit: 200_000,
-      rpcWorkQueueWorkers: 1,
-      zkSyncWorkQueueWorkers: 1,
+      rpcWorkQueueWorkers: 100,
```
krzkaczor, line 27: do you think we should just make all of these optional in config?

sin6pi7: IMO, ideally all these tx counting updaters run by default in dead-simple mode i.e. single worker, <= 60 calls per minute - this way running it locally won't cause rate limiting issues and is simple to debug.

sin6pi7 (approved):

PR #759, opened 2022-10-05
Fix timezone-specific Events tests

sin6pi7:


sz-piotr (changes requested):
The real fix would be to move away from timestamps with timezones in our database.

According to docs `table.datetime` uses timezones, so a migration that changes those columns should fix the issues: https://knexjs.org/guide/schema-builder.html#datetime

sin6pi7:
> The real fix would be to move away from timestamps with timezones in our database.
> 
> According to docs `table.datetime` uses timezones, so a migration that changes those columns should fix the issues: https://knexjs.org/guide/schema-builder.html#datetime

You're right - I'm gonna do that for the whole codebase, because we have to cast to utc in other places as well.

PR #765, opened 2022-10-05
Ethereum in activity table

michalsidzej:
Resolves L2B-193
Screenshot from 2022-10-05 15-41-21


packages/frontend/src/pages/scaling-activity/props/getActivityView.ts
```diff
@@ -41,3 +47,28 @@ export function getActivityViewEntry(
     transactionsWeeklyCount,
   }
 }
+
+function getEthereumActivityViewEntry(
+  ethereum: Ethereum,
```
sz-piotr, line 52: Why do we need to pass this all the way from the config package? I think having `name: 'Ethereum'` and `slug: 'ethereum'` hardcoded will be much cleaner

packages/frontend/src/pages/scaling-activity/view/ActivityView.tsx
```diff
@@ -48,9 +58,20 @@ export function ActivityView({ items, className }: ActivityViewProps) {
...
+  const rows: RowConfig<ActivityViewEntry> = {
+    getProps: (entry) =>
+      entry.name === 'Ethereum'
+        ? {
+            className: 'accent',
+          }
+        : {},
+  }
+
+  console
```
sz-piotr, line 70: Hello :)

packages/frontend/src/components/ProjectLink.tsx
```diff
@@ -19,33 +19,33 @@ export function ProjectLink({ project, type }: Props) {
       : `/bridges/projects/${project.slug}`
   return (
     <>
-      <a className="ProjectLink" href={href}>
+      <a className="ProjectCell--link" href={href}>
```
sz-piotr, line 22: If you are not planning on rewriting the styles of `ProjectLink` in tailwind I would just leave them untouched.

packages/frontend/src/components/EthreumCell.tsx
```diff
@@ -0,0 +1,23 @@
...
+  return (
+    <>
+      <div className="ProjectCell">
+        <img
+          className="ProjectCell-Icon"
+          src={`/icons/${props.project.slug}.png`}
+          alt={`${props.project.name} logo`}
+        />
+        {props.project.name}
+      </div>
```
sz-piotr, line 20: Please style all new components using tailwind

sz-piotr (changes requested):

packages/frontend/src/components/EthreumCell.tsx
michalsidzej: Oooh, ok, so the SCSS is just a leftover from old days?

michalsidzej: Ethereum cell is sharing all the styling with ProjectLink tho, excluding the hover effect, looks to me like a good idea to use SCSS to keep it consistent here.

sz-piotr: > Ethereum cell is sharing all the styling with ProjectLink tho, excluding the hover effect, looks to me like a good idea to use SCSS to keep it consistent here.

I'll let it slide now, but I'm going to definitely enforce it more after I implement the new designs and reduce SCSS use even more

packages/frontend/src/build/config/common.ts
```diff
@@ -18,4 +18,5 @@ export const common = {
   layer2s,
   bridges,
   tokens: tokenList,
+  ethereum,
```
sz-piotr, line 21: This is not exported now, right?

sz-piotr (changes requested):

packages/frontend/src/components/table/EthereumCell.tsx
```diff
@@ -0,0 +1,23 @@
...
+interface EthereumCellProps {
+  project: {
+    slug: string
+    name: string
+  }
+}
+
+export function EthereumCell({ project }: EthereumCellProps) {
+  return (
+    <>
```
sz-piotr, line 12: The fragment is unnecessary :)

sz-piotr (changes requested):

sz-piotr (approved):

PR #774, opened 2022-10-06
Track events in TaskQueues

sin6pi7:
Introduce a simple, in-memory `EventTracker` and use it in `TaskQueue` - it counts how many started, successful, errored and retried tasks have happened every second. In order not to introduce a memory leak I capped recorded seconds to be a part of last hour (i.e. 3600 seconds maximum per queue will be kept in memory). I considered following information useful and exposed it over http endpoint for every `TransactionCounter`:
- last second count,
- average count/second during last minute,
- average count/second during last hour

Resolves L2B-189


krzkaczor (changes requested):

packages/common/src/tools/queue/TaskQueueMonitor.ts
```diff
@@ -0,0 +1,66 @@
...
+    this.removeOldCounters(now)
+  }
+
+  private removeOldCounters(now: UnixTime) {
+    Array.from(this.timestampCounters.keys())
+      .filter((t) => now.add(-1, 'hours').gt(new UnixTime(t)))
+      .forEach((t) => this.timestampCounters.delete(t))
+  }
+
+  record(when: UnixTime, type: 'success' | 'retry' | 'error') {
```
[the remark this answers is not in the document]
sin6pi7: This is currently visible through `busyWorkers` in the queue.

packages/common/src/tools/queue/TaskQueueMonitor.ts
```diff
@@ -0,0 +1,66 @@
...
+export class TaskQueueMonitor {
+  private readonly timestampCounters = new Map<number, Counters>()
+
+  constructor() {
+    setInterval(() => this.handleNewSecond(), 1_000)
+  }
+
+  private handleNewSecond() {
+    const now = UnixTime.now()
+    if (!this.timestampCounters.has(+now)) {
```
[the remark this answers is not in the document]
sin6pi7: The more I think about it we don't need to store empty info here indeed. And have the interval run every hour to clean up - I like your suggestion 👍

packages/common/src/tools/queue/Monitor.ts
```diff
@@ -0,0 +1,86 @@
...
+export class Monitor {
+  private readonly counters: Record<TaskEvent, Counter> = {
+    success: new Map(),
+    retry: new Map(),
+    error: new Map(),
+    started: new Map(),
+  }
+
+  constructor() {
+    setInterval(() => this.pruneOldCounts(), 1_000 * 60 * 60)
```
krzkaczor, line 18: please extract magic number to a constant

packages/common/src/tools/queue/Monitor.ts
```diff
@@ -0,0 +1,86 @@
+import { UnixTime } from '@l2beat/types'
+
+type Counts = Record<TaskEvent, number>
+
+type TaskEvent = 'success' | 'retry' | 'error' | 'started'
+
+type Counter = Map<number, number>
+
+export class Monitor {
+  private readonly counters: Record<TaskEvent, Counter> = {
```
krzkaczor, line 10: IMO I would try to make this generic ie. store any event. This way we could calculate the occurrence of any event in codebase in the future.

krzkaczor: @sz-piotr opionions?

sin6pi7: Do you rather see it being parametrised on creation i.e. `class Monitor<MonitorEvent> { ... }` or rather allow to record any arbitrary string? I think I'd make it parametrised on creation - wdyt?

packages/common/test/tools/EventTracker.test.ts
```diff
@@ -0,0 +1,65 @@
...
+        something: 1 / 60,
+      },
+      lastHourAverage: {
+        something: 1 / 3_600,
+      },
+    })
+  })
+
+  it('groups event', async () => {
+    const tracker = new EventTracker()
```
krzkaczor, line 33: don't we want to mock timers in tests? i am pretty sure we already do it across the codebase

packages/common/src/tools/EventTracker.ts
```diff
@@ -0,0 +1,65 @@
...
+  }
+
+  private enablePruning() {
+    const intervalId = setInterval(
+      () => this.pruneOldEvents(),
+      this.historySize,
+    )
+    // Let tests end without explicitly stopping the interval
+    // Ref: https://nodejs.org/api/timers.html#timers_timeout_unref
+    intervalId.unref()
```
krzkaczor, line 37: how about adding a public method `stop()`?

krzkaczor (changes requested):

krzkaczor (approved):

PR #775, opened 2022-10-06
Increase container width on main pages

sz-piotr:


krzkaczor (approved):
I just checked how the page looks now. And it looks great 🔥

PR #787, opened 2022-10-07
Add a small badge to indicate L2s

sz-piotr:
Resolves L2B-227

Requires #785
Screenshot 2022-10-07 at 09 20 03

### DIFF
--- a/packages/frontend/src/components/table/ProjectCell.tsx
+++ b/packages/frontend/src/components/table/ProjectCell.tsx
@@ -14,6 +14,7 @@ export interface ProjectCellProps {
     slug: string
     provider?: Layer2['technology']['provider']
     warning?: string
+    showTypeBadge?: boolean
   }
   type: 'layer2' | 'bridge'
 }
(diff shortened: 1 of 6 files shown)


packages/frontend/src/components/table/ProjectCell.tsx
```diff
@@ -35,6 +36,16 @@ export function ProjectCell({ project, type }: ProjectCellProps) {
...
           alt={`${project.name} logo`}
         />
+        {project.showTypeBadge && (
+          <div
+            role="img"
+            aria-label={type}
+            className="absolute -bottom-1 left-2.5 text-3xs font-bold px-0.5 rounded-sm bg-gray-800 dark:bg-gray-200 text-white dark:text-black"
+          >
+            {type === 'layer2' && 'L2'}
+            {type === 'bridge' && 'BR'}
```
michalsidzej, line 46: This will be always false for the current implementation, I understand the badge will also be used to indicate L2s in the future.

luke7211 (approved):

PR #789, opened 2022-10-07
Fix TaskQueue flaky tests

sin6pi7:
Use `sinonjs/fake-timers` to mock timers. This way we make tests deterministic as well as speed them up a lot.

### DIFF
--- a/packages/common/test/tools/queue/TaskQueue.test.ts
+++ b/packages/common/test/tools/queue/TaskQueue.test.ts
@@ -154,5 +154,6 @@ describe(TaskQueue.name, () => {
   })
 })
 
+// Unfortunately fake-timers do not work well with timers/promises
 const wait = (delay: number) =>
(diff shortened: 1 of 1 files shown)


packages/common/test/tools/queue/TaskQueue.test.ts
```diff
@@ -136,8 +146,13 @@ describe(TaskQueue.name, () => {
...
 
+    await time.runAllAsync()
+
     await queue.waitTilEmpty()
 
     expect(queue.length).toEqual(0)
   })
 })
+
+const wait = (delay: number) =>
```
sin6pi7, line 157: Unfortunately `fake-timers` do not work well with `timers/promises`, so I had to make our own promisified version.

krzkaczor: this should be in-code comment

krzkaczor (approved):

PR #790, opened 2022-10-07
Add Gravity

antooni:


luke7211 (changes requested):
Check current token list and add major tokens missing (USDT, WBTC, maybe PSTAKE)

sz-piotr:
I have update the risk view descriptions to be in line with #794. Please check if the changes are correct before proceeding

antooni:
I decided not to include PSTAKE because it is currently 635th by mc on coingecko, so it fells below our top-300 coingecko criteria

packages/config/src/bridges/gravity.ts
```diff
@@ -9,20 +10,94 @@ export const gravity: Bridge = {
...
+      },
+    ],
+    risks: [],
+    isIncomplete: true,
+  },
+  permissions: [
+    {
+      name: 'Cosmos Validators',
+      description:
+        'Control Gravity contract on Ethereum, funds cannot be transfer without the signature of at least 2/3 of the validators set.',
```
sz-piotr, line 99: How does Ethereum know who the 2/3 are

luke7211 (approved):

PR #793, opened 2022-10-08
Xdai bridge

adamiak:
Resolves L2B-232


sz-piotr:
I have update the risk view descriptions to be in line with https://github.com/l2beat/l2beat/pull/794. Please check if the changes are correct before proceeding

sz-piotr (approved):

PR #798, opened 2022-10-10
Hide news

antooni:
I decided to hide news section if there are no news, I know the scope of this task was to just hide news on bridges but I thought it is a good idea to hide it overall if there are no news.
image

Resolves L2B-167


adamiak (approved):

packages/frontend/src/components/project/NewsSection.tsx
```diff
@@ -16,6 +16,10 @@ export interface NewsItem {
 }
 
 export function NewsSection({ news }: NewsSectionProps) {
+  if (news.length == 0) {
```
sz-piotr, line 19: ===. Why doesn't linter check this?

sz-piotr (changes requested):

sz-piotr (approved):

PR #802, opened 2022-10-10
Simplify activity config

sin6pi7:
- allow rpc url overriding from backend config,
- make updaters creation simpler by always using `JsonRpcProvider`


packages/backend/src/config/config.local.ts
```diff
@@ -33,24 +33,46 @@ export function getLocalConfig(): Config {
     projects: layer2s.map(layer2ToProject).concat(bridges.map(bridgeToProject)),
     syncEnabled: !getEnv.boolean('SYNC_DISABLED', false),
     freshStart: getEnv.boolean('FRESH_START', false),
-    tvlReportSync: true,
+    tvlReportSync: false,
```
sin6pi7, line 36:
```suggestion
    tvlReportSync: true,
```

michalsidzej (approved):
I quite like those changes

PR #811, opened 2022-10-11
Add materialized view for getting transactions.block daily count

michalsidzej:
Resolves L2B-244


packages/backend/src/core/transaction-count/MaterializedViewRefresher.ts
```diff
@@ -0,0 +1,32 @@
...
+  start() {
+    this.logger.info('Started')
+    this.queue.addIfEmpty()
+    return this.clock.onNewHour(() => {
+      this.queue.addIfEmpty()
+    })
+  }
+
+  async update() {
+    await this.blockTransactionCountRepository.refresh()
```
sin6pi7, line 30: Would be nice to log that it started and finished

sin6pi7 (approved):

PR #812, opened 2022-10-11
Add Synapse

antooni:
Resolves L2B-215


packages/update-monitor/src/types.ts
```diff
@@ -42,6 +43,12 @@ export interface ArbitrumProxyUpgradeability {
   userImplementation: string
 }
 
+export interface EIP1967ProxyUpgradeability {
+  type: 'eip1967 proxy'
+  adminSlot: string
+  implementationSlot: string
```
adamiak, line 49: I don't think these should be named "Slots", because they're the values of those slots, so "admin" and "implementation".

packages/update-monitor/src/projects/synapse/contracts/timelock.ts
```diff
@@ -0,0 +1,48 @@
...
+    addresses.multisig,
+  )
+  const isMultisigExecutor = await timelock.hasRole(
+    EXECUTOR_ROLE,
+    addresses.multisig,
+  )
+  const isMultisigProposer = await timelock.hasRole(
+    PROPOSER_ROLE,
+    addresses.multisig,
+  )
```
sz-piotr, line 29: Maybe we can get all Executors, Timelock admins and Proposers through events?

packages/update-monitor/README.md
```diff
@@ -3,3 +3,38 @@
 Verifies permissioned addresses against the blockchain.
 
 To run please use `yarn start`. Display help with `yarn start --help`.
+
+## How to add project ?
```
sz-piotr, line 7:
```suggestion
## How to add a new project?
```

packages/update-monitor/src/projects/synapse/contracts/synapseBridge.ts
```diff
@@ -0,0 +1,52 @@
...
+  provider: providers.JsonRpcProvider,
+): Promise<ContractParameters> {
+  const synapseBridge = SynapseBridge__factory.connect(
+    addresses.synapseBridge,
+    provider,
+  )
+
+  const ADMIN_ROLE = await synapseBridge.DEFAULT_ADMIN_ROLE()
+  const GOVERNANCE_ROLE = await synapseBridge.GOVERNANCE_ROLE()
+  const NODEGROUP_ROLE = await synapseBridge.NODEGROUP_ROLE()
```
sz-piotr, line 21: One other idea is since the Role pattern from OpenZeppelin is quite common, maybe we can have some common utils to get all the members?

packages/update-monitor/src/projects/synapse/index.ts
```diff
@@ -0,0 +1,35 @@
...
+
+import { getGnosisSafe } from '../../common/gnosisSafe'
+import { DiscoveryEngine } from '../../discovery/DiscoveryEngine'
+import { ProjectParameters } from '../../types'
+import { addresses } from './constants'
+import { getProxyAdmin } from './contracts/proxyAdmin'
+import { getSynapseBridge } from './contracts/synapseBridge'
+import { getTimelock } from './contracts/timelock'
+
+export const SYNAPSE_NAME = 'synapse'
```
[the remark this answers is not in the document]
antooni: I agree, it is misleading, we should rethink naming or folder structure, but it should not be a part of this PR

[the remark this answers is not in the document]
adamiak: I have been adding: 'orbitBridge', 'omniBridge', 'polynetworkBridge' in my PRs. Synapse is the only one that is not using this convention.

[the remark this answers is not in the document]
adamiak: This is only in update-monitor

packages/update-monitor/src/projects/synapse/contracts/synapseBridge.ts
```diff
@@ -0,0 +1,49 @@
...
+      GOVERNANCE_ROLE,
+      governors: [await synapseBridge.getRoleMember(GOVERNANCE_ROLE, 0)],
+      governorsCount: (
+        await synapseBridge.getRoleMemberCount(GOVERNANCE_ROLE)
+      ).toNumber(),
+      NODEGROUP_ROLE,
+      nodes: [await synapseBridge.getRoleMember(NODEGROUP_ROLE, 0)],
+      nodesCount: (
+        await synapseBridge.getRoleMemberCount(NODEGROUP_ROLE)
+      ).toNumber(),
```
sz-piotr, line 46: This code should be refactored to be like timelock.ts

packages/update-monitor/src/projects/synapse/contracts/timelock.ts
```diff
@@ -0,0 +1,93 @@
...
+    values: {
+      DEFAULT_ADMIN_ROLE,
+      EXECUTOR_ROLE,
+      PROPOSER_ROLE,
+      TIMELOCK_ADMIN_ROLE,
+      minDelaySeconds,
+      timelockAdmins: timelockAdmins.map((t) => t.toString()),
+      defaultAdmins: defaultAdmins.map((t) => t.toString()),
+      executors: executors.map((t) => t.toString()),
+      proposers: proposers.map((t) => t.toString()),
```
sz-piotr, line 86: why do we need to map?

sz-piotr (changes requested):

packages/update-monitor/src/common/accessControl.ts
```diff
@@ -0,0 +1,66 @@
...
+  const revokedFilter = accessControl.filters.RoleRevoked(role)
+
+  const logsRevoked = await provider.getLogs({
+    ...revokedFilter,
+    fromBlock,
+  })
+
+  const revoked = logsRevoked
+    .filter((log) => log.topics[1] === role)
+    .map((log) => log.topics[2])
```
adamiak, line 48: I think it would look simpler to inline logic of removing grantees here and remove `getMembers()` function below completely.

```suggestion
    .map((log) => log.topics[2])
    .forEach((revokedAddress) => {
        const index = granted.indexOf(revokedAddress)
        if (index !== -1) {
            granted.splice(index, 1)
        }
    })
```

But if you prefer to keep `getMembers()` instead then I'm ok with it, but then I'd prefer for it to not modify passed parameter (`revoked`), as it's not a good practice. Personally I prefer the inlined version above.

packages/update-monitor/src/common/accessControl.ts
```diff
@@ -0,0 +1,66 @@
...
+
+  const revokedFilter = accessControl.filters.RoleRevoked(role)
+
+  const logsRevoked = await provider.getLogs({
+    ...revokedFilter,
+    fromBlock,
+  })
+
+  const revoked = logsRevoked
+    .filter((log) => log.topics[1] === role)
```
adamiak, line 47: Do we need to filter here if `revokedFilter` already returns events for the single `role` we want?

packages/update-monitor/src/common/accessControl.ts
```diff
@@ -0,0 +1,66 @@
...
+
+  const grantedFilter = accessControl.filters.RoleGranted(role)
+
+  const logsGranted = await provider.getLogs({
+    ...grantedFilter,
+    fromBlock,
+  })
+
+  const granted = logsGranted
+    .filter((log) => log.topics[1] === role)
```
adamiak, line 36: Do we need to filter here if `grantedFilter` already returns events for the single `role` we want?

antooni: yeah thanks, i didnt remove it after refactoring

packages/update-monitor/src/common/accessControl.ts
antooni: good idea, thanks, I will leave the inline function for now, maybe in the future it would be a good idea to make a pure function out of it for easier testing

packages/update-monitor/src/projects/synapse/contracts/timelock.ts
antooni: thanks, I missed this one after refactoring

adamiak (approved):

packages/update-monitor/src/projects/synapse/contracts/synapseBridge.ts
```diff
@@ -0,0 +1,58 @@
...
+    values: {
+      ADMIN_ROLE,
+      admins,
+      adminsCount,
+      GOVERNANCE_ROLE,
+      governors,
+      governorsCount,
+      NODEGROUP_ROLE,
+      nodes,
+      nodesCount,
```
sz-piotr, line 55: Why only first admin, first governor and first node? Just return arrays like in timelock

sz-piotr (changes requested):

sz-piotr (approved):

luke7211 (changes requested):
- add support for gOHM - it has massive TVL on that bridge
- add support for AMMs - they should also add to TVL
- list AMMs in the contract list

luke7211 (approved):

PR #813, opened 2022-10-11
Api activity improvements

michalsidzej:
After creating a view for blocks, zksync table seems to be the bottleneck. This PR reduces local /api/activity response time (with backend sync on) from ~2s to ~400ms.

### DIFF
--- a/packages/backend/src/peripherals/database/ZksyncTransactionRepository.ts
+++ b/packages/backend/src/peripherals/database/ZksyncTransactionRepository.ts
@@ -113,25 +113,20 @@ export class ZksyncTransactionRepository extends BaseRepository {
     return _.zip(noNextBlockNumbers, noPrevBlockNumbers) as [number, number][]
   }
 
-  async getDailyTransactionCount(maxTimestamp: UnixTime) {
+  async refresh() {
     const knex = await this.knex()
-    const { rows } = (await knex.raw(
-      `
-      SELECT
-        date_trunc('day', unix_timestamp) AS unix_timestamp,
-        count(*) as count
-      FROM
-        transactions.zksync
-      WHERE
-        unix_timestamp < ?
-      GROUP BY
-        date_trunc('day', unix_timestamp)
-      ORDER BY unix_timestamp
-    `,
-      maxTimestamp.toDate(),
-    )) as unknown as {
-      rows: { unix_timestamp: Date; count: string }[]
-    }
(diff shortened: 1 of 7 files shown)


packages/backend/src/peripherals/database/ZksyncTransactionRepository.ts
```diff
@@ -113,25 +113,20 @@ export class ZksyncTransactionRepository extends BaseRepository {
     return _.zip(noNextBlockNumbers, noPrevBlockNumbers) as [number, number][]
   }
 
-  async getDailyTransactionCount(maxTimestamp: UnixTime) {
+  async refresh() {
```
sin6pi7, line 116: Please rename these `refresh` methods to something more specific e.g. `refreshDailyTransacionCount()` - this applies to both repositories.

sin6pi7 (changes requested):

packages/backend/src/peripherals/database/ZksyncTransactionRepository.ts
michalsidzej: Done

packages/backend/src/peripherals/database/ZksyncTransactionRepository.ts
```diff
@@ -113,25 +113,20 @@ export class ZksyncTransactionRepository extends BaseRepository {
...
+
+  async getDailyTransactionCount(
+    maxTimestamp: UnixTime,
+  ): Promise<{ timestamp: UnixTime; count: number }[]> {
+    const knex = await this.knex()
+    const rows = (await knex
+      .withSchema('transactions')
+      .from('zksync_count_view')
+      .where('unix_timestamp', '<', maxTimestamp.toDate())
+      .orderBy('unix_timestamp')) as { unix_timestamp: Date; count: string }[]
```
sz-piotr, line 129: Why not use types? Also the row type for the other view is named `...View` instead of `...ViewRow`

sz-piotr (changes requested):

packages/backend/src/peripherals/database/ZksyncTransactionRepository.ts
michalsidzej: Fixed types

sz-piotr (approved):

sin6pi7 (approved):

PR #820, opened 2022-10-12
Add StarkNet transaction count updater

sin6pi7:
`StarkNetClient` uses `starknet` package to talk to StarkNet infrastructure. Decided to go with this one, because of implementation simplicity. The other way is to query rpc nodes on our own. Wdyt @michalsidzej @sz-piotr ?

Resolves L2B-235


packages/backend/src/peripherals/StarknetClient.ts
```diff
@@ -0,0 +1,35 @@
...
+      const rateLimiter = new RateLimiter({
+        callsPerMinute: opts.callsPerMinute,
+      })
+      this.getBlockNumber = rateLimiter.wrap(this.getBlockNumber.bind(this))
+      this.getBlock = rateLimiter.wrap(this.getBlock.bind(this))
+    }
+  }
+
+  async getBlockNumber() {
+    const result = await this.provider.getBlock('latest')
```
sin6pi7, line 23: The other way to go would be to create an instance of `RpcProvider` and use `getBlockNumber` method (ref: https://www.starknetjs.com/docs/API/provider#methods-2) - not sure what are the differences, so I decided to choose a simpler route. Please let me know your thoughts @michalsidzej @sz-piotr

packages/backend/package.json
```diff
@@ -39,11 +39,14 @@
...
+    "rollbar": "^2.25.2",
     "source-map-support": "^0.5.21",
-    "zod": "^3.17.3",
-    "rollbar": "^2.25.2"
+    "starknet": "^4.4.2",
+    "zod": "^3.17.3"
   },
   "devDependencies": {
+    "@types/bn.js": "^5.1.1",
+    "@types/elliptic": "^6.4.14",
```
sin6pi7, line 49: Had to add these two, because it was not automatically added when installing `starknet`.

packages/backend/src/setup/createRpcTransactionUpdaters.ts
```diff
@@ -83,3 +84,24 @@ export function createEthereumTransactionUpdater(
...
   return updater
 }
+
+export function createStarkNetTransactionUpdater(
+  config: TransactionCountSyncConfig,
+  blockTransactionCountRepository: BlockTransactionCountRepository,
+  clock: Clock,
+  logger: Logger,
+) {
+  const client = new StarkNetClient({ callsPerMinute: 60 * 5 })
```
michalsidzej, line 94: Shouldn't callsPerMinute come from config?

michalsidzej: Have you tested how many calls per minute can the provider handle?

packages/backend/src/peripherals/StarknetClient.ts
michalsidzej: It is interesting that the provider doesn't have this method.

packages/backend/src/setup/createRpcTransactionUpdaters.ts
sin6pi7: Yes, this value is causing `Too Many Requests` error sometimes, but it keeps on going.

PR #822, opened 2022-10-13
Update Orbit bridge description

adamiak:
Resolves L2B-204


packages/update-monitor/src/projects/orbitBridge/contracts/governance.ts
```diff
@@ -0,0 +1,28 @@
...
+    provider,
+  )
+
+  return {
+    name: 'Bridge with Governance',
+    address: addresses.bridge,
+    upgradeability: {
+      type: 'proxy',
+      implementation: await getEip897Implementation(provider, addresses.bridge),
+    },
```
sz-piotr, line 22: I have checked and this bridge does not follow https://eips.ethereum.org/EIPS/eip-897, because it does not have the `proxyType()` function. Maybe `Eip897` should be renamed to `Generic` in our code?

packages/update-monitor/src/common/eip897.ts
```diff
@@ -0,0 +1,32 @@
...
+    })
+    return bytes32ToAddress(value)
+  } catch (e) {
+    if (
+      e instanceof Error &&
+      e.message.includes('Transaction reverted without a reason string')
+    ) {
+      // This is a hacky way of noticing that contract doesn't have "implementation()"
+      // function. The proper way would be to check ABI, but in current function flow
+      // we don't have access to ABI yet.
```
sz-piotr, line 26: I disagree. Calling the contract is the proper way :)

packages/update-monitor/src/common/address.ts
```diff
@@ -1,5 +1,7 @@
-import { utils } from 'ethers'
+import { constants, utils } from 'ethers'
 
 export function bytes32ToAddress(bytes32: string) {
-  return utils.getAddress('0x' + bytes32.slice(-40))
+  return bytes32 === '0x'
+    ? constants.AddressZero
```
sz-piotr, line 5: I think that this will lead to problems down the line. If you call an EOA then `0x` will always be a return value which will silently return `0x00000...` as a valid address instead of throwing and alerting us.

sz-piotr (changes requested):

packages/update-monitor/src/projects/orbitBridge/contracts/governance.ts
```diff
@@ -0,0 +1,28 @@
...
+    values: {
+      validators: await bridgeProxy.getOwners(),
+      threshold: (await bridgeProxy.required()).toNumber(),
```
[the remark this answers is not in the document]
adamiak: Thanks, resolved!

packages/update-monitor/src/common/eip897.ts
adamiak: Ok, comment removed :)

packages/update-monitor/src/projects/orbitBridge/contracts/governance.ts
adamiak: I saw that but also noticed that Etherscan calls such proxies (without `proxyType()`) EIP-897 anyway. Of about 5 random EIP-897 proxies on Etherscan I checked, only one had `proxyType()`. So I reluctantly decided to check only for `implementation()` in my code, although it didn't feel quite right. But, since you noticed this as well, I agree we should stick to the EIP requirements. I have updated the code to check for `proxyType()` too.

As for the Orbit Bridge contract it actually gets even more challenging than that. It has a lot of functions already and calls the proxy contract only as a fallback. So it's more of an "extension" mechanism rather than full proxy.

packages/update-monitor/src/common/address.ts
adamiak: Makes sense, reverted. I moved the check to the specific function that needs it.

luke7211 (changes requested):
- add ORC token
- add warning about farms
- add warning about farms contract not being verified (e.g. 0x378F1CD69e1012cfe8FbeAfFeC02630190fda4d9)
- add major farms contracts, see how much value is managed there, research if it should be added to TVL
- add TODO to track automatically farms in the future

sz-piotr (approved):

PR #824, opened 2022-10-14
Show pretty tvl-activity toggle

michalsidzej:
Resolves L2B-191

PR #825, opened 2022-10-14
Restore activity time controls

michalsidzej:
Requires https://github.com/l2beat/l2beat/pull/824
Resolves L2B-236

### DIFF
--- a/packages/frontend/src/components/chart/RangeControls.tsx
+++ b/packages/frontend/src/components/chart/RangeControls.tsx
@@ -2,19 +2,26 @@ import React from 'react'
 
 import { RadioGroup } from './RadioGroup'
 
-export function RangeControls({ days }: { days: number }) {
+export function RangeControls({
+  days,
+  type,
+}: {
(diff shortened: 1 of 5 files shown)


packages/frontend/src/components/chart/RangeControls.tsx
```diff
@@ -2,19 +2,26 @@ import React from 'react'
...
+  const options = [
+    { value: '30D', checked: days === 30 },
+    { value: '90D', className: '!hidden sm:!block' },
+    { value: '180D', className: '!hidden sm:!block' },
+    { value: '1Y' },
+    { value: 'MAX' },
+  ]
+
+  if (type !== 'activity') {
+    options.unshift({ value: '7D', checked: days === 7 })
```
sz-piotr, line 21: I would add hidden instead

sz-piotr (changes requested):

sz-piotr (approved):

PR #826, opened 2022-10-14
Show proper scaling factor description

michalsidzej:
Resolves L2B-239
Screenshot 2022-10-17 at 16 19 19
Screenshot 2022-10-17 at 16 19 03


packages/frontend/src/components/header/ActivityHeader.tsx
```diff
@@ -20,8 +23,25 @@ export function ActivityHeader(props: ActivityHeaderProps) {
...
+        <p
+          className={cx(
+            'text-gray-500 dark:text-gray-600 text-xs md:text-base text-right w-full md:w-auto',
+            'flex items-center gap-1.5',
+          )}
+        >
+          Observed over the last 7 days
+          <span
+            className="Tooltip inline-block"
+            title="More txs on Ethereum because of L2s. Formula: (L2 txs + ETH txs) / ETH txs"
```
sz-piotr, line 36:
```suggestion
            title="How many more transactions are settled by Ethereum if we take into account the L2 solutions. Formula: (L2 txs/7d + ETH txs/7d) / ETH txs/7d"
```

michalsidzej: Changed the tooltip so it looks like this: 
Screenshot 2022-10-18 at 09 06 16

sz-piotr (approved):

PR #829, opened 2022-10-15
Refactor chart logic based on elm architecture

sz-piotr:
Left to do:
- "More" tokens button
- flesh out initial message
- alternative tvl endpoint


michalsidzej (commented):
Seems like there's some bug with the axis labels and Eth Mainnet Transactions display
Screenshot 2022-10-17 at 08 46 11
2b6-d919fd229729.png">

michalsidzej:
You're using the old `compatibleApi` endpoints to fetch the chart data. This endpoint is exposing daily TPS data, not daily transactions. That's where the bug comes from.

packages/frontend/src/components/chart/configure/controls/onMouseMove.ts
```diff
@@ -0,0 +1,32 @@
...
+  let wasInside = false
+
+  window.addEventListener('mousemove', (e) => onEvent(e))
+  window.addEventListener('touchstart', (e) => onEvent(e.touches[0]))
+  window.addEventListener('touchmove', (e) => onEvent(e.touches[0]))
+
+  function onEvent(e: MouseEvent | Touch) {
+    const rect = view.getBoundingClientRect()
+    const isInside =
+      e.clientX >= rect.left - 16 &&
```
sin6pi7, line 16: `16` seems pretty arbitrary - what's the rationale behind it?

packages/frontend/src/components/chart/configure/effects/handleEffect.ts
```diff
@@ -0,0 +1,106 @@
...
+    // eslint-disable-next-line @typescript-eslint/no-unsafe-assignment
+    (data) => ({ type: 'ActivityLoaded', requestId, data }),
+    () => ({ type: 'ActivityFailed', requestId }),
+  )
+}
+
+function fetchThenDispatch(
+  url: string,
+  dispatch: (message: Message) => void,
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
```
sin6pi7, line 85: Not necessary now, but would be nice to not have this - maybe a zod(-like) parser that makes sure types are correct?

packages/frontend/src/components/chart/configure/render/render.ts
```diff
@@ -0,0 +1,81 @@
...
+    elements.view.tokenControlsWrapper?.classList.toggle('hidden', !isTvl)
+    elements.controls.showEthereum?.parentElement?.classList.toggle(
+      'hidden',
+      isTvl,
+    )
+  }
+
+  if (state.view !== previousState.view) {
+    const labels = elements.view.labels
+    for (let i = 0; i < 5; i++) {
```
sin6pi7, line 57: Let's give `5` here a name.

packages/frontend/src/components/chart/configure/render/render.ts
```diff
@@ -0,0 +1,81 @@
...
+import { State } from '../state/State'
+import { renderChart } from './renderChart'
+import { renderHover } from './renderHover'
+
+export function render(
+  elements: ChartElements,
+  previousState: State,
+  state: State,
+) {
+  if (previousState === state) {
```
sin6pi7, line 12: Is ref comparison here intentional?

sin6pi7 (approved):
Awesome work. No red flags, only syntactic changes and future todos from my side.

packages/frontend/src/components/chart/configure/render/render.ts
sz-piotr: Yes. There is a super rare edge case where multiple updates happen in the same animation frame and the previous render has already drawn the latest state, so we can skip the work. Potentially we can just remove this check

sin6pi7: Sounds good, let's keep it then.

packages/frontend/src/components/chart/configure/effects/handleEffect.ts
sz-piotr: I agree that it would be a nice to have in the future, but this PR is already large enough. I can add a task to the backlog

packages/frontend/src/components/chart/configure/controls/onMouseMove.ts
sz-piotr: It is totally arbitrary. This basically solves for the fact that the first and last value of the chart have 2x less hover space, because every other value has some space to the left and to the right but the edge values only have one of those. So this 16px is compensation.

I can extract it to some named constant.

sin6pi7: I see, let's give it a name and put a comment with your explanation 👍

packages/frontend/src/components/chart/configure/render/render.ts
sz-piotr: I've added a comment explaining it too :)

michalsidzej (approved):

PR #830, opened 2022-10-16
Sollet bridge

adamiak:
Resolves L2B-208


luke7211 (changes requested):
- fix big discrepancy in TVL (SRM token)
- add warning

luke7211 (approved):

PR #838, opened 2022-10-17
Add Allbridge

luke7211:


packages/config/src/bridges/allbridge.ts
```diff
@@ -0,0 +1,64 @@
...
+  riskView: {
+    validatedBy: {
+      value: 'Third Party',
+      description: '???',
+      sentiment: 'bad',
+    },
+    sourceUpgradeability: {
+      value: '???',
+      description: '', //TODO: fill
+      sentiment: 'bad',
```
adamiak, line 61: I suggest to remove `sourceUpgradeability` so that "No info" displays, rather than red "???".

PR #840, opened 2022-10-17
Activity fully synced logic

sin6pi7:
This PR makes sure that we only return fully synced data on the `/api/activity` endpoint. This means:
- for all non-starkex projects we track a tip i.e. first row in the database that does not have a following row (i.e. with distance of 1 day|block to the previous row),
- materialized views include only rows with `timestamp <= tip.timestamp.toStartOf('day')`,
- since updaters run asynchronously every project can have different tip - when building `/api/activity` response we find the smallest tip of all projects and return only data until that tip.

Resolves L2B-262


packages/backend/src/api/controllers/ActivityController.ts
```diff
@@ -22,46 +23,62 @@ export class ActivityController {
...
+    const fullySyncedEthereumCounts = this.limitCounts(ethereumCounts, tip)
 
     return {
-      combined: this.toCombinedActivity(layer2Counts),
-      projects: this.toProjectsActivity(layer2Counts),
-      ethereum: this.countsToChart(ethereumCounts),
+      combined: this.toCombinedActivity(fullySyncedLayer2s),
+      projects: this.toProjectsActivity(fullySyncedLayer2s),
+      ethereum: this.countsToChart(fullySyncedEthereumCounts),
     }
```
sz-piotr, line 41: All of this, including the private functions is pure. It would be easier to write and test if it was extracted outside of the class.

Also the tests for this are really lacking, once its extracted it would be nice to see that the tip is calculated correctly for example.

packages/backend/src/peripherals/database/ZksyncTransactionRepository.ts
```diff
@@ -25,39 +28,60 @@ export class ZksyncTransactionRepository extends BaseRepository {
     super(database, logger)
 
     /* eslint-disable @typescript-eslint/unbound-method */
-
-    this.add = this.wrapAdd(this.add)
     this.addMany = this.wrapAddMany(this.addMany)
     this.deleteAll = this.wrapDelete(this.deleteAll)
```
sz-piotr, line 32: What about the new methods? Shouldn't they be wrapped?

packages/backend/src/peripherals/database/migrations/035_tipped_transaction_count_view_zksync.ts
```diff
@@ -0,0 +1,38 @@
...
+export async function up(knex: Knex) {
+  await oldView.down(knex)
+  await knex.schema.raw(
+    `
+  CREATE MATERIALIZED VIEW transactions.zksync_count_view AS
+    SELECT
+      date_trunc('day', zksync.unix_timestamp) unix_timestamp,
+      count(*) count
+    FROM transactions.zksync zksync
+    INNER JOIN transactions.block_tip tip ON tip.project_id = '${ProjectId.ZKSYNC.toString()}'
```
sz-piotr, line 28: Please use bindings

packages/backend/src/peripherals/database/BlockTransactionCountRepository.ts
```diff
@@ -50,15 +57,52 @@ export class BlockTransactionCountRepository extends BaseRepository {
...
   }
 
+  async refreshProjectTip(projectId: ProjectId) {
+    const knex = await this.knex()
+    const currentTip = await this.getTipByProject(projectId)
+    const tipNumber =
+      (await this.getFirstBlockNumberWithoutNext(
+        projectId,
+        currentTip?.block_number,
+      )) ?? (await this.getMaxBlockNumber(projectId))
```
sz-piotr, line 67: What is the edge case for this.getMaxBlockNumber? When will this.getFirstBlockNumberWithoutNext return undefined but this.getMaxBlockNumber not?

packages/backend/src/peripherals/database/StarkexTransactionCountRepository.ts
```diff
@@ -128,6 +136,47 @@ export class StarkexTransactionCountRepository extends BaseRepository {
     const knex = await this.knex()
     return await knex('transactions.starkex').delete()
   }
+
+  async getTipTimestamp(projectId: ProjectId) {
```
sz-piotr, line 140: Why is there no tip refresh here?

sz-piotr (changes requested):

packages/backend/src/peripherals/database/migrations/035_tipped_transaction_count_view_zksync.ts
sin6pi7: I wanted to use that, but `knex.schema.raw` does not accept bindings. Now that I checked it's just types (https://github.com/knex/knex/issues/4069) - will update

packages/backend/src/peripherals/database/BlockTransactionCountRepository.ts
sin6pi7: `getFirstBlockNumberWithoutNext` returns first row without next. If there are no gaps then there will be no such row, because `max` is the tip and hence we return `max`.

packages/backend/src/peripherals/database/BlockTransactionCountRepository.ts
```diff
@@ -50,15 +57,52 @@ export class BlockTransactionCountRepository extends BaseRepository {
...
+
+    if (!tipNumber) {
+      await knex('transactions.block_tip')
+        .delete()
+        .where('project_id', projectId.toString())
+      return undefined
+    } else {
+      const tip = await knex('transactions.block')
```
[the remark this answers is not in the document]
sin6pi7: I implemented that in the beginning, but it resulted in more code, especially types. I can approach this once again and simplify it as much as possible.

sz-piotr (approved):
How to improve this PR:

## 1. Remove code duplication

Before: (same code is in Loopring, Rpc, Starkex and Zksync)
```ts
const [boundaries, gaps, latestBlock] = await Promise.all([
  this.blockTransactionCountRepository.findBoundariesByProject(
    this.projectId,
  ),
  this.blockTransactionCountRepository.getGapsByProject(this.projectId),
  this.loopringClient.getFinalizedBlockNumber(),
])
this.latestBlock = latestBlock

if (!boundaries) {
  gaps.push([1, this.latestBlock])
} else {
  if (boundaries.max < this.latestBlock) {
    gaps.push([boundaries.max + 1, this.latestBlock])
  }
  if (boundaries.min > 1) {
    gaps.push([1, boundaries.min - 1])
  }
}

for (const [start, end] of gaps) {
  for (let i = start; i <= end; i++) {
    this.blockQueue.addToBack(i)
  }
}
```

After:
```ts
const latestBlock = await this.loopringClient.getFinalizedBlockNumber()
const gaps = await this.blockTransactionCountRepository.getGapsByProject(
  this.projectId,
  latestBlock
)

for (const [start, end] of gaps) {
  for (let i = start; i <= end; i++) {
    this.blockQueue.addToBack(i)
  }
}
```

This is achieved by combining the functionality from `findBoundariesByProject` and `getGapsByProject` together with the boundary edge case logic into one function.

I also expect the common logic to be refactored into a simple pure function and used throughout the 3 repositories.

## 2. Simplify tip updates

Currently the tip is used for 3 things
1. Optimizing the gaps query
2. Setting the upper bound for the materialized view
3. Showing up in status

The logic behind tip updates is as follows:
1. Updated every time `getGapsByProject` is called
2. Updated every time `refreshDailyCounts` is called

Updating the tip also requires selecting the current tip, calculating a new one and then doing a weird remove / update combo. In Rpc it's complicated, in Zksync its simpler but should be wrapped in a transaction!

The calculation of the new tip also duplicates the SQL from the gaps query.

I propose to simplify this radically, by doing the following:
1. Show gaps instead of tip in status
2. Set the tip based on the result from gaps query instead of having a separate query
3. Update the tip like in Zksync, but wrap in a transaction

sz-piotr (changes requested):
I pressed approve and wanted to request changes

sz-piotr (approved):

PR #845, opened 2022-10-18
New Project Details header

michalsidzej:
Resolves L2B-237

Percent change arrows from the design will be a part of [L2B-268](https://linear.app/l2beat/issue/L2B-268/add-arrows-to-procent-change)

Happy to hear some advice about the bridge version

Screenshot 2022-10-18 at 14 24 05
Screenshot 2022-10-18 at 14 23 51
Screenshot 2022-10-18 at 14 41 23
Screenshot 2022-10-18 at 14 56 42
Screenshot 2022-10-18 at 14 57 01
Screenshot 2022-10-18 at 14 57 16


packages/frontend/src/utils/tvl/getTvlWitchChange.ts
```diff
@@ -0,0 +1,16 @@
+import { Bridge, Layer2 } from '@l2beat/config'
+import { TvlApiResponse } from '@l2beat/types'
+
+import { getPercentageChange } from '../utils'
+
+export function getTvlWithChange(
```
sz-piotr, line 6: You should probably use this on tvl pages as well, because now we have duplicated functionality

packages/frontend/src/utils/activity/getTransactionMonthlyCount.ts
```diff
@@ -0,0 +1,11 @@
+import { ActivityApiChartPoint } from '@l2beat/types'
+
+export function getTransactionMonthlyCount(
```
sz-piotr, line 3: I thought we already had this?

sz-piotr (changes requested):

packages/frontend/src/utils/activity/getTransactionMonthlyCount.ts
michalsidzej: Renamed the file but forgot to delete the old one. Now updated, so `getTransactionCount` can count both monthly and weekly

packages/frontend/src/pages/scaling-projects/props/getProjectHeader.ts
```diff
@@ -0,0 +1,55 @@
...
+    case 'Immutable X':
+      return 'long'
+    case 'OMG Network':
+    case 'Layer2.Finance':
+    case 'ZKSwap V2':
+    case 'Polygon Hermez':
+    case 'Metis Andromeda':
+      return 'very-long'
+  }
+}
```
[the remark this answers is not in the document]
michalsidzej: It is duplicated with the file getProjectHeader which should be deleted after we turn on the activity feature flag. This will not be used anywhere else, so IMO it makes sense to keep this function inside this file. WYT?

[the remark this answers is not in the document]
sz-piotr: Ok :)

sz-piotr (approved):

PR #846, opened 2022-10-18
Add Celer

antooni:


packages/update-monitor/README.md
```diff
@@ -17,7 +17,9 @@ To run please use `yarn start`. Display help with `yarn start --help`.
 
 3. run `yarn start discover [project]`
 
-4. Discovery Engine will generate file with values from ABI calls inside `dist/discovery` and ABI per-contract in the folder `dist/discovery/abi/[contractAddress].json`
+   - \*if your script is stuck at some method there probably is a need to skip it by providing configuration parameter `skipMethods`, see example [arbitrum](src/projects/arbitrum/index.ts)
```
sz-piotr, line 20:
```suggestion
   - if your script is stuck at some method there probably is a need to skip it by providing configuration parameter `skipMethods`, see example [arbitrum](src/projects/arbitrum/index.ts)
```

packages/update-monitor/src/common/utils/getMembers.ts
```diff
@@ -0,0 +1,12 @@
+export function getMembers(granted: string[], revoked: string[]): string[] {
+  const members: string[] = [...granted]
+
+  revoked.forEach((revokedAddress) => {
+    const index = members.indexOf(revokedAddress)
+    if (index !== -1) {
+      members.splice(index, 1)
+    }
+  })
```
sz-piotr, line 9: I like this design for its simplicity, but unfortunately it fails for a very weird and uncommon edge case:

- Add Joe
- Remove Joe
- Remove Joe (again, because the smart contract just sets false and emits)
- Add Joe

This function should return `['Joe']` but will return `[]`

I think that block numbers should be taken into account

sz-piotr (changes requested):

packages/update-monitor/src/common/utils/getMembers.ts
antooni: I thought about it, there is also logic inside smart contract to prevent such case. Maybe the more robust method can be coded in the next PRs? because this one handles our case without a problem, WDYT

luke7211 (changes requested):
Consider adding WXT. WireX is one of the better known crypto cards, they made a big announcement on their integration with cBridge: https://wirexapp.com/blog/post/partnering-with-celer-to-bridge-wxt-between-networks 

Also it is worth mentioning Open Canonical Token Starndard that cBridge/Celer is trying to promote

packages/update-monitor/src/common/utils/getMembers.ts
antooni: @sz-piotr

sz-piotr:
@antooni I talked with @luke7211 and we decided to not include WXT for now, but it would be great if you could reach out to Wirex and ask them about the coingecko situation and the marketcap mismatch.

sz-piotr (approved):

PR #847, opened 2022-10-18
Handle too early activity blocks

sin6pi7:
```
l2beat-staging::DATABASE=> select * from transactions.block where unix_timestamp = '1970-01-01 00:00:00+00';
 project_id  |     unix_timestamp     | count | block_number 
-------------+------------------------+-------+--------------
 arbitrum    | 1970-01-01 00:00:00+00 |     0 |            0
 nova        | 1970-01-01 00:00:00+00 |     0 |            0
 metis       | 1970-01-01 00:00:00+00 |     0 |            0
 optimism    | 1970-01-01 00:00:00+00 |     0 |            0
 bobanetwork | 1970-01-01 00:00:00+00 |     0 |            0
(5 rows)
```
Resolves L2B-267


sz-piotr (approved):
I am approving this on the condition that you have checked and each removed block had 0 transactions

PR #850, opened 2022-10-19
Add Eth chart line icon

michalsidzej:
Screenshot 2022-10-19 at 07 41 55


packages/frontend/src/components/icons/chart/EthereumLineIcon.tsx
```diff
@@ -0,0 +1,15 @@
...
+import { Icon } from '../Icon'
+
+export function EthereumLineIcon(props: SVGAttributes<SVGElement>) {
+  return (
+    <Icon
+      aria-label="Ethereum line icon"
+      viewBox="0 0 10 6"
+      className={props.className}
+    >
+      <rect width="10" height="6" fill="#53A2FF" />
```
sz-piotr, line 12: we can specify the fill using a tailwind class like `fill-blue-???`

sz-piotr (changes requested):

sin6pi7 (approved):

sz-piotr (approved):

PR #858, opened 2022-10-20
Refactor proxy detection and unify code

sz-piotr:
This PR updates discovery in update-monitor in such a way that it uses the same code for proxies as analysis. It also streamlines the process by making proxy detection do a bit more work reducing the need to do a large `if` over the upgradeability discriminated union.


antooni (changes requested):
`verify` function does not work in arbitrum, nova and zksync

antooni (approved):

PR #860, opened 2022-10-23
Polynetwork bridge

adamiak:
Resolves L2B-207

Polynetwork Bridge uses a few proxies with unverified source code, which makes update-monitor not able to find implementation contracts. It makes update-monitor code a bit more complex than usual.


packages/update-monitor/src/projects/polynetworkBridge/index.ts
```diff
@@ -0,0 +1,72 @@
...
+  provider: providers.Provider,
+): Promise<string> {
+  // Since `managerProxyContract` is not verified on Etherscan,
+  // discovery won't find the manager address itself. We get it manually:
+  const proxyContract = bytes32ToAddress(
+    await provider.call({ to: lockProxy, data: '0xd798f881' }), // $ cast sig "managerProxyContract()"
+  )
+  return bytes32ToAddress(
+    await provider.call({ to: proxyContract, data: '0x87939a7f' }), // $ cast sig "getEthCrossChainManager()"
+  )
```
sz-piotr, line 24: use getCallResult

packages/update-monitor/src/discovery/DiscoveryEngine.ts
```diff
@@ -9,7 +9,7 @@ import { DiscoveryOptions } from './DiscoveryOptions'
 
 export class DiscoveryEngine {
   constructor(
-    private readonly provider: providers.Provider,
+    readonly provider: providers.Provider,
```
sz-piotr, line 12: Why?

sz-piotr (changes requested):

luke7211 (changes requested):
Poly is AMB that is used by more than one bridge. We need to identify these bridges, at least the biggest ones and list them separately, possibly with different PRs

packages/update-monitor/src/projects/polynetworkBridge/index.ts
adamiak: 👍

packages/update-monitor/src/discovery/DiscoveryEngine.ts
adamiak: Related to your previous comment. During discovery(!), we need `provider` to call a contract that is not verified (a proxy that returns implementation, so we need to get it somehow). That's why I had to expose provider from DiscoverEngine which is the only object discover process receives.

luke7211 (approved):

sz-piotr (approved):

PR #867, opened 2022-10-25
Fix activity db queries

sin6pi7:
For some reason heroku postgres does not like simple (without `as`) aliasing for activity queries near or around `next` alias. I unified it to use `as` in all these queries.


michalsidzej (approved):
Interesting

sz-piotr (approved):

PR #871, opened 2022-10-25
Get VerificationStatus

antooni:
This PR aims to add method able to read `verified.json` file contents. Method has the parameter `isEnabled` for the later feature flag support. Additionally I added new zod type to parse the file contents.

Resolves L2B-285

### DIFF
--- a/packages/frontend/src/build/getVerificationStatus.ts
+++ b/packages/frontend/src/build/getVerificationStatus.ts
@@ -4,14 +4,7 @@ import fs from 'fs'
 const PATH = '../config/build/verified.json'
 
 export function getVerificationStatus(isEnabled: boolean): VerificationStatus {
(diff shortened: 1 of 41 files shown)


packages/frontend/src/build/getVerificationStatus.ts
```diff
@@ -0,0 +1,24 @@
+import { VerificationStatus } from '@l2beat/types'
+import fs from 'fs'
+
+const PATH = '../config/build/verified.json'
+
+export function getVerificationStatus(isEnabled: boolean): VerificationStatus {
+  if (!isEnabled) {
```
adamiak, line 7:
```suggestion
  if (!isEnabled || !fs.existsSync(PATH)) {
```

sz-piotr (approved):

PR #873, opened 2022-10-25
Find unverified contracts

adamiak:
Resolves L2B-279


packages/config/scripts/checkVerifiedContracts/tasks.ts
```diff
@@ -0,0 +1,54 @@
...
+  logger.info(`Processing ${addresses.length} addresses.`)
+  const result: VerificationMap = {}
+  let errorOccured = false
+
+  // Copy previously verified contracts directly to result
+  addresses
+    .filter((a) => previouslyVerified.has(a))
+    .forEach((a) => (result[a.toString()] = true))
+
+  const taskQueue = new TaskQueue(
```
krzkaczor, line 23: IMO this could be extracted from here to something like PromiseAllThrotthled(?) but with a limited parallelism. It could be just a separate function in this file for now but we also could move it to a @common lib.

packages/config/scripts/checkVerifiedContracts/output.ts
```diff
@@ -0,0 +1,58 @@
...
+}
+
+export async function saveResult(
+  filePath: string,
+  addressVerificationMap: VerificationMap,
+  projectVerificationMap: VerificationMap,
+) {
+  // We need to have a deterministic order of addresses to avoid unnecessary git diff changes in the file.
+  // There's no simple way to do that with JSON.stringify for nested map, so we sort each internal map.
+  const output = [
```
krzkaczor, line 40: this is so ugly ;d @sz-piotr any ideas?

krzkaczor (changes requested):
linter fails

packages/config/scripts/checkVerifiedContracts/output.ts
```diff
@@ -0,0 +1,58 @@
...
+    const data = await readFile(filePath, 'utf-8')
+    const parsed = FileStructure.parse(JSON.parse(data))
+    const addresses = Object.keys(parsed.contracts)
+    const result = new Set(
+      addresses.filter((a) => parsed.contracts[a]).map(EthereumAddress),
+    )
+    console.log(`Loaded ${result.size} previously verified contracts.`)
+    return result
+  } catch (e) {
+    console.log('Unable to load previously verified contracts.')
```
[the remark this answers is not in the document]
adamiak: thx, good point. We need to support file not existing, but we should crash on anything else. I'll update the code.

packages/config/scripts/checkVerifiedContracts/tasks.ts
adamiak: I tried to do that, but taskQueue has different semantics than Promise.all, because taskQueue is fed *data*, while Promise.all is fed *promises*. I'd have to implement something like `PromiseAllThrottled()` from scratch, so I decided to stick to what we already use.

Saying that, I have just refactored this call to a separate function - it does look cleaner.

packages/config/scripts/checkVerifiedContracts/output.ts
adamiak: I stopped short from importing js lib that does that. Given that this is a temporary solution I kept it as it is now.

Saying that, I have just refactored it a bit to look cleaner.

krzkaczor (approved):

PR #880, opened 2022-10-26
Check if transaction counters work properly

sin6pi7:
Resolves L2B-264


packages/backend/src/core/transaction-count/MaterializedViewRefresher.ts
```diff
@@ -33,9 +36,69 @@ export class MaterializedViewRefresher {
...
     )
     await this.zksyncTransactionRepository.refreshDailyCounts()
     this.logger.info('Refresh finished')
+    await this.checkIfFullySynced()
+  }
+
+  private async checkIfFullySynced() {
+    const lastHour = this.clock.getLastHour()
+    const startOfDay = lastHour.toStartOf('day')
+    const timeToCheck = lastHour.equals(
```
michalsidzej, line 49: Shouldn't it be gte? 
```suggestion
    const timeToCheck = lastHour.gte(
```

sz-piotr (approved):

michalsidzej (approved):

PR #887, opened 2022-10-27
Prepare backend activity production config

sin6pi7:
Environment variables have been copied over from staging api. Performance values have been cut in half to enable both staging and production to run using the same keys. This is feasible since we're already fully synced and are only catching up on the tip.

Resolves L2B-291

### DIFF
--- a/packages/backend/src/config/config.staging.ts
+++ b/packages/backend/src/config/config.staging.ts
@@ -2,7 +2,6 @@ import { Knex } from 'knex'
 
 import { Config } from './Config'
 import { getProductionConfig } from './config.production'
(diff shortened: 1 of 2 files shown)


packages/backend/src/config/config.staging.ts
```diff
@@ -16,32 +15,5 @@ export function getStagingConfig(): Config {
       application_name: name,
     },
     tvlReportSync: true,
-    transactionCountSync: {
```
michalsidzej, line 19: Wouldn't that change turn off activity sync on staging?

michalsidzej (approved):

PR #890, opened 2022-10-28
Meta images sanity check

michalsidzej:
Resolves L2B-294
In the end, I decided to check if a meta image has the L2beat logo in it. This will give us the most confidence that the image is right.


packages/frontend/src/build/buildMetaImages.ts
```diff
@@ -4,12 +4,18 @@ import { mkdirpSync } from 'fs-extra'
...
 
+import { getConfig } from './config'
+
 main().catch((e) => {
   console.error(e)
   process.exit(1)
 })
 
 async function main() {
+  const env = process.env.DEPLOYMENT_ENV ?? 'production'
```
krzkaczor, line 15: we already have this in the buildPages, do we need it here too?

packages/frontend/src/build/buildMetaImages.ts
```diff
@@ -47,3 +60,11 @@ async function main() {
     server.close((err) => (err ? reject(err) : resolve())),
   )
 }
+
+async function sanityCheck(page: puppeteer.Page, slug: string) {
```
krzkaczor, line 64: please, provide explcit return type here.

packages/frontend/src/build/buildMetaImages.ts
```diff
@@ -47,3 +60,11 @@ async function main() {
...
+  const l2BeatLogo = await page.$('[aria-label="L2BEAT logo"]')
+  if (!l2BeatLogo) {
+    const string = `Meta image for ${slug} did not build properly!`
+    throw new Error(string)
```
krzkaczor, line 68: is this "string" var really better than inlining it? ;d

packages/frontend/src/build/buildMetaImages.ts
michalsidzej: `buildPages` runs in a separate process, so yes, I think we need it here too.

krzkaczor (approved):

PR #891, opened 2022-10-28
Enable activity on frontend

krzkaczor:


packages/frontend/src/build/config/common.ts
```diff
@@ -13,7 +13,7 @@ export const common = {
     banner: true,
     gitcoinOption: false,
     bridges: true,
-    activity: false,
+    activity: true,
```
sin6pi7, line 16: You can also remove obsolete `activity: true` in staging and local configs :)

PR #895, opened 2022-10-28
Fix warnings on frontend build

michalsidzej:
Fixing `Warning: Invalid DOM property 'clip-path'. Did you mean 'clipPath'` like warnings.


adamiak (approved):

krzkaczor:
@michalsidzej can we somehow make such warnings fail the build?

PR #898, opened 2022-10-28
Add verification status to main page

antooni:
image

image

image

image

Feature is hidden behind the feature flag `highlightUnverified`, and is set to false on production, which means it will not be shown.

This change will add highlight to projects with unverified contracts in the main table on following pages:
- scaling/tvl
- scaling/risk 
- scaling/activity
- bridges/tvl
- bridges/risk

There is also a change to the way we pass arguments to functions in `renderPages`, the new type `PagesData` was introduced to wrap all parameters into one object. This will be especially useful in the future when we add more and more features, then they all would be wrapped inside this type.

Resolves L2B-281


packages/frontend/src/pages/bridges-risk/BridgesRiskView.tsx
```diff
@@ -79,13 +96,22 @@ export function BridgesRiskView({ items }: BridgesRiskViewProps) {
...
+      const result: Record<string, string | boolean> = {
+        className: '',
+      }
+
+      if (entry.isVerified === false) {
+        result.className += cx(UNVERIFIED_LIGHT_CX, UNVERIFIED_DARK_CX)
+      }
+
+      if (entry.type !== 'bridge') {
+        result.className += cx(result.className, 'hidden')
```
michalsidzej, line 109: adding the `result.className` twice

packages/frontend/src/pages/scaling-tvl/props/getScalingTvlView.ts
```diff
@@ -12,16 +12,25 @@ export function getScalingTvlView(
...
+      ),
+    ),
   }
 }
 
 function getScalingTvlViewEntry(
   project: Layer2,
   tvlApiResponse: TvlApiResponse,
   aggregateTvl: number,
+  isVerified: boolean | undefined,
```
michalsidzej, line 33: Nit pick
```suggestion
  isVerified?: boolean,
```

packages/frontend/src/pages/scaling-activity/view/ActivityView.tsx
```diff
@@ -109,15 +122,24 @@ export function ActivityView({ items }: ActivityViewProps) {
...
-            className: cx(
-              'bg-blue-400 hover:bg-blue-400 border-b-blue-600',
-              'dark:bg-blue-900 dark:border-b-blue-500 dark:hover:bg-blue-900',
-            ),
-          }
-        : {},
+    getProps: (entry) => {
+      let className = ''
+
+      if (entry.name === 'Ethereum') {
```
michalsidzej, line 128: There's no way that Ethereum is unverified. I think I'd like explicit returns in each of cases, there would be no need to create `className` variable then

packages/frontend/src/pages/scaling-activity/props/getProps.ts
```diff
@@ -10,16 +9,24 @@ import { getPageMetadata } from './getPageMetadata'
 
 export function getProps(
   config: Config,
-  activityApiResponse: ActivityApiResponse,
+  pagesData: PagesData,
 ): Wrapped<ActivityPageProps> {
+  // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
```
sin6pi7, line 14: This is not a good practice and is not necessary here - we are invoking this function only if `activityApiResponse` is defined. I assume it was necessary, because `PagesData['activityApiResponse']` is optional. Simpler solution is to define `pagesData` here with only required properties that are necessary to use in this function.

packages/frontend/src/pages/scaling-activity/view/ActivityView.tsx
```diff
@@ -38,7 +44,14 @@ export function ActivityView({ items }: ActivityViewProps) {
       name: '#',
       alignRight: true,
       minimalWidth: true,
-      getValue: (_, index) => index + 1,
+      getValue: (entry, index) => {
+        if (entry.isVerified === false) {
+          return (
+            <UnverifiedWarning message="This project includes unverified contracts" />
```
sin6pi7, line 50: Is there any place in the code which uses a different message? If not, let's make `"This project includes unverified contracts"` part of `UnverifiedWarning.tsx` component in order to remove possible discrepancies in this message across the codebase.

packages/frontend/src/pages/scaling-tvl/props/getScalingTvlView.ts
```diff
@@ -12,16 +12,25 @@ export function getScalingTvlView(
   projects: Layer2[],
   tvlApiResponse: TvlApiResponse,
   tvl: number,
+  verificationStatus: VerificationStatus,
 ): ScalingTvlViewProps {
   return {
-    items: projects.map((x) => getScalingTvlViewEntry(x, tvlApiResponse, tvl)),
+    items: projects.map((x) =>
+      getScalingTvlViewEntry(
+        x,
```
sin6pi7, line 20: Please rename `x` to `project`.

packages/frontend/src/pages/scaling-tvl/view/ScalingTvlView.tsx
```diff
@@ -29,13 +36,25 @@ export interface ScalingTvlViewEntry {
   technology: string
 }
 
+export const UNVERIFIED_LIGHT_CX =
+  'bg-red-100 hover:bg-red-100 border-b-red-200'
+export const UNVERIFIED_DARK_CX =
+  'dark:bg-red-900 dark:hover:bg-red-900 dark:border-b-red-200'
```
sin6pi7, line 42: 1. These two are never used separately, always together. For that reason there is no point in exporting them separately,
2. I don't like this being exported from here since it's not the responsibility of this component.

Export this as `getRowVerificationClassNames(entry: { isVerified?: boolean }): string` from `components/table/getRowVerificationClassNames.ts` and encapsulate the whole logic of checking `isVerified` there.

packages/frontend/src/pages/bridges-risk/BridgesRiskView.tsx
```diff
@@ -35,9 +42,19 @@ export function BridgesRiskView({ items }: BridgesRiskViewProps) {
       minimalWidth: true,
       getValue: (entry, index) => (
         <>
-          <span data-bridges-only>{onlyBridges.indexOf(entry) + 1}</span>
+          <span data-bridges-only>
+            {entry.isVerified === false ? (
+              <UnverifiedWarning message="This project includes unverified contracts" />
+            ) : (
+              onlyBridges.indexOf(entry) + 1
+            )}
```
sin6pi7, line 50: This is being repeated over and over. There could be a separate component `components/table/IndexCell.tsx`, with `interface IndexCellProps { index: number, isVerified: boolean }` and would encapsulate this ternary operator.

sin6pi7 (changes requested):

packages/frontend/src/pages/bridges-risk/BridgesRiskView.tsx
```diff
@@ -35,9 +39,14 @@ export function BridgesRiskView({ items }: BridgesRiskViewProps) {
       minimalWidth: true,
       getValue: (entry, index) => (
         <>
-          <span data-bridges-only>{onlyBridges.indexOf(entry) + 1}</span>
+          <span data-bridges-only>
+            <IndexCell
+              entry={entry}
+              index={onlyBridges.indexOf(entry) + 1}
+            ></IndexCell>
```
sin6pi7, line 46: No need for a closing tag if you're not passing children.

packages/frontend/src/pages/bridges-risk/BridgesRiskView.tsx
```diff
@@ -79,13 +88,22 @@ export function BridgesRiskView({ items }: BridgesRiskViewProps) {
...
-        : {
-            ['data-combined-only']: true,
-            className: 'hidden',
-          },
+    getProps: (entry) => {
+      const result: Record<string, string | boolean> = {
+        className: '',
+      }
+
+      if (entry.isVerified === false) {
```
sin6pi7, line 96: This `if` statement is unnecessary - you're checking this in `getRowVerificationClassNames` already.

packages/frontend/src/pages/bridges-tvl/BridgesTvlView.tsx
```diff
@@ -42,9 +46,14 @@ export function BridgesTvlView({ items }: BridgesTvlViewProps) {
...
-          <span data-bridges-only>{onlyBridges.indexOf(entry) + 1}</span>
+          <span data-bridges-only>
+            <IndexCell
+              entry={entry}
+              index={onlyBridges.indexOf(entry) + 1}
+            ></IndexCell>
+          </span>
           <span data-combined-only className="hidden">
-            {index + 1}
+            <IndexCell entry={entry} index={index + 1}></IndexCell>
```
sin6pi7, line 56: Same as above.

packages/frontend/src/pages/bridges-tvl/BridgesTvlView.tsx
```diff
@@ -107,13 +116,22 @@ export function BridgesTvlView({ items }: BridgesTvlViewProps) {
...
-            ['data-combined-only']: true,
-            className: 'hidden',
-          },
+    getProps: (entry) => {
+      const result: Record<string, string | boolean> = {
+        className: '',
+      }
+
+      if (entry.isVerified === false) {
+        result.className += getRowVerificationClassNames(entry)
```
sin6pi7, line 125: Same as above.

packages/frontend/src/components/table/UnverifiedWarning.tsx
```diff
@@ -0,0 +1,25 @@
+import cx from 'classnames'
+import React from 'react'
+
+import { InfoIcon } from '../icons'
+
+export function UnverifiedWarning() {
```
sin6pi7, line 6: Let's make it more specific - `UnverifiedContractsWarning`

sin6pi7 (changes requested):

packages/frontend/src/pages/bridges-risk/BridgesRiskView.tsx
```diff
@@ -79,13 +88,22 @@ export function BridgesRiskView({ items }: BridgesRiskViewProps) {
...
+        result.className += getRowVerificationClassNames(entry)
+      }
+
+      if (entry.type !== 'bridge') {
+        result.className += cx(' hidden')
+        result['data-combined-only'] = true
+      }
+
+      return result
```
sin6pi7, line 105:
```suggestion
     return {
       className: cx(getRowVerificationClassNames(entry), entry.type !== 'bridge' && 'hidden'),
       'data-combined-only': entry.type !== 'bridge' ? true : undefined,
     }
```

packages/frontend/src/pages/bridges-tvl/BridgesTvlView.tsx
```diff
@@ -107,13 +116,22 @@ export function BridgesTvlView({ items }: BridgesTvlViewProps) {
...
+      }
+
+      if (entry.type !== 'bridge') {
+        result.className += cx(' hidden')
+        result['data-combined-only'] = true
+      }
+
+      return result
+    },
```
sin6pi7, line 134:
```suggestion
      return {
        className: cx(
          getRowVerificationClassNames(entry),
          entry.type !== 'bridge' && 'hidden',
        ),
        'data-combined-only': entry.type !== 'bridge' ? true : undefined,
      }
```

sin6pi7 (approved):

sin6pi7 (changes requested):
Please remove the bottom border from highlighted rows.

sin6pi7 (approved):

PR #902, opened 2022-10-31
Handle missing days in transaction counting

sin6pi7:
Two improvements to daily counts returned from the api:
1. Add zeroes to daily counts if:
- a day passed without a new block on a blockchain,
- no blocks were added to a blockchain for a while and we are fully synced,
2. A day has passed and some transactions happened yesterday, but today is pretty idle and no traffic has been present on the source. Our server fetched all the yesterdays data. Prior to this PR we would not consider yesterday synced until a new block today (or in the future) takes place, which resulted in yesterday not being included in the api response. This is far from ideal and this PR changes it - yesterday is included as long as we fetched everything from the source.

Resolves L2B-287


packages/common/src/services/cache/InMemoryCacheBackend.ts
```diff
@@ -0,0 +1,18 @@
...
+import { NestedDict } from './NestedDict'
+
+export class InMemoryCacheBackend implements CacheBackend {
+  dict: NestedDict
+
+  constructor(data: NestedDict['data'] = {}) {
+    this.dict = new NestedDict(data)
+  }
+
+  read() {
```
michalsidzej, line 11: We spoke recently about explicitly stating the return type of the function. I'm not sure if it also applies to a short function like this one, just pointing this out.

packages/backend/src/peripherals/loopring/LoopringClient.ts
```diff
@@ -28,9 +28,12 @@ export class LoopringClient {
     }
   }
 
-  async getFinalizedBlockNumber() {
+  async getFinalizedBlock() {
```
michalsidzej, line 31: Return type here as well.

packages/common/src/services/cache/InMemoryCacheBackend.ts
sin6pi7: This is interface implementation, so I don't think that's necessary - types are already specified there.

packages/backend/src/peripherals/loopring/LoopringClient.ts
```diff
@@ -4,10 +4,15 @@ import {
   Logger,
   RateLimiter,
 } from '@l2beat/common'
+import { UnixTime } from '@l2beat/types'
 
-import { LoopringResponse } from './schemas'
+import { LoopringBlock } from './schemas'
 
-type Block = number | 'finalized'
+interface Block {
```
sz-piotr, line 11: This type is returned from LoopringClient but is not exported and has a generic name. Maybe it should be named LoopringBlock and the schema renamed to LoopringBlockResponse

packages/backend/src/core/transaction-count/getFilledDailyCounts.ts
```diff
@@ -0,0 +1,20 @@
...
+import { fillMissingDailyCounts } from './fillMissingDailyCounts'
+import { DailyTransactionCount } from './TransactionCounter'
+
+export function getFilledDailyCounts(
+  counts: DailyTransactionCount[],
+  lastTimestamp: UnixTime,
+): DailyTransactionCount[] {
+  const tip = counts.at(-1)?.timestamp
+  const isFullySynced = tip && lastTimestamp.lte(tip)
+  const startOfDay = UnixTime.now().toStartOf('day')
```
sz-piotr, line 12: Shouldn't this rely on the clock?

packages/backend/src/core/transaction-count/RpcTransactionUpdater.ts
```diff
@@ -52,6 +53,8 @@ export class RpcTransactionUpdater implements TransactionCounter {
     this.assessCount = opts?.assessCount ?? identity
     this.startBlock = opts?.startBlock ?? 0
     this.workQueueSizeLimit = opts?.workQueueSizeLimit ?? 200_000
+    const cache = new Cache(new InMemoryCacheBackend())
+    cache.wrapMethod(rpcClient, 'getBlockAtOrBefore')
```
sz-piotr, line 57: I am not a fan of using Cache here because it was generally unused and only kept around so that some scripts for optimism are still working and InMemoryCacheBackend was only created so that it it can be tested.

I think having an in memory map here would be more verbose but better. Also adding `// @deprecated` to cache would make this clear for anyone in the future

sz-piotr (changes requested):
Looks good, left some minor comments

packages/backend/src/peripherals/loopring/LoopringClient.ts
```diff
@@ -28,16 +33,17 @@ export class LoopringClient {
     }
   }
 
-  async getFinalizedBlockNumber() {
-    const block = await this.call('finalized')
-    return block.blockId
+  async getFinalizedBlock(): Promise<Block> {
```
krzkaczor, line 36: maybe `getLatestFinalizedBlock`?

packages/backend/src/config/config.local.ts
```diff
@@ -51,21 +51,21 @@ export function getLocalConfig(): Config {
...
-            callsPerMinute: 60 / 5,
+            callsPerMinute: 60 * 10,
             url: getEnv(
               'ACTIVITY_OPTIMISM_URL',
               'https://mainnet.optimism.io/',
             ),
           },
           arbitrum: {
-            callsPerMinute: 60 / 5,
+            callsPerMinute: 60 * 10,
```
sin6pi7, line 68: These has been increased to speed up latest block bin search locally - otherwise we would have maximum 12 calls per minute.

packages/backend/src/core/transaction-count/getFilledDailyCounts.ts
sin6pi7: Passing it to the function indeed makes it easier to test. I'm not convinced we need clock here - could you provide an example of why lack of clock would introduce an error?

PR #903, opened 2022-10-31
Add layer zero omnichain tokens

luke7211:


packages/config/src/bridges/lzOmnichain.ts
```diff
@@ -0,0 +1,182 @@
...
+          category: 'Funds can be stolen if',
+          text: 'omnichain token owner changes Oracle/Relayer pair for their own.',
+          isCritical: true,
+        },
+      ],
+      isIncomplete: true,
+    },
+  },
+  config: {
+    escrows: [
```
sz-piotr, line 78: Note to someone. The escrow list should somehow be autofetched (similar maybe to multichain).

One way is to find trades going through LayerZero and find unique contracts

PR #909, opened 2022-11-03
Check correctness of addresses in permissions. Fix.

adamiak:


packages/config/test/projects.test.ts
```diff
@@ -44,6 +44,20 @@ describe('projects', () => {
           }
         }
       })
+
+      describe('permissions', () => {
+        for (const project of [...layer2s, ...bridges]) {
+          const permissions = project.permissions ?? []
+          const permissionAddresses = permissions
+            .flatMap((permission) => permission.accounts)
+            .filter((account) => account.type !== 'EOA')
```
antooni, line 53:
```suggestion
```

antooni: Can you elaborate why there is this filter? This way I can make a typo in EOA and it will pass.

antooni (changes requested):

michalsidzej (approved):

packages/config/test/projects.test.ts
michalsidzej: This will not be an issue, look at the `type` type definition   
```
export interface ProjectPermissionedAccount {
  address: string
  type: 'EOA' | 'MultiSig' | 'Contract'
}
```

michalsidzej: But that's a good point, it probably makes sense to test all the addresses

adamiak: Good point, EOAs should be checked as well. Fixing.

PR #912, opened 2022-11-04
Auto add risk when unverified

adamiak:
Resolves L2B-282

Since the verification checking is behind a flag, we cannot remove RISK_UNVERIFIED from project configs yet! If we did, all those risks would disappear on production. 

(I made an exception for Poly Bridge to make testing possible)


packages/config/src/bridges/polynetwork.ts
```diff
@@ -206,7 +206,7 @@ export const polynetwork: Bridge = {
           'Contract unverified on Etherscan. Used to proxy requests from LockProxy to EthCrossChainManager.',
       },
     ],
-    risks: [CONTRACTS.UNVERIFIED_RISK, CONTRACTS.UPGRADE_NO_DELAY_RISK],
+    risks: [CONTRACTS.UPGRADE_NO_DELAY_RISK],
```
krzkaczor, line 209: what about all other projects with this risk? shouldn't they be updated too?

krzkaczor (changes requested):

packages/config/src/bridges/polynetwork.ts
krzkaczor: nvm i just read PR description

krzkaczor (approved):
So either remove polynetwork bridge change or please add `CONTRACTS.UNVERIFIED_RISK` only when feature flag is off.

PR #913, opened 2022-11-06
Support manual list of verified contracts.

adamiak:
Resolves L2B-322


packages/config/scripts/checkVerifiedContracts/tasks.ts
```diff
@@ -7,18 +7,14 @@ import { VerificationMap } from './output'
 export async function verifyContracts(
   addresses: EthereumAddress[],
   previouslyVerified: Set<EthereumAddress>,
+  mamnuallyVerified: Set<EthereumAddress>,
```
krzkaczor, line 10: typo. I recommend installing spell check plugin if you use vscode :)

krzkaczor (changes requested):

packages/config/scripts/checkVerifiedContracts/tasks.ts
adamiak: thx! fixed.

krzkaczor (approved):

PR #914, opened 2022-11-06
Run check-verified-contracts on ci and deployment.

adamiak:
Resolves L2B-284


.github/workflows/deploy-site.yml
```diff
@@ -17,6 +17,8 @@ jobs:
         with:
           node-version: 16.x
       - run: yarn --frozen-lockfile
+      - run: yarn build:config
```
sz-piotr, line 20: If you build:config and then build:frontend you build config and common twice. Maybe the solution is explicitly building the dependencies here.

sz-piotr (changes requested):

.github/workflows/deploy-site.yml
adamiak: Yes, we build `common`, `types` and `config` twice, which takes about 20-30 additional seconds :( On top of around 15 additional seconds it takes to "re-verify" contracts.

On the other hand, this approach runs `build` on all necessary dependencies (`wsun -r`), so it's future-proof. Explicitly specifying dependencies would cause this script to fail as soon as new dependency is added to e.g. frontend.

I explored other solutions, like adding verification to `gulp`, but then it would trigger on local machines at each build, which is unnecessary and would be annoying.

Ultimately I don't have a better idea on how to integrate it into build, cron-build, but not local build. So given it only adds around 20-30 seconds to the deployment and the solution is simple and future-proof, I decided to do it this way.

krzkaczor: > I explored other solutions, like adding verification to gulp, but then it would trigger on local machines at each build, which is unnecessary and would be annoying.

We could write gulp task in a way that it has to be forced with some special flag OR it's always on when `CI == true` (which is set on CIs ;) )

sz-piotr (approved):

krzkaczor (approved):

PR #917, opened 2022-11-07
Add activity tracking for aztec projects

sin6pi7:
Resolves L2B-64


packages/backend/src/peripherals/aztec/AztecClient.ts
```diff
@@ -0,0 +1,72 @@
...
+  Rollup,
+} from './schemas'
+
+export interface Block {
+  number: number
+  timestamp: UnixTime
+  transactionCount: number
+}
+
+const API_URLS = {
```
sz-piotr, line 17: Maybe those urls should come from `@l2beat/config`

sz-piotr (changes requested):

packages/config/src/layer2s/aztecconnect.ts
```diff
@@ -61,6 +61,11 @@ export const aztecconnect: Layer2 = {
         sinceTimestamp: new UnixTime(1654638194),
       },
     ],
+    transactionApi: {
+      type: 'rpc',
```
sz-piotr, line 65: Why not have the type be `aztec`, then you wouldn't have to filter for those project ids specifically

sz-piotr (changes requested):

sz-piotr (changes requested):
Please add a flag to prevent aztec syncing from holding other apis

packages/backend/src/setup/ActivityModule.ts
```diff
@@ -131,7 +131,13 @@ export function getActivityModule(
       ...layer2BlockTransactionUpdaters,
       ...starkexTransactionUpdaters,
       zksyncTransactionUpdater,
-    ],
+    ].filter((updater) =>
+      config.projects.find(
```
sz-piotr, line 135: use .some instead of .find

sz-piotr (approved):

PR #918, opened 2022-11-07
Add unverified contracts highlight

antooni:
image

image

image

image

Resolves L2B-280


packages/frontend/src/components/icons/symbols/BulletIcon.tsx
```diff
@@ -0,0 +1,17 @@
...
+
+import { Icon } from '../Icon'
+
+export function BulletIcon(props: SVGAttributes<SVGElement>) {
+  return (
+    <Icon
+      viewBox="0 0 20 20"
+      width="16"
+      height="20"
+      aria-label="Unverified contract icon"
```
sz-piotr, line 11: this text is incorrect. Also this icon doesn't change colour on light mode

packages/config/src/verified.json
```diff
@@ -61,7 +60,6 @@
     "0x051AD3F020274910065Dcb421629cd2e6E5b46c4": true,
     "0x063e871f8DB991CEAd34B557A00B157B360084cc": true,
     "0x070cB1270A4B2bA53c81CeF89d0FD584Ed4F430B": true,
-    "0x07245eEa05826F5984c7c3C8F478b04892e4df89": true,
```
sz-piotr, line 64: are all those changes in this file warranted?

packages/frontend/src/components/table/IndexCell.tsx
```diff
@@ -11,7 +11,12 @@ export interface IndexCellProps {
 
 export function IndexCell({ entry, index }: IndexCellProps) {
   if (entry.isVerified === false) {
-    return <UnverifiedContractsWarning />
+    return (
+      <UnverifiedContractsWarning
+        className={'absolute top-0 -translate-y-1/2 left-[6px]'}
+        tooltip="This project includes unverified contracts."
```
sz-piotr, line 17: Is this the only version of the tooltip?

packages/frontend/src/components/project/WarningBar.tsx
```diff
@@ -2,42 +2,42 @@ import cx from 'classnames'
...
             {text.slice(-1)}
           </>
         ) : (
           text
-        )}
-      </p>
-    </div>
+        )
+      }
+    ></Callout>
```
sz-piotr, line 41: Why don't we have a lint rule for this?
```suggestion
    />
```

packages/frontend/src/components/project/PermissionsSection.tsx
```diff
@@ -1,47 +1,49 @@
...
-              ))}
-            </span>
-            {permission.description && (
-              <p className="text-gray-860 dark:text-gray-400">
-                {permission.description}
-              </p>
-            )}
-          </li>
+      <div className="flex flex-wrap gap-4 my-4">
+        {permissions.map(toContract).map((contract, i) => (
```
sz-piotr, line 23: Can't we do the mapping in getProps?

packages/frontend/src/components/project/ContractsSection.tsx
```diff
@@ -48,30 +39,16 @@ export function ContractsSection(props: ContractsSectionProps) {
...
       </h3>
-      <ul className="list-disc my-4 pl-8 space-y-4">
+      <div className="flex flex-wrap gap-4 my-4">
         {props.contracts.map((contract, i) => (
-          <li key={i}>
-            <strong>{contract.name}</strong>{' '}
-            <EtherscanLink
-              address={contract.address}
-              className="text-xs md:text-base"
+          <React.Fragment key={i}>
```
sz-piotr, line 44: There is no need to wrap this in Fragment just use ContractEntry directly

sz-piotr (changes requested):

krzkaczor (commented):
Icons changed on main screen too ([according to visual regression report](https://l2beat_l2beat_7f2f4bd00faa141c2386_reg.artifact.superactions.io/report/index.html?id=changed-Pages/Scaling/TvlView/Tvl%20View.png)). Was this expected?

packages/frontend/src/components/project/PermissionsSection.tsx
```diff
@@ -1,47 +1,33 @@
...
-          </li>
+      <div className="flex flex-wrap gap-4 my-4">
+        {permissions.map((permission, i) => (
+          <React.Fragment key={i}>
+            <ContractEntry
+              contract={permission}
+              verificationStatus={verificationStatus}
+            />
+          </React.Fragment>
         ))}
```
michalsidzej, line 29: No need to use `React.Fragment`. BTW, you can also use shor syntax for fragments if you need them https://reactjs.org/docs/fragments.html#short-syntax
```suggestion
        {permissions.map((permission, i) => (
          <ContractEntry
            key={i}
            contract={permission}
            verificationStatus={verificationStatus}
          />
        ))}
```

michalsidzej (changes requested):

michalsidzej (dismissed):
The bullet points seem to be a bit off.
Screenshot 2022-11-09 at 10 19 08
Try setting `cy="10"` so the circle is in the center of SVG and then add classes `h-6 md:h-[27px]`, so the height of the SVG is equal to line-height on the right

krzkaczor (approved):
Approved (once you're done with @michalsidzej comments)

packages/frontend/src/components/project/ContractsSection.tsx
michalsidzej: +1

PR #921, opened 2022-11-08
Create /health endpoint

michalsidzej:
Resolves L2B-340
Skipped the status field as it will be implicit from `response status 200`

### DIFF
--- a/packages/backend/src/config/config.production.ts
+++ b/packages/backend/src/config/config.production.ts
@@ -66,7 +66,8 @@ export function getProductionConfig(): Config {
       },
     },
     health: {
-      startedAt: getEnv('HEROKU_RELEASE_CREATED_AT', new Date().toISOString()),
+      releasedAt: getEnv('HEROKU_RELEASE_CREATED_AT'),
+      startedAt: new Date().toISOString(),
       commitSha: getEnv('HEROKU_SLUG_COMMIT', getGitCommitHash()),
     },
   }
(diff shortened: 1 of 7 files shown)


packages/backend/src/config/config.production.ts
```diff
@@ -64,5 +64,9 @@ export function getProductionConfig(): Config {
         },
       },
     },
+    health: {
+      startedAt: getEnv('HEROKU_RELEASE_CREATED_AT'),
+      commitSha: getEnv('HEROKU_SLUG_COMMIT'),
+    },
```
sz-piotr, line 70: Can we get those values in a local config too?

sz-piotr (changes requested):

packages/backend/src/config/config.production.ts
```diff
@@ -64,5 +64,9 @@ export function getProductionConfig(): Config {
         },
       },
     },
+    health: {
+      startedAt: getEnv('HEROKU_RELEASE_CREATED_AT'),
```
krzkaczor, line 68: How is this value exactly defined? I thought that we should just get date.now() when creating config and this will be correct.

krzkaczor (changes requested):

packages/backend/src/config/config.production.ts
michalsidzej: https://devcenter.heroku.com/articles/dyno-metadata#dyno-metadata

michalsidzej: Ok, going for date.now()

krzkaczor (approved):

krzkaczor:
Once deployed, please verify that it works correctly on heroku.

krzkaczor (approved):

PR #923, opened 2022-11-08
Fix aztec connect

sin6pi7:


michalsidzej (approved):
I think the problem was with original Aztec, not Aztec Connect ;)

PR #932, opened 2022-11-10
Exclude contracts's admin from verification

antooni:
Adding the field `isAdmin` allows to differentiate it on the frontend and exclude it from the checks.

Resolves L2B-371


sz-piotr (approved):
I am also unsure if this change wouldn't make more sense in the script that checks verification

PR #934, opened 2022-11-10
Move discovery code to backend.

adamiak:
Resolves L2B-353


packages/backend/src/core/discovery/proxies/GnosisSafe.ts
```diff
@@ -0,0 +1,66 @@
+import { providers } from 'ethers'
+
+import { ContractParameters } from '../types'
+import { bytes32ToAddress } from '../utils/address'
+import { getCallResult } from '../utils/getCallResult'
+import { getStorage } from '../utils/getStorage'
+import { GnosisSafe__factory } from './typechain/GnosisSafe__factory'
```
sz-piotr, line 7: I think I would rewrite this file to not use typechain and just declare the abi on the top of the file and the call and cast the values from any to whatever is needed. Yes it is less type safe, but we would also not have to commit autogenerated files from a dependency that is no longer used.

packages/backend/src/core/discovery/DiscoveryEngine.ts
```diff
@@ -0,0 +1,58 @@
...
+          meta: undefined,
+        })),
+    }
+
+    await writeFile(
+      `discovery/${name}/discovered.json`,
+      JSON.stringify(project, null, 2),
+    )
+
+    // Temporarily commenting out saving abis to separate files.
```
sz-piotr, line 47: We should just remove this code. No need to comment it out.

packages/backend/src/index.ts
```diff
@@ -13,6 +14,11 @@ main().catch((e) => {
 
 async function main() {
   const cli = getCliParameters()
+  const env =
+    process.env.DEPLOYMENT_ENV ??
+    (process.env.NODE_ENV === 'production' ? 'production' : 'local')
+  console.log('Loading config for:', env)
+  const config = getConfig(env)
```
sz-piotr, line 21: Maybe a task to rethink what things are required in the config now needs to be added

sz-piotr (changes requested):
Left some small comments

packages/backend/src/index.ts
sz-piotr: Added a task in linear

sz-piotr (approved):

PR #944, opened 2022-11-14
Enable highlightUnverified flag for production.

adamiak:
Resolves L2B-382


antooni (approved):

sz-piotr (changes requested):
Please also hide the banner

sz-piotr (approved):

PR #948, opened 2022-11-15
Unify timestamp and timezone handling

michalsidzej:
Resolves L2B-407


packages/backend/src/peripherals/database/shared/Database.ts
```diff
@@ -2,8 +2,11 @@ import { Logger } from '@l2beat/common'
 import KnexConstructor, { Knex } from 'knex'
 import path from 'path'
 
+import { configureUtc } from './configureUtc'
 import { PolyglotMigrationSource } from './PolyglotMigrationSource'
 
+configureUtc()
```
sz-piotr, line 8: I would move this into database constructor

sz-piotr (changes requested):
If we migrate `unix_timestamp` to `timestamp` then why are we keeping the name? Either the code must change or the PR name should change. I'm for the PR name to just say `Unify timestamp and timezone handling`

sz-piotr (approved):

PR #950, opened 2022-11-15
Update Stargate contracts

antooni:
This PR will:
- copy Relayer and Oracle contracts from Aptos to Stargate
- copy permissions to corresponding contracts
- reorder and rename some of the contracts
- remove duplicate contracts
- update Aptos multisig
- update descriptions

image


packages/config/src/bridges/aptos.ts
```diff
@@ -166,24 +148,34 @@ export const aptos: Bridge = {
...
-          type: 'Contract',
+          address: '0x971016EF5Bd9C71fA4ff34D731974d03cEFf5F05',
+          type: 'MultiSig',
         },
       ],
-      name: 'Aptos Bridge Relayer',
+      name: 'Aptos Multisig',
       description:
-        'Contract authorized to relay messages and - as a result - withdraw funds from the bridge.',
+        'Bridge owner, can setup tokens, fees, WETH token address (potentially malicious). Can withdraw all the funds from the Escrow after unlocking emergency withdrawal with 1 week delay.',
```
krzkaczor, line 157: Here we could mention the possibility of instantly changing config of L0 and spoofing withdrawals...

krzkaczor (approved):

PR #953, opened 2022-11-15
Fix `assert` imports

sin6pi7:


adamiak (approved):

sz-piotr (approved):

sz-piotr:
Can we also somehow ban such imports?

PR #957, opened 2022-11-15
Refactor unverified contracts to use `promiseAllPlus`

krzkaczor:


packages/config/scripts/checkVerifiedContracts/tasks.ts
```diff
@@ -1,4 +1,9 @@
-import { EtherscanClient, Logger, TaskQueue } from '@l2beat/common'
+import {
+  EtherscanClient,
+  Logger,
+  promiseAllPlus,
+  TaskQueue,
```
adamiak, line 5: `TaskQueue` is no longer used.

packages/config/scripts/checkVerifiedContracts/tasks.ts
```diff
@@ -13,67 +18,22 @@ export async function verifyContracts(
...
-      workers: workersCount,
-      shouldRetry: (_) => {
-        queueStatus.errorOccurred = true
-        return { retry: false }
-      },
-    },
-  )
-
-  return { taskQueue, queueStatus }
+  return Object.fromEntries(verification)
```
adamiak, line 38: My IDE screams that `verification` has `any` type. I'm not sure why our lint doesn't pick it up here.

adamiak (approved):

packages/config/scripts/checkVerifiedContracts/tasks.ts
krzkaczor: rebuild the project, perhaps. Works for me (TM)

adamiak: Restarting IDE helped :/

PR #962, opened 2022-11-16
Add DiscoveryProvider to help with future tests

sz-piotr:
This PR adds an abstraction over ethers.js and Etherscan so that a mocked version can be passed in tests.
Unfortunately this PR doesn't yet include those tests :(

It also refactors the code to use `EthereumAddress` and `Bytes` instead of `string`.


packages/backend/src/core/discovery/DiscoveryEngine.ts
```diff
@@ -1,28 +1,23 @@
...
-import { providers } from 'ethers'
+import { EthereumAddress } from '@l2beat/types'
 import { writeFile } from 'fs/promises'
 
 import { discover } from './discover'
 import { DiscoveryOptions } from './DiscoveryOptions'
+import { DiscoveryProvider } from './provider/DiscoveryProvider'
 import { ProjectParameters } from './types'
 
 export class DiscoveryEngine {
```
adamiak, line 9: It's getting a bit confusing to have `DiscoveryModule`, `DiscoveryEngine` and `DiscoveryProvider`. 

Since `DiscoveryEngine` became so short (that it doesn't even seem to warrant using a class), I suggest removing `DiscoveryEngine` completely and transforming it into a function somewhere.

adamiak: This comment is for the whole file, not the highlighted line :)

adamiak (approved):

PR #970, opened 2022-11-17
Add new client for AztecConnect

michalsidzej:
AztecConnect changed its API, but unfortunately Aztec still has the graphql version. 

Co-authored-by: Jacek <sin6pi7@users.noreply.github.com>


packages/backend/src/peripherals/aztec/schemas.ts
```diff
@@ -5,18 +5,21 @@ import { stringAs } from '../../tools/types'
...
+export const Rollup = z.object({
   id: z.number().int(),
   mined,
   numTxs: z.number().int(),
 })
 
+export type Rollups = z.infer<typeof Rollups>
+export const Rollups = z.array(Rollup.extend({ mined: mined.nullable() }))
+
 export type GetRollupsResponseBodySchema = z.infer<
```
sin6pi7, line 17:
```suggestion
export type AztecGetRollupsResponseBody = z.infer<
```

packages/backend/src/peripherals/aztec/schemas.ts
```diff
@@ -5,18 +5,21 @@ import { stringAs } from '../../tools/types'
...
   typeof GetRollupsResponseBodySchema
 >
 export const GetRollupsResponseBodySchema = z.object({
```
sin6pi7, line 20:
```suggestion
export const AztecGetRollupsResponseBody = z.object({
```

sin6pi7: Same goes for `GetRollupResponseBodySchema` -> `AztecGetRollupResponseBody`

sin6pi7 (changes requested):

packages/backend/src/peripherals/aztec/utils.ts
```diff
@@ -0,0 +1,9 @@
+import { assert } from '@l2beat/common'
+
+import { Block, Rollup, Rollups, toBlock } from './schemas'
+
+export function findLatestBlock(rollups: Rollups): Block {
+  const latestMined = rollups.find((r): r is Rollup => r.mined !== null)
+  assert(latestMined, 'No mined block found in first 10 blocks')
```
sz-piotr, line 7: It isn't obvious why this error occurs. Maybe the name of the function has to change? Also, if the blocks aren't mined shouldn't we just query for more?

sin6pi7: > It isn't obvious why this error occurs. Maybe the name of the function has to change?
`findLatestBlockOrThrow` would work?

packages/backend/src/peripherals/aztec/schemas.ts
```diff
@@ -28,3 +31,32 @@ export const GetRollupResponseBodySchema = z.object({
     rollup: Rollup,
   }),
 })
+
+export function parseWithSchema<T extends z.ZodType<unknown>>(
```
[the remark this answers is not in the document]
sin6pi7: Re-thrown error includes also the original `data` - zod only tells us what was different than expected, does not provide the whole input. In this case where we don't really know what can come from the api we'd rather see the whole body, because it gives more context for debugging.

packages/backend/src/peripherals/aztec/utils.ts
sin6pi7, line 7:
```suggestion
  assert(latestMined, `No mined block found in first ${rollups.length} blocks`)
```

packages/config/src/layer2s/types/Layer2TransactionApiV2.ts
```diff
@@ -18,7 +18,7 @@ export interface StarknetTransactionApiV2 {
 }
 
 export interface AztecTransactionApiV2 {
-  type: 'aztec'
+  type: 'aztec' | 'aztecconnect'
```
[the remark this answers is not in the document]
sz-piotr: I think that for any other system if they had a different api it would have its own distinct function. The only reason they are together right now is the fact that both systems are created by the same company, so I would be for splitting the logic. Duplication should be fine in this case.

packages/backend/src/peripherals/aztec/schemas.ts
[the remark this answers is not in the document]
sz-piotr: Okay. Let's keep it for now then

sz-piotr (changes requested):
I'd split the creation of Aztec and Aztec Connect processors, but otherwise it's ok

packages/config/src/layer2s/types/Layer2TransactionApiV2.ts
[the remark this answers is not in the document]
sin6pi7: Yeah, I realised that too actually, so they are split now.

sin6pi7:
> I'd split the creation of Aztec and Aztec Connect processors, but otherwise it's ok

It's actually split already :)

sz-piotr (approved):

PR #973, opened 2022-11-18
Ensure proper db version

sin6pi7:
Whenever a connection to postgres server is acquired check if this server is of required version.

### DIFF
--- a/packages/backend/src/peripherals/database/shared/Database.ts
+++ b/packages/backend/src/peripherals/database/shared/Database.ts
@@ -2,6 +2,7 @@ import { assert, Logger } from '@l2beat/common'
 import KnexConstructor, { Knex } from 'knex'
 import path from 'path'
 import { Client } from 'pg'
+import { callbackify } from 'util'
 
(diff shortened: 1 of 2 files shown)


packages/backend/src/peripherals/database/shared/Database.ts
```diff
@@ -35,6 +52,9 @@ export class Database {
           path.join(__dirname, '..', 'migrations'),
         ),
       },
+      pool: {
+        afterCreate: callbackify(this.ensureServerVersion.bind(this)),
+      },
```
michalsidzej, line 57: Do we need to check it after creating every connection? Doesn't really look optimal for me. Is there a way to do it only on startup of the app maybe?

sin6pi7: As far as my understanding goes it does not happen that often - there's only a handful of connections to postgres server and they are pooled and reused by node process. But in fact just checking it in `Application.ts` will simplify the code and will utilise simpler concepts.

michalsidzej (approved):

PR #982, opened 2022-11-21
Enable starkex to repeat last 7 days

michalsidzej:
Resolves L2B-388

Decided not to add this functionality to SequenceProcessor. This patch is applicable only to Starkex flaky API rn.


packages/backend/src/config/Config.ts
```diff
@@ -79,6 +79,7 @@ export interface ActivityV2Config {
   readonly starkexApiKey: string
   readonly starkexApiDelayHours: number
   readonly starkexCallsPerMinute: number
+  readonly starkexRepeatLastDays: number
```
antooni, line 82: wdyt?
```suggestion
  readonly starkexResyncLastDays: number
```

packages/backend/src/core/activity/processors/StarkexProcessor.ts
```diff
@@ -25,18 +26,20 @@ export function createStarkexProcessor(
...
     sequenceProcessorRepository,
     {
       batchSize,
-      startFrom: options.sinceTimestamp.toStartOf('day').toDays(),
+      startFrom: startDay,
       // eslint-disable-next-line @typescript-eslint/require-await
       getLatest: async () =>
         getStarkexLastDay(clock.getLastHour(), options.starkexApiDelayHours),
       processRange: async (from, to, trx) => {
+        from = Math.max(from - options.starkexRepeatLastDays, startDay)
```
antooni, line 42: this line is a bit misleading for me, I assume `from` is the amount of days
maybe adding a custom type and explicitly typing it is a solution, I dunno tbh

antooni (approved):

packages/backend/src/core/activity/processors/StarkexProcessor.ts
sz-piotr: Yeah, I also wouldn't choose this as a solution. It is definitely not intuitive and also affects syncing from scratch as opposed to just getting new values.

sz-piotr:
@sin6pi7 I like this suggestion. The only issue is adding complexity to SequenceProcessor

sin6pi7:
> @sin6pi7 I like this suggestion. The only issue is adding complexity to SequenceProcessor

The other solution would be to extract my suggestion outside SequenceProcessor by adding a handler to `ALL_PROCESSED` event when creating a processor for starkex and doing sofr resync work there:

```
  const processorId = projectId.toString()
  const processRange = (from, to, trx?) => { ... } 
  const processor = new SequenceProcessor(processorId, sequenceProcessorRepository, { processRange }...)
  processor.on(ALL_PROCESSED_EVENT, async () => {
    const state = await sequenceProcessorRepository.getById(processorId)
    assert(state)
    await processRange(state.lastProcessed - starkexResyncLastDays, state.lastProcessed)
  })
  ```

**Review by sin6pi7 — CHANGES_REQUESTED** [review]
Current implementation is not correct (it will not resync when sequencer already processed everything every hour) and does too much work.

File: packages/backend/src/config/Config.ts [hunk 3]
```diff
@@ -79,6 +79,7 @@ export interface ActivityV2Config {
   readonly starkexApiKey: string
   readonly starkexApiDelayHours: number
   readonly starkexCallsPerMinute: number
+  readonly starkexResyncLastDays: number
```
sin6pi7, line 82: I don't see a reason why this would change depending on the environment. That being said I think it will be cleaner and simpler to add this to `StarkexLayer2TransactionApiV2` and define per project (even if at this point all of them will have the same value).

packages/backend/src/config/Config.ts
```diff
@@ -79,6 +79,7 @@ export interface ActivityV2Config {
   readonly starkexApiKey: string
   readonly starkexApiDelayHours: number
```
sin6pi7, line 80: Do we need this now? We still need to define a delay for refreshing view, but this should be renamed now to `dailyViewRefreshDelayHours`

sin6pi7 (changes requested):

packages/backend/src/core/activity/processors/StarkexProcessor.ts
```diff
@@ -25,33 +27,52 @@ export function createStarkexProcessor(
...
+
+  processor.on(ALL_PROCESSED_EVENT, () => {
+    void (async function () {
+      const state = await sequenceProcessorRepository.getById(processor.id)
+      assert(state)
+      await processRange(
+        state.lastProcessed - options.starkexResyncLastDays,
+        state.lastProcessed,
+      )
+    })()
```
sin6pi7, line 72: This has two issues:
1. `void (async () => {})()` reduces readability (I do understand `EventEmitter.on` expects void as a return type though),
2. Async error thrown here is swallowed.

Instead I suggest the following:
```
const handleAllProcessedEvent = async () => {
      const state = await sequenceProcessorRepository.getById(processor.id)
      assert(state)
      await processRange(
        state.lastProcessed - options.starkexResyncLastDays,
        state.lastProcessed,
      )
}

processor.on(ALL_PROCESSED_EVENT, () => {
    handleAllProcessedEvent().catch(logger.error.bind(logger))
})
```

This way we don't fight with type system and get proper error handling.

sin6pi7 (changes requested):

sin6pi7 (approved):

PR #983, opened 2022-11-21
Skip often eth_chainId calls during activity sync

michalsidzej:
Resolves L2B-443

https://docs.ethers.io/v5/api/providers/jsonrpc-provider/#StaticJsonRpcProvider

### DIFF
--- a/packages/backend/src/core/activity/processors/RpcProcessor.ts
+++ b/packages/backend/src/core/activity/processors/RpcProcessor.ts
@@ -24,7 +24,7 @@ export function createRpcProcessor(
   const batchSize = getBatchSizeFromCallsPerMinute(callsPerMinute)
   const url = transactionApi.url
   assert(url, 'Url for rpc client must be defined')
-  const provider = new providers.JsonRpcProvider({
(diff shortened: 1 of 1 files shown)


packages/backend/src/core/activity/processors/RpcProcessor.ts
```diff
@@ -24,7 +24,7 @@ export function createRpcProcessor(
   const batchSize = getBatchSizeFromCallsPerMinute(callsPerMinute)
   const url = transactionApi.url
   assert(url, 'Url for rpc client must be defined')
-  const provider = new providers.JsonRpcProvider({
+  const provider = new providers.StaticJsonRpcProvider({
```
sz-piotr, line 27: Can we do that for activity v2 also?

sz-piotr (changes requested):

packages/backend/src/core/activity/processors/RpcProcessor.ts
michalsidzej: This is actually the v2, updated v1 as well

sz-piotr (approved):

PR #984, opened 2022-11-21
Add milestones type to config

antooni:
This PR aims to introduce new field to project's config: `milestones`, responsible for holding information about milestones achieved by a given project. 

Resolves L2B-420


packages/config/src/common/Milestone.ts
```diff
@@ -0,0 +1,6 @@
+export interface Milestone {
+  name: string
+  description: string
```
sz-piotr, line 3: If we are going to have empty descriptions it is better to make them optional

sz-piotr (changes requested):

sz-piotr (approved):

PR #985, opened 2022-11-21
Remove news section

antooni:
This PR aims to remove news section and code responsible for it. The initial idea was to hide it behind feature flag but IMO it is better to delete because:
- if we want to retrieve it, we can do it via git
- news section was not updated recently, it just looks outdated
- removal is not possible behind feature flag, so faced with the decision whether to pass empty array in getProps or hide on UI, I decided that it will be best to just delete it, code will be cleaner, and we will no longer worry about it

Resolves L2B-422


sz-piotr (approved):

luke7211:
We we are removing news section ? :(

PR #989, opened 2022-11-22
fix: update zkspace config for the verified status of contracts

ZKSwap-fe:
The pair manager has been verified: https://etherscan.io/address/0x5f3bE7846efC473552C5619b929F7d4aa640fb54#code


ZKSwap-fe:
@sz-piotr I think you missed the change of the contract's description in this PR. The pages still says "The source code of this contract is not verified on Etherscan." while it has been verified.

PR #995, opened 2022-11-22
Make sure invert tool looks into upgradeability

adamiak:
Resolves L2B-447

### DIFF
--- a/packages/update-monitor/src/inversion/index.ts
+++ b/packages/update-monitor/src/inversion/index.ts
@@ -44,15 +44,8 @@ export function invertAndPrint(
   }
 
   for (const contract of project.contracts) {
-    const upgradeabilityValues = contract.upgradeability as unknown as Record<
-      string,
-      ContractValue
-    >
-
-    add(contract.address)
-    const values: Record<string, ContractValue | undefined> = {
-      ...contract.values,
-      ...upgradeabilityValues,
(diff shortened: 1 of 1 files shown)


packages/update-monitor/src/inversion/index.ts
```diff
@@ -43,17 +38,45 @@ export function invertAndPrint(
...
+      ...contract.values,
+      ...upgradeabilityValues,
+
+      // We don't want to show the implementation as nodes in the diagram
+      implementation: undefined,
+      callImplementation: undefined,
+      adminImplementation: undefined,
+      userImplementation: undefined,
+      implementations: undefined,
+    }
```
sz-piotr, line 63: We can make this a bit more robust by making sure we don't override values.
```suggestion
    const upgradeabilityValues = {
      ...contract.upgradeability as unknown as Record<string, ContractValue>

      // We don't want to show the implementation as nodes in the diagram
      implementation: undefined,
      callImplementation: undefined,
      adminImplementation: undefined,
      userImplementation: undefined,
      implementations: undefined,
    }

    add(contract.address)
    const values: Record<string, ContractValue | undefined> = {
      ...upgradeabilityValues,
      ...contract.values,
    }
```

sz-piotr (changes requested):

sz-piotr (approved):

PR #998, opened 2022-11-22
Add Godwoken L2

Pumpitza:


luke7211:
It looks like this is L2 built on top of Nervos L1 - L2beat right now supports only L2s built on top of Ethereum

Pumpitza:
> It looks like this is L2 built on top of Nervos L1 - L2beat right now supports only L2s built on top of Ethereum

I see, thank you for the info, will wait until we could push this.

PR #1013, opened 2022-11-24
Refactor DailyTransactionCountService

sin6pi7:
Debugging sync check made me realise it's hard to reason about `DailyTransactionCountService`, because it's too big and vague. I decided to split this class into smaller ones with specific responisbilities:
- refreshing sql view,
- monitoring sync correctnes,
- postprocessing db counts.

The last one can probably be just included in the controller, but I did not want to complicate this PR any further.

I decided to introduce `TransactionCounter` abstraction, which is a wrapper on top of `SequenceProcessor`:
- it exposes only necessary properties from `SequenceProcessor` relevant for transaction counting,
- it exposes information about syncing status time-wise


packages/backend/src/core/activity/postprocessCounts.ts
```diff
@@ -15,7 +15,7 @@ export function postprocessCounts(
   counts: DailyTransactionCount[],
   processedAll: boolean,
   now?: UnixTime,
-) {
+): DailyTransactionCount[] {
   const today = (now ?? UnixTime.now()).toStartOf('day')
```
sz-piotr, line 19: Why this doesn't use the clock?

sin6pi7: Maybe I am not getting something, but I don't see a reason why it should use clock - `DailyTransactionCount.timestmap` is an actual timestamp, not clock-relative and in this function we are interested in relation between timestamp and actual time.

sz-piotr (approved):

packages/backend/src/core/activity/postprocessCounts.ts
sz-piotr: Yes and I thought we are using clock to get the actual time and it returns the offset, so that the entire system can work with that.

PR #1028, opened 2022-11-29
Move Milestones on mobile

antooni:
This PR aims to move up Milestones component on the mobile view, just above description.
image

Resolves L2B-475


sz-piotr (approved):

sz-piotr:
I'm approving this, but I still think that eventually we probably would want to move risks above the milestones

PR #1031, opened 2022-12-01
Add sequencer and proposer to Optimism config

luke7211:


packages/update-monitor/src/projects/optimism/contracts/getStateCommitmentChain.ts
```diff
@@ -12,18 +13,27 @@ export async function getStateCommitmentChain(
...
   )
 
+  const addressManager: string = await stateCommitmentChain.libAddressManager()
+
+  const bondManager = await getCallResultWithRevert<string>(
+    provider,
+    addressManager,
+    'function getAddress(string implementationName) view returns(address)',
+    ['BondManager'],
+  )
```
sz-piotr, line 23: I'm going to approve, but why have this here instead of it being on the addressManager?

sz-piotr (approved):

PR #1035, opened 2022-12-02
Add milestones to chart

antooni:
Warning; I recommend testing it on Arbitrum page (180D scale),  be aware that not all of the projects have milestones  in config so it will not show up

This PR aims to place milestone icon on the chart, the edited code is mostly passing milestones in props, changing types and integrating it with current chart flow. The most interesting file is `packages/frontend/src/components/chart/configure/render/milestones/renderMilestones.ts` responsible for logic of the display.

image

Resolves L2B-390


packages/frontend/src/components/chart/configure/render/milestones/getMilestoneHtml.tsx
```diff
@@ -0,0 +1,19 @@
...
+  return renderToStaticMarkup(
+    <div className="absolute z-50" style={{ left: x - OFFSET, top: -OFFSET }}>
+      <MilestoneIcon
+        width={ICON_SIZE}
+        height={ICON_SIZE}
+        viewBox={`0 0 ${ICON_SIZE} ${ICON_SIZE}`}
+      />
+    </div>,
+  )
+}
```
sz-piotr, line 19: As much as I like the code clarity here this solution imports `react` as well as `react-dom/server` which weren't previously imported in frontend scripts. For this one edge case please use template strings.

sz-piotr (changes requested):

sz-piotr (approved):

PR #1036, opened 2022-12-02
Update Across Bridge

adamiak:
Resolves L2B-502


sz-piotr (changes requested):
Add to monitor

packages/update-monitor/src/projects/acrossBridge/contracts/hubPool.ts
```diff
@@ -0,0 +1,42 @@
...
+      paused: await contract.paused(),
+      protocolFeeCaptureAddress: await contract.protocolFeeCaptureAddress(),
+      protocolFeeCapturePct: (
+        await contract.protocolFeeCapturePct()
+      ).toNumber(),
+      timerAddress: await contract.timerAddress(),
+      weth: await contract.weth(),
+      crossChainContracts: (
+        await Promise.all(
+          new Array(20).fill(0).map(async (_, i) => {
```
sz-piotr, line 32: is the number 20 a "good enough large number" here?

sz-piotr (approved):

packages/update-monitor/src/projects/acrossBridge/contracts/hubPool.ts
adamiak: I assume so but couldn't find a proper way to find the max. I'll research further, but separately from this PR.

PR #1042, opened 2022-12-07
Add ArrayHandler

sz-piotr:
Resolves L2B-516, L2B-358

This new array handler is quite smart, here are some examples:
```jsonc
{
  "one": {
    // simplest mode, finds the method one(uint) from abi and calls it
    // until it reverts
    "type": "array"
  },
  "two": {
    // specifies a different method foo(uint) from abi
    "type": "array",
    "method": "foo"
  },
  "three": {
    // specifies a specific abi that doesn't have to exist on the contract
    "type": "array",
    "method": "function bar(uint256 i) view returns (uint256)"
  },
  "four": {
    // specifies a specific abi that doesn't have to exist on the contract
    "type": "array",
    "method": "function bar(uint256 i) view returns (uint256)"
  },
  "five": {
    // specifies a maximum length
    "type": "array",
    "length": 1
  },
  "six": {
    // specifies a maximum length as a reference
    "type": "array",
    "length": "{{ foo }}"
  },
  "seven": {
    // everything, everywhere all at once
    "type": "array",
    "method": "function bar(uint256 i) view returns (uint256)",
    "length": "{{ foo }}"
  }
}
```

### DIFF
--- a/packages/backend/src/core/discovery/handlers/user/ArrayHandler.ts
+++ b/packages/backend/src/core/discovery/handlers/user/ArrayHandler.ts
@@ -16,7 +16,10 @@ export const ArrayHandlerDefinition = z.strictObject({
   type: z.literal('array'),
   method: z.optional(z.string()),
   length: z.optional(z.union([z.number().int().nonnegative(), Reference])),
+  maxLength: z.optional(z.number().int().nonnegative()),
 })
+
+const DEFAULT_MAX_LENGTH = 100
 
 export class ArrayHandler implements Handler {
   readonly dependencies: string[] = []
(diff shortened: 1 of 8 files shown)


packages/backend/src/core/discovery/handlers/user/ArrayHandler.ts
```diff
@@ -1,29 +1,100 @@
...
+      const errorResult = results.find((x) => x.error)
+      if (errorResult) {
+        return { field: this.field, error: errorResult.error }
+      }
+      // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
+      const value = results.map((x) => x.value!)
+      return { field: this.field, value }
+    } else {
+      const value: ContractValue[] = []
+      for (let i = 0; ; i++) {
```
adamiak, line 82: I'm worried about infinite loop here, especially when running in watch mode. This can happen when someone changes implementation. We would not get any error and it would stop our watch mode.

I suggest we introduce a limit here (200?) and return an error if crossed. Or maybe require a maxLenght to always be passed. Anything that will guarantee we never end up in infinite loop.

adamiak (approved):

packages/backend/src/core/discovery/handlers/user/ArrayHandler.ts
sz-piotr: I've added a `maxLength` field with a `100` default

PR #1043, opened 2022-12-07
Add milestone hover on chart

antooni:
This PR aims to add hover functionality to milestone on the chart. It is achieved via current hover functionality, plugging into the hoverIndex calculation with custom logic.
image

Resolves L2B-424


packages/frontend/src/components/chart/configure/controls/onMouseMove.ts
```diff
@@ -22,13 +22,14 @@ export function onMouseMove(
       e.clientX >= rect.left - HOVER_AREA_EXTENSION_PX &&
       e.clientX <= rect.right + HOVER_AREA_EXTENSION_PX &&
       e.clientY >= rect.top &&
-      e.clientY <= rect.bottom
+      e.clientY <= rect.bottom + 10
```
sz-piotr, line 25: 15 px would also be ok :)

packages/frontend/src/components/chart/configure/render/renderHover.ts
```diff
@@ -137,3 +164,42 @@ function renderTpsRow(value: number, source: 'L2' | 'ETH') {
...
+    case 1:
+      return 'st'
+    case 2:
+      return 'nd'
+    case 3:
+      return 'rd'
+    default:
+      return 'th'
+  }
+}
```
sz-piotr, line 205: We should have consistent date formatting. Either use this new format everywhere or use the old format here.

packages/frontend/src/components/chart/configure/update/view/getMilestoneHover.ts
```diff
@@ -0,0 +1,20 @@
...
+export function getMilestoneHover(
+  mouseX: number,
+  mouseY: number,
+  points: { x: number; y: number; milestone?: Milestone }[] | undefined,
+) {
+  if (points === undefined) {
+    return undefined
+  }
+  for (const [index, point] of points.entries()) {
+    if (point.milestone) {
```
sz-piotr, line 12: You should find closest instead of first to the left

sz-piotr (changes requested):

sz-piotr (approved):

PR #1069, opened 2022-12-15
Add zksync2 to new update monitor

antooni:
This PR aims to achieve two things:
- add script for automatic detection of diamond proxies (EIP-2535)
- add zkSync2 to update monitor

Resolves L2B-553


packages/backend/src/core/discovery/proxies/auto/Eip2535Proxy.ts
```diff
@@ -0,0 +1,33 @@
...
+
+  if (facets === undefined) {
+    return
+  }
+
+  return {
+    implementations: facets,
+    relatives: [],
+    upgradeability: {
+      type: 'EIP2535 proxy diamond',
```
sz-piotr, line 29:
```suggestion
      type: 'EIP2535 diamond proxy',
```

packages/backend/src/core/discovery/proxies/auto/Eip2535Proxy.ts
```diff
@@ -0,0 +1,33 @@
...
+import { ProxyDetection } from '../types'
+
+export async function detectEip2535proxy(
+  provider: DiscoveryProvider,
+  address: EthereumAddress,
+): Promise<ProxyDetection | undefined> {
+  const facets = await getCallResult<EthereumAddress[]>(
+    provider,
+    address,
+    'function facetAddresses() external view returns (address[] memory facetAd)',
```
sz-piotr, line 18: This will only work for diamonds with `IDiamondLoupe`
https://eips.ethereum.org/EIPS/eip-2535#inspecting-facets--functions

We don't know if a diamond has this so we should rely on the standard events instead:
https://eips.ethereum.org/EIPS/eip-2535#addingreplacingremoving-functions

sz-piotr (changes requested):

packages/backend/src/core/discovery/proxies/auto/Eip2535Proxy.ts
antooni: I agree this is different interface, but the EIP-2535 specification states: `Diamonds must support inspecting facets and functions by implementing the IDiamondLoupe interface`. Can't we assume that all Diamonds will implement this interface?

antooni: @sz-piotr

sz-piotr: I missed that completely when reading the standard. I think that's fine then. We'll only revisit this if we encounter future exceptions :)

sz-piotr (approved):

PR #1075, opened 2022-12-20
Create metrics module

michalsidzej:
Resolves L2B-564, L2B-565

### DIFF
--- a/packages/backend/src/Metrics.ts
+++ b/packages/backend/src/Metrics.ts
@@ -12,7 +12,7 @@ import {
 } from 'prom-client'
 
 export class Metrics {
-  init(): void {
+  collectDefaultMetrics(): void {
     collectDefaultMetrics()
   }
 
--- a/packages/backend/src/Metrics.ts
+++ b/packages/backend/src/Metrics.ts
@@ -1,0 +1,39 @@
+import {
+  collectDefaultMetrics,
+  Counter,
+  CounterConfiguration,
+  Gauge,
+  GaugeConfiguration,
+  Histogram,
+  HistogramConfiguration,
+  register,
+  Summary,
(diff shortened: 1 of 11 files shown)


packages/backend/src/Metrics.ts
```diff
@@ -0,0 +1,39 @@
...
+  async getMetrics() {
+    // we could also not use the global register, but that doesn't matter so much
+    return await register.metrics()
+  }
+
+  createCounter(configuration: CounterConfiguration<string>) {
+    return new Counter(configuration)
+  }
+
+  createGauge(configuration: GaugeConfiguration<string>) {
```
krzkaczor, line 28: explicit return types, maybe? 🤟

krzkaczor (approved):

packages/backend/src/Metrics.ts
```diff
@@ -0,0 +1,39 @@
...
+  Gauge,
+  GaugeConfiguration,
+  Histogram,
+  HistogramConfiguration,
+  register,
+  Summary,
+  SummaryConfiguration,
+} from 'prom-client'
+
+export class Metrics {
```
antooni, line 14: Why is this a class? You are neither holding any state inside it, nor injecting other services inside it?  Do you plan to do it in the future? If not, maybe a single export would be better?
 ```
export const Metrics = {
  init,
  getMetrics
}
```

antooni (changes requested):

packages/backend/src/Metrics.ts
michalsidzej: The idea behind it being a class is that 

1. it allows us to change the client library easily in the future
2. it's trivial to mock for tests

krzkaczor (approved):

PR #1076, opened 2022-12-20
Metrics auth

michalsidzej:
Resolves L2B-565

Before merging to master set env variables on staging and production

### DIFF
--- a/packages/backend/src/api/routers/MetricsRouter.ts
+++ b/packages/backend/src/api/routers/MetricsRouter.ts
@@ -8,6 +8,12 @@ import { Metrics } from '../../Metrics'
 export function createMetricsRouter(config: Config, metrics: Metrics) {
   const router = new Router()
 
+  if (!config.metricsAuth) {
+    console.warn(
+      '/metrics accessible without any authorization. This is fine for local environment but not for production',
+    )
+  }
+
   router.get('/metrics', async (ctx) => {
     const credentials = auth(ctx.req)
 
--- a/packages/backend/src/api/routers/MetricsRouter.ts
+++ b/packages/backend/src/api/routers/MetricsRouter.ts
@@ -2,6 +2,7 @@ import Router from '@koa/router'
 import auth from 'basic-auth'
 
 import { Config } from '../../config'
+import { MetricsAuthConfig } from '../../config/Config'
 import { Metrics } from '../../Metrics'
 
 export function createMetricsRouter(config: Config, metrics: Metrics) {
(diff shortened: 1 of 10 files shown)


packages/backend/src/config/Config.ts
```diff
@@ -65,3 +66,8 @@ export interface DiscoveryConfig {
   readonly alchemyApiKey: string
   readonly etherscanApiKey: string
 }
+
+export interface MetricsConfig {
```
krzkaczor, line 70: i think it would be nice to have authentication-free version for local environments. This probably should print out warning in the console.

krzkaczor (changes requested):

packages/backend/src/api/routers/MetricsRouter.ts
```diff
@@ -1,13 +1,37 @@
...
 
+import { Config } from '../../config'
+import { MetricsAuthConfig } from '../../config/Config'
 import { Metrics } from '../../Metrics'
 
-export function createMetricsRouter(metrics: Metrics) {
+export function createMetricsRouter(config: Config, metrics: Metrics) {
   const router = new Router()
 
   router.get('/metrics', async (ctx) => {
```
krzkaczor, line 11: this could use a simple test

krzkaczor (changes requested):

packages/backend/src/api/routers/MetricsRouter.ts
```diff
@@ -1,13 +1,37 @@
...
     ctx.body = await metrics.getMetrics()
   })
 
   return router
 }
+
+function check(metricsAuth: MetricsAuthConfig, name: string, pass: string) {
+  let valid = true
+
+  valid = name === metricsAuth.user && valid
```
krzkaczor, line 33: Wouldn't it be much more readable to write:

```typescript
return name === metrics.auth.user && pass === metrics.auth.pass
```

krzkaczor (changes requested):

packages/backend/src/api/routers/MetricsRouter.ts
michalsidzej: Much better in fact

krzkaczor (approved):

packages/backend/src/test/testApiServer.ts
```diff
@@ -0,0 +1,10 @@
+import Router from '@koa/router'
+import { Logger } from '@l2beat/common'
+import { agent } from 'supertest'
+
+import { ApiServer } from '../../src/api/ApiServer'
+
+export function createTestApiServer(routers: Router[]) {
```
krzkaczor, line 7: dont we have such helper somewhere already?

michalsidzej: I don't think so, that's search result on master 
Screenshot 2022-12-21 at 15 36 12

michalsidzej:
Because of changes in add metrics module, closing this PR and cherry-picking those commits to https://github.com/l2beat/l2beat/pull/1075

PR #1079, opened 2022-12-20
Bunch of random fixes

krzkaczor:
Mostly stuff that I noticed while running backend locally:
* fixed logger names for the activity
* untangled StarkexCounter code by introducing `uncertaintyBuffer` directly to SequenceProcessor
* Fix bug in StarkexCountRepository
* Improved start_db.sh script

### DIFF
--- a/packages/backend/src/core/SequenceProcessor.ts
+++ b/packages/backend/src/core/SequenceProcessor.ts
@@ -22,6 +22,7 @@ export interface SequenceProcessorOpts {
     trx: Knex.Transaction,
     logger: Logger, // logger with properly set name
   ) => Promise<void>
+  uncertaintyBuffer?: number // resync from lastProcessed - uncertaintyBuffer
   scheduleIntervalMs?: number
 }
 
@@ -48,6 +49,7 @@ interface State {
 export class SequenceProcessor extends EventEmitter {
   private readonly processQueue: TaskQueue<void>
   private readonly scheduleInterval: number
+  private readonly uncertaintyBuffer: number
   private readonly logger: Logger
   private state?: State
   private refreshId: NodeJS.Timer | undefined
@@ -77,6 +79,7 @@ export class SequenceProcessor extends EventEmitter {
       },
     )
     this.scheduleInterval = opts.scheduleIntervalMs ?? HOUR
+    this.uncertaintyBuffer = opts.uncertaintyBuffer ?? 0
   }
 
   async start(): Promise<void> {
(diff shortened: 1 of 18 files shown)


packages/backend/src/core/SequenceProcessor.ts
```diff
@@ -117,9 +121,14 @@ export class SequenceProcessor extends EventEmitter {
   private async process(): Promise<void> {
```
antooni, line 121: This function is pretty big and hard to read without greater context. Definietely some comment at the top would be useful, or to be more clean splitting it into smaller pieces to be more self-explainatory. But this one should not be in the scope of this PR.

packages/backend/src/core/SequenceProcessor.ts
```diff
@@ -117,9 +121,14 @@ export class SequenceProcessor extends EventEmitter {
...
     this.logger.debug('Processing started')
 
+    let firstRun = true
```
antooni, line 124: This variable looks a bit ugly at the first sight, but after reading this code I must say this solution is very clean.

packages/backend/src/core/SequenceProcessor.ts
```diff
@@ -47,6 +49,7 @@ interface State {
 export class SequenceProcessor extends EventEmitter {
```
antooni, line 49: contructor of this class is a bit unexpected for me, I thought that in our codebase we treat constructor as only property setting function. In this case inside the constructor `process()` is added to queque. IMO we should try to have the same behaviour repo-wise and change this flow to sth like: 
```
create object
start()
```

This should not be in the scope of this PR

antooni (approved):

michalsidzej (approved):
LGTM, but maybe @sz-piotr has something to add

packages/backend/src/peripherals/database/shared/types.ts
```diff
@@ -94,3 +94,7 @@ declare module 'knex/types/tables' {
     'activity.daily_count_view': DailyTransactionCountRow
   }
 }
+
+// Some aggregations return not empty row with null values. Use this type to explicitly type them.
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+export type OptionalDict<T = any> = Record<string, T | null>
```
sz-piotr, line 100: Maybe NullableDict instead?

packages/backend/src/core/SequenceProcessor.ts
```diff
@@ -117,9 +121,14 @@ export class SequenceProcessor extends EventEmitter {
   private async process(): Promise<void> {
     this.logger.debug('Processing started')
 
+    let firstRun = true
     // eslint-disable-next-line @typescript-eslint/no-unnecessary-condition
     processing: while (true) {
-      const lastProcessed = this.state?.lastProcessed
+      let lastProcessed = this.state?.lastProcessed
+      // we need to adjust starting block if its first run and uncertaintyBuffer is set
+      if (firstRun && lastProcessed && this.uncertaintyBuffer > 0) {
```
sz-piotr, line 129: This implicitly relies on heroku's behavior of restarting the process every day. Maybe if uncertainty buffer is set we should always set `lastProcessed = actualLastProcessed - uncertaintyBuffer`? Or only if we're on the tip?

krzkaczor: I am not sure if I understand, `process()` gets called on hourly basis so it should work fine. Please ping me if I am wrong but I am gonna merge it now.

PR #1082, opened 2022-12-21
Add api metrics middleware

michalsidzej:
Resolves L2B-566

During the review please mind the code repetitions between `createApiLogger` and `createApiMetrics`. Alternatively, we can use a approach analogical to the following: 
```
// logger

app.use(async (ctx, next) => {
  await next();
  const rt = ctx.response.get('X-Response-Time');
  console.log(`${ctx.method} ${ctx.url} - ${rt}`);
});

// x-response-time

app.use(async (ctx, next) => {
  const start = Date.now();
  await next();
  const ms = Date.now() - start;
  ctx.set('X-Response-Time', `${ms}ms`);
});
```
This would give us less accurate timings but would be more readable and we would have the same exact time logged and saved as a metric.

### DIFF
--- a/packages/backend/src/api/middleware/metrics.ts
+++ b/packages/backend/src/api/middleware/metrics.ts
@@ -24,10 +24,12 @@ export function createApiMetrics(metrics: Metrics): Middleware {
 
       const timeMs = Date.now() - start
 
+      // eslint-disable-next-line @typescript-eslint/no-unsafe-assignment
+      const path = (ctx._matchedRoute as string | undefined) ?? ctx.path
+
       apiHistogam
         .labels({
-          // TODO: normalizePath
-          path: ctx.path,
+          path,
           method: ctx.method,
           status_code: ctx.status,
         })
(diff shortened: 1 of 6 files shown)


packages/backend/src/api/middleware/metrics.ts
```diff
@@ -0,0 +1,45 @@
...
+    const { res } = ctx
+
+    const done = () => {
+      res.removeListener('finish', done)
+      res.removeListener('close', done)
+
+      const timeMs = Date.now() - start
+
+      // eslint-disable-next-line @typescript-eslint/no-unsafe-assignment
+      const path = (ctx._matchedRoute as string | undefined) ?? ctx.path
```
krzkaczor, line 31: matched route would include parameter ranges like `user/:id` right? i think this could be explained in a comment

packages/backend/src/api/middleware/metrics.ts
```diff
@@ -0,0 +1,45 @@
+import { Context, Middleware, Next } from 'koa'
+
+import { Metrics } from '../../Metrics'
+
+export function createApiMetrics(metrics: Metrics): Middleware {
+  const labels = ['path', 'method', 'status_code']
+  const apiHistogam = metrics.createHistogram({
+    name: 'api_request_duration_seconds_sum',
```
krzkaczor, line 8: should `_sum` postifx be part of histogram metric?

packages/backend/src/api/middleware/metrics.ts
```diff
@@ -0,0 +1,45 @@
...
+    help:
+      'duration histogram of  api http responses labeled with: ' +
```
krzkaczor, line 10: double space

krzkaczor (approved):

PR #1085, opened 2022-12-22
Measure repository methods execution times

michalsidzej:
Resolves L2B-90

Many files changed, but mostly passing the `histogram` object down the pipe.

Simple (maybe naive?) approach to adding repository methods execution times measurements. We want all the repositories to use the same `Histogram` instance, so we should instantiate it on the top level and pass it to each module (probably could use a function to extract this logic from `Application.ts`. I was thinking about adding this metric to `Metrics` class, but I'm not sure if it should be aware of this kind of logic. Help here is appreciated.

TODO: add test to BaseRepository to check if we get a measurement

### DIFF
--- a/packages/backend/src/Application.ts
+++ b/packages/backend/src/Application.ts
@@ -10,7 +10,6 @@ import { createDiscoveryModule } from './modules/discovery/DiscoveryModule'
 import { createHealthModule } from './modules/health/HealthModule'
 import { createMetricsModule } from './modules/metrics/MetricsModule'
 import { createTvlModule } from './modules/tvl/TvlModule'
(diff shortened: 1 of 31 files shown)


packages/backend/src/Application.ts
```diff
@@ -30,12 +31,31 @@ export class Application {
     )
 
     const metrics = new Metrics()
+    const repositoryHistogram: RepositoryHistogram = metrics.createHistogram({
```
sz-piotr, line 34: Move this inside Metrics constructor and expose something like `getRepositoryHistogram`.

krzkaczor (approved):

PR #1086, opened 2022-12-22
Add TwitterClient

antooni:
This PR aims to add TwitterClient able to create tweets from authorized account. 

I decided to solve communication with Twitter API using [twitter-api-v2](https://www.npmjs.com/package/twitter-api-v2?activeTab=explore) package, which handles OAuth 1.0a authentication out of the box. This API does not add any external dependencies, what it does is: it takes your message, constructs request, generate OAuth header and send request using node's `http`. Using this library introduces security risks, we are trusting the library maintainer to not extract our API keys via malicious code upgrade. But without it I would have to spend X amount of time to construct HTTP request myself and test it well. Provided this is an experimental feature, spending too much time on it is not a desired outcome.

I have done some research to do this more 'low-level' either via constructing Authentication header in code, or using the library for OAuth 1.0, but both cases would have produced harder to maintain code IMO.

Resolves L2B-557

### DIFF
--- a/packages/backend/package.json
+++ b/packages/backend/package.json
@@ -44,6 +44,7 @@
     "react-dom": "^18.1.0",
     "rollbar": "^2.25.2",
     "source-map-support": "^0.5.21",
+    "twitter-api-v2": "^1.12.10",
     "zod": "^3.17.3"
   },
   "devDependencies": {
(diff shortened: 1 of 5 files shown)


michalsidzej (commented):
LGTM

PR #1089, opened 2023-01-03
Update coingeckoId for Fantom

antooni:

PR #1090, opened 2023-01-03
Hide dydx activity

michalsidzej:
There's some problem with starkex activity API which causes our website to not build. Let's hide dydx and investigate the API issue with starkware.


antooni (approved):

krzkaczor (commented):
As i explained here: https://github.com/l2beat/l2beat/pull/1089 i am not sure about fantom coingecko id change.

PR #1095, opened 2023-01-04
Fix CoingeckoAPI id errors

antooni:
This PR aims to solve two things:
- build time errors, to be more precise CI/CD, currently the test do not pass although the coingeckoId is still supported, simple change to the test resolved it
- runtime errors, when the server started and during the runtime Coingecko decides to change id, the fetchnig function will catch the error, ask for the list of supported tokens, and based on the address will try to find new matching id

Code is a bit ugly, I focused more on the PoC, if it is a good direction I will sure refactor it.

Resolves L2B-581

### DIFF
--- a/packages/common/src/services/coingecko/CoingeckoClient.ts
+++ b/packages/common/src/services/coingecko/CoingeckoClient.ts
@@ -1,4 +1,5 @@
 import { CoingeckoId, UnixTime } from '@l2beat/types'
+import { getTokenByCoingeckoId } from '@l2beat/config'
 
 import { RateLimiter } from '../../tools/RateLimiter'
 import { HttpClient } from '../HttpClient'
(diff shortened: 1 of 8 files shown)


packages/common/src/services/coingecko/CoingeckoClient.ts
```diff
@@ -49,33 +49,73 @@ export class CoingeckoClient {
...
-      `/coins/${coinId.toString()}/market_chart/range`,
-      {
-        vs_currency: vs_currency.toLowerCase(),
-        from: from.toString(),
-        to: to.toString(),
-      },
-    )
-
-    const parsedData = CoinMarketChartRangeResult.parse(data)
+    try {
```
sz-piotr, line 54: I like the idea, dislike the implementation. Ideally this function is super simple, like 3 lines: 1. call `market_chart/range` 2. if error get new coingeckoId 3. call `market_chart/range` with new id

Steps 1 and 3 get a new function, step 2 gets a new function.

The function for getting new coingecko id should ideally cache the result and only re-fetch every hour (configurable).

sz-piotr (changes requested):

packages/common/src/services/coingecko/CoingeckoClient.ts
```diff
@@ -90,10 +131,18 @@ export class CoingeckoClient {
     }
     const res = await this.httpClient.fetch(url, { timeout: this.timeoutMs })
     if (!res.ok) {
+      const body = JSON.stringify(await res.json())
```
sz-piotr: u
```suggestion
      const body = await res.text()
```

packages/common/src/services/coingecko/CoingeckoClient.ts
```diff
@@ -79,6 +96,30 @@ export class CoingeckoClient {
...
+      return (
+        addr?.toLocaleLowerCase() === address.toString().toLocaleLowerCase()
+      )
+    })
+    if (coingeckoSupported?.id === undefined) {
+      throw new Error(
+        `Server responded with non-2XX result: Could not fetch the prices for ${oldId.toString()}`,
+      )
+    }
+    return coingeckoSupported.id
```
sz-piotr, line 120: You are not remembering the new id, so next time you have to query it again. I'd keep an id map storing the latest ids so that you can only query once every change

sz-piotr (changes requested):

sz-piotr (approved):

PR #1098, opened 2023-01-04
Change tests location

torztomasz:
Resolves L2B-518 and L2B-519.

It moves the test folder inside the src directory. Now the test directory contains only some setup/utils files and tests got moved to the directory where the file that is being tested is located.
It also improves the .eslintrc.js file of each package.


packages/frontend/gulpfile.js
```diff
@@ -46,7 +46,7 @@ function watchStyles() {
 }
 
 function copyStatic() {
-  return gulp.src('src/static/**/*').pipe(gulp.dest('build'))
+  return gulp.src(['src/static/**/*']).pipe(gulp.dest('build'))
```
krzkaczor, line 49: why change this?

krzkaczor (commented):
Didn't this require changes to tsconfig that we discussed?

packages/frontend/gulpfile.js
torztomasz: I wanted to close this task and then move to the tsconfig changes

krzkaczor (approved):

PR #1101, opened 2023-01-05
Backend watch service MVP

antooni:
This PR aims to add module responsible for automatic trigger of discovery to our backend server. 
From the birds-eye-view it just runs discovery over all of the configs located inside discovery directory.
When looking closer it involves some refactoring, mostly to hide logs.
Resolves L2B-575


packages/backend/src/core/discovery/handlers/system/ErrorHandler.ts
```diff
@@ -1,12 +1,16 @@
...
 
 export class ErrorHandler implements Handler {
   readonly dependencies = []
   private readonly errorMessage: string
 
-  constructor(readonly field: string, error: unknown) {
+  constructor(
+    readonly field: string,
+    error: unknown,
+    readonly discoveryLogger: DiscoveryLogger,
```
sz-piotr, line 12: Does it need a logger?

packages/common/src/tools/DiscoveryLogger.ts
```diff
@@ -0,0 +1,48 @@
+import chalk from 'chalk'
+
+interface LoggerOptions {
+  enabled: boolean
+}
+
+export class DiscoveryLogger {
```
sz-piotr, line 7: I am not sure this should be in common. This is only used in backend discovery.

sz-piotr (changes requested):

sz-piotr (approved):

PR #1105, opened 2023-01-09
Ignore unnecessary files in build output

torztomasz:
Resolves L2B-582

Ignore unnecessary files in build output (node_modules, build output dir, test files). I am 99% sure that frontend already excluded these unnecessary files.


packages/backend/tsconfig.json
```diff
@@ -2,5 +2,6 @@
   "extends": "../../tsconfig.json",
   "compilerOptions": {
     "outDir": "build"
-  }
+  },
+  "exclude": ["build", "node_modules"]
```
sz-piotr, line 6: Why do we need to modify the original tsconfig?

packages/types/tsconfig.json
```diff
@@ -2,5 +2,6 @@
   "extends": "../../tsconfig.json",
   "compilerOptions": {
     "outDir": "build"
-  }
+  },
+  "exclude": ["build"]
```
sz-piotr, line 6: Do we also need node_modules here?

packages/backend/tsconfig.json
torztomasz: There is no reason to include build and node_modules when using typecheck or smth similiar.

packages/types/tsconfig.json
torztomasz: There is no node_modules in types package

sz-piotr: I think I'd still add this a) for consistency, but also b) because if some dependencies are added node_modules might appear and break builds

sz-piotr (approved):

PR #1106, opened 2023-01-09
Update optimism.ts

Joxess:
Fixing explorer URL and adding OPLabs social media


Joxess:
Fixing explorer URL and adding OPLabs social media

sz-piotr (approved):

PR #1112, opened 2023-01-10
Fix styling issues

torztomasz:
Resolves L2B-586, L2B-545

This PR contains a lot of small UI fixes.


packages/frontend/src/components/header/DetailsHeader.stories.tsx
```diff
@@ -32,7 +32,7 @@ function Template({ projectId }: TemplateProps) {
       value: <StatWithChange stat="2.21" change="-10.23%" />,
     },
     {
-      title: '30d Transaction Count',
+      title: '30D ts count',
```
sz-piotr, line 35:
```suggestion
      title: '30D tx count',
```

packages/frontend/src/pages/scaling-projects/view/ProjectHeader.tsx
```diff
@@ -47,7 +47,7 @@ export function ProjectHeader(props: ProjectHeaderProps) {
         ),
     },
     {
-      title: '30d transaction count',
+      title: '30D ts count',
```
sz-piotr, line 50:
```suggestion
      title: '30D tx count',
```

packages/frontend/src/components/HorizontalSeparator.tsx
```diff
@@ -0,0 +1,16 @@
+import cx from 'classnames'
+import React from 'react'
+export interface SeparatorProps {
+  className?: string
+}
+export function HorizontalSeparator(props: SeparatorProps) {
+  const { className } = props
```
sz-piotr, line 7: We never use this convention anywhere. We do:
- either destructure in the function signature, e.g. `function Component({ a, b }: ComponentProps`
- or just use `props.a` and `props.b` directly

sz-piotr: You can also update KnowledgeNuggetItem to follow this :) as I missed it during your previous PR

sz-piotr (changes requested):

packages/frontend/src/components/HorizontalSeparator.tsx
torztomasz: Okay, I have to change my habits from previous projects 👍

sz-piotr (approved):

PR #1114, opened 2023-01-10
Add database per request logging, improve existing wrap methods

krzkaczor:
I decided to make mutating methods of repositories log to info level instead of debug.

btw. it should be possible to replace manual wrapping with "automatic" one in the constructor of `BaseRepository`.


packages/backend/src/peripherals/database/shared/BaseRepository.ts
```diff
@@ -62,19 +62,24 @@ export class BaseRepository {
...
-      const count =
-        typeof idsOrCount === 'number' ? idsOrCount : idsOrCount.length
-      this.logger.debug({ method: method.name, count })
+
       return idsOrCount
     }
     Object.defineProperty(fn, 'name', { value: method.name })
-    return fn
+
+    return this.wrap(fn, log)
```
krzkaczor, line 82: addMany methods weren't tracked in metrics 😨 CC: @michalsidzej

packages/backend/src/peripherals/database/shared/BaseRepository.ts
```diff
@@ -49,7 +49,7 @@ export class BaseRepository {
     method: AddMethod<T, R>,
   ): AddMethod<T, R> {
     return this.wrap(method, (id) =>
-      this.logger.debug({ method: method.name, id: id.valueOf() }),
+      this.logger.info({ method: method.name, id: id.valueOf() }),
```
sz-piotr, line 52: Using info will produce too many logs for papertrail, so it's actually not useful because we'll get cut off early

sz-piotr (changes requested):

packages/backend/src/Application.ts
```diff
@@ -64,6 +64,10 @@ export class Application {
           await database.rollbackAll()
         }
         await database.migrateToLatest()
+
+        if (config.logger.logLevel >= LogLevel.DEBUG) {
```
sz-piotr, line 68: Move this check to `database.enableQueryLogging()`

sz-piotr (changes requested):

sz-piotr (approved):

PR #1115, opened 2023-01-11
Activity metrics

michalsidzej:
Resolves L2B-594


krzkaczor (approved):

packages/backend/src/Metrics.ts
```diff
@@ -12,16 +12,52 @@ import {
...
+      labelNames: ['project'],
+    })
+
+    this.activityLatest = this.createGauge({
+      name: 'activity_latest',
+      help: 'Latest existing unit to be synced to',
+      labelNames: ['project'],
+    })
+
+    this.activityIncludedInApi = this.createGauge({
```
torztomasz, line 45: I don't know if there is some use case in Grafana for activityIncludedInApi One idea that comes to my mind is that there could be some graph to show when we excluded some activity from API but I do not know if it is useful information. Also, the only use case for activity config is to show the config in Grafana but I would assume that it is unnecessary. @krzkaczor @sz-piotr do you agree?

torztomasz (changes requested):

packages/backend/src/Metrics.ts
michalsidzej: @torztomasz We agreed to keep that. It's cheep to store it in Prometheus and will give us a retrospective view in the application config.

torztomasz (approved):

PR #1118, opened 2023-01-16
Create Dadudida Web3 SECURITY.md

Dadudidas:


sz-piotr:
Hey @Dadudidas. Thanks for your contribution. As our project is not a dependency of other project it doesn't make sense for us to support past versions with security updates or have vulnerability reporting policies.

PR #1121, opened 2023-01-17
Truncate Discord message

antooni:
This PR aims to cut discord message into pieces when it exceedes Discord API`BASE_TYPE_MAX_LENGTH` (2000 characters). Message is semantically cut inside `diffToMessage`, additionally the check inside `DiscordClient` was added.

Resolves L2B-635

### DIFF
--- a/packages/backend/src/core/DiscoveryWatcher.ts
+++ b/packages/backend/src/core/DiscoveryWatcher.ts
@@ -94,8 +94,10 @@ export class DiscoveryWatcher {
     )
 
     if (diff.length > 0) {
-      const message = diffToMessage(name, diff)
-      await this.notify(message)
+      const messages = diffToMessage(name, diff)
+      for (const message of messages) {
+        await this.notify(message)
+      }
     }
   }
 
--- a/packages/backend/src/core/DiscoveryWatcher.ts
+++ b/packages/backend/src/core/DiscoveryWatcher.ts
@@ -71,11 +71,6 @@ export class DiscoveryWatcher {
         this.logger.error(error)
       }
     }
-
-    await this.notify(
-      `Run discovery for all projects | block_number = ${blockNumber}`,
-    )
-
     this.logger.info('Update finished', { blockNumber })
   }
 
@@ -95,13 +90,11 @@ export class DiscoveryWatcher {
 
     if (diff.length > 0) {
       const messages = diffToMessage(name, diff)
-      for (const message of messages) {
-        await this.notify(message)
-      }
+      await this.notify(messages)
     }
   }
 
-  async notify(message: string) {
+  async notify(messages: string[]) {
     if (!this.discordClient) {
       // TODO: maybe only once? rethink
       this.logger.info(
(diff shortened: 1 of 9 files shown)


packages/backend/src/core/discovery/utils/diffToMessage.ts
```diff
@@ -1,26 +1,27 @@
 import { DiscoveryDiff } from './diffDiscovery'
 
-export function diffToMessage(name: string, diff: DiscoveryDiff[]): string[] {
-  const beginning = `Detected changes for ${name}\n\n\`\`\`diff\n`
-  const end = '```'
+export function diffToMessage(name: string, diffs: DiscoveryDiff[]): string[] {
+  const prefix = `Detected changes for ${name}\n\n\`\`\`diff\n`
+  const postfix = '```'
```
torztomasz, line 5: This ``` in prefix and postfix (by the way it should probably be suffix) could be extracted to a function called wrapCodeBlock(message: string). This would improve readability.

packages/backend/src/core/DiscoveryWatcher.ts
```diff
@@ -94,8 +94,10 @@ export class DiscoveryWatcher {
     )
 
     if (diff.length > 0) {
-      const message = diffToMessage(name, diff)
-      await this.notify(message)
+      const messages = diffToMessage(name, diff)
+      for (const message of messages) {
```
torztomasz, line 98: Maybe instead of calling notify multiple times, we also accept string[] as an argument of notify and handle message arrays inside the function

packages/backend/src/core/discovery/utils/diffToMessage.ts
antooni: but what about overheadLenght?

torztomasz: hmmm, maybe you can extract it to prefix and content where content would be wrapCodeBlock(diff) then there is no need for suffix

packages/backend/src/core/discovery/utils/diffToMessage.ts
```diff
@@ -1,30 +1,33 @@
 import { DiscoveryDiff } from './diffDiscovery'
 
 export function diffToMessage(name: string, diffs: DiscoveryDiff[]): string[] {
```
torztomasz, line 3: I was thinking more about something like this:
```
export function diffToMessage(name: string, diffs: DiscoveryDiff[]): string[] {
  const header = wrapBold(`Detected changes for ${name}`)
  const messages: string[][] = [[]]
  let index = 0

  for (const diff of diffs) {
    const currentLength = header.length + wrapCodeBlock(messages[index]).length
    const nextDiff = diffToString(diff)

    if (currentLength + nextDiff.length >= 2000) {
      index += 1
      messages[index] = []
    }

    messages[index].push(nextDiff)
  }

  const result = messages.map((m) => wrapCodeBlock(m))
  return result
}

export function wrapBold(message: string) {
  return `**${message}**`
}

export function wrapCodeBlock(messages: string[]) {
  const prefix = '```\n'
  const postfix = '\n```'

  return `${prefix}${messages.join('\n')}${postfix}`
}
```

torztomasz (changes requested):

packages/backend/src/core/discovery/utils/diffToMessage.ts
torztomasz: We can get on call and talk about this 👍

antooni: I wanted to wrap all the cahnges connected with a project into one block of code, I think it would be more readable, WDYT?

torztomasz (approved):

PR #1122, opened 2023-01-17
Add generic wrapping for all methods in repo class

torztomasz:
Resolves L2B-602

This PR aims to minimize the manual wrapping of repository class methods by wrapping them based on the naming convention.


sz-piotr:
I spent some time thinking about this and I have some solution. The code below will not compile if any of the methods have wrong signatures. If this approach makes sense, then `CheckConvention` and `autoWrap` will go to `BaseRepository`.

@krzkaczor I'd love to get your feedback on this.

```ts
import { AssetId, UnixTime } from '@l2beat/types'
import { BaseRepository } from './shared/BaseRepository'

type AddMethod<T, R> = (record: T) => Promise<R>
type AddManyMethod<T, R> = (records: T[]) => Promise<R[] | number>
type SaveMethod<T> = (record: T) => Promise<boolean>
type GetMethod<A extends unknown[], T> = (...args: A) => Promise<T[]>
type FindMethod<A extends unknown[], T> = (...args: A) => Promise<T | undefined>
type DeleteMethod<A extends unknown[]> = (...args: A) => Promise<number>

type Keys<T, U> = Extract<keyof T, U>
type Match<T, U> = T extends U ? T : never

/* eslint-disable @typescript-eslint/no-explicit-any */

type CheckConvention<T> = {
  [K in Keys<T, `add${string}`>]: Match<T[K], AddMethod<any, any>>
} & {
  [K in Keys<T, `addMany${string}`>]: Match<T[K], AddManyMethod<any, any>>
} & {
  [K in Keys<T, `save${string}`>]: Match<T[K], SaveMethod<any>>
} & {
  [K in Keys<T, `find${string}`>]: Match<T[K], FindMethod<any[], any>>
} & {
  [K in Keys<T, `get${string}`>]: Match<T[K], GetMethod<any[], any>>
} & {
  [K in Keys<T, `delete${string}`>]: Match<T[K], DeleteMethod<any[]>>
}

/* eslint-enable @typescript-eslint/no-explicit-any */

export class TestRepository extends BaseRepository {
  constructor(...args: ConstructorParameters<typeof BaseRepository>) {
    super(...args)
    // eslint-disable-next-line @typescript-eslint/unbound-method
    this.wrapAny(this.thingThatShouldBeWrappedAny)
    this.autoWrap<CheckConvention<TestRepository>>(this)
  }

  autoWrap<T>(obj: T) {}

  async addOne(a: string): Promise<number> {
    return 1
  }

  async addTwo(a: string): Promise<number> {
    return 1
  }

  async addManyWithCount(a: string[]): Promise<number> {
    return 1
  }

  async addManyWithIds(a: string[]): Promise<number[]> {
    return [1]
  }

  async save(a: number): Promise<boolean> {
    return true
  }

  async getAll(): Promise<string[]> {
    return []
  }

  async getByFoo(foo: string): Promise<string[]> {
    return []
  }

  async findById(a: string, b: string): Promise<Map<AssetId, UnixTime>> {
    return new Map()
  }

  async findByMagic(a: string, b: string): Promise<string | undefined> {
    return undefined
  }

  async deleteAll(): Promise<number> {
    return 1
  }

  async deleteBetween(a: string, b: string): Promise<number> {
    return 1
  }

  async thingThatShouldBeWrappedAny() {
    return 1
  }
}
```

krzkaczor:
Yes, this is perfect @sz-piotr . @torztomasz you gotta **believe** in TS's type system 😆

sz-piotr (dismissed):

packages/backend/src/peripherals/database/shared/BaseRepository.ts
```diff
@@ -20,9 +20,25 @@ type FindMethod<A extends unknown[], T> = (...args: A) => Promise<T | undefined>
...
+  [K in Keys<T, `find${string}`>]: Match<T[K], FindMethod<any[], any>>
+} & {
+  [K in Keys<T, `get${string}`>]: Match<T[K], GetMethod<any[], any>>
+} & {
+  [K in Keys<T, `delete${string}`>]: Match<T[K], DeleteMethod<any[]>>
+}
+/* eslint-enable @typescript-eslint/no-explicit-any */
 
-export class BaseRepository {
+export abstract class BaseRepository {
```
krzkaczor, line 41: Please, add a comment to this class explaining the conventions used here. Also, what exactly is the difference between findX and getX? This should be explained in the comment too! CC: @sz-piotr

krzkaczor (changes requested):

packages/backend/src/peripherals/database/shared/BaseRepository.ts
torztomasz: Sure, gonna do that in a minute. The difference is that get is returning list of records and find may return single record or undefined.

krzkaczor: That's what I thought... i was confused by the signature in CheckConvetion (`any[]` type). It's clear after reading `FindMethod` and `GetMethod` types 👍 But anyway, comment is needed.

packages/backend/src/peripherals/database/shared/BaseRepository.ts
```diff
@@ -20,9 +20,48 @@ type FindMethod<A extends unknown[], T> = (...args: A) => Promise<T | undefined>
...
+  [K in Keys<T, `find${string}`>]: Match<T[K], FindMethod<any[], any>>
+} & {
+  [K in Keys<T, `get${string}`>]: Match<T[K], GetMethod<any[], any>>
+} & {
+  [K in Keys<T, `delete${string}`>]: Match<T[K], DeleteMethod<any[]>>
+}
+/* eslint-enable @typescript-eslint/no-explicit-any */
+
+/* 
+  This class enforces its child classes to persist given naming convention of methods and wrap them with logger and metrics.
```
krzkaczor, line 42: enforces -> enables?

krzkaczor (approved):

PR #1125, opened 2023-01-18
Discovery polygonpos

Wendrowiec13:


packages/config/src/bridges/polygonpos.ts
```diff
@@ -136,24 +136,21 @@ export const polygonpos: Bridge = {
     isIncomplete: true,
     addresses: [
       {
-        address: '0x40ec5B33f54e0E8A33A975908C5BA1c14e5BbbDf',
```
michalsidzej, line 139: Is the change in contracts order an intended change? This will be reflected on the frontend

Wendrowiec13: Yes, it was consulted with @antooni

michalsidzej (approved):

PR #1128, opened 2023-01-20
Fix diff discovery

michalsidzej:
Fix `diffDiscovery` taking into account `{key: undefined}`


packages/backend/src/core/DiscoveryWatcher.ts
```diff
@@ -85,11 +85,11 @@ export class DiscoveryWatcher {
     overrides?: Record<string, DiscoveryContract>,
   ) {
     const committed = await this.configReader.readDiscovery(name)
-    const discoveredAsCommitted = prepareDiscoveryFile(discovered)
+    const parsedDiscovery = parseDiscoveryOutput(discovered)
```
antooni, line 88: shouldnt there be prepareDiscoveryFile?

antooni (approved):

PR #1134, opened 2023-01-23
Add Gitcoin banner

torztomasz:
Resolves L2B-643

### DIFF
--- a/packages/frontend/src/build/config/common.ts
+++ b/packages/frontend/src/build/config/common.ts
@@ -12,7 +12,7 @@ export const common: Omit<Config, 'backend'> = {
     forum: 'https://gov.l2beat.com/',
   },
   features: {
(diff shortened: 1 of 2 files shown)


sz-piotr (approved):

packages/frontend/src/build/config/common.ts
```diff
@@ -12,7 +12,7 @@ export const common: Omit<Config, 'backend'> = {
     forum: 'https://gov.l2beat.com/',
   },
   features: {
-    banner: false,
+    banner: true,
```
sz-piotr, line 15: actually can you set this to automatically hide at the end of the round?

sz-piotr (changes requested):

sz-piotr (approved):

PR #1137, opened 2023-01-24
Update project dependencies

michalsidzej:
Resolves L2B-667


torztomasz (approved):

krzkaczor (approved):

packages/common/src/tools/promiseAllPlus.ts
```diff
@@ -59,6 +59,7 @@ export async function promiseAllPlus<T>(
   await taskQueue.waitTilEmpty()
 
   if (permanentError) {
+    // eslint-disable-next-line @typescript-eslint/no-throw-literal
```
krzkaczor, line 62: Possibly a stupid rule that we could disable for the whole codebase.

michalsidzej: @krzkaczor Agree, disabled

PR #1140, opened 2023-01-24
Add TVL metrics

michalsidzej:
Resolves L2B-580


packages/common/src/tools/queue/TaskQueue.ts
```diff
@@ -153,3 +160,15 @@ export class TaskQueue<T> {
     }
   }
 }
+
+function wrapAndMeasure<T extends unknown[], U>(
```
krzkaczor, line 164: i wonder can this be shared across this and BaseRepository's wrap

packages/backend/src/Metrics.ts
```diff
@@ -58,6 +60,13 @@ export class Metrics {
         'uncertaintyBuffer',
       ],
     })
+
+    this.tvlHistogram = this.createHistogram({
+      name: 'tvl_sync_histogram',
```
krzkaczor, line 65: I think this deserves a better description. 
`tvl_sync_histogram` -> `tvl_sync_duration_histogram`
`Histogram showing tvl core sync speed` -> `Histogram showing TVL sync duration`

packages/backend/src/core/reports/ReportUpdater.ts
```diff
@@ -26,12 +27,19 @@ export class ReportUpdater {
...
     private readonly logger: Logger,
+    metrics: Metrics,
   ) {
     this.logger = this.logger.for(this)
     this.configHash = getReportConfigHash(projects)
     this.taskQueue = new TaskQueue(
       (timestamp) => this.update(timestamp),
       this.logger.for('taskQueue'),
+      {
+        metrics: {
```
krzkaczor, line 38: Maybe we need something like `this.metrics.for(this)`.

packages/common/src/tools/queue/TaskQueue.ts
michalsidzej: That was probably my goal looking at the generic types here. Will do

krzkaczor (approved):
Accepted but this should be refactored in the future.

PR #1141, opened 2023-01-24
Increase StarkEx API delay

krzkaczor:
Build failures are connected with StarkEx API processing delay. This results in APIs returning zeroes when we expect a non-zero value for a given day. This PR increases the delay, which *should* fix the problem because API won't return "unaligned" data (when one of the processors is lagging behind).

image


packages/backend/src/core/activity/counters/StarkexCounter.ts
```diff
@@ -62,5 +62,5 @@ export function createStarkexCounter(
 }
 
 function getStarkexLastDay(timestamp: UnixTime) {
-  return timestamp.toStartOf('day').add(-1, 'days').toDays()
+  return timestamp.toStartOf('day').add(-1, 'days').add(-4, 'hours').toDays() // StarkEx APIs have inherent delay
```
antooni, line 65: This fix looks great. I only wonder if we could make this offset even bigger? I am not aware about the amount of computation that it will include, so maybe it is not the best idea. But I am also worried that some day they change the delay and our code will stop working.

antooni (approved):

packages/backend/src/core/activity/counters/StarkexCounter.ts
michalsidzej: This does not increase the amount of computation, it just makes StarkexSequenceProcessors start the sync 4 hours later.

michalsidzej (approved):

packages/backend/src/core/activity/counters/StarkexCounter.ts
krzkaczor: This will make update to all processors delayed by 4 hours aka we will show new data on 4am in the morning. That's why I want to keep this value minimal.

PR #1144, opened 2023-01-25
Add job link to navbar

torztomasz:
Resolves L2B-644


packages/frontend/src/components/badge/Badge.tsx
```diff
@@ -0,0 +1,57 @@
...
+  ERROR = 'ERROR',
+  GRAY = 'GRAY',
+  BRIGHT_YELLOW = 'BRIGHT_YELLOW',
+  PURPLE = 'PURPLE',
+}
+
+export enum BadgeSize {
+  MD = 'MD',
+  SM = 'SM',
+}
```
michalsidzej, line 15: Maybe we can use Union Types for those? It will result in slightly shorter code, and we'll get type safety anyway.

michalsidzej (commented):
Were slight the changes is all badges intentional? Checkout the storybook diff report.

packages/frontend/src/components/badge/Badge.tsx
torztomasz: I was also unsure about this one, I think we can do it the way you are talking about.

packages/frontend/src/components/badge/Badge.tsx
```diff
@@ -0,0 +1,43 @@
...
+
+export interface BadgeProps {
+  type: BadgeType
+  className?: string
+  children: ReactNode
+  title?: string
+  oneSize?: boolean
+}
+
+const badgeClassnames: { [key in BadgeType]: string } = {
```
michalsidzej, line 14:
```suggestion
const badgeClassnames: Record<BadgeType, string> = {
```

michalsidzej (approved):

PR #1145, opened 2023-01-25
Implement log throttling mechanism

torztomasz:
Resolves L2B-604


packages/common/src/tools/Logger.ts
```diff
@@ -83,6 +88,9 @@ export class Logger {
   }
 
   private print(level: string, parameters: LoggerParameters) {
+    if (this.loggerHistory.includes(parameters)) return
```
krzkaczor, line 91: i think we need a more complicated criterion for skipping logs here. 

It should work so that if there is a repeated log for x times in y seconds, we throttle it for z seconds. Furthermore, once we are done with throttling it would be amazing to write a summary like "message A was throttled b times during last z seconds". @sz-piotr thoughts? If this should be better defined i can write a longer explanation in the ticket.

krzkaczor (changes requested):

packages/common/src/tools/LogThrottler.ts
```diff
@@ -0,0 +1,71 @@
...
+  count: number
+  throttleCount: number
+  isThrottling: boolean
+}
+
+export class LogThrottler {
+  private readonly recentLogs: Map<string, LogInfo>
+
+  constructor(
+    private readonly threshold: number,
```
krzkaczor, line 11: pass it as a single config object to increase redability

packages/common/src/tools/Logger.ts
```diff
@@ -20,7 +21,11 @@ export interface LoggerOptions {
 export type LoggerParameters = Record<string, json>
 
 export class Logger {
-  constructor(private readonly options: LoggerOptions) {}
+  private readonly logThrottler: LogThrottler
+
+  constructor(private readonly options: LoggerOptions) {
+    this.logThrottler = new LogThrottler(5, 10000, 20000)
```
krzkaczor, line 27: make these parameters part of the logger options along side "throttling: boolean" to enable/disable this feature

packages/common/src/tools/LogThrottler.ts
```diff
@@ -0,0 +1,71 @@
...
+    }
+
+    this.incrementCount(logInfo)
+
+    if (logInfo.count >= this.threshold && !logInfo.isThrottling) {
+      this.throttle(logKey, logInfo)
+    }
+  }
+
+  isThrottling(logKey: string) {
```
krzkaczor, line 37: explicit return types please

packages/common/src/tools/LogThrottler.ts
```diff
@@ -0,0 +1,71 @@
...
+    return this.recentLogs.get(logKey)?.isThrottling
+  }
+
+  private throttle(logKey: string, logInfo: LogInfo) {
+    logInfo.isThrottling = true
+
+    setTimeout(() => {
+      if (logInfo.throttleCount !== 0) {
+        console.log(
+          `[LOG THROTTLER] ${logKey} was logged ${
```
krzkaczor, line 47: we could use "our" logger here but with throttling disabled.

packages/common/src/tools/LogThrottler.ts
```diff
@@ -0,0 +1,71 @@
+interface LogInfo {
+  count: number
+  throttleCount: number
+  isThrottling: boolean
+}
+
+export class LogThrottler {
```
krzkaczor, line 7: I think we need tests for this. However, more importantly, I am worried that this implementation might be too resource heavy -- it creates a timer for every single thing logged. 

Maybe we can create something with a slightly different behavior but much more performant. What if we count logs only that happened last minute ie. clear `recentLogs` every minute. (This allows you to get rid of `setTimeout` for every log message). From practical PoV, I think it should work the same way.

krzkaczor (changes requested):

packages/common/src/tools/LogThrottler.ts
torztomasz: Yeah, I was thinking about something similar to the idea you proposed. The only drawback of this approach is that instead of checking if x logs have been called during y time, we would check if x logs have been called within y time span. But regarding the performance, this would be better and should not cause any issues.

packages/common/src/tools/LogThrottler.ts
torztomasz: I implemented this before and got lint errors
image

packages/backend/src/config/config.local.ts
```diff
@@ -34,6 +34,11 @@ export function getLocalConfig(cli: CliParameters): Config {
     logger: {
       logLevel: getEnv.integer('LOG_LEVEL', LogLevel.INFO),
       format: 'pretty',
+      throttlerOptions: {
```
krzkaczor, line 37: i think we might not need throttling locally.

packages/common/src/tools/LogThrottler.ts
```diff
@@ -0,0 +1,81 @@
...
+
+interface LogInfo {
+  count: number
+  throttleCount: number
+  isThrottling: boolean
+}
+
+export interface LogThrottlerOptions {
+  threshold: number
+  thresholdTime: number
```
krzkaczor, line 11: Is it in seconds? I think the name should clearly indicate the unit.

packages/common/src/tools/LogThrottler.ts
```diff
@@ -0,0 +1,81 @@
...
+  throttleTime: number
+}
+
+export class LogThrottler {
```
krzkaczor, line 15: This class could use a two sentence comment. Why we are throttling and what exactly these options do.

packages/common/src/tools/LogThrottler.ts
```diff
@@ -0,0 +1,81 @@
...
+  private readonly DEFAULT_LOG_INFO = {
+    count: 0,
+    isThrottling: false,
+    throttleCount: 0,
+  }
+
+  private readonly recentLogs: Map<string, LogInfo>
+  private readonly logger: Logger = new Logger({
```
krzkaczor, line 23: We should create this logger based on logger options (log level, log style). Just pass it in constructor.

packages/common/src/tools/LogThrottler.ts
```diff
@@ -0,0 +1,81 @@
...
+  isThrottling(logKey: string): boolean | undefined {
+    return this.recentLogs.get(logKey)?.isThrottling
+  }
+
+  private throttle(logKey: string, logInfo: LogInfo): void {
+    logInfo.isThrottling = true
+
+    setTimeout(() => {
+      if (logInfo.throttleCount !== 0) {
+        this.logger.info(
```
krzkaczor, line 58: just ignore eslint warning aboud cirucular deps, here it should work just fine in the runtime.

krzkaczor (commented):
This starts to look good but we need tests for it anyway :P

packages/common/src/tools/LogThrottler.ts
```diff
@@ -0,0 +1,86 @@
...
+export class LogThrottler {
+  private readonly recentLogs: Map<string, LogInfo>
+  private readonly logger: Logger
+
+  constructor(
+    private readonly options: LogThrottlerOptions,
+    loggerOptions: LoggerOptions,
+  ) {
+    this.recentLogs = new Map<string, LogInfo>()
+    this.logger = new Logger(loggerOptions).for(LogThrottler.name)
```
krzkaczor, line 30: Why can't we just inject it as parameter? in application.ts we could create a "fresh" logger just for this class but in tests we could easily mock it.

torztomasz: We can do a separate logger in application.ts just for LogThrottler, I was thinking about it but it would look kinda ugly.

krzkaczor: I think it's fine -- this is really only sane way to test it. Dependency Injection all the way down :D

packages/common/src/tools/LogThrottler.ts
```diff
@@ -0,0 +1,84 @@
...
+    }
+
+    this.incrementCount(logInfo)
+
+    if (!logInfo.isThrottling && logInfo.count >= this.options.threshold) {
+      this.throttle(logKey, logInfo)
+    }
+  }
+
+  isThrottling(logKey: string): boolean | undefined {
```
krzkaczor, line 52: I think it's much cleaner to have simply `boolean` return type here (and do !! inside the function) so callers don't have to worry about handling `undefined`.

krzkaczor (approved):

packages/common/src/tools/LogThrottler.ts
torztomasz: that's totally right

packages/common/src/tools/LogThrottler.ts
```diff
@@ -0,0 +1,84 @@
+// eslint-disable-next-line import/no-cycle
+import { Logger } from './Logger'
```
sz-piotr, line 2: Instead of disabling `import/no-cycle` add an `ILogger` interface have `Logger implements ILogger` and `LogThrottler` receive `ILogger`

sz-piotr (changes requested):

sz-piotr (approved):

PR #1149, opened 2023-01-26
Activity API improvements

krzkaczor:
* dropped Aztec Connect activity because API returns 404 :O
* increase connection pool size to maybe fix our problems 
* remove delay from starex API as we re-record data from past 7 days anyway


packages/config/src/layer2s/aztecconnect.ts
```diff
@@ -61,11 +61,6 @@ export const aztecconnect: Layer2 = {
         sinceTimestamp: new UnixTime(1654638194),
       },
     ],
-    transactionApi: {
```
sz-piotr, line 64: Aztec connect too?

packages/backend/src/core/activity/counters/StarkexCounter.ts
```diff
@@ -63,6 +63,5 @@ export function createStarkexCounter(
 
 /** @internal */
 export function getStarkexLastDay(timestamp: UnixTime) {
-  // StarkEx APIs have inherent delay
-  return timestamp.add(-1, 'days').add(-4, 'hours').toStartOf('day').toDays()
+  return timestamp.toStartOf('day').toDays()
```
michalsidzej, line 66: Isn't this gonna show 0 tx for the past day tps all starkex projects? Also, we'll have our frontend stop building because of the sanity check.

torztomasz (approved):

packages/config/src/layer2s/aztecconnect.ts
krzkaczor: Fixed: actually only aztec connect doesnt work.

packages/backend/src/core/activity/counters/StarkexCounter.ts
krzkaczor: That's why we have `uncertaintyBuffer`for, right? We can record data as fast as possible and we will simply update what's in the database.

sz-piotr (approved):

packages/backend/src/core/activity/counters/StarkexCounter.ts
michalsidzej: Yes, it will sync eventually, but the sanity check will probably be never passed

michalsidzej (approved):

PR #1150, opened 2023-01-30
Update across discovery config

torztomasz:
Resolves L2B-687


krzkaczor (approved):

packages/backend/discovery/across/config.jsonc
```diff
@@ -33,14 +33,21 @@
...
     "0x226726Ac52e6e948D1B7eA9168F9Ff2E27DbcbB5": {
-      "ignoreMethods": ["bondedProposals"]
+      "ignoreMethods": ["bondedProposals", "getCurrentTime"]
     },
     "0x4D9079Bb4165aeb4084c526a32695dCfd2F77381": {
-      "ignoreMethods": ["rootBundles"]
+      "ignoreMethods": ["rootBundles", "getCurrentTime", "numberOfDeposits"]
+    },
+    "0x04Fa0d235C4abf4BcF4787aF4CF447DE572eF828": {
+      "ignoreInWatchMode": ["totalSupply"] // Maybe we should also ignore it in discovery?
```
krzkaczor, line 50: Current approach is that we monitor (almost) everything in discovery, and ignore in watchmode things that change too often.

torztomasz: If you say so then should I move getCurrentTime and numberOfDeposits from ignoreMethods to ignoreInWatchMode?

michalsidzej: Yes, that's what we currently do. I like your approach, but right now `ignoreMethods` are used so `discovery` doesn't go too deep.

packages/backend/discovery/across/discovered.json
```diff
@@ -234,26 +229,28 @@
       }
     },
     {
+      "_comment": "I checked the code and the changes are not malicious or smth, but I would like to confirm it with someone :)",
```
michalsidzej, line 232: It looks fine.

PR #1155, opened 2023-01-31
Update discovery configs

antooni:
This PR aims to update current discovery configs, to ignore methods in watch mode.

Resolves L2B-690


packages/backend/discovery/metis/config.jsonc
```diff
@@ -114,7 +114,16 @@
...
-      }
+      },
+      "ignoreInWatchMode": [
+        "getLastSequencerTimestamp",
+        "getTotalBatches",
+        "getTotalElements",
+        "getLastSequencerTimestampByChainId",
+        "getTotalBatchesByChainId",
+        "getTotalElementsByChainId",
+        ""
```
michalsidzej, line 125: Is this supposed to be here?

michalsidzej (approved):
One comment, otherwise LGTM

krzkaczor (approved):

packages/backend/src/core/discovery/utils/diffContracts.test.ts
```diff
@@ -45,7 +45,7 @@ describe(diffContracts.name, () => {
         E: 'ignoreMePlease',
       },
     }
-    const ignoreInWatchMode = ['E']
+    const ignoreInWatchMode = ['random-key', 'E']
```
krzkaczor, line 48: don't we want to warn (or throw an error)  when unknown key is ignored here?

PR #1156, opened 2023-01-31
Updating how to add new L2 project

claudioantonio:
The location for files and folders mentioned in the current guideline is not  correct.


michalsidzej (approved):
Thanks for the contribution

PR #1157, opened 2023-02-01
Change job link

torztomasz:
Resolves L2B-727

### DIFF
--- a/packages/frontend/src/components/navbar/Navbar.tsx
+++ b/packages/frontend/src/components/navbar/Navbar.tsx
@@ -138,7 +138,7 @@ export function Navbar(props: NavbarProps) {
               <li className="h-full">
                 <OutLink
                   className="flex items-center h-full px-2 font-medium"
-                  href="https://www.notion.so/native/l2beat/We-are-hiring-Work-at-L2BEAT-e4e637265ae94c5db7dfa2de336b940f"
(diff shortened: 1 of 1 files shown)


packages/frontend/src/components/navbar/Navbar.tsx
```diff
@@ -138,7 +138,7 @@ export function Navbar(props: NavbarProps) {
               <li className="h-full">
                 <OutLink
                   className="flex items-center h-full px-2 font-medium"
-                  href="https://www.notion.so/native/l2beat/We-are-hiring-Work-at-L2BEAT-e4e637265ae94c5db7dfa2de336b940f"
+                  href="https://www.notion.so/We-are-hiring-Work-at-L2BEAT-e4e637265ae94c5db7dfa2de336b940f"
```
sz-piotr, line 141: I checked it again and it should actually be:
https://l2beat.notion.site/We-are-hiring-Work-at-L2BEAT-e4e637265ae94c5db7dfa2de336b940f

sz-piotr (changes requested):

sz-piotr (approved):

packages/frontend/src/components/navbar/Navbar.tsx
krzkaczor: @torztomasz please, next time test things like this in "private tabs" in your browser. It's easy to find such mistakes.

PR #1161, opened 2023-02-01
Add hiring banner

michalsidzej:
Resolves L2B-734

### DIFF
--- a/packages/frontend/src/components/navbar/Banner.tsx
+++ b/packages/frontend/src/components/navbar/Banner.tsx
@@ -3,11 +3,11 @@ import React from 'react'
 export function Banner() {
   return (
     <a
-      className="block text-center bg-pink-200 dark:bg-purple-700 text-xs font-medium px-4 py-1.5"
+      className="block text-center bg-pink-900 text-white text-xs font-medium px-4 py-1.5"
       href="https://l2beat.notion.site/We-are-hiring-Work-at-L2BEAT-e4e637265ae94c5db7dfa2de336b940f"
       target="_blank"
     >
-      We are hiring! 
+      We are hiring!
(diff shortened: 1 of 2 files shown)


packages/frontend/src/components/navbar/Banner.tsx
```diff
@@ -3,12 +3,12 @@ import React from 'react'
 export function Banner() {
   return (
     <a
-      className="block text-center bg-blue-500 text-black text-xs font-bold px-4 py-1.5"
-      href="https://grant-explorer.gitcoin.co/#/round/1/0xe575282b376e3c9886779a841a2510f1dd8c2ce4/0x4a8d8bb79a51a2b0e3d136c15b672ea87c9fd7dc5b0c3cd76674d5ebe4f7dac0-0xe575282b376e3c9886779a841a2510f1dd8c2ce4"
+      className="block text-center bg-pink-200 dark:bg-purple-700 text-xs font-medium px-4 py-1.5"
```
sz-piotr, line 6: Can we use the same color scheme as the Jobs `Hiring` badge?

sz-piotr (changes requested):

packages/frontend/src/components/navbar/Banner.tsx
michalsidzej: Yeah, I tried it before with `font-bold` and didn't like it so that's why I changed the colors. Actually, with font-medium, it looks good. Fixed

michalsidzej: Yeah, I tried it before with font-bold and didn't like it so that's why I changed the colors. Actually, with font-medium, it looks good. Fixed

sz-piotr (approved):

PR #1162, opened 2023-02-01
Banner hotfix 

michalsidzej:
Old arrow looks bad on mobile 
image
So I fixed it 
Screenshot 2023-02-01 at 18 11 34

### DIFF
--- a/packages/frontend/src/components/navbar/Banner.tsx
+++ b/packages/frontend/src/components/navbar/Banner.tsx
@@ -11,7 +11,7 @@ export function Banner() {
     >
       We are hiring!
       <span className="underline ml-2">Join L2BEAT</span>
-      <ArrowRightIcon className="fill-white inline-block w-3 h-3 ml-0.5" />
+      <ArrowRightIcon className="fill-white w-3 h-3 ml-0.5" />
     </a>
   )
 }
--- a/packages/frontend/src/components/navbar/Banner.tsx
+++ b/packages/frontend/src/components/navbar/Banner.tsx
@@ -1,14 +1,17 @@
 import React from 'react'
 
+import { ArrowRightIcon } from '../icons'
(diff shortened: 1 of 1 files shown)


packages/frontend/src/components/navbar/Banner.tsx
```diff
@@ -1,14 +1,17 @@
...
     <a
-      className="block text-center bg-pink-900 text-white text-xs font-medium px-4 py-1.5"
+      className="text-center bg-pink-900 text-white text-xs font-medium px-4 py-1.5 flex items-center justify-center"
       href="https://l2beat.notion.site/We-are-hiring-Work-at-L2BEAT-e4e637265ae94c5db7dfa2de336b940f"
       target="_blank"
     >
       We are hiring!
-      <span className="underline ml-2">Join L2BEAT &rarr;</span>
+      <span className="underline ml-2">Join L2BEAT</span>
+      <ArrowRightIcon className="fill-white w-3 h-3 ml-0.5" />
```
torztomasz, line 14: why exactly do we need w-3 h-3?

torztomasz (approved):

packages/frontend/src/components/navbar/Banner.tsx
michalsidzej: Because otherwise the arrow is bigger

torztomasz: So maybe we should constrain the size in SVG or className inside the ArrowRightIcon?

michalsidzej: But we use this arrow in some other places

michalsidzej: And it's bigger there

PR #1164, opened 2023-02-01
Add tests for discovery configs

antooni:
This PR aims to test values saved in `ignoreInWatchMode`, thanks to the newly added test, which will throw if you specify value name not present in `discovered.json`

Resolves L2B-730


packages/backend/src/core/discovery/utils/config.test.ts
```diff
@@ -0,0 +1,57 @@
+import { expect } from 'earljs'
+
+import { ConfigReader } from '../ConfigReader'
+
+export {}
```
krzkaczor, line 5: is this needed?

krzkaczor (changes requested):

packages/backend/src/core/discovery/utils/config.test.ts
antooni: no, in the beginning file was in different folder and needed this one, but I decided to move it, and forgot to remove this statement, thanks

krzkaczor (approved):

michalsidzej (approved):

PR #1170, opened 2023-02-02
Add config hash to discovery

antooni:
Resolves L2B-732


packages/backend/src/core/discovery/runDiscovery.ts
```diff
@@ -1,20 +1,22 @@
...
+import { getDiscoveryConfigHash } from './utils/getDiscoveryConfigHash'
 
 export async function runDiscovery(
   provider: providers.AlchemyProvider,
   etherscanClient: MainnetEtherscanClient,
   configReader: ConfigReader,
-  config: DiscoveryConfig,
+  config: DiscoveryModuleConfig,
 ) {
   const projectConfig = await configReader.readConfig(config.project)
```
torztomasz, line 18: I am thinking about extracting configReader to common package as I am also reading discovery in deriving

torztomasz (changes requested):

torztomasz (approved):

packages/backend/src/core/discovery/utils/diffDiscovery.ts
```diff
@@ -63,3 +61,19 @@ export function diffDiscovery(
...
+    return []
+  }
+
+  // eslint-disable-next-line @typescript-eslint/no-unnecessary-condition
+  if (overrides[address.toString()] === undefined) {
+    return []
+  }
+
+  return overrides[address.toString()].ignoreInWatchMode ?? []
+}
```
michalsidzej, line 79:
```suggestion
function getIgnored(
  address: EthereumAddress,
  overrides?: Partial<Record<string, DiscoveryContract>>,
): string[] {
  if (overrides === undefined) {
    return []
  }

  const override = overrides[address.toString()]
  if (override === undefined) {
    return []
  }

  return override.ignoreInWatchMode ?? []
}
```

packages/backend/src/core/discovery/utils/diffDiscovery.ts
```diff
@@ -30,10 +30,8 @@ export function diffDiscovery(
       })
       continue
     }
-    // eslint-disable-next-line @typescript-eslint/no-unnecessary-condition
-    const ignored: string[] = overrides[committedContract.address.toString()]
-      ? overrides[committedContract.address.toString()].ignoreInWatchMode ?? []
-      : []
+
+    const ignored: string[] = getIgnored(committedContract.address, overrides)
```
michalsidzej, line 34:
```suggestion
    const ignored = getIgnored(committedContract.address, overrides)
```

packages/backend/src/core/DiscoveryWatcher.ts
```diff
@@ -84,15 +88,30 @@ export class DiscoveryWatcher {
...
+          overrides,
+        )
+      } else {
+        return diffDiscovery(
+          databaseEntry.discovery.contracts,
+          discovery.contracts,
+          overrides,
+        )
+      }
+    }
```
michalsidzej, line 114: No need to use `if...else` if you are returning

michalsidzej (changes requested):
Comments above

packages/backend/src/core/DiscoveryWatcher.ts
```diff
@@ -84,15 +88,26 @@ export class DiscoveryWatcher {
...
-      : (await this.configReader.readDiscovery(name)).contracts
+    if (databaseEntry === undefined) {
+      const committed = await this.configReader.readDiscovery(name)
+      return diffDiscovery(committed.contracts, discovery.contracts, overrides)
+    }
+
+    if (databaseEntry.configHash !== configHash) {
+      const committed = await this.configReader.readDiscovery(name)
+      return diffDiscovery(committed.contracts, discovery.contracts, overrides)
+    }
```
michalsidzej, line 104: this could be one statement with logical alternative

michalsidzej (approved):

packages/backend/src/core/discovery/utils/getDiscoveryConfigHash.ts
```diff
@@ -0,0 +1,46 @@
...
+          }
+        }
+      } else {
+        // @ts-expect-error cannot determine type, generic sorting
+        // eslint-disable-next-line @typescript-eslint/no-unsafe-assignment
+        acc[key] = object[key]
+      }
+      return acc
+    }, {})
+}
```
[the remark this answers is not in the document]
sz-piotr: How about a different approach: In config reader whenever you read the config the file contents are both parsed and hashed and the result is `{ ...parsed, hash}`. That way the config hash is actually just the hash of the config file including comments, whitespace, etc. Any change to the config file requires re-running discovery.
@michalsidzej @antooni

PR #1176, opened 2023-02-06
Make navbar not grayed out

torztomasz:
Resolves L2B-745


packages/frontend/src/components/icons/pages/ActivityIcon.tsx
```diff
@@ -5,18 +5,22 @@ import { Icon } from '../Icon'
 export function ActivityIcon(props: SVGAttributes<SVGElement>) {
   return (
     <Icon aria-label="Activity icon" {...props}>
-      <g clipPath="url(#clip0_1221_33203)">
+      <g clipPath="url(#clip0_18_177)">
```
michalsidzej, line 8: What about the TVL icon? Shouldn't both of them be updated? Please confirm with @sz-piotr and Janek

michalsidzej (approved):

PR #1184, opened 2023-02-07
Make whole table row clickable for projects

torztomasz:
Resolves L2B-749


packages/frontend/src/components/table/TableView.tsx
```diff
@@ -59,32 +60,34 @@ export function TableView<T>({ items, columns, rows }: Props<T>) {
...
-                    j !== columns.length - 1 && 'pr-3 md:pr-4',
-                  )}
-                >
-                  {column.getValue(item, i)}
-                </td>
-              ))}
-            </tr>
-          ))}
+          {items.map((item, i) => {
+            return (
```
michalsidzej, line 64: return is not necessary here

michalsidzej (approved):

PR #1186, opened 2023-02-07
Metrics refactor

antooni:
This PR aims to remove Metrics class, which was instantiated in Application.ts and later passed as an argument to different modules. This solution did not age well, was painful during testing and overall required more code than was needed.

That is why we decided to remove it, and instead create metrics "inline", as close to the "metrified" object as possible.

Resolves L2B-739


packages/backend/src/core/activity/DailyTransactionCountViewRefresher.ts
```diff
@@ -13,11 +13,15 @@ export class DailyTransactionCountViewRefresher {
     private readonly logger: Logger,
   ) {
     this.logger = logger.for(this)
-    this.refreshQueue = new TaskQueue<void>(async () => {
-      this.logger.info('Refresh started')
-      await this.viewRepository.refresh()
-      this.logger.info('Refresh finished')
-    }, this.logger.for('refreshQueue'))
+    this.refreshQueue = new TaskQueue<void>(
```
michalsidzej, line 16: Nice catch

packages/backend/src/peripherals/database/shared/BaseRepository.ts
```diff
@@ -62,16 +62,22 @@ export type CheckConvention<T extends BaseRepository> = {
...
 */
+const repositoryHistogram = new Histogram({
+  name: 'repository_method_duration_seconds',
+  help: 'duration histogram of repository methods',
+  labelNames: ['repository', 'method'],
+})
+
 export abstract class BaseRepository {
-  protected histogram: RepositoryHistogram
+  protected histogram: Histogram
```
michalsidzej, line 72: This RepositoryHistogram type is better, because it will not allow labels that are not in `labelNames` on initialization.

packages/backend/src/core/DiscoveryWatcher.ts
```diff
@@ -29,6 +29,9 @@ export class DiscoveryWatcher {
     this.taskQueue = new TaskQueue(
```
michalsidzej, line 29: I'm thinking if we could have a method like `TaskQueue.for(this, ...constructorArgs)` Then we could get the ID from `constructor.name` like in logger. WYT?

packages/backend/src/core/activity/TransactionCountingMonitor.ts
```diff
@@ -27,10 +27,14 @@ export class TransactionCountingMonitor {
...
-      this.logger.info('Sync check triggered')
-      await this.checkIfSynced()
-    }, this.logger.for('checkQueue'))
+    this.checkQueue = new TaskQueue<void>(
+      async () => {
+        this.logger.info('Sync check triggered')
+        await this.checkIfSynced()
+      },
+      this.logger.for('checkQueue'),
+      { taskQueueId: TransactionCountingMonitor.name },
```
michalsidzej, line 36: Honestly, the more I look at this, I'm not a fan of making the `taskQueueId` required and measuring every single call here. I think we should focus on the metrics that matter to us.

packages/backend/src/core/activity/counters/RpcCounter.ts
```diff
@@ -65,7 +62,9 @@ export function createRpcCounter(
           }
         })
 
-        const blocks = await promiseAllPlus(queries, logger)
+        const blocks = await promiseAllPlus(queries, logger, {
+          taskQueueId: `${projectId.toString()}-${createRpcCounter.name}`,
```
michalsidzej, line 66:
```suggestion
          taskQueueId: `${projectId.toString()}-activityRpcCounter`,
```
But anyway, I don't think we need it that much. We have metrics for activity in the `SequenceProcessor`

packages/backend/src/modules/activity/createTransactionCounters.ts
```diff
@@ -20,13 +20,18 @@ import { SequenceProcessorRepository } from '../../peripherals/database/Sequence
 import { Database } from '../../peripherals/database/shared/Database'
 import { StarkexClient } from '../../peripherals/starkex'
 
+const activityIncludedInApi = new Gauge({
+  name: 'activity_included_in_api',
+  help: 'Boolean value 1 when this project is included in activity api response',
+  labelNames: ['project'],
+})
+
```
michalsidzej, line 28: Maybe move it just above the function that is using this?

michalsidzej (commented):
Do you think we should have a metric for every single task queue?

packages/backend/src/core/activity/counters/AztecConnectCounter.ts
```diff
@@ -49,7 +46,9 @@ export function createAztecConnectCounter(
           }
         })
 
-        const blocks = await promiseAllPlus(queries, logger)
+        const blocks = await promiseAllPlus(queries, logger, {
+          metricsId: `AztecConnectBlockCounter[${projectId.toString()}]`,
```
michalsidzej, line 50: There's only one AztecConnect ;)
```suggestion
          metricsId: 'AztecConnectBlockCounter',
```

packages/backend/src/core/activity/counters/AztecCounter.ts
```diff
@@ -49,7 +46,9 @@ export function createAztecCounter(
           }
         })
 
-        const blocks = await promiseAllPlus(queries, logger)
+        const blocks = await promiseAllPlus(queries, logger, {
+          metricsId: `AztecBlockCounter[${projectId.toString()}]`,
```
michalsidzej, line 50:
```suggestion
          metricsId: 'AztecBlockCounter',
```

packages/backend/src/core/activity/counters/LoopringCounter.ts
```diff
@@ -46,7 +43,9 @@ export function createLoopringCounter(
           }
         })
 
-        const blocks = await promiseAllPlus(queries, logger)
+        const blocks = await promiseAllPlus(queries, logger, {
+          metricsId: `LoopringBlockCounter[${projectId.toString()}]`,
```
michalsidzej, line 47:
```suggestion
          metricsId: 'LoopringBlockCounter',
```

packages/backend/src/core/activity/counters/StarknetCounter.ts
```diff
@@ -56,7 +53,9 @@ export function createStarknetCounter(
           }
         })
 
-        const blocks = await promiseAllPlus(queries, logger)
+        const blocks = await promiseAllPlus(queries, logger, {
+          metricsId: `StarkNetBlockCounter[${projectId.toString()}]`,
```
michalsidzej, line 57:
```suggestion
          metricsId: 'StarkNetBlockCounter',
```

packages/backend/src/core/queue/promiseAllPlus.ts
```diff
@@ -49,6 +50,7 @@ export async function promiseAllPlus<T>(
       }
       return shouldRetry
     },
+    metricsId: opts?.metricsId ?? 'undefined',
```
michalsidzej, line 53: I don't think this is a good approach. Either this id should be required or not.

packages/backend/src/peripherals/database/shared/BaseRepository.ts
```diff
@@ -62,16 +62,24 @@ export type CheckConvention<T extends BaseRepository> = {
...
   protected histogram: RepositoryHistogram
 
   constructor(
     protected readonly database: Database,
     protected readonly logger: Logger,
-    readonly metrics: Metrics,
   ) {
     this.logger = logger.for(this)
-    this.histogram = metrics.repositoryHistogram
+
```
michalsidzej, line 81: Is this new line intentional?

michalsidzej (approved):

PR #1187, opened 2023-02-07
Fix staging start error

antooni:


torztomasz (approved):

torztomasz (commented):
Maybe we can force node to always put index.js in build dir? This way we will not have to worry about having ts files outside src dir.

Or we can force node to keep files in build/src even if there is no ts files outside src dir.

PR #1193, opened 2023-02-08
Discovery watcher metrics

antooni:
This PR aims to add metrics buckets related to DiscoveryWatcher

Resolves L2B-367


packages/backend/src/core/discovery/DiscoveryEngine.ts
```diff
@@ -26,7 +34,16 @@ export class DiscoveryEngine {
...
-    const discovered = await discover(discoveryProvider, config, this.logger)
+    const discovered = await wrapAndMeasure(
+      async () => await discover(discoveryProvider, config, this.logger),
+      {
+        histogram: discoveryHistogram,
+        labels: {
+          project: config.name,
+        },
+      },
+    )()
```
michalsidzej, line 45: Nit pick: Maybe save the function from wrapAndMeasure in a separate variable (e.g. `measuredDiscovery`). It will make the code a bit more readable

michalsidzej (approved):
Good metrics.

PR #1201, opened 2023-02-10
Add more discovery metrics

antooni:


packages/backend/src/core/DiscoveryWatcher.ts
```diff
@@ -136,3 +134,36 @@ export class DiscoveryWatcher {
...
+
+const changesDetected = new Gauge({
+  name: 'discovery_watcher_changes_detected',
+  help: 'Value showing the amount of changes detected by DiscoveryWatcher',
+})
+
+const syncDuration = new Gauge({
+  name: 'discovery_watcher_sync_duration',
+  help: 'Value showing how long does it take to sync all the projects',
+})
```
michalsidzej, line 151: This should be a histogram

packages/backend/src/core/DiscoveryWatcher.ts
```diff
@@ -75,6 +72,7 @@ export class DiscoveryWatcher {
         if (diff.length > 0) {
           const messages = diffToMessages(projectConfig.name, diff)
           await this.notify(messages)
+          changesCounter += diff.length
```
michalsidzej, line 75: What do you think about saving the changes per project?

packages/backend/src/core/DiscoveryWatcher.ts
```diff
@@ -48,13 +43,15 @@ export class DiscoveryWatcher {
   }
 
   async update() {
+    const metricsDone = initMetrics()
```
michalsidzej, line 46: Is this approach necessary? We could just start the timer here and stop it at the end. I'm not sure, but I have to admit I actually quite like how it looks here.

michalsidzej (approved):

PR #1210, opened 2023-02-14
Add aztecconnect discovery

michalsidzej:
Added `startIndex` and `ignoreRelative` to `ArrayHandler` in discovery. This allows us to get all the Aztec's DefiBridges, without running discovery on them. WYT?


packages/backend/src/core/discovery/handlers/user/ArrayHandler.test.ts
```diff
@@ -277,6 +279,7 @@ describe(ArrayHandler.name, () => {
       expect<unknown>(result).toEqual({
         field: 'owners',
         value: owners,
+        ignoreRelative: undefined,
```
sz-piotr, line 282: add a single test for `ignoreRelative: true`

sz-piotr (changes requested):

sz-piotr (approved):

PR #1215, opened 2023-02-15
Track ETH deposits on Fuel V1

FossilFrank:
The Fuel V1 adapter doesn't track the ETH deposited into Fuel, which is around 0.25 ETH


krzkaczor (approved):

FossilFrank:
@krzkaczor @antooni I still don't see it updated on L2Beat.com, is there a delay until the data gets ingested?

antooni:
Yes, there is a delay. When the staging backend will be promoted to production, then updated ETH balance will be visible.

PR #1218, opened 2023-02-16
Update ProjectDetail page

antooni:
Problem: currently when viewing on a window larger than mobile breakpoint, but not wide enough to fit both columns, the project details page is hard to read
image

This PR aims to change the breakpoint, so the column split happens only from window larger than 1200px
image

### DIFF
--- a/packages/frontend/src/styles/components/_ProjectDetails.scss
+++ b/packages/frontend/src/styles/components/_ProjectDetails.scss
@@ -6,7 +6,7 @@
   column-gap: 16px;
   row-gap: 3rem;
 
-  @media (min-width: 1200px) {
+  @media (min-width: 1024px) {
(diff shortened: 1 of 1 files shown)


packages/frontend/src/styles/components/_ProjectDetails.scss
```diff
@@ -6,7 +6,7 @@
   column-gap: 16px;
   row-gap: 3rem;
 
-  @media (min-width: 750px) {
+  @media (min-width: 1200px) {
```
michalsidzej, line 9:
```suggestion
  @media (min-width: 1024px) {
```

michalsidzej (changes requested):

michalsidzej (approved):

PR #1222, opened 2023-02-16
zkSync rebrand

michalsidzej:


sz-piotr (approved):
It would be great to have milestones for when the rebrand happened, but this can be done later

PR #1225, opened 2023-02-17
Fix unverified contracts icon

michalsidzej:


antooni (approved):
When doing such a small UI improvement, it is a good practice to provide screenshot in PR description, this makes life of a reviewer much easier

PR #1232, opened 2023-02-21
Implement Archived Projects

michalsidzej:
Resolves L2B-629, L2B-846

Please review if the Archived Projects list is truthful


packages/config/src/layer2s/hermez.ts
```diff
@@ -16,6 +16,7 @@ import { Layer2 } from './types'
 export const hermez: Layer2 = {
   type: 'layer2',
   id: ProjectId('hermez'),
+  isArchived: true,
```
luke7211, line 19: I don't think Hermez is archived at this stage

luke7211: No, Fuel is definitely not archived, it's up and running as we have experienced

packages/frontend/src/components/table/props/getBridgesRowProps.ts
```diff
@@ -18,8 +19,10 @@ export function getBridgesRowProps(entry: BridgeTableEntry) {
     className: cx(
       getRowVerificationClassNames(entry),
       entry.type !== 'bridge' && 'hidden',
+      entry.isArchived && 'hidden opacity-50',
     ),
-    'data-combined-only': entry.type !== 'bridge' ? true : undefined,
+    'data-layer2': entry.type !== 'bridge' ? true : undefined,
```
antooni, line 24: A comment would be useful there

antooni (changes requested):
Is this view on mobile desired?
image

packages/frontend/src/components/table/props/getBridgesRowProps.ts
michalsidzej: Changed it, do you like it now?

antooni (approved):

PR #1237, opened 2023-02-22
Fix failing tests

michalsidzej:
I'm not sure if this will actually fix the failing tests, but for sure it will run them as we should run them. In the previous approach, because `time.runToLastAsync()` breaks the event loop, the `SequenceProcessor`'s interval function was running every time on failed job (so also after the last attempt). `time.tickAsync` fixes that.

### DIFF
--- a/packages/backend/src/core/SequenceProcessor.test.ts
+++ b/packages/backend/src/core/SequenceProcessor.test.ts
@@ -220,10 +220,6 @@ describe(SequenceProcessor.name, () => {
       await sequenceProcessor.start()
       await waitForErrorReport(time, reportErrorMock)
 
-      await Promise.all([
-        sequenceProcessor.stop(),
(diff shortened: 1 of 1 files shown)


krzkaczor (approved):

packages/backend/src/core/SequenceProcessor.test.ts
```diff
@@ -249,15 +247,13 @@ describe(SequenceProcessor.name, () => {
...
         processRange: processRangeMock,
         reportError: reportErrorMock,
+        // Because we use fake-timers, we need to disable the refresh interval
+        refreshInterval: Infinity,
       })
 
       await sequenceProcessor.start()
       await waitForErrorReport(time, reportErrorMock)
 
-      await Promise.all([
```
krzkaczor, line 257: i think this was needed because refresh Interval was not infirinity before.

PR #1245, opened 2023-02-24
Add multiple addresses support in contracts

antooni:
image

Resolves L2B-917


packages/config/scripts/checkVerifiedContracts/addresses.ts
```diff
@@ -82,3 +89,18 @@ export function areAllProjectContractsVerified(
...
+    return c.multipleAddresses
+  }
+}
+
+function isSingleAddress(
+  c: ProjectContract,
+): c is ProjectContractSingleAddress {
+  // eslint-disable-next-line @typescript-eslint/no-unnecessary-condition
+  return (c as ProjectContractSingleAddress).address !== undefined
+}
```
michalsidzej, line 106: This TG is repeated

packages/frontend/src/components/project/ContractEntry.tsx
```diff
@@ -52,6 +52,12 @@ export function ContractEntry({
       <BulletIcon className="h-6 md:h-[27px]" />
     )
 
+  const isAddressUnverified = (address: string) => {
```
michalsidzej, line 55: Either this name should be different or only check the verification status.

michalsidzej (changes requested):

packages/config/scripts/checkVerifiedContracts/addresses.ts
antooni: I know, but they are in differnet packages with no clear code path to export (the one used in config is used inside script)

michalsidzej (approved):

PR #1262, opened 2023-03-01
Update tokens list

antooni:
This PR aims to update our outdated token list, I went through coingecko TOP100 by MCap and cross checked the missing tokens.
Addidtionally, I have written a script that will automate the process of adding new tokens

Resolves L2B-849


packages/config/scripts/tokens/add/getTokenInfo.ts
```diff
@@ -0,0 +1,140 @@
...
+
+import { TokenInfo } from '../../../src/tokens/types'
+import { getEnv } from '../../checkVerifiedContracts/utils'
+
+export async function getTokenInfo(
+  provider: providers.JsonRpcProvider,
+  address: EthereumAddress,
+  category: z.infer<typeof TokenInfo.shape.category>,
+): Promise<TokenInfo> {
+  const name = await getName(provider, address)
```
torztomasz, line 19: I guess we can improve this by using Promise.all

packages/config/scripts/tokens/add/getTokenInfo.ts
```diff
@@ -0,0 +1,140 @@
...
+
+async function getCoingeckoId(address: EthereumAddress) {
+  const http = new HttpClient()
+  const coingeckoClient = new CoingeckoClient(http, undefined)
+
+  const coinList = await coingeckoClient.getCoinList({
+    includePlatform: true,
+  })
+
+  const coin = coinList.find((coin) => {
```
torztomasz, line 99: isn't that the same?
```tsx
  const coin = coinList.find((coin) => {
    coin.platforms.ethereum?.toLowerCase() === address.toString().toLowerCase()
  })
  ```

File: packages/config/scripts/tokens/add/getTokenInfo.ts [hunk 3]
```diff
@@ -0,0 +1,140 @@
...
+      coin.platforms.ethereum === undefined
+    ) {
+      return false
+    }
+    return (
+      coin.platforms.ethereum.toLowerCase() === address.toString().toLowerCase()
+    )
+  })
+
+  if (coin === undefined || coin.id === undefined) {
```
torztomasz, line 112:
```suggestion
    if(!coin || !coin.id) {
```

packages/config/scripts/tokens/add/getTokenInfo.ts
```diff
@@ -0,0 +1,140 @@
...
+  const data = (await response.json()) as unknown as {
+    result: { txHash: string }[]
+  }
+
+  const txHash = data.result[0].txHash
+
+  const tx = await provider.getTransaction(txHash)
+
+  // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
+  const block = await provider.getBlock(tx.blockNumber!)
```
torztomasz, line 137: Maybe we should handle this case instead using non null assertion

packages/config/scripts/tokens/add/index.ts
```diff
@@ -0,0 +1,78 @@
...
+import { z } from 'zod'
+
+import { tokenList } from '../../../src'
+import { TokenInfo } from '../../../src/tokens/types'
+import { getEnv } from '../../checkVerifiedContracts/utils'
+import { getTokenInfo } from './getTokenInfo'
+
+async function main() {
+  const [address, category] = handleCLIParameters()
+  if (!address || !category) {
```
torztomasz, line 15: You can probably move this check to handleCLIParameters and return non undefined variables from it

packages/config/src/tokens/types.ts
```diff
@@ -0,0 +1,32 @@
...
+export type TokenInfo = z.infer<typeof TokenInfo>
+export const TokenInfo = z.object({
+  /** Internal token id. Usually ticker-name */
+  id: branded(z.string(), (s) => AssetId(s)),
+  /** Token name as dictated by the token contract */
+  name: z.string(),
+  /** Token Coingecko API id. Used to fetch prices */
+  coingeckoId: branded(z.string(), (s) => CoingeckoId(s)),
+  /** Token address. Only Ether has no address */
+  address: z.optional(branded(z.string(), (s) => EthereumAddress(s))),
```
torztomasz, line 19:
```suggestion
  address: branded(z.string(), (s) => EthereumAddress(s)).optional(),
```

packages/config/src/tokens/types.ts
```diff
@@ -0,0 +1,32 @@
+import {
+  AssetId,
+  branded,
+  CoingeckoId,
+  EthereumAddress,
+  UnixTime,
+} from '@l2beat/shared'
+import { z } from 'zod'
+
```
torztomasz, line 9: On Starkex we have helper functions for branded types
```tsx
export function stringAs<T>(Brand: (s: string) => T) {
  return z
    .string()
    .refine((s) => {
      try {
        Brand(s)
        return true
      } catch {
        return false
      }
    })
    .transform(Brand)
}
```

Then you would use it like this
```tsx
id: stringAs(AssetId)
```

Maybe you can add it on L2Beat also

torztomasz (changes requested):

packages/config/src/tokens/types.ts
torztomasz: You can find those helper functions there: starkex-explorer/packages/shared/src/types.ts

packages/config/scripts/tokens/add/index.ts
antooni: doing it this way allows me to handle this kind of error with human readable console.log, if I throw then the error will be less readable

torztomasz (approved):

PR #1267, opened 2023-03-06
Change l2fees.info bar

vitali-l2beat:
Resolves L2B-940

https://linear.app/l2beat/issue/L2B-940/implement-changes-for-l2feesinfo-bar-on-the-main-page

Screenshot 2023-03-06 at 11 28 58
Screenshot 2023-03-06 at 11 29 09


krzkaczor:
image

Can u improve the mobile view for this component?

packages/frontend/src/components/otherSites/OtherSites.tsx
```diff
@@ -0,0 +1,34 @@
...
+
+import { OutLink } from '../OutLink'
+
+export function OtherSites() {
+  return (
+    <div
+      className="
+      mt-[32px] flex flex-row items-center
+      justify-between rounded-[12px] bg-gradient-to-r
+      from-[#7E41CC99]
```
antooni, line 11: avoid inlining colors, add them to tailwind config instead

antooni:
image
Agree with @krzkaczor, maybe the font should be smaller?

antooni (changes requested):
Code looks great, with one minor exception (see the comment). Please update the mobile view.

packages/frontend/src/components/otherSites/OtherSites.tsx
torztomasz: yup, and if there is really similar color then use this one instead

packages/frontend/src/components/otherSites/OtherSites.tsx
```diff
@@ -0,0 +1,34 @@
+import React from 'react'
+
+import { OutLink } from '../OutLink'
+
+export function OtherSites() {
```
torztomasz, line 5: rename directory from otherSites to other-sites as this is the convention we use

packages/frontend/src/components/otherSites/OtherSites.tsx
```diff
@@ -0,0 +1,34 @@
...
+  return (
+    <div
+      className="
+      mt-[32px] flex flex-row items-center
```
torztomasz, line 9: mt-[32px] = mt-8

packages/frontend/src/components/otherSites/OtherSites.tsx
```diff
@@ -0,0 +1,34 @@
...
+      justify-between rounded-[12px] bg-gradient-to-r
```
torztomasz, line 10: rounded-[12px] = rounded-3

packages/frontend/src/components/otherSites/OtherSites.tsx
```diff
@@ -0,0 +1,34 @@
...
+      from-[#7E41CC99] 
+      via-[#FF46C099] to-[#EE2C0199] px-[40px] py-[24px]
```
torztomasz, line 12: do not use the [40px] if you can use px-10 for example, same for py

packages/frontend/src/components/otherSites/OtherSites.tsx
```diff
@@ -0,0 +1,34 @@
...
+    "
+    >
+      <div className="flex flex-col justify-between text-2xl text-white">
+        <span>Ethereum Layer 1 is expensive.</span>
+        <span>How much does it cost to use Layer 2?</span>
+      </div>
+      <OutLink
+        className="
+        flex items-center 
+        rounded-[8px] bg-white
```
torztomasz, line 23: same here about using px

packages/frontend/src/components/otherSites/OtherSites.tsx
```diff
@@ -0,0 +1,34 @@
...
+        from-[#7E41CC]
+        to-[#FF46C0]
+        px-[48px] py-[19px] text-lg dark:bg-gradient-to-r"
```
torztomasz, line 26: and here

torztomasz (changes requested):

michalsidzej (commented):
Are the light mode colors supposed to look like that? 
Screenshot 2023-03-07 at 07 05 33

packages/frontend/src/components/other-sites/OtherSites.tsx
```diff
@@ -0,0 +1,31 @@
...
+      to-red-200/40 px-10
+      py-6 md:flex-row
+    "
+    >
+      <div className="text-center text-lg text-white md:text-left">
+        Ethereum Layer 1 is expensive. How much does it cost to use Layer 2? 💸
+      </div>
+      <OutLink
+        className="
+        mt-3 ml-0 min-w-[90px] self-center text-center
```
michalsidzej, line 21:
```suggestion
        mt-3 ml-0 min-w-[90px]
```
Correct me if I'm wrong, but it seems to me that those classes are not necessary (we're centering this link by flex already).

vitali-l2beat: Yeah, "self-center" is superfluous, but "text-center" is needed

packages/frontend/src/components/other-sites/OtherSites.tsx
```diff
@@ -0,0 +1,31 @@
+import React from 'react'
+
+import { OutLink } from '../OutLink'
+
+export function OtherSites() {
+  return (
+    <div
+      className="
+      mt-8 flex flex-col items-center justify-between
```
torztomasz, line 9: avoid multilines of classnames, if line is too long then maybe use cx module

krzkaczor: don't we have autoformatting for this? :O

torztomasz (approved):

antooni (approved):

PR #1281, opened 2023-03-09
Add first upcoming rollup

vitali-l2beat:
Resolves L2B-979

This PR will add first upcomin rollup to config (Scroll), as well as change some frontend logic to include upcoming projects in the table.


sz-piotr:
@vitali-l2beat can you make the pull request name nicer? Something that is man made and not autogenerated by linear+github

packages/config/src/layer2s/common/index.ts
```diff
@@ -8,3 +8,10 @@ export * from './nuggets'
 export * from './operator'
 export * from './riskView'
 export * from './stateCorrectness'
+
+export const EMPTY_TECHNOLOGY_CHOICE = {
+  name: '–',
+  description: 'Not provided',
+  references: [],
+  risks: [],
+}
```
sz-piotr, line 17: The index file ideally shouldn't have any content outside of exports. If we need to create a new file just for this, fine, let's do it.

packages/config/src/layer2s/index.ts
```diff
@@ -49,6 +51,7 @@ export const layer2s: Layer2[] = [
   nova,
   omgnetwork,
   optimism,
+  scroll,
```
sz-piotr, line 54: I'm not seeing this project even after checking the upcoming rollups checkbox. This is probably related to the API returning 0 tvl for it.

packages/config/src/layer2s/scroll.ts
```diff
@@ -0,0 +1,62 @@
...
+    validatedBy: RISK_VIEW.EMPTY,
+  }),
+  technology: {
+    category: 'ZK Rollup',
+    stateCorrectness: EMPTY_TECHNOLOGY_CHOICE,
+    dataAvailability: EMPTY_TECHNOLOGY_CHOICE,
+    operator: EMPTY_TECHNOLOGY_CHOICE,
+    forceTransactions: EMPTY_TECHNOLOGY_CHOICE,
+    exitMechanisms: [],
+  },
```
sz-piotr, line 57: Is this what we actually want for upcoming projects? I thought that the purpose is to actually write something. At the very least we should mention that scroll is a general purpose zk rollup.

antooni: I discussed this with @luke7211, we decided to leave it empty, to avoid doubling the research work, because this will change when they deploy (or at least will need a re-check)

antooni: Next time such discussions should take place on #research discord channel, so everyon will be on the same page

packages/config/src/layer2s/common/index.ts
antooni: I will move it to other file, let's not break the pattern from our codebase

packages/config/src/layer2s/index.ts
antooni: yeap, I have addressed this issue in the recent commits

PR #1284, opened 2023-03-10
Add table of contents

vitali-l2beat:
Resolves L2B-963

### DIFF
--- a/packages/frontend/src/pages/faq/FaqPage.tsx
+++ b/packages/frontend/src/pages/faq/FaqPage.tsx
@@ -27,7 +27,7 @@ export function FaqPage(props: FaqPageProps) {
           </span>
         </span>
         <div className="mt-12 flex flex-row justify-between">
-          <aside className="flex max-w-[288px] flex-col">
+          <aside className="hidden max-w-[288px] md:flex md:flex-col">
             <a
               className={linkClasses}
               href="#what-is-the-overall-purpose-of-this-site"
--- a/packages/frontend/src/pages/faq/FaqPage.tsx
+++ b/packages/frontend/src/pages/faq/FaqPage.tsx
@@ -11,14 +11,117 @@ export interface FaqPageProps {
 }
 
 export function FaqPage(props: FaqPageProps) {
+  const linkClasses =
+    'mt-4 first:mt-0 text-gray-850 hover:text-pink-900 dark:text-white/80 dark:hover:text-pink-200 font-semibold text-base'
   return (
     <>
       <Navbar {...props.navbar} />
-      <PageContent narrow>
-        <article
-          className="Faq"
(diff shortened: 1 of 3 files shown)


packages/frontend/src/pages/faq/FaqPage.tsx
```diff
@@ -11,14 +11,117 @@ export interface FaqPageProps {
...
   return (
     <>
       <Navbar {...props.navbar} />
-      <PageContent narrow>
-        <article
-          className="Faq"
-          dangerouslySetInnerHTML={{ __html: props.content }}
-        />
+      <PageContent>
+        <span className="mt-16 flex flex-col border-b border-gray-850 pb-12">
```
sz-piotr, line 20: The FAQ should be autogenerated from the markdown. This can be achieved using a toc plugin for our markdown renderer I think

sz-piotr (changes requested):

antooni (changes requested):
Mobile view should be changed, although I do not see a design for it. Maybe let's just hide this navigation on mobile 
image

packages/frontend/src/styles/components/_Faq.scss
```diff
@@ -3,27 +3,22 @@
...
+    line-height: 29px;
+
+    &:first-child {
+      margin-top: 0;
+    }
   }
 
   p {
-    margin: 16px 0 0 0;
+    margin-top: 16px;
```
antooni, line 17: We are phasing out scss, use Tailwind instead. If it possible delete this .scss altogether and style everything in tailwind.

packages/frontend/src/pages/faq/FaqPage.tsx
vitali-l2beat: Are you saying we need to implement the content table within MD? I don't believe it's possible

packages/frontend/src/styles/components/_Faq.scss
vitali-l2beat: It's impossible to delete scss because it styles an auto generated code

packages/frontend/src/pages/faq/FaqPage.tsx
sz-piotr: We are using the `markdown-it` library. It supports plugins. We could use for example https://github.com/cmaas/markdown-it-table-of-contents and have the toc be created for us automatically.

packages/frontend/src/styles/components/_Faq.scss
sz-piotr: It is possible, we did something similar for definitions. Look at packages/frontend/src/pages/definitions/style.css, it's imported in packages/frontend/src/styles/style.css and uses tailwind.

PR #1285, opened 2023-03-10
Change the link color

vitali-l2beat:
Resolves L2B-962


krzkaczor (approved):

antooni (approved):
When you are submitting a PR changing the UI, it is a good practice to post a screenshot of change inside description. It makes job easier for the reviewer (even though we have visual regression I still believe it is a good practice)

sz-piotr (approved):
It would be nice if we'd use such opportunities to further remove scss from our codebase. Ideally this task would've be to remove the `--link` and `--link-rgb` variables, add a color to tailwind, and use that color for all links around the codebase

PR #1286, opened 2023-03-10
Include upcoming rollups 

antooni:
This PR aims to change some frontend logic to include upcoming projects in the table.


packages/frontend/src/pages/bridges-tvl/BridgesTvlView.tsx
```diff
@@ -26,13 +26,14 @@ export interface BridgesTvlViewEntry {
...
-  oneDayChange: string
-  sevenDayChange: string
-  bridgesMarketShare: string
-  combinedMarketShare: string
+  tvl?: string
+  tvlBreakdown?: TVLBreakdownProps
+  oneDayChange?: string
+  sevenDayChange?: string
+  bridgesMarketShare?: string
+  combinedMarketShare?: string
```
michalsidzej, line 36: What do you think about making some sort of type logic here? So those would be undefined for Upcoming Rollup, but required for existing.

michalsidzej (approved):
Maybe type logic would be useful here, otherwise LGTM

packages/frontend/src/pages/bridges-tvl/BridgesTvlView.tsx
antooni: I would stick to my solution just for now, in the future we can do the refactor

PR #1289, opened 2023-03-13
Update Nuggets thumbnails

antooni:
This PR aims to change thumbnails for knowledge nuggets, currently they are downloaded from URL. This change will add static assets to our repo and generate webpage from those assets. When thumbnail is not specified, then the default would be used.

Additionally the test have been written to ensure static asset usage is correct. Test are inside config and frontend. Inside config there is a test checking that only static assets are used for thumbnails, and that they resolve to correct paths. Inside frontend there is a check that the path resolve to the correct folder. 

Resolves L2B-1056


krzkaczor (approved):

michalsidzej (approved):
LGTM

PR #1291, opened 2023-03-14
Update discovery Across

antooni:
This PR will:
- update discovery config for accross
- change the logic of daily updates (to include projects with errors)


michalsidzej (approved):
Maybe our internal bot could notify us about errors in the project? Or do we have a test for that?

antooni:
> Maybe our internal bot could notify us about errors in the project? Or do we have a test for that?

We have a test for that

PR #1293, opened 2023-03-14
Use discovery result for new Starknet proof verifier

adamiak:
* Update discovery config due to GpsStatementVerifier contract implementation change.
* Use discovered result on the website.


packages/config/src/layer2s/common/contracts.ts
```diff
@@ -29,17 +32,7 @@ export const SHARP_VERIFIER_CONTRACT: ProjectContractSingleAddress = {
   description:
     'Starkware SHARP verifier used collectively by StarkNet, Sorare, Immutable X and rhino.fi. It receives STARK proofs from the Prover attesting to the integrity of the Execution Trace of these four Programs including correctly computed L2 state root which is part of the Program Output.',
   address: EthereumAddress('0x47312450B3Ac8b5b8e247a6bB6d523e7605bDb60'),
```
michalsidzej, line 34:
```suggestion
  address: discovery.getContract('CallProxy').address,
```

michalsidzej (approved):

PR #1298, opened 2023-03-16
Update Arbitrum upgradeability risk factor

PlasmaPower:
Arbitrum One and Arbitrum Nova are now both owned by the Arbitrum DAO, which has a 13 day delay on upgrades going through after approval. However, there is a 9/12 security council multisig that can make instant upgrades, as noted here.


michalsidzej:
Thanks for the PR! As it is a significant update, we'll handle it in https://github.com/l2beat/l2beat/pull/1300

PR #1301, opened 2023-03-18
Decode constructor arguments

krzkaczor:
Closes: L2B-1000, L2B-1136


packages/backend/src/core/discovery/handlers/user/ConstructorArgsHandler.test.ts
```diff
@@ -0,0 +1,149 @@
...
+const sampleAbi = ['constructor(string,string,uint8,bytes32,uint256)']
+
+describe(ConstructorArgsHandler.name, () => {
+  describe(ConstructorArgsHandler.prototype.execute.name, () => {
+    it('works', async () => {
```
antooni, line 19: correctly fetches contructor arguments

PR #1305, opened 2023-03-20
Discover L1CustomGateway for Arbitrum

michalsidzej:


packages/backend/discovery/arbitrum/config.jsonc
```diff
@@ -56,6 +57,22 @@
...
+    "0xE6841D92B0C345144506576eC13ECf5103aC7f49": {
+      "ignoreMethods": ["l2Timelock"]
+    },
+    // L1Erc20Gateway
+    "0xa3A7B6F88361F48403514059F1F16C8E78d60EeC": {
+      "ignoreMethods": ["counterpartGateway"]
+    },
+    // L1GatewayRouter
+    "0x72Ce9c846789fdB6fC1f34aC4AD25Dd9ef7031ef": {
+      "ignoreMethods": ["counterpartGateway"]
```
luke7211, line 71: These should be "ignoreDiscovery" rather than "ignoreMethods"

luke7211 (changes requested):

packages/backend/discovery/arbitrum/config.jsonc
michalsidzej: Ignored methods are ignored in discovery. Ignore discovery refers to the key (address). I intentionally ignore this method so when potentially this address changes, we will still skip the discovery of it.

luke7211 (approved):

PR #1306, opened 2023-03-20
Add hover to rosette

michalsidzej:
Resolves L2B-1131


sz-piotr (commented):
Seems to work really good now

sz-piotr (approved):

PR #1310, opened 2023-03-22
Updating zkSync Era information

githubdoramon:
This PR will be valid on March 24th. I'll change its state then


packages/config/src/layer2s/zksyncera.ts
```diff
@@ -21,15 +21,14 @@ export const zksyncera: Layer2 = {
   display: {
     name: 'zkSync Era',
     slug: 'zksync-era',
-    warning: 'Currently only registered addresses can bridge to zkSync Era.',
```
michalsidzej, line 24: Is this true? I'm trying to bridge to zksync era using https://bridge.zksync.io/ and I keep getting this prompt 
Screenshot 2023-03-23 at 11 59 26

githubdoramon: @michalsidzej this PR is supposed to go live tomorrow :) - Then, it will be true

michalsidzej: Cool, make sure you include it into description or mark the PR as draft :) Thanks for contributing tho 👍

githubdoramon: good point... added now, and will change tomorrow

githubdoramon:
@michalsidzej PR is now valid and ready for review :)

PR #1313, opened 2023-03-23
Update to earl 0.3 canary

sz-piotr:
Resolves L2B-1225
The goal is for L2BEAT to dogfood earl's new 0.3 version. There are currently no docs for this version except the doc comments on functions.


packages/backend/src/api/controllers/BlocksController.test.ts
```diff
@@ -15,7 +15,7 @@ describe(BlocksController.name, () => {
       },
     })
     const blocksController = new BlocksController(blockNumberRepository)
-    expect<unknown>(await blocksController.getAllBlocks()).toEqual([
+    expect(await blocksController.getAllBlocks()).toEqual([
```
antooni, line 18: Can you provide the context in PR description why `<unknown>` is no longer needed? (I prefer description because it will stay as a "documentation" for this migration)

packages/backend/src/core/BlockNumberUpdater.test.ts
```diff
@@ -47,27 +47,32 @@ describe(BlockNumberUpdater.name, () => {
       await blockNumberUpdater.start()
 
       await waitForExpect(() => {
-        expect(etherscan.getBlockNumberAtOrBefore).toHaveBeenCalledExactlyWith([
```
antooni, line 50: I am more than happy that we got rid of this matcher!

krzkaczor (approved):

packages/backend/src/api/controllers/BlocksController.test.ts
sz-piotr: Done

antooni (approved):
LGTM, i left some small comments

PR #1317, opened 2023-03-23
Make ARB associated Token for Arbitrum

michalsidzej:


packages/config/src/layer2s/arbitrum.ts
```diff
@@ -53,6 +53,7 @@ export const arbitrum: Layer2 = {
     activityDataSource: 'Blockchain RPC',
   },
   config: {
+    associatedTokens: ['ARB'],
     escrows: [
       {
         address: EthereumAddress('0x8315177aB297bA92A06054cE80a67Ed4DBd7ed3a'),
```
luke7211, line 59: ARB token on L1 is a wrapper for ARB token on L2, no ? So you will be counting this token twice

luke7211 (changes requested):

packages/config/src/layer2s/arbitrum.ts
michalsidzej: No we don't, as the token is locked on L2 and minted on L1

luke7211 (approved):

PR #1322, opened 2023-03-24
Discovery config names

antooni:
This PR aims to add custom names map to the discovery config

e.g.
image

Resolves L2B-1144

### DIFF
--- a/packages/backend/discovery/config.schema.json
+++ b/packages/backend/discovery/config.schema.json
@@ -23,7 +23,7 @@
       "description": "Specific configuration for discovered contracts",
       "unevaluatedProperties": false,
       "patternProperties": {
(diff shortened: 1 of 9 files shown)


packages/backend/src/api/controllers/status/discovery/view/components/Header.tsx
```diff
@@ -16,7 +16,7 @@ export function Header(props: HeaderProps) {
   return (
     <blockquote>
       <h4>
-        <ContractName contract={props.contract} />
+        <ContractName contract={props.contract} showDerivedName={true} />
```
sz-piotr, line 19: Nitpick: you can omit `={true}`

sz-piotr (approved):

PR #1325, opened 2023-03-24
Update the Arbitrum One and Nova logos and links

bfreshhb:
Wanted to replace the old Arbitrum One and Nova chain logos with the new ones

As well make sure all the links are up to date on both!


antooni (changes requested):
Can you provide me any proof that those icons are indeed the new visual identification of Arbitrum? because I cannot find them being used anywhere @bfreshhb

PR #1341, opened 2023-03-27
Update scroll.ts

OrestTa:


OrestTa:
Hm, looks like only legacy YouTube handles are supported by this check:

```
          } else if (link.includes('youtube')) {
            if (!link.includes('playlist')) {
              expect(link).toMatchRegex(
                /^https:\/\/youtube\.com\/(c|channel)\/[\w-]+$/,
              )
            }
```
Removing Scroll's YouTube for now

PR #1345, opened 2023-03-28
Add descriptions to discovery

antooni:
This PR aims to add the functionality of adding descriptions to contracts and methods in discovery. This PR will not change the discovery logic, only the way the output is displayed on the dashboard

Resolves L2B-1138
Resolves L2B-1139


michalsidzej (changes requested):
An example of usage?

packages/backend/src/api/controllers/status/discovery/props/dashboardContracts/getDescription.ts
```diff
@@ -0,0 +1,11 @@
+import { EthereumAddress } from '@l2beat/shared'
+
+import { DiscoveryConfig } from '../../../../../../core/discovery/DiscoveryConfig'
+
+export function getDescription(
+  field: string,
+  address: EthereumAddress,
+  config: DiscoveryConfig,
+): string | undefined {
```
michalsidzej, line 9:
```suggestion
export function getDescription(
  config: DiscoveryConfig,
  address: EthereumAddress,
  field: string,
): string | undefined {
```
It is a nit pick, but I think we should have a consistent approach to arguments order

packages/backend/src/api/controllers/status/discovery/props/dashboardContracts/getIgnoreInWatchMode.ts
```diff
@@ -19,6 +20,7 @@ export function getIgnoreInWatchMode(
         return {
           name: getFieldName(field, viewABI),
           values: getValues(discovery, contract, field),
+          description: getDescription(field, contract.address, config),
```
michalsidzej, line 23: Looking at getValues here it seems to me this order is more consistent
```suggestion
          values: getValues(discovery, contract, field),
          description: getDescription(config, contract.address, field),
```

packages/backend/src/api/controllers/status/discovery/props/dashboardContracts/getIgnoredMethods.ts
```diff
@@ -17,6 +18,7 @@ export function getIgnoredMethods(
       ignoreMethods = override.ignoreMethods.map((field) => {
         return {
           name: getFieldName(field, viewABI),
+          description: getDescription(field, contract.address, config),
```
michalsidzej, line 21: But here it's the opposite

packages/backend/src/api/controllers/status/discovery/props/dashboardContracts/getDescription.ts
antooni: I agree

antooni:
> I'm not a fan of this approach of having 2 objects that refer to the same contract (one in `overrides` and one in `descriptions`. IMO this can prove to be a bad DX. It's fine for now though

Yeah, I have a similiar feeling

PR #1346, opened 2023-03-28
Improve upcoming risks display

antooni:
image

image

image


michalsidzej (approved):
Please add an example of UPCOMING to storybook, otherwise LGTM

PR #1349, opened 2023-03-28
Polygon zkevm activity

krzkaczor:


michalsidzej (approved):

michalsidzej (commented):
I left it for a bit and it synced up to 2.8k block. Should be fine

PR #1369, opened 2023-04-03
Update chart tooltip & control colors

michalsidzej:
Resolves L2B-1110


antooni (approved):
Dark mode visual regression is something we need ;p

PR #1373, opened 2023-04-05
Derive escrow section

krzkaczor:
1. New escrow format introduced, with fields like upgradability. 
2. Derive escrows from discovery.
3. Drop duplicated entries from smart contract section UNLESS given escrow has another usecase (like `ArbitrumOneBridge`)

### DIFF
--- a/packages/backend/discovery/arbitrum/config.jsonc
+++ b/packages/backend/discovery/arbitrum/config.jsonc
@@ -100,7 +100,6 @@
         "counterpartGateway",
         "inbox",
         "l1Dai",
-        "l1Escrow",
         "l1Router",
         "l2Counterpart",
         "l2Dai"
--- a/packages/config/src/discovery/ProjectDiscovery.ts
+++ b/packages/config/src/discovery/ProjectDiscovery.ts
@@ -76,7 +76,7 @@ export class ProjectDiscovery {
     let contract: ContractParameters
     if (path) {
       const address = this.getAddressFromValue(identifier, path)
-      contract = this.getContractByAddress(address)
+      contract = this.getContractByAddress(address.toString())
     } else {
       contract = this.getContract(identifier)
     }
--- a/packages/config/src/discovery/ProjectDiscovery.ts
+++ b/packages/config/src/discovery/ProjectDiscovery.ts
@@ -10,7 +10,11 @@ import fs from 'fs'
 import { isArray, isString } from 'lodash'
 import path from 'path'
 
-import { ProjectPermission, ProjectPermissionedAccount } from '../common'
+import {
+  ProjectEscrowFromDiscovery,
+  ProjectPermission,
+  ProjectPermissionedAccount,
+} from '../common'
 import {
   ProjectContract,
   ProjectContractSingleAddress,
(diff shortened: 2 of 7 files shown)


packages/config/src/discovery/ProjectDiscovery.ts
```diff
@@ -64,6 +68,28 @@ export class ProjectDiscovery {
     }
   }
 
+  getEscrowDetails(
+    identifier: string,
+    description?: string,
+    path?: string,
```
antooni, line 74: rename path to key

packages/config/src/layer2s/arbitrum.ts
```diff
@@ -58,31 +58,41 @@ export const arbitrum: Layer2 = {
     associatedTokens: ['ARB'],
     escrows: [
       {
-        address: discovery.getContract('ArbitrumOneBridge').address,
+        ...discovery.getEscrowDetails('ArbitrumOneBridge'),
         sinceTimestamp: new UnixTime(1661450734),
```
antooni, line 62: after more thought I think more readable solution would be something like this:
```
        ...discovery.getEscrowDetails({
          name: 'ArbitrumOneBridge',
          sinceTimestamp: new UnixTime(1661450734),
          tokens: ['ETH'],
          description:
            'Contract managing Inboxes and Outboxes. It escrows ETH sent to L2.',
        }),
    ```

File: packages/backend/discovery/arbitrum/config.jsonc [hunk 3]
```diff
@@ -100,7 +100,6 @@
         "counterpartGateway",
         "inbox",
         "l1Dai",
-        "l1Escrow",
```
antooni, line 103: why there is need to discover this contract?

packages/frontend/src/utils/project/getContractSection.ts
```diff
@@ -23,6 +28,23 @@ export function getContractSection(
     return makeTechnologyContract(contract, project, isUnverified)
   })
 
+  const escrows = project.config.escrows
+    .sort(moreTokensFirst)
+    .map((escrow) => {
```
antooni, line 33: maybe a better pattern would be to filter out the hidden and not newFormat after the sort, because currently it is a bit hard to understand at first glance

packages/backend/discovery/arbitrum/config.jsonc
antooni: fair enough

packages/frontend/src/utils/project/getContractSection.ts
```diff
@@ -227,3 +253,39 @@ function isContractUnverified(
...
+  },
+): boolean {
+  return verificationStatus.contracts[address.toString()] === false
+}
+
+function escrowToProjectContract(
+  escrow: ProjectEscrow,
+): ProjectContract | undefined {
+  // project config was not yet migrated to the new format
+  if (!escrow.newVersion || escrow.hidden) {
```
antooni, line 271: you can get rid of this condition, it is checked in the loop in the function above

krzkaczor: I left in on purpose because it's not obvious that this is checked above. Either it should be asserted (And then we can get rid of undefined in the sig) or it should be the way i tweaked it.

packages/frontend/src/utils/project/getContractSection.ts
```diff
@@ -187,7 +212,8 @@ function makeTechnologyContract(
     const tokens = project.config.escrows.find(
       (x) => x.address === item.address,
     )?.tokens
-    if (tokens) {
+    // if contract is an escrow we already tweak it's name so we don't need to add this
```
antooni, line 215: Is this logic really necessary? If I understand correctly it will update the descriptions for contracts that are also inside escrows, but do not follow the newFormat, right?

I feel like this adds additional logic to the already comlpex function and does not provide much value. Maybe let's just assume that we update escrows to the new format as soon as possible, WDYT?

antooni (changes requested):
please address my comments before merging, otherwise LGTM

antooni (approved):

PR #1376, opened 2023-04-06
Add Taiko as an L2 project

d1onys1us:
## Summary of changes
- [x] Add Taiko as described in CONTRIBUTING.md.
- [x] Reorder some of the improperly ordered exports.


packages/config/src/layer2s/index.ts
```diff
@@ -1,18 +1,19 @@
...
-import { HOMEPAGE_MILESTONES } from './common'
-import { linea } from './consensys'
 import { dydx } from './dydx'
 import { fuelv1 } from './fuelv1'
 import { gluon } from './gluon'
 import { hermez } from './hermez'
 import { immutablex } from './immutablex'
 import { layer2finance } from './layer2finance'
 import { layer2financezk } from './layer2financezk'
+import { linea } from './consensys'
```
michalsidzej, line 16: We have an eslint rule for imports. Changing them manually will cause the CI to fail

antooni (approved):

PR #1377, opened 2023-04-06
Use immutablex discovery

michalsidzej:


packages/config/src/layer2s/immutablex.ts
```diff
@@ -56,7 +72,7 @@ export const immutablex: Layer2 = {
   riskView: makeBridgeCompatible({
     stateValidation: RISK_VIEW.STATE_ZKP_ST,
     dataAvailability: RISK_VIEW.DATA_EXTERNAL_DAC,
-    upgradeability: RISK_VIEW.UPGRADE_DELAY('14 days'),
+    upgradeability: RISK_VIEW.UPGRADE_DELAY(delay),
```
antooni, line 75: <3

antooni (approved):
LGTM

krzkaczor (approved):

packages/config/src/layer2s/immutablex.ts
michalsidzej: This is not the final form, as it potentially shows '0 days' but warning. Not a dealbraker rn, this will be solved here https://linear.app/l2beat/issue/L2B-1305/get-upgrade-delay-risk-from-discovery

PR #1379, opened 2023-04-06
Add discovery dry run

michalsidzej:
Resolves L2B-1300


antooni (approved):
My only concern is whether one day is long enough time period, otherwise LGTM

PR #1381, opened 2023-04-06
Add force transaction risk to zkSync Era

michalsidzej:


packages/config/src/layer2s/zksyncera.ts
```diff
@@ -118,7 +119,7 @@ export const zksyncera: Layer2 = {
       description:
         'If a user is censored by L2 Sequencer, they can try to force transaction via L1 queue. Right now there is no mechanism that forces L2 Sequencer to include\
         transactions from L1 queue in an L1 block.',
-      risks: [],
+      risks: FORCE_TRANSACTIONS.NO_MECHANISM.risks,
```
antooni, line 122: if it is possible add link to the source code with force transactions inside references

antooni (approved):
LGTM, adding a reference would a nice addition

PR #1382, opened 2023-04-07
Update rosette colors

antooni:
Resolves L2B-1311


michalsidzej (changes requested):
The upcoming rosette should be gray
Screenshot 2023-04-07 at 11 03 03

antooni:
> It seems to me that all the rosettes have the `isBrightGreen` prop set to true. Is that intentional?

yeah, rosette is bright in most cases, co probably a flag isDarkGreen would be more handy

PR #1388, opened 2023-04-11
Update discovery for starkex

michalsidzej:


packages/backend/discovery/apex/config.jsonc
```diff
@@ -47,6 +47,9 @@
           "type": "constructorArgs"
         }
       }
+    },
+    "0x21F9eC47b19d95b5C2DDFB6Ae5D4F92fAdacAEc4": {
```
antooni, line 51: fair enough, but would be prettier with custom name ;p

antooni (approved):

PR #1394, opened 2023-04-14
Update zkSync Era config

antooni:
This PR will:
- update zkSync Era .ts config to use new discovery features
- add references to riskView
- update functionality of source code fetching

Resolves L2B-1255


packages/config/src/discovery/values/hardcoded.ts
```diff
@@ -19,4 +19,13 @@ export const HARDCODED = {
...
     SET_SEQUENCER_COUNT: 3,
   },
+  ZKSYNC_2: {
+    FACETS: [
+      '0xF1fB730b7f8E8391B27B91f8f791e10E4a53CEcc',
+      '0x6df4A6D71622860dcc64C1FD9645d9a5BE96f088',
+      '0x2a2d6010202B93E727b61a60dfC1d5CF2707c1CE',
+      '0xc796A402E1B26eCd2cd38F23e05A2f904504ec89',
+      '0x389a081BCf20e5803288183b929F08458F1d863D',
+    ],
```
michalsidzej, line 29: Why are those hardcoded?

packages/config/src/discovery/values/projects/zksync2/ZKSYNC_2.ts
```diff
@@ -0,0 +1,12 @@
+import { ProjectRiskViewEntry } from '../../../../common'
+
+const UPGRADEABILITY_STRING: ProjectRiskViewEntry = {
+  value: 'Yes',
+  description:
+    'The code that secures the system can be instantly upgraded without any notice.',
+  sentiment: 'bad',
+}
```
michalsidzej, line 8: How does this differ from RISK_VIEW.UPGRADABLE_YES? We should be really careful while creating custom risk per project. This can cause some differences in how we categorise the same risk.

packages/config/src/layer2s/zksyncera.ts
```diff
@@ -73,10 +77,54 @@ export const zksyncera: Layer2 = {
...
+      references: [
+        'https://etherscan.io/address/0x3dB52cE065f728011Ac6732222270b3F2360d919#code#F5#L89',
+        'https://etherscan.io/address/0x389a081BCf20e5803288183b929F08458F1d863D#code#F10#L254',
+        'https://etherscan.io/address/0xF1fB730b7f8E8391B27B91f8f791e10E4a53CEcc#code#F7#L24',
+        'https://etherscan.io/address/0x473b1887d45d61efd87731a1d8ec3590b93c565d#code#F5#L227',
+        'https://era.zksync.io/docs/dev/developer-guides/transactions/transactions.html#transaction-types',
+        'https://era.zksync.io/docs/dev/developer-guides/system-contracts.html#executorfacet',
+      ],
+      contracts: ['ValidatorTimelock', 'DiamondProxy', 'Verifier'],
+    },
```
michalsidzej, line 93: WYT about spreading value and description from a common file and adding references on top of that? This will prevent us from typos and differing descriptions for the same risks

packages/config/src/discovery/values/hardcoded.ts
antooni: when any of the facets changes then the upgradability can change, so it should be re-evaluated

luke7211 (approved):

PR #1397, opened 2023-04-17
Add discovery to Gravity bridge

luke7211:
Add discovery to Gravity Bridge


michalsidzej (approved):
This hardly watches anything, but LGTM

PR #1398, opened 2023-04-17
Improve getProxyGovernors stability

michalsidzej:
Throw when proxyIsGovernor reverted to prevent false negatives


michalsidzej:
This would not solve the problem

PR #1406, opened 2023-04-19
Update readme

michalsidzej:


packages/backend/src/core/DiscoveryWatcher.ts
```diff
@@ -42,6 +42,7 @@ export class DiscoveryWatcher {
 
   start() {
     this.logger.info('Started')
+    this.taskQueue.addToFront(UnixTime.now())
```
antooni, line 45: why this one? we are only running the bot every hour

antooni (changes requested):

packages/backend/src/core/DiscoveryWatcher.ts
michalsidzej: Oupsi dupsi

antooni (approved):

PR #1407, opened 2023-04-19
Add sanity check to discord bot

antooni:
Sometimes Alchemy reverts without a reason, when it should not, or the Etherscan does not return the source code when it should. This PR aims to add a sanity check to prevent such malfunctions to influence our bot stability.

When a bot detects the chane, it re-runs the discovery to see whether the result is the same, if not then there obviously was an 3rd party API error and should notify this one to our internal discord.

Resolves L2B-1353

### DIFF
--- a/packages/backend/src/core/DiscoveryWatcher.ts
+++ b/packages/backend/src/core/DiscoveryWatcher.ts
@@ -1,5 +1,6 @@
 import { Hash256, Logger, ProjectParameters, UnixTime } from '@l2beat/shared'
 import { providers } from 'ethers'
+import { isEqual } from 'lodash'
 import { Gauge, Histogram } from 'prom-client'
 
 import { DiscoveryWatcherRepository } from '../peripherals/database/discovery/DiscoveryWatcherRepository'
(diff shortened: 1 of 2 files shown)


michalsidzej (approved):

packages/backend/src/core/DiscoveryWatcher.ts
```diff
@@ -214,6 +217,28 @@ export class DiscoveryWatcher {
...
+    discovery: ProjectParameters,
+    projectConfig: DiscoveryConfig,
+    blockNumber: number,
+  ) {
+    const secondDiscovery = await this.discoveryEngine.run(
+      projectConfig,
+      blockNumber,
+    )
+
+    if (!isEqual(discovery, secondDiscovery)) {
```
sz-piotr, line 233: I think this makes sense, but I'd like to get a third opinion on whether this check doesn't run into some weird edge cases because for some reason our discovery code is non-deterministic @michalsidzej

sz-piotr (approved):

PR #1408, opened 2023-04-19
 Upgrade risk function from seconds

michalsidzej:
Resolves L2B-1305


packages/config/src/layer2s/common/contracts.ts
```diff
@@ -24,6 +26,14 @@ function UPGRADE_WITH_DELAY_RISK(delay: string): ProjectRisk {
   }
 }
 
+function UPGRADE_WITH_DELAY_SECONDS_RISK(delaySeconds: number): ProjectRisk {
```
antooni, line 29: this constant is not used anywhere

antooni (approved):
LGTM

PR #1409, opened 2023-04-19
Refactor analyzeItem into AddressAnalyzer

sz-piotr:
Resolves L2B-1360


packages/backend/src/core/discovery/discover.ts
```diff
@@ -1,16 +1,19 @@
...
+import { SourceCodeService } from './source/SourceCodeService'
 import { DiscoveryLogger } from './utils/DiscoveryLogger'
 
 export async function discover(
   provider: DiscoveryProvider,
   config: DiscoveryConfig,
   logger: DiscoveryLogger,
 ) {
-  const resolved = new Map<EthereumAddress, AnalyzedData>()
+  const resolved = new Map<EthereumAddress, Analysis>()
```
michalsidzej, line 16: This introduces a change to `discovered.json` (added key `type: "Contract"` to all contracts). Honestly I think that this is probably a good change (fewer interfaces, more logical data path through code), but adding this requires from us either updating all discovered.json OR agreeing that we'll see our bot to go mad. 

BTW, I wouldn't have spotted that if I haven't started the bot locally. Try adding that to your workflow, I can help you out in setting it up.

michalsidzej: I'm leaning towards `yarn discover --all --preserve-block-number`. Looks like an easy fix, but this script will take some time :/.

sz-piotr: I even thought about needing to remove that key, but later forgot. Good catch!

michalsidzej (approved):

PR #1429, opened 2023-04-26
Refactor discover function

sz-piotr:
This PR continues the work of refactoring discovery to use dependency injection and add tests.

This time it concerns itself with the `discover` function and the core discovery loop.


packages/backend/src/core/discovery/engine/DiscoveryEngine.ts
```diff
@@ -0,0 +1,40 @@
...
+
+  async discover(config: DiscoveryConfig) {
+    const resolved: Analysis[] = []
+    const stack = new DiscoveryStack()
+    stack.push(config.initialAddresses, 0)
+
+    while (!stack.isEmpty()) {
+      const item = stack.pop()
+
+      const reason = shouldSkip(item, config)
```
torztomasz, line 21: can't we put it as a method in DiscoveryStack?

torztomasz (commented):
The only thing that looks kinda ugly is keeping discover.ts and runDiscover.ts, can't we put theirs logic inside DiscoveryRunner?

packages/backend/src/core/discovery/engine/DiscoveryEngine.ts
sz-piotr: It's a pure function, I don't see a particular need.

sz-piotr:
> The only thing that looks kinda ugly is keeping discover.ts and runDiscover.ts, can't we put theirs logic inside DiscoveryRunner?

Yeah, the plan was to refactor bottom up, so I started a few PRs ago with AddressAnalyzer and so on, got to DiscoveryEngine and I'll keep going up. It's just a lot to do in one step, so its multiple PRs

torztomasz (approved):

PR #1435, opened 2023-04-27
Refactor DI in discovery

sz-piotr:
Requires  #1434


packages/backend/src/core/discovery/config/ConfigReader.ts
```diff
@@ -35,14 +35,14 @@ export class ConfigReader {
...
+  async readDiscovery(name: string): Promise<DiscoveryOutput> {
     const contents = await readFile(
       `discovery/${name}/discovered.json`,
       'utf-8',
     )
 
     const parsed: unknown = JSON.parse(contents)
 
-    return parsed as ProjectParameters
+    return parsed as DiscoveryOutput
```
torztomasz, line 46: maybe we should consider defining DiscoveryOutput in Zod? Then we would just do `DiscoveryOutput.parse(JSON.parse(contents))`

torztomasz (approved):

packages/backend/src/core/discovery/config/ConfigReader.ts
sz-piotr: I think it's out of scope of this PR but I'm going to add a task for it for the future

PR #1437, opened 2023-04-28
Update bridge details page

torztomasz:
This PR aims to update bridge details page to the latest changes of scaling details page.


packages/frontend/src/components/project/navigation/DesktopProjectNavigation.tsx
```diff
@@ -11,7 +12,7 @@ interface Project {
 }
 interface ProjectNavigationProps {
   project: Project
-  sections: ProjectDetailsSection[]
+  sections: (ScalingProjectDetailsSection | BridgeProjectDetailsSection)[]
```
sz-piotr, line 15: How about slightly shorter names?
```suggestion
  sections: (ScalingDetailsSection | BridgeDetailsSection)[]
```

sz-piotr (changes requested):

packages/frontend/src/components/project/navigation/DesktopProjectNavigation.tsx
torztomasz: That's a good catch

sz-piotr (approved):

PR #1439, opened 2023-04-28
Make milestones less crowded on mobiles

torztomasz:
Resolves L2B-848


sz-piotr (approved):
This sure makes them less crowded, but doesn't resolve the ease of clicking a specific one. There should be another task for this.

PR #1442, opened 2023-05-01
Make checkbox match with design

torztomasz:
Resolves L2B-1004


packages/frontend/src/components/Checkbox.tsx
```diff
@@ -1,6 +1,8 @@
 import cx from 'classnames'
 import React, { ReactNode } from 'react'
 
+import { CheckIcon } from './icons/symbols/CheckIcon'
```
sz-piotr, line 4: import from `./icons`

sz-piotr (changes requested):

sz-piotr (approved):

PR #1443, opened 2023-05-01
Improve reg-viz stability

torztomasz:
Resolves L2B-983


packages/frontend/package.json
```diff
@@ -20,9 +20,9 @@
     "test": "mocha",
     "typecheck:ci": "yarn typecheck",
     "typecheck": "tsc --noEmit",
-    "storybook": "start-storybook -p 6006 --no-open",
-    "storybook:build": "(cd ../.. && wsrun -p @l2beat/{shared,config} -recsm build --module es6) && build-storybook",
-    "storybook:screenshot": "yarn storycap --serverCmd \"npx http-server ./storybook-static -p 9001\" http://localhost:9001 -V 2560x1440 -V 390x844"
+    "storybook": "storybook dev -p 6006 --no-open",
+    "storybook:build": "(cd ../.. && wsrun -p @l2beat/{shared,config} -recsm build --module es6) && storybook build",
+    "storybook:screenshot": "yarn storycap --serverCmd \"npx http-server ./storybook-static -p 9001\" http://localhost:9001 --captureTimeout 50000 --parallel 8"
```
krzkaczor, line 25: 50s is quite long. Have you tried shorter timeouts?

packages/frontend/package.json
```diff
@@ -66,9 +66,11 @@
     "prettier": "^2.6.2",
     "prettier-plugin-tailwindcss": "^0.2.2",
     "sass": "^1.52.2",
-    "storybook-dark-mode": "^1.1.0",
+    "storybook": "^7.0.9",
+    "storybook-dark-mode": "^3.0.0",
     "storycap": "^3.1.9",
     "tailwindcss": "^3.1.0",
+    "ts-node": "^10.9.1",
```
krzkaczor, line 73: new dep?

krzkaczor (approved):

packages/frontend/package.json
torztomasz: Unfortunately, charts are taking so long to set up (~40000ms). I will revisit chart stories to try to optimize it but I tried once and could not do it :(

packages/frontend/package.json
torztomasz: When I started using ts files for storybook (main and preview), build started to fail, adding this as a dev dependency fixed the issue

krzkaczor:
Summary of the problems with storybook building:
* the core of the issue is that storybook in devmode renders things in the browser, but our frontend always does `renderToString`,
* in our `shared` package, we have modules that don't work in the browser (we even import `prom-client` there!)


Solutions:
* try to extract pure frontend modules from `shared` and hope that none of the components that we render with storybook uses something nodejs specific,
* try to extract things related to vir-reg rendering stability from this PR and ignore bigger issues with the setup for now (maybe we need to rethink whole frontend architecture with such issues in mind ie. migrate to next or qwik or something else).

PR #1446, opened 2023-05-04
Fix project details issues

torztomasz:
This PR aims to remove bugs that came up after updating project details page


claudioantonio:
The title for the nugget knowledge section has a black background and foreground that does not allow to read the section title.
image

If you highlight where it was supposed to have the section title, you can see it.

image

sz-piotr (approved):

PR #1450, opened 2023-05-07
Add navigation to faq page

torztomasz:
Resolves L2B-963


krzkaczor (approved):

antooni (changes requested):
Maybe we should make the navigation menu sticky, the same as on the project detail?

torztomasz:
> Maybe we should make the navigation menu sticky, the same as on the project detail?

The navigation is overflowing on smaller screens. That's why it is not possible.

antooni (approved):

PR #1457, opened 2023-05-10
add degate

BergDev22:
This PR is to add DeGate. DeGate is an app-specific ZK rollup that enables a Trustless, fast and low-fee decentralized order book exchange. https://degate.com/


packages/config/src/layer2s/degate.ts
```diff
@@ -0,0 +1,265 @@
...
+)} ETH`
+
+export const degate: Layer2 = {
+  type: 'layer2',
+  id: ProjectId('degate'),
+  display: {
+    name: 'DeGate',
+    slug: 'degate',
+    description:
+      'DeGate is an app-specific ZK rollup that enables a Trustless, fast and low-fee decentralized order book exchange, helping users to trade easy and sleep easy.',
```
lucadonnoh, line 46:
```suggestion
      'DeGate is an app-specific ZK rollup that enables a trustless, fast and low-fee decentralized order book exchange, helping users to trade easy and sleep easy.',
```

packages/config/src/layer2s/degate.ts
```diff
@@ -0,0 +1,265 @@
...
+        'https://youtube.com/@degatedex1718',
+        'https://medium.com/degate',
+        'https://mirror.xyz/0x078a601f492043C8e7D0E15B0F8815f58b4c342f',
+      ],
+    },
+  },
+  config: {
+    associatedTokens: ['DG'],
+    escrows: [
+      discovery.getEscrowDetails({
```
lucadonnoh, line 66: escrows should be hardcoded, see here: https://www.notion.so/l2beat/Config-progress-ec546917540243fbb0ec599a7af646f3

packages/config/src/layer2s/degate.ts
```diff
@@ -0,0 +1,265 @@
...
+
+        const permissionedAccount = discovery.formatPermissionedAccount(owner1)
+
+        // if it was updated, we should add multisig participants
+        assert(permissionedAccount.type === 'EOA')
+
+        return [permissionedAccount]
+      })(),
+      description:
+        'This address is the owner of the following contracts: LoopringIOExchangeOwner, LoopringV3, DefaultDepositContract, BlockVerifier. Can add or remove block submitters. Can change the forced withdrawal fee. Can change a way that balance is calculated per contract during the deposit, allowing the support of non-standard tokens.',
```
lucadonnoh, line 220: I would specify here that there is a max fee value (and the actual value), it is an important difference compared to Loopring

packages/config/src/layer2s/degate.ts
lucadonnoh: I would also add that it is a fork of Loopring, to give context the same way we do for example for Boba and Metis

PR #1459, opened 2023-05-11
Add delay sanity check

michalsidzej:
Resolves L2B-1438


antooni (changes requested):
Add tests for both sanity checks

packages/frontend/src/build/sanityCheck.ts
```diff
@@ -0,0 +1,204 @@
...
+  const bridgesInApi = bridges.filter((x) => projectsInApi.includes(x.id))
+  const layer2sInApi = layer2s.filter((x) => projectsInApi.includes(x.id))
+
+  if (layer2sInApi.length / layer2s.length < 0.8) {
+    throw new Error(
+      'The API has returned an insufficient number of layer2s tvl',
+    )
+  }
+
+  if (bridgesInApi.length / bridges.length < 0.8) {
```
antooni, line 25: calculate this value via filtering the upcoming, and then compare with 0.8

packages/frontend/src/build/sanityCheck.test.ts
```diff
@@ -0,0 +1,220 @@
...
+      'projectB',
+      [
+        [today.add(-1, 'days'), 3],
+        [today, 4],
+      ],
+    ],
+  ]
+
+  describe(checkIfEmptyActivityCharts.name, () => {
+    it('happy path', () => {
```
antooni, line 123: rename to "data fully synced"

packages/frontend/src/build/sanityCheck.test.ts
```diff
@@ -0,0 +1,220 @@
...
+        checkIfZeroTpsProjects(allProjects, importantProjects),
+      ).not.toThrow()
+    })
+
+    it('throws if zero tps important project exists', () => {
+      const allProjects: ActivityProjectData[] = [
+        [
+          'projectA',
+          [
+            [today.add(-1, 'days'), 1],
```
antooni, line 159: return only one row with zero value, test will be more readable

antooni (changes requested):
address requested changes and change "happy case" to something more descriptive 
change delay to 2days5hours and make a comment inside activity backend

PR #1460, opened 2023-05-11
Add SKALE

payvint:
Hey L2beat,

Just recently we found out your platform and were impressed!
And we want to add SKALE Network(which is EVM multichain network and currently has 20 chains and the number would be increased) to L2 and bridges comparison.
But we did not find so much resources or templates how we can add it.
Can you please show and help us to move in a right direction to add SKALE to your platform?:)

Thanks!


luke7211:
Thank you @payvint for the PR. It looks really good ! We will hook the discovery engine and we will merge it shortly

payvint:
Hey @luke7211 if you can send some example how to hook the discovery engine, I will do it!

antooni (approved):

PR #1461, opened 2023-05-12
Align activity Ethereum chart

michalsidzej:
Resolves L2B-1427


packages/frontend/src/build/createApi.ts
```diff
@@ -77,13 +77,27 @@ function getActivityChart(
   apiChart: ActivityApiChart,
   ethereumChart: ActivityApiChart,
 ): FrontendActivityChart {
-  const length = Math.min(apiChart.data.length, ethereumChart.data.length)
+  const lastProjectTimestamp = apiChart.data.at(-1)?.[0]
```
antooni, line 80: write this logic as a pure function and add a test for it

packages/frontend/src/build/createApi.ts
```diff
@@ -77,13 +77,27 @@ function getActivityChart(
...
+    (x) => x[0].toNumber() === lastProjectTimestamp.toNumber(),
+  )
+  if (ethChartTimestampIndex === -1) {
+    throw new Error('No matching timestamp in ethereum chart')
+  }
+  const alignedEthChartData = ethereumChart.data.slice(
+    0,
+    ethChartTimestampIndex + 1,
+  )
+  const length = Math.min(apiChart.data.length, alignedEthChartData.length)
```
antooni, line 94: are we sure that both Ethereum and projects data start from the same timestamp?

antooni (changes requested):
Add tests

packages/frontend/src/build/createApi.ts
michalsidzej: Fair point, tested, works :)

packages/frontend/src/build/createApi.ts
```diff
@@ -73,17 +73,31 @@ export function outputCharts(
...
   return {
```
antooni, line 95: this function would be much more readable if it only said: cut the data and return, write everything above return as a seperate function

antooni (changes requested):
Good work, please address some small issues.

antooni (approved):
LGTM

PR #1462, opened 2023-05-12
Update dydx delay risk

michalsidzej:
Resolves L2B-1431

### DIFF
--- a/packages/backend/src/core/discovery/proxies/auto/StarkWareProxy.ts
+++ b/packages/backend/src/core/discovery/proxies/auto/StarkWareProxy.ts
@@ -126,6 +126,7 @@ export async function detectStarkWareProxy(
 const coder = new utils.Interface([
   'event Upgraded(address indexed implementation)',
   'event ImplementationUpgraded(address indexed implementation, bytes initializer)',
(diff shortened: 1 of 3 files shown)


packages/backend/src/core/discovery/proxies/auto/StarkWareProxy.ts
```diff
@@ -169,14 +170,41 @@ async function getStarkWareDiamond(
...
     ) {
       throw e
     }
-    console.log(
-      'Failed to decode upgradeTo data, falling back to decoding logs parameters',
-    )
-    data = `0x${lastUpgrade.data.slice(130)}`
+    console.log('Failed to decode upgradeTo data')
+
+    if (lastUpgrade.topics[0] === coder.getEventTopic('Upgraded')) {
```
antooni, line 180: this works for now, but tbh it really is becoming a pain point in our discovery, maybe all starkex deployments share the same storage layout? and we can somehow access the values?

antooni (approved):

packages/backend/src/core/discovery/proxies/auto/StarkWareProxy.ts
michalsidzej: I really don't think they do, but that's maybe with checking

PR #1463, opened 2023-05-12
Split shared package to common and services

michalsidzej:
Resolves L2B-1435


krzkaczor:
@torztomasz I think it would make sense to rebase your previous PR on top of that and see if things work now

PR #1469, opened 2023-05-12
Fix DB transaction bug in activity 

antooni:
Activity is up-to-date now, but the issue still exists and was solved via manual work.

IMO the problems lies in this function
https://github.com/l2beat/l2beat/blob/master/packages/backend/src/core/SequenceProcessor.ts#L194

somehow those two functions processRange and  setState do not run in the same transaction 
image

because we end up with the DB records that show that only processRange has been correctly executed
image

This PR aims to resolve the issue by changing the `addMany` method inside repositories. As @adamiak has found out during his work on `starkex-explorer`,  the `insert()` method from knex does always create an atomic transaction when passing array of rows to insert.
Now, `addMany` just calls `add` in a loop, so everything can happen inside one transaction.

Resolves L2B-1447


michalsidzej (changes requested):
One small thing here

michalsidzej (approved):

PR #1472, opened 2023-05-15
Update order in discovered.json

antooni:
WARNING: merge only after running the script on the newest discovery JSONs

This PR:
- adds sorting logic
- sorts JSON via script (no discovery rerun, just sort the array)

Resolves L2B-1471


michalsidzej (approved):
Nice one

PR #1473, opened 2023-05-15
Update frontend build logs

antooni:
Resolves L2B-1476


michalsidzej (commented):
Honestly, I don't know if that's necessary. Is there any benefit of moving this logic out? It's extracted to a function anyway, IMO moving it to a separate file will worsen the readability.

michalsidzej (approved):
GIT

PR #1475, opened 2023-05-16
Fix Optimism badge

michalsidzej:
Fixes a bug of missing op mask in optimism badge 
Screenshot 2023-05-16 at 07 57 51
This was caused by multiple svgs in the document that had the same maskId (#OP). Changing this to unique maskIds is good enough IMO.


torztomasz (approved):

michalsidzej:
> LGTM, although why do we need a mask there?
I want to have a transparent cutout.

PR #1489, opened 2023-05-17
Update Loopring discovery

lucadonnoh:


packages/backend/src/core/discovery/handlers/user/index.ts
```diff
@@ -64,6 +69,13 @@ export function getUserHandler(
       return new AccessControlHandler(field, definition, abi, logger)
     case 'arrayFromOneEvent':
       return new ArrayFromOneEventHandler(field, definition, abi, logger)
+    case 'arrayFromOneEventWithArgs':
```
antooni, line 72: add usage to Notion docs https://www.notion.so/l2beat/Handlers-8fd74177dde44b0bbc42f24f11aab812

packages/config/src/layer2s/loopring.ts
```diff
@@ -76,8 +80,21 @@ export const loopring: Layer2 = {
...
+    sequencerFailure: {
+      ...RISK_VIEW.SEQUENCER_FORCE_EXIT_L1,
+      description:
+        RISK_VIEW.SEQUENCER_FORCE_EXIT_L1.description +
+        ' The sequencer can censor individual deposits, but in such case users can get their funds back.',
+      references: [
+        'https://etherscan.io/address/0x26d8Ba776a067C5928841985bCe342f75BAE7E82#code#L7252',
+      ],
+    },
+    validatorFailure: {
```
antooni, line 92: maybe we want to add delay here as well?

packages/config/src/layer2s/loopring.ts
```diff
@@ -76,8 +80,21 @@ export const loopring: Layer2 = {
...
     upgradeability: RISK_VIEW.UPGRADABLE_YES,
-    sequencerFailure: RISK_VIEW.SEQUENCER_FORCE_EXIT_L1,
-    validatorFailure: RISK_VIEW.VALIDATOR_ESCAPE_MP,
+    sequencerFailure: {
+      ...RISK_VIEW.SEQUENCER_FORCE_EXIT_L1,
+      description:
+        RISK_VIEW.SEQUENCER_FORCE_EXIT_L1.description +
+        ' The sequencer can censor individual deposits, but in such case users can get their funds back.',
+      references: [
+        'https://etherscan.io/address/0x26d8Ba776a067C5928841985bCe342f75BAE7E82#code#L7252',
```
antooni, line 89: it is the function processConditionalTransactions(), right? is there a delay?

packages/config/src/layer2s/loopring.ts
```diff
@@ -114,12 +131,12 @@ export const loopring: Layer2 = {
       ...OPERATOR.CENTRALIZED_OPERATOR,
       references: [
         {
-          text: 'ExchangeV3.sol#L315-L322 - Loopring source code',
-          href: 'https://github.com/Loopring/protocols/blob/master/packages/loopring_v3/contracts/core/impl/ExchangeV3.sol#L315-L322',
+          text: 'ExchangeV3.sol#L315-L322 - Etherscan source code, submitBlocks function',
```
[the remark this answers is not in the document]
lucadonnoh: I'm following the same structure as other projects, if you prefer the shorter version I can do it for all the projects in another PR, I don't have strong opinions on this.

packages/backend/src/core/discovery/handlers/user/index.ts
lucadonnoh: done: https://www.notion.so/l2beat/Array-from-one-event-with-args-5d6fc2eb50414dce867ce2009b95d89e

packages/config/src/layer2s/loopring.ts
lucadonnoh: that function it's the one that shows that the sequencer can ignore deposits, I'll add the reference to the function to withdraw the deposit and the delay associated with it

packages/backend/src/core/discovery/handlers/user/ArrayFromOneEventWithArgHandler.ts
```diff
@@ -0,0 +1,93 @@
...
+  event: z.string(),
+  valueKey: z.string(),
+  flagKey: z.optional(z.string()),
+  invert: z.optional(z.boolean()),
+  ignoreRelative: z.optional(z.boolean()),
+  arg: z.string(),
+  argValue: z.string(),
+})
+
+export class ArrayFromOneEventWithArgsHandler implements Handler {
```
michalsidzej, line 26:
```suggestion
export class ArrayFromOneEventWithArgHandler implements Handler {
```

michalsidzej (commented):
One small naming change, otherwise LGTM

packages/config/src/layer2s/loopring.ts
lucadonnoh: done

PR #1491, opened 2023-05-17
Add NotificationManager to UpdateMonitor

antooni:
This PR will extract the notification logic to the other class from UpdateMonitor, this way, the interface for the UpdateMonitor would be to call NotificationManager.detectedChanges(diff).

I have removed the logic of "suppressing" the notification about the errors in discovered.json, because they should be displayed, but only on internal channels.

Also I have removed sending discord notification when 3rd party API is malfunctioning


packages/backend/src/core/discovery/NotificationManager.test.ts
```diff
@@ -0,0 +1,168 @@
...
+      expect(discordClient.sendMessage).toHaveBeenNthCalledWith(
+        2,
+        '***project-a*** | detected changes```diff\nContract | ' +
+          address.toString() +
+          '\n\nA\n- 1\n+ 2\n\n```',
+        'PUBLIC',
+      )
+    })
+
+    it('filters errors field changes', async () => {
```
michalsidzej, line 84:
```suggestion
    it('sends errors only to internal channel', async () => {
```

michalsidzej (approved):

PR #1495, opened 2023-05-18
Update discovery

antooni:
Optimism:
- rerun discovery, add nonce to ignored in watch mode

pNetwork:
- run discovery, the implementation of vault has been once again upgraded, seems like something connected with GALA token


lucadonnoh (approved):
looked a bit into pNetwork and yes it's used for migrating the GALA v1 tokens to v2. Not sure about the internal mechanisms of GALA (e.g. the double transfer), but I've added some notes on Notion. LGTM.

PR #1497, opened 2023-05-19
Move discovery into a package

sz-piotr:
This PR extracts the discovery code into its own package. Crucially the discovery CLI is now exclusively in that package and so backend once again doesn't have a CLI.

This change is intended to resolve the cyclic dependency between discovery and the config package. Now discovery no longer depends on config, rather backend depends on discovery and config and config depends on discovery.

This PR intentionally changes almost no functionality. The discovery is still run via the same `yarn discover` command in the backend package, the config.jsonc files also remain the the backend.

In the future I envision the config.jsonc files moving closer to [project].ts files.


packages/discovery/src/discovery/runDiscovery.ts
```diff
@@ -76,14 +74,7 @@ export async function dryRunDiscovery(
...
   if (diff.length > 0) {
-    const messages = diffToMessages(
-      projectConfig.name,
-      await findDependents(projectConfig.name, configReader),
-      diff,
-    )
-    for (const message of messages) {
-      console.log(message)
-    }
+    console.log(diff)
```
sz-piotr, line 77: Dear reviewer. This is the only real change in functionality. Because diffToMessages and findDependents were heavily dependent on external factors like discord message size and the configs being in backend I decided to drop nicer formatting for now. This was the easiest to do while preserving a logical separation of the two packages.

packages/discovery/src/discovery/config/ConfigReader.test.ts
```diff
@@ -2,8 +2,17 @@ import { expect } from 'earl'
 import { readdirSync } from 'fs'
 
 import { ConfigReader } from './ConfigReader'
+import path from 'path'
 
 describe(ConfigReader.name, () => {
+  // FIXME: This is a temporary hack to make the tests while configs are in backend
```
sz-piotr, line 8: This is here to make this PR work with minimal code changes. I expect the location of config.jsonc files to change in the future.

packages/backend/package.json
```diff
@@ -13,15 +13,16 @@
     "lint:fix": "yarn lint --fix",
     "lint": "eslint --ext .ts,.tsx --max-warnings 0 src",
     "start:dev": "node -r esbuild-register src server",
-    "start": "node -r source-map-support/register build/src/index.js server",
-    "discover": "node -r esbuild-register src/discoveryCli.ts discover",
-    "invert": "node -r esbuild-register src/discoveryCli.ts invert",
+    "start": "node -r source-map-support/register build/src/index.js",
+    "discover": "discovery discover",
+    "invert": "discovery invert",
```
sz-piotr, line 18: This corresponds to the package.json `"bin"` in `@l2beat/discovery`

krzkaczor (approved):

PR #1499, opened 2023-05-19
Update Loopring discovery

lucadonnoh:
There is an additional address in the state that was not picked up by the discovery that is allowed to arbitrarily set the forced withdrawal fee and change the block verifier address


packages/config/src/layer2s/loopring.ts
```diff
@@ -25,6 +25,12 @@ const maxAgeDepositUntilWithdrawable = formatSeconds(
     'getMaxAgeDepositUntilWithdrawable',
   ),
 )
+const forcedWithdrawalFee = discovery.getContractValue<number>(
+  'LoopringV3',
+  'forcedWithdrawalFee',
+)
+
+const forcedWithdrawalFeeString = `${forcedWithdrawalFee / 1e18} ETH`
```
sz-piotr, line 33: utils.formatEther would probably make more sense here

antooni (approved):

PR #1500, opened 2023-05-19
Notification nonce

antooni:
Resolves L2B-1354


packages/backend/src/peripherals/database/discovery/NotificationManagerRepository.ts
```diff
@@ -0,0 +1,79 @@
...
+export interface NotificationManagerRecord {
+  id: number
+  createdAt: UnixTime
+  updatedAt: UnixTime
+  projectName: string
+  blockNumber: number
+  diff: DiscoveryDiff[]
+}
+
+export class NotificationManagerRepository extends BaseRepository {
```
sz-piotr, line 17: Maybe rename to NotificationRepository or UpdateNotificationRepository?
If you go for UpdateNotificationRepository then NotificationManager could be renamed to UpdateNotifier.

sz-piotr (changes requested):

sz-piotr (approved):

Issue #1447, opened 2023-05-04
Automatically discover destination chains for across-v2


sz-piotr:
Currently the config and discovery for across have hardcoded chain ids and that's how it knows about chain specific contracts.

The task is to configure the discovery for across-v2 in such a way that the output includes a list of supported chain ids. This can be achieved either by using existing handlers if it turns out that they are sufficient or by writing a special purpose handler that can read this information from on chain.

compojoom:
Ok, I've been playing around a little and wanted to share my progress and ask for advise. 

The current config definiton is like:

```
"crossChainMainnet": {
          "type": "call",
          "method": "crossChainContracts",
          "args": [1]
},
"crossChainOptimism": {
          "type": "call",
          "method": "crossChainContracts",
          "args": [10]
}
...
```

That's fine and good, but if we want to have it dynamic it's hard when the chainname is part of the prop key. 

I've looked into the across code and since the crossChainContracts are stored in a struct it is impossible to read the state for them. I saw however that when they update the struct, they trigger an event:
https://github.com/across-protocol/contracts-v2/blob/a231a21df4ee58378d3f0fc8148daf5bc4791e3c/contracts/HubPool.sol#LL342C14-L342C36

I played around and modified the `ArrayFromOneEventHandler` to support multiple keys for values. 

Basically I have this definition:
```
  "overrides": {
    "HubPool": {
      "ignoreMethods": ["crossChainContracts", "getCurrentTime"],
      "fields": {
        "crossFields": {
          "type": "stateFromEvent",
          "event": "CrossChainContractsSet",
          "valueKeys": ["l2ChainId","adapter", "spokePool"],
        },
       }
}
```

whichj resulted in:
```
"crossFields": [
          {
            "l2ChainId": 1,
            "adapter": "0x527E872a5c3f0C7c24Fe33F2593cFB890a285084",
            "spokePool": "0x4D9079Bb4165aeb4084c526a32695dCfd2F77381"
          },
          {
            "l2ChainId": 42161,
            "adapter": "0x937958992799bF4A9a656E6596FD10d7Da5c2216",
            "spokePool": "0xA44A832B994f796452e4FaF191a041F791AD8A0A"
          },
          {
            "l2ChainId": 288,
            "adapter": "0x33B0Ec794c15D6Cc705818E70d4CaCe7bCfB5Af3",
            "spokePool": "0xcA5BcA331b8137fD15c2fa6BA34655FA58de5a0B"
          },
          {
[...]

PR #1503, opened 2023-05-22
Add StateFromEvent handler

compojoom:
The handler can return either all emitted event in an array structure (basically outlining what modifications were made), or it can return a single object that is overwritten with the latest value. In addition to this one could specify an objectKey by which to group the returned object.

fixes: #1447


packages/discovery/src/discovery/handlers/user/StateFromEvent.ts
```diff
@@ -0,0 +1,123 @@
+import { ContractValue, EthereumAddress } from '@l2beat/shared'
```
sz-piotr, line 1: This file should be named StateFromEventHandler

packages/discovery/src/discovery/handlers/user/StateFromEvent.ts
```diff
@@ -0,0 +1,123 @@
...
+      address,
+      [this.abi.getEventTopic(this.fragment)],
+      0,
+      blockNumber,
+    )
+
+    const values = new Set<ContractValue>()
+    for (const log of logs) {
+      const parsed = this.abi.parseLog(log)
+      const value = this.definition.eventKeys
```
sz-piotr, line 73: Because eventKeys will now be used to filter you can get all event keys by going through `this.fragment.inputs`

sz-piotr (changes requested):

packages/discovery/src/discovery/handlers/user/StateFromEvent.ts
```diff
@@ -0,0 +1,123 @@
...
+import { DiscoveryProvider } from '../../provider/DiscoveryProvider'
+import { Handler, HandlerResult } from '../Handler'
+import { getEventFragment } from '../utils/getEventFragment'
+import { toContractValue } from '../utils/toContractValue'
+
+export type StateFromEventDefinition = z.infer<typeof StateFromEventDefinition>
+export const StateFromEventDefinition = z.strictObject({
+  type: z.literal('stateFromEvent'),
+  event: z.string(),
+  eventKeys: z.array(z.string()),
```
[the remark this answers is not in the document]
compojoom: Ah well, to be honest I got events mixed up with function overloading. So I thought that one could have:
```
event MyEvent(uint256 indexed id);
event MyEvent(uint256 indexed id, string name);
```

But that actually doesn't work. So the whole eventKeys is I guess pointless. I want the handler to be active whenever we see the event. If we do filter keys - it would mean that we want to extract only part of the event arguments.

event: "MyEvent"
filterKeys: ['id']

This will only then return the id, but then we have also the case where no filterKeys is provided - it would then return all the arguments.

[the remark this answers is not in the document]
compojoom: No, I'll add them. I think it is nice. You could have events that return booleans or bytes or other stuff we are not interested in.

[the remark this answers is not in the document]
compojoom: I want for returnParams, as I want the event params in the return object. I didn't like `filterKeys` for this. If you don't like `returnParams` I can live with `onlyKeys`.

packages/discovery/src/discovery/handlers/user/StateFromEvent.ts
```diff
@@ -0,0 +1,123 @@
...
+    }
+
+    if (
+      // eslint-disable-next-line @typescript-eslint/no-unnecessary-condition
+      this.definition.returnAs === 'object' &&
+      typeof this.definition.groupByObjectKey !== 'undefined'
+    ) {
+      const result = [...values].reduce<Record<string, ContractValue>>(
+        (acc, curr) => {
+          const groupByKey = String(
```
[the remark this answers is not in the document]
compojoom: Ok, went for reduce from lodash. GroupBy was just grouping by the prop, but it was adding all event changes to an array, instead of overriding them.

compojoom:
Hey guys! What are we going to do with this? Shall I re-base again? Do you want me to change something else in the code?

michalsidzej:
> Hey guys! What are we going to do with this? Shall I re-base again? Do you want me to change something else in the code?

Hi. Please rebase, we'll try our best to merge this PR asap

compojoom:
Ok, done.

michalsidzej (dismissed):

compojoom:
ok, removed the ignoreRelative

packages/discovery/src/discovery/handlers/user/StateFromEventHandler.ts
```diff
@@ -0,0 +1,94 @@
...
+      )
+      values.add(params)
+    }
+
+    if (typeof this.definition.groupBy !== 'undefined') {
+      const result = reduce(
+        [...values],
+        (acc, value) => {
+          const groupByKey = String(
+            value[this.definition.groupBy as keyof typeof value],
```
michalsidzej, line 73: Let's check that on runtime. We have no guarantee that `this.definition.groupBy` is in fact `keyof typeof value`

packages/discovery/src/discovery/handlers/user/StateFromEventHandler.ts
```diff
@@ -0,0 +1,94 @@
...
+        (acc, value, key) => {
+          acc[key] = toContractValue(value)
+          return acc
+        },
+        {} as Record<string, ContractValue>,
+      )
+      values.add(params)
+    }
+
+    if (typeof this.definition.groupBy !== 'undefined') {
```
michalsidzej, line 68:
```suggestion
    if (typeof this.definition.groupBy) {
```

packages/backend/discovery/across-v2/config.jsonc
```diff
@@ -16,30 +16,11 @@
...
-        "crossChainBoba": {
-          "type": "call",
-          "method": "crossChainContracts",
-          "args": [288]
-        },
-        "crossChainArbitrum": {
-          "type": "call",
-          "method": "crossChainContracts",
-          "args": [42161]
+        "CrossChainContractsFromEmittedEvents": {
```
michalsidzej, line 19:
```suggestion
        "CrossChainContracts": {
```

michalsidzej (changes requested):
Good job overall. Few suggestions to fit our code style. Thanks a lot for the contribution!

michalsidzej (dismissed):
Good job overall. Few suggestions to fit our code style. Thanks a lot for the contribution!

sz-piotr (approved):

PR #1504, opened 2023-05-22
Display detailed upgrade information

sz-piotr:
Resolves L2B-1514

Check contracts in zksync lite in preview

### DIFF
--- a/packages/config/src/layer2s/zksynclite.ts
+++ b/packages/config/src/layer2s/zksynclite.ts
@@ -33,6 +33,13 @@ const securityCouncilMembers = discovery.getContractValue<string[]>(
 
 const securityCouncil = `${securityCouncilThreshold} of ${securityCouncilMembers.length}`
 
+const upgrades = {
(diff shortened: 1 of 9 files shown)


packages/config/src/layer2s/zksynclite.ts
```diff
@@ -167,18 +174,20 @@ export const zksynclite: Layer2 = {
...
+        ...upgrades,
+      }),
+      discovery.getMainContractDetails('Verifier', {
+        description: 'Implements zkProof verification logic.',
+        ...upgrades,
+      }),
+      discovery.getMainContractDetails('Governance', {
+        description:
+          'Keeps a list of block producers, NFT factories and whitelisted tokens.',
+        ...upgrades,
```
michalsidzej, line 189: I'm just thinking if this information should be here or in the permission addresses section... That's a lot of repeated info.

michalsidzej (approved):

PR #1511, opened 2023-05-23
Show discovery logs on server

antooni:
Resolves L2B-1520


packages/discovery/src/discovery/DiscoveryLogger.ts
```diff
@@ -1,16 +1,34 @@
...
-  static SILENT = new DiscoveryLogger({ enabled: false })
+  constructor(
+    private readonly logger: Logger,
+    private readonly options: LoggerOptions,
+  ) {}
+
+  static SILENT = new DiscoveryLogger(Logger.SILENT, { enabled: false })
+  static CLI = new DiscoveryLogger(Logger.SILENT, { enabled: true })
+  static SERVER = (logger: Logger) =>
+    new DiscoveryLogger(logger, { enabled: false })
```
sz-piotr, line 19: What about a slightly different solution? I dislike the coupling of the two loggers.
Instead I suggest two flags: enabled and buffered.
If enabled you console log.
If buffered you concat to `this.bufferedLogs`.
Both can be enabled or both can be disabled too.

Then the function `flush` just returns the buffer and clears it.
In the discovery then you use the regular logger to log the flushed messages.

sz-piotr (changes requested):

PR #1512, opened 2023-05-23
Update starknet

michalsidzej:


lucadonnoh:
there are some typos in `EXIT.STARKNET` but the description should be changed completely

lucadonnoh:
The `PROVER_DOWN` description in RiskView should match the `EXITS.STARKNET` one and that risk (validator failure) should be changed from yellow to red

lucadonnoh:
Also StarkWare guys asked us to rename every occurrence of StarkNet in the page into Starknet

packages/config/src/layer2s/common/riskView.ts
```diff
@@ -292,7 +292,7 @@ export const VALIDATOR_WHITELISTED_BLOCKS: ProjectRiskViewEntry = {
 export const PROVER_DOWN: ProjectRiskViewEntry = {
   value: 'No mechanism',
   description:
-    'There is no generic escape hatch as StarkNet cannot be frozen. Application-specific escape hatches can be built.',
+    'There is no generic escape hatch as Starknet cannot be frozen. Application-specific escape hatches can be built.',
```
lucadonnoh, line 295:
```suggestion
    'There is no generic escape hatch mechanism as Starknet cannot be forced by users into a frozen state. Note that a freezing mechanism on L2, to be secure, requires anti-censorship protection.',
```

packages/config/src/layer2s/common/riskView.ts
```diff
@@ -292,7 +292,7 @@ export const VALIDATOR_WHITELISTED_BLOCKS: ProjectRiskViewEntry = {
...
   sentiment: 'warning',
```
lucadonnoh, line 296:
```suggestion
  sentiment: 'bad',
```

lucadonnoh (changes requested):

PR #1515, opened 2023-05-24
Add pausable in contract details

lucadonnoh:
Update: I changed it to a semi-automatic approach as proposed by Piotr 

Old: I tried to add the info that a contract is pausable in the contract details, but I'm not sure this is the best approach. An issue I see is the `paused` name could be used by some contract with some other functionality (idk if checkable) -- also I tried to include the owner but for some contracts there is no name associated and the address looks ugly (e.g. EOAs)


antooni (approved):
Looks good. Although the config is getting more and more complicated, so we will need to think about refactoring and UI redesing in the future, especially for risks and contracts section

PR #1523, opened 2023-05-25
Update source code references in riskView

antooni:
Source code references started as an experimental feature and it stayed with us. Unfortunately the interface was unintuitive so in many cases it was wrongly used, which lead to skipping the tests that user was intending to have while putting the source code reference.

In the old behaviour one needed to provide `references` as well as `contracts` arrays, which were then checked, all the contracts needed to have at least on reference. Unfortunately in many cases there was no array contracts, so the test will not run.

In this change the interface changes, so now user needs to provide references to the certain contract, enforcing him to pass both of those values.

Resolves L2B-1516


packages/config/src/layer2s/apex.ts
```diff
@@ -78,11 +78,20 @@ export const apex: Layer2 = {
...
-      references: [
-        'https://etherscan.io/address/0xdD5f42B087C1D2F73a2b443249b7D3DbE148a859#code#F36#L174',
-        'https://etherscan.io/address/0x23cab3cf1aa7b929df5e9f3712aca3a6fb9494e4#code#F1#L84',
+      sourceCodeReferences: [
+        {
+          contractIdentifier: 'StarkExchange',
+          references: [
+            'https://etherscan.io/address/0xdD5f42B087C1D2F73a2b443249b7D3DbE148a859#code#F36#L174',
+          ],
+        },
```
michalsidzej, line 87: WYT of this naming? A bit more concise
```suggestion
      sources: [
        {
          contract: 'StarkExchange',
          references: [
            'https://etherscan.io/address/0xdD5f42B087C1D2F73a2b443249b7D3DbE148a859#code#F36#L174',
          ],
        },
```

packages/config/src/bridges/index.test.ts
```diff
@@ -18,6 +21,47 @@ describe('bridges', () => {
       }
     })
   })
+  describe('references', () => {
+    describe('every contract has risk view code references', () => {
```
michalsidzej, line 25:
```suggestion
    describe('point to an existing implementation', () => {
```

packages/config/src/bridges/index.test.ts
```diff
@@ -18,6 +21,47 @@ describe('bridges', () => {
...
+          for (const [riskName, riskEntry] of Object.entries(
+            bridge.riskView ?? {},
+          )) {
+            const risk = riskEntry as ProjectRiskViewEntry
+            if (risk.sourceCodeReferences === undefined) continue
+
+            it(`${bridge.id.toString()} : ${riskName}`, () => {
+              for (const sourceCodeReference of risk.sourceCodeReferences ??
+                []) {
+                const referencedAddresses = getReferencedAddresses(
```
michalsidzej, line 39: This would make debugging easier
```suggestion
            describe(`${bridge.id.toString()} : ${riskName}`, () => {
              for (const sourceCodeReference of risk.sourceCodeReferences ??
                []) {
                it(sourceCodeReference.contractIdentifier, () => {
                  const referencedAddresses = getReferencedAddresses(
```

michalsidzej (changes requested):
I'm not the biggest fan of this approach. It looks like it's bloating the config. But it seems it works, so could we just make the names a bit shorter?

michalsidzej (approved):
LGTm

PR #1527, opened 2023-05-26
Update SKALE-IMA

payvint:
Changes:
- Add Razor token to statistics and TVL
- Modify riskView from Third Party to Destination Chain(explanation below)

Explanation:
Each message delivered to SKALE chain A is signed by 11/16 validators nodes which are operating SKALE chain A. And it would be verified in smart contracts by BLS common public key of this SKALE chain A. And these keys are using to sign and verify blocks on SKALE chain A. If messages are sent to SKALE chain B - it can not be sent to SKALE chain A or can not be sign by SKALE chain A keys and delivered to SKALE chain B


packages/config/src/bridges/skaleIMA.ts
```diff
@@ -45,7 +45,7 @@ export const skaleIMA: Bridge = {
         address: EthereumAddress('0x8fB1A35bB6fB9c47Fb5065BE5062cB8dC1687669'),
```
antooni, line 45: This token does not fit coingecko TOP300 criteria

antooni (dismissed):

packages/config/src/bridges/skaleIMA.ts
payvint: Hmm that's true, but I did not find any limits in code that only 300 tokens are operating.
And also the ranking is changing everyday(today this token is not in top300 but tomorrow it can be)

I have also checked a TokenList.json file, and I found that you already have some tokens which are not in top300 (tenset, agEUR, ANY(which is rebranded to MULTI), AURORA ...)

Maybe it would be possible to add this token to the list, because it is a big part of SKALE IMA bridge, and without this token comparison is not fair?

luke7211: Tokens occasionally fall out of Top300 and they are not taken out. The intent is to make sure we don't list inflated tokens like SHIBA etc... with outrageous valuation. If RAZOR is legitimate token from legit team, then exceptions can be made. I have looked briefly at the token and the project and it looks ok to me

payvint: Thank you very much, for understanding!
Razor Network is a legitimate project - https://razor.network
They are doing on oracle network on SKALE(and in v2 which is coming in this quarter it is possible to expand the Oracle to other skale chains and other network)

PR #1535, opened 2023-05-31
Update dYdX config

lucadonnoh:


packages/config/src/layer2s/dydx.ts
```diff
@@ -225,6 +225,28 @@ export const dydx: Layer2 = {
       description:
         'Allowed to update state of the rollup. When Operator is down the state cannot be updated.',
     },
+    {
+      name: 'Short Timelock Executor',
+      accounts: [discovery.getPermissionedAccount('Treasury', 'owner')],
```
michalsidzej, line 230: What if this one changes? The owner field will be pointing to a different contract

michalsidzej: Also, I don't think this is a permissioned account. This is a contract, no?

adamiak (approved):

lucadonnoh (commented):
upgradability info in contracts is still missing, right?

lucadonnoh (commented):
apart from the comments it lgtm

PR #1548, opened 2023-06-06
Update zkSync

michalsidzej:
Resolves L2B-1529

Skipping display of Multisig 1, 2 & 3. Those are members of the security council. I don't think that there's value in adding them to permissioned addresses, IMO it will just bloat the page.

Skipping approvedUpgradeNoticePeriod and upgradePreparationActivationTime. This sounds like adding a new feature (upgrade is coming or something) 

Please consider those two points while reviewing.


lucadonnoh (approved):
lgtm but maybe wait for piotr since he knows better

sz-piotr (approved):
LGTM

sz-piotr (approved):
LGTM

PR #1549, opened 2023-06-06
Change Bedrock upgrade warning text to undergoing

lucadonnoh:
DO NOT MERGE: meant to be merged at 16:00 UTC (18:00 CET)


sz-piotr (approved):
LGTM, but maybe we can make it so if date > something we automatically mark it as under review?

lucadonnoh:
> LGTM, but maybe we can make it so if date > something we automatically mark it as under review?

i thought about it but i didn't know the best way to do it, will do next time

PR #1553, opened 2023-06-07
Update Proposer failure column

lucadonnoh:
- change Validator failure to Proposer failure
- change the Proposer assessments to the new ones


luke7211 (approved):

packages/config/src/layer2s/aztec.ts
```diff
@@ -131,8 +131,8 @@ export const aztec: Layer2 = {
         },
       ],
     },
-    validatorFailure: {
-      ...RISK_VIEW.VALIDATOR_PROPOSE_BLOCKS_ZKP,
+    proposerFailure: {
+      ...RISK_VIEW.SELF_PROPOSE_ZK,
```
michalsidzej, line 135: Nit pick: I like how previously the name for the `validatorFailure` always started with `VALIDATOR`. It kinda gives it structure. With so many risks I think it makes sense to keep that, WYT?

PR #1565, opened 2023-06-13
Split shared to shared and shared-pure

michalsidzej:
Resolves L2B-1435


packages/shared-pure/README.md
```diff
@@ -0,0 +1 @@
+# @l2beat/shared-pure
```
antooni, line 1:
```suggestion
# @l2beat/shared-pure

### Warning
This package will be imported inside `frontend`, put only pure JS dependencies there.
```

antooni (approved):
Looks good, gj with linter. The only thing that I am not sure about is putting RateLimiter and Retires inside shared-pure, the intent is that this could be used for the calls on frontend?

PR #1566, opened 2023-06-13
Add typecheck to CI

michalsidzej:
We had some typecheck problems that were not caught by the CI. This PR adds typecheck as part of CI. Should not take much longer than current workflow, but will add a bit more safety

The problem:
- we had a script `typecheck:ci` that was responsible for running tsc on CI
- some packages missed this script and were not checked

Solution:
change CI to run `typecheck`


antooni (commented):
so what exactly was the problem there? some packages were missing `typecheck:ci` and were not checked, I do not quite get this change

krzkaczor (approved):

PR #1568, opened 2023-06-13
Persist Chart's settings in localStorage

### LINKED ISSUE #260: Make the "Log" vs. "Linear" selection for the graph axis persistent
A month or two ago, L2beat added a selection button to enable a _Logarithmic_ scale on the y-axis.  This is super useful.  Thanks.

But as it stands, the web app resets the axis view to _Linear_ every time the page is opened, or refreshed.

This forces the user to have to manually reselect the "Log" view each time if they prefer logarithmic scales.  

(_personal opinion_:  log scales are vastly better than linear scales for properly analyzing exponential growth, which a very great deal of blockchain adoption data is.  E.g., growth in overall L2 TVL in the recent week/month; or the growth of Arbitrum L2 in any reasonable time period now.)
But hey, let's make it user-selectable, _and persistent_.

Could you consider **adding a way to store the user choice used on the L2beat.com website to be persistent?**  A non-tracking cookie?  Or something else?

WojciechKo:
Resolves #260


antooni:
Hey @WojciechKo,
thanks for the contribiution <3

Currently our pattern to persist the state is via local storage, see `/frontend/src/components/navbar/configureDarkThemeToggle.ts`

Can you implement your PR using localStorage?

antooni (changes requested):
implement using localStorage

WojciechKo:
> Can you implement your PR using localStorage?
Is it supposed to use localStorage only or in combination with query string as well to enable more specific link sharing?

antooni:
For now let's just implement only localStorage

WojciechKo:
What do you think about this approach? It stores `isLogScale` value separate for each chart.
Now it's quite trivial to persist `days`, `currency` and `showEthereum` alongside.

packages/frontend/src/components/chart/configure/persistedSettings.ts
```diff
@@ -0,0 +1,35 @@
...
+
+interface PersistedSettings {
+  isLogScale: boolean
+}
+
+const DEFAULT_SETTINGS = {
+  isLogScale: false,
+}
+
+export function persistSettings(state: State) {
```
antooni, line 11: there is no need to pass the whole State object

packages/frontend/src/components/chart/configure/persistedSettings.ts
```diff
@@ -0,0 +1,35 @@
...
+    chartStorageKey(state.controls.chartId),
+    JSON.stringify(settings),
+  )
+}
+
+export function getPersistedSettings(chartId: string) {
+  return getPersitedSettings(chartId) ?? DEFAULT_SETTINGS
+}
+
+function getPersitedSettings(chartId: string) {
```
antooni, line 25: there is the typo in the name

packages/frontend/src/components/chart/configure/persistedSettings.ts
```diff
@@ -0,0 +1,35 @@
...
+  const settings: PersistedSettings = {
+    isLogScale: state.controls.isLogScale,
+  }
+  localStorage.setItem(
+    chartStorageKey(state.controls.chartId),
+    JSON.stringify(settings),
+  )
+}
+
+export function getPersistedSettings(chartId: string) {
```
antooni, line 21: maybe let's just make it one function

antooni (changes requested):
@WojciechKo kudos,
I like the direction we took, I propose some changes:
- get rid of chartId (it is a good idea, but not for the scope of this PR)
- rename PersistedSettings to UserChartSettings
- save also time range and currency to the localStorage

antooni:
Also I have merged your branch with current master

packages/frontend/src/components/chart/configure/persistedSettings.ts
WojciechKo: Not now, but I'll implement other settings like days and currency soon. Do you prefer to pass them explicitly rather than the whole State?

WojciechKo:
If we get rid of `chartId` the settings will be shared across every chart. It seems to be within the scope of this PR if we want to have a separated chart's settings.
Can you suggest other approach that will allow to achieve this?

antooni:
The main idea was to preserve "isLog", so we will extend this one and preserve time range and currency. So next time user visits/refreshes will see the same time period, currency and logarithmic scale.
Dividing settings per chart would require rethinking in terms of the UX, you can create issue for that.

WojciechKo:
In terms of UX, shared configuration might be confusing for users
1. User enters scaling tvl page, sees chart in lin scale
2. User enters bridge tvl page, sees chart in lin scale and selects log
3. User enters scaling tvl page, sees chart in log scale

The same applies for time range and currency.

Just want to confirm if we want to move forward with described approach, despite the fact the support for separated settings is already implemented.

antooni:
okay, we misunderstood each other, I thought you wanted to persist the scale when switching between the time periods.

let's keep chartId then, and add more values to the state

antooni:
@WojciechKo if this will simplify the code and not break any current functionality I think you can try implementing that

antooni:
@WojciechKo it looks good, kudos

I have only left some small suggestions to improve the code,
after resolving them we can merge it!

Overall really great work and cooperation, 
if you are interested in more contributions please contact me on discord `.antooni`

packages/frontend/src/components/chart/configure/userChartSettings.ts
```diff
@@ -0,0 +1,61 @@
...
+function userChartSettingsKey(pagePathname: string) {
+  return `${pagePathname}chartSettings`
+}
+
+const DEFAULT_CHART_SETTINGS = {
+  isLogScale: false,
+  days: 365,
+  currency: 'usd',
+  showEthereum: true,
+} as const
```
antooni, line 61: I do not like this `as const` here, please try to type it more explicitly

packages/frontend/src/components/chart/configure/userChartSettings.ts
```diff
@@ -0,0 +1,61 @@
...
+    showEthereum: state.controls.showEthereum,
+  }
+
+  localStorage.setItem(
+    userChartSettingsKey(state.controls.pagePathname),
+    JSON.stringify(chartSettings),
+  )
+}
+
+export function getUserChartSettings(pagePathname: string) {
```
antooni, line 32:
```suggestion
export function getUserChartSettings(pagePathname: string): UserChartSettings {
```

antooni (approved):

PR #1570, opened 2023-06-14
Add ZKEX as an upcoming L2 project and zkLink as technical provider

zkTyral:
ZKEX.com is an omni-chain decentralized cryptocurrency and derivatives trading platform secured with zero-knowledge proofs.


lucadonnoh:
The solution doesn't classify as an L2 since DA is not on Ethereum. My understanding is that while it might not be possible to steal funds, the oracle network can double spend by allowing two different roots, meaning it's not a Validium as well. Nevertheless, the risk profile is very interesting and unique and it can be listed as a bridge when it will be live on mainnet. If I'm missing something, please let me know!

PR #1571, opened 2023-06-14
Add Kroma to upcoming rollups

lucadonnoh:
Added changes from #1569


lucadonnoh:
currently it's a fork of Optimism Bedrock with no fraud proofs, then they want to implement the a zk fraud proof system where the bisection games decides on a disagreement block, which is then proven using Scroll zkEVM (somehow). Since this is only done in case of a dispute, it still qualifies as a optimistic rollup. 

Later they want to completely move away from fraud proofs and become a zk rollup.

antooni (approved):

PR #1573, opened 2023-06-14
Update OP token balance

michalsidzej:
Resolves L2B-1598


packages/backend/src/core/reports/custom/report.ts
```diff
@@ -15,12 +15,15 @@ function createCustomTokenReport({
...
   tokenPriceUsd: number
   tokenDecimals: number
-  tokenBalance: bigint
+  tokenBalance: bigint | ((t: UnixTime) => bigint)
   tokenId: AssetId
   projectId: ProjectId
   ethPriceUsd: number
   timestamp: UnixTime
 }): ReportRecord {
+  tokenBalance =
```
antooni, line 24: I am not sure about this line, have you tested it? Maybe tokenBalance could always be a function?

antooni (changes requested):
Please resolve conversation about tokenBalance

antooni (approved):

PR #1574, opened 2023-06-15
Implement Stages UI

michalsidzej:
Resolves L2B-1581


packages/config/src/layer2s/common/stages/types.ts
```diff
@@ -20,21 +20,30 @@ export type ChecklistTemplate<T extends StageBlueprint> = {
   }
 }
 
+// TODO: maybe it shouldn't be undefined
+export type Stage = 'Stage 0' | 'Stage 1' | 'Stage 2' | undefined
```
antooni, line 24: 'Candidate' ?

packages/frontend/src/components/stages/StageDisclaimer.tsx
```diff
@@ -0,0 +1,27 @@
+import cx from 'classnames'
+import React from 'react'
+
+import { InfoIcon } from '../icons'
+
+export function StageDisclaimer({
```
antooni, line 6: use `Callout` component for this

packages/frontend/src/components/stages/StageTooltip.tsx
```diff
@@ -1,34 +1,83 @@
...
-            {item.missing.requirements.map((requirement, i) => (
-              <li className="w-60" key={i}>
-                ❌ {requirement}
-              </li>
-            ))}
-          </ul>
-        </div>
+      {item.stage === 'UnderReview' ? (
+        <>
+          Projects under review might present uncompleted information & data.
```
antooni, line 25: this text seems a bit too long, but it is a nitpick

antooni (changes requested):
kudos, gj

please resolve comments before merging

packages/config/src/layer2s/common/stages/types.ts
michalsidzej: Yeah, I was thinking about that, but we do not have any designs ready for this case. We could ignore it for now IMO

packages/frontend/src/components/stages/StageTooltip.tsx
michalsidzej: What do you propose then?

antooni: Let's leave it for now, I do not have an idea

antooni (approved):

PR #1577, opened 2023-06-16
Finalize stage assessment

lucadonnoh:


packages/config/src/layer2s/aztec.ts
```diff
@@ -152,17 +152,17 @@ export const aztec: Layer2 = {
       callsItselfRollup: true,
       stateRootsPostedToL1: true,
       dataAvailabilityOnL1: true,
-      rollupNodeOpenSource: true,
+      rollupNodeOpenSource: ['UnderReview', 'The rollup node is open source.'],
```
michalsidzej, line 155: Are we sure we want to display it this way? (This is just the `positive` wording.) If so, slight change in the logic could make it easier, we could just say `UnderReview` here, and display the positive sentece. This would take away the control though, so not really sure if it's necessary.

michalsidzej: If we leave it like this, please fix ZkSpace.ts

lucadonnoh: zkspace is already fixed, right? but yeah i'm not sure tbh, i don't want to convey that we actively looked and didn't find one, we just need more research and maybe contact the teams. wdyt? the positive description was first suggested by Piotr

michalsidzej: I just mean that we can easily do it as part of a logic rather than manually.

lucadonnoh: i agree, can you take care of it?

PR #1582, opened 2023-06-19
Link to section id from the table view

michalsidzej:
Resolves L2B-1626


krzkaczor:
What about linking header section?

image

michalsidzej:
> What about linking header section?
> 
> image

This has been done in https://github.com/l2beat/l2beat/commit/623e1a8215e318199fd93dceda468c062e82f7d8

PR #1592, opened 2023-06-21
TaskQueue error handling strategy improvements

krzkaczor:
Resolves L2B-398.

Current state: if an error happens during task execution, we are retrying it ad infinitum, emitting `log.warnings` at the same time. This means that in case of a persisting error (revoked API keys or simply programmer mistake in the new piece of logic) not only we will never stop retrying it but also we will never propagate it to sentry (it happens only when the retry strategy gives up and `log.error` is emitted). This means that even right now one of the queues can be failing indefinitely and we even don't know about it. Emitting errors on every retry is not advised as random failures do happen, and we don't want to be flooded with notifications. 

Proposed solution: change the default behavior of the task queue to a confined number of retries after which an error is propagated and finally, the whole queue is halted. This means that it will require manual intervention from the dev team (fixing the error and restarting the service), but at least errors are now **visible**.

### Alternative:
An alternative solution would be detailed monitoring of queue failure rates via metrics and alerting the dev team based on that.


antooni:
I think proposed solution is the best for now, it is also similiar to what we do with sanity checks (we tried monitoring but no one is checking so right now we make the build fail and everyone notices ;p).

In the future we would need a solution with the monitoring, but I do not think we are ready for that right now.

packages/backend/src/core/queue/TaskQueue.test.ts
```diff
@@ -62,6 +62,74 @@ describe(TaskQueue.name, () => {
     expect(completed).toEqual([0, 1, 2, 3, 4, 5, 6, 7, 8, 9])
   })
 
+  it('can handle permanent failure', async () => {
```
antooni, line 65:
```suggestion
  it('can handle task failure', async () => {
```

packages/backend/src/core/queue/TaskQueue.test.ts
```diff
@@ -62,6 +62,74 @@ describe(TaskQueue.name, () => {
...
+    }
+
+    await time.runAllAsync()
+    await queue.waitTillEmpty()
+
+    expect(completed).toEqual([0, 2]) // 1 was dropped
+    expect(eventTracker.record).toHaveBeenCalledWith('error')
+  })
+
+  it('can halt on permanent failure', async () => {
```
antooni, line 99:
```suggestion
  it('can halt on repeating task failure', async () => {
```

antooni (approved):
LGTM,
the only thing I am not sure about is the wording in the test names, I have proposed changes

PR #1597, opened 2023-06-21
Make navbar sticky to the top

WojciechKo:
I have no idea why class `bg-inherit` is not working.
I believe it's because it's never been used before. Still, building the project again with `yarn build:frontend` doesn't help here. Any idea?


antooni:
@WojciechKo is this `bg-inherit` crucial for the functionality? maybe let's skip it for now and try to make navbar sticky first, and style later

antooni (changes requested):
Looks good, I only need to discuss with the designer for final approval

PR #1600, opened 2023-06-22
Introduce chainId to block_numbers table

antooni:
Resolves L2B-1682

- [x] download staging DB and check indexes usage on queries


packages/backend/src/peripherals/database/BlockNumberRepository.test.ts
```diff
@@ -1,10 +1,13 @@
 import { Logger } from '@l2beat/shared'
-import { UnixTime } from '@l2beat/shared-pure'
+import { ChainId, UnixTime } from '@l2beat/shared-pure'
 import { expect } from 'earl'
 
 import { setupDatabaseTestSuite } from '../../test/database'
 import { BlockNumberRepository } from './BlockNumberRepository'
 
+const ETH = ChainId.ETHEREUM
+const ARB = ChainId.ARBITRUM
```
sz-piotr, line 9: Why not use those constants directly like everywhere else?

packages/backend/src/peripherals/database/migrations/058_block_numbers_chain_id.ts
```diff
@@ -0,0 +1,40 @@
...
+
+*/
+
+import { ChainId } from '@l2beat/shared-pure'
+import { Knex } from 'knex'
+
+export async function up(knex: Knex) {
+  await knex.schema.alterTable('block_numbers', function (table) {
+    table
+      .string('chain_id')
```
sz-piotr, line 20: why string? Can't we have integers like everyone? 1 is ethereum, 42 optimism i think and so on.

packages/shared-pure/src/types/ChainId.ts
```diff
@@ -0,0 +1,18 @@
...
+export function ChainId(value: string): ChainId {
+  if (!CHAIN_IDS.includes(value)) {
+    throw new TypeError('Invalid ChainId')
+  }
+  return value as unknown as ChainId
+}
+
+const CHAIN_IDS = [
+  '0x1', // Ethereum Mainnet
+  '0xa4b1', // Arbitrum One
```
sz-piotr, line 14: We should store those as integers

sz-piotr (changes requested):

sz-piotr (approved):

PR #1613, opened 2023-06-28
Add beamer bridge discovery

compojoom:
Hey guys,

I gave adding a discovery config for beamer bridge a shot. Since currently l2beat can't index contracts on other chains this only index the beamer bridge contracts on ethereum.

I've added all other contract addresses with the networks they are deployed to, but they all have "ignoreDiscovery": true

~~I've also tried to create a beamer bridge config for the frontend, but not sure if I've done a good job here. Is there a trick that would allow me to view the generated page?~~

Ok, the trick is:
1. modify packages/frontend/src/build/buildPages.ts after the tvlResponse attach your new project with assigned tvlApiResponse from a different project.
2. yarn test - make sure that everything is green
3. yarn build
3. start the frontend with yarn start:staging


compojoom:
I ran the verify contracts config, but it treats the optimism, arbitrum contracts as non-verified.
For example this one: `0x13bAF73f48FCF6A8aAb8431CA3A38c624cdfd8F3` is not a contract on mainnet, but on arbitrum one:
https://arbiscan.io/address/0x13baf73f48fcf6a8aab8431ca3a38c624cdfd8f3

What am I supposed to do with such contracts? Set them to manually verified?

compojoom:
Also the volume is calculated only based on the RequestManager contract, but that is not the true volume. 
The LiquidityProvider addresses actually hold the biggest part of the Volume, but it seems that we can't read their balances?

compojoom:
I don't understand why the build is failing now, but it certainly works fine locally?

lucadonnoh:
hey @compojoom i made some edits, please take a look :) we don't list L2 contracts so just Ethereum contracts are fine. About TVL, we usually just count funds in escrow contracts but i'll discuss this with the team

also, could you please provide a description on the Messenger contracts? thanks!

lucadonnoh (approved):

compojoom:
@lucadonnoh thanks for the updates. They look good. I added info on the Messenger contracts.

PR #1614, opened 2023-06-28
Print errors during discovery 

michalsidzej:
Resolves L2B-1691

### DIFF
--- a/packages/discovery/src/discovery/engine/DiscoveryEngine.ts
+++ b/packages/discovery/src/discovery/engine/DiscoveryEngine.ts
@@ -40,6 +40,21 @@ export class DiscoveryEngine {
     }
 
     this.logger.flush(config.name)
+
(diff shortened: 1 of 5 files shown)


packages/discovery/src/discovery/engine/DiscoveryEngine.ts
```diff
@@ -40,6 +40,21 @@ export class DiscoveryEngine {
...
 
     this.logger.flush(config.name)
+
+    this.checkErrors(resolved)
+
     return resolved
   }
+
+  private checkErrors(resolved: Analysis[]) {
+    let errors = 0
```
antooni, line 50: what about storing array of strings, where each string is <contract>.<value>

antooni (approved):
Looks good, If you have time you can add this more detailed error message at the end, but it is not required for now IMO

PR #1572, opened 2023-06-14
Add Symbiosis to bridges section

allush:

PR #1617, opened 2023-06-30
Add Symbiosis bridge

lucadonnoh:
prev #1572 

https://l2beat-staging-git-symbiosis-finance-master-l2beat.vercel.app/bridges/projects/symbiosis

How to add a preview?
https://www.notion.so/l2beat/New-project-preview-9a20e970b89b4f7da008349ab12ca00a


packages/config/src/bridges/symbiosis.ts
```diff
@@ -0,0 +1,167 @@
...
+
+export const symbiosis: Bridge = {
+  type: 'bridge',
+  id: ProjectId('symbiosis'),
+  display: {
+    name: 'Symbiosis',
+    slug: 'symbiosis',
+    category: 'Hybrid',
+    description:
+      'Symbiosis is a cross-chain AMM DEX that pools together liquidity from different networks, whether they use EVM technology or not.',
```
antooni, line 18: rewrite the desciption, make it sound more professional, also include some info about how it is validated

packages/config/src/bridges/symbiosis.ts
```diff
@@ -0,0 +1,167 @@
...
+    links: {
```
antooni, line 19: maybe also link to the app?

antooni (approved):

packages/config/src/bridges/symbiosis.ts
fastchain: Hi! 
Thank for the feedback.


`make it sound more professional`
Could you please provide a link to any example that sounds really professional?


`include some info about how it is validated`
I'm afraid I don't get it, what you mean by validated?   It works this way right now, so everyone can go to the app and validate it by his own hands.

lucadonnoh: i'll take care of it :)

packages/config/src/bridges/symbiosis.ts
IngeFX: If you think it will be convenient to add it here, then: https://app.symbiosis.finance/

PR #1621, opened 2023-06-30
Enable incremental build

antooni:
Resolves L2B-1698

Open question: 

Is it possible to write such a script that will:
1. run from e.g. frontend
2. go to root directory and build all the deps (not with the script where we specify the list of them, but somehow yarn will figure it out)
3. go back to frontend and run

right now we have this `build:frontend` but it builds the dependencies and the frontend itself so we will build the frontend twice when runing it

### DIFF
--- a/packages/backend/package.json
+++ b/packages/backend/package.json
@@ -6,6 +6,7 @@
   "types": "./index.d.ts",
   "scripts": {
     "build": "tsc -p tsconfig.build.json",
+    "build:dependencies": "cd ../.. && yarn build:dependencies && cd -",
     "clean": "rm -rf build",
     "format:fix": "prettier --write .",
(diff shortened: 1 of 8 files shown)


michalsidzej (approved):
LGTM

packages/backend/package.json
```diff
@@ -6,7 +6,7 @@
   "types": "./index.d.ts",
   "scripts": {
     "build": "tsc -p tsconfig.build.json",
-    "prebuild": "yarn clean",
+    "build:dependencies": "cd ../.. && yarn build:dependencies && cd -",
```
sz-piotr, line 9: There is no need to go back in this script. I think the working directory stays the same outside.

sz-piotr (approved):

PR #1622, opened 2023-06-30
Create new risk sentiments, 'good' and 'neutral'

michalsidzej:
Right now, the undefined sentiment is ambiguous, it can mean 'good' or 'neutral' or 'i forgot to add sentiment actually'. This PR makes it more explicit, making it easier to work with type system changes (yes, 'In review' is coming, this is part of the untangling).

### DIFF
--- a/packages/frontend/src/utils/risks/color.ts
+++ b/packages/frontend/src/utils/risks/color.ts
@@ -1,23 +1,25 @@
-import { Sentiment } from './types'
+import { Sentiment } from '@l2beat/config'
 
-export function sentimentToFillColor(sentiment?: Sentiment): string {
+export function sentimentToFillColor(sentiment: Sentiment): string {
   switch (sentiment) {
     case 'bad':
       return 'fill-orange-600'
     case 'warning':
       return 'fill-yellow-200'
-    case undefined:
+    case 'good':
+    case 'neutral':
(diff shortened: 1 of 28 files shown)


packages/frontend/src/utils/risks/color.ts
```diff
@@ -1,23 +1,25 @@
...
 
-export function sentimentToFillColor(sentiment?: Sentiment): string {
+export function sentimentToFillColor(sentiment: Sentiment): string {
   switch (sentiment) {
     case 'bad':
       return 'fill-orange-600'
     case 'warning':
       return 'fill-yellow-200'
-    case undefined:
+    case 'good':
```
lucadonnoh, line 9: so 'good' will be green and 'neutral' will be gray, right? what's the difference between 'neutral' and 'upcoming'?

michalsidzej: Yes, that's it. Upcoming risks are neutral because they do not exist yet. Theres also `Destination` "risk" in bridges that should be consider grey if any (imo should not be part of risks, but this is part of refining bridges page that we'll have to do some time in the future). There will also be 'In review' risk which will be different than 'neutral'. Screenshot 2023-07-03 at 09 08 14

antooni (approved):

PR #1623, opened 2023-06-30
Introduce `chainId` to `asset_balances` table

antooni:
Resolves L2B-1724

### DIFF
--- a/packages/backend/src/core/balances/BalanceUpdater.test.ts
+++ b/packages/backend/src/core/balances/BalanceUpdater.test.ts
@@ -3,6 +3,7 @@ import { Logger } from '@l2beat/shared'
 import {
   AssetId,
   Bytes,
+  ChainId,
   CoingeckoId,
   EthereumAddress,
   ProjectId,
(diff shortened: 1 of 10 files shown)


packages/backend/src/core/balances/BalanceUpdater.test.ts
```diff
@@ -298,3 +307,17 @@ describe(BalanceUpdater.name, () => {
     }
   }
 })
+function mockBalance(
+  assetId: AssetId,
+  timestamp: UnixTime,
+  holderAddress: EthereumAddress,
+  chainId: ChainId,
+): BalanceRecord {
```
michalsidzej, line 315: Nit pick: mockBalance always get the same timestamp, so it can be ommited as an argument here

packages/backend/src/core/balances/BalanceUpdater.ts
```diff
@@ -39,6 +40,7 @@ export class BalanceUpdater {
     private readonly clock: Clock,
     private readonly projects: BalanceProject[],
     private readonly logger: Logger,
+    private readonly chainId: ChainId,
```
michalsidzej, line 43: Hmm, I'm thinking if actually checking the chainId against other services (multicall in particular here) wouldn't be a good idea here. Or maybe even the multicall can hold this information? This would make sure thet we're downloading the data from the correct chain.

Issue #1307, opened 2023-03-20
UX Suggestion: Update Risk Analysis chart colors


dewanshparashar:
Great work with the Risk analysis component on L2Beat. It's very insightful!

While the colors yellow/red are good for warning, as a small enhancement, it'd be amazing to see the good/ok parts for all L2s green instead of grey. That would give a much better sense of approval to the audience as green is widely associated to be a happy indicator.

<img width="390" alt="image" src="https://user-images.githubusercontent.com/7558499/226458062-80b2cdbb-ef4a-4935-8cfd-4cdc3de5a31f.png">

lucadonnoh:
wip #1631

PR #1627, opened 2023-07-04
Implement Under Review project state and update rosette colors

michalsidzej:
Rosette colors was included here to emphasize the difference between 'good' and 'under review' state on rosette 

Resolves L2B-1467, L2B-1311
Closes https://github.com/l2beat/l2beat/pull/1631, Fixes https://github.com/l2beat/l2beat/issues/1307

https://l2beat-staging-git-under-review-sections-l2beat.vercel.app/scaling/tvl


antooni:
IMO if there is at least one field under review we should mark it as under review

other solution is to have severity of the field, so e.g. risks have to be all filled, but lets say technology can have under review and the project will not be marked under review

I am just worried that if we decide on anything other than strict rule that everything should be reviewed, then we will end up in the state similiar to the bridges, where we forget what is needed for the review, 

idea:

maybe our UpdateMonitor bot can be pulled into this workflow and remind us about under-review projects?

packages/config/src/layer2s/index.test.ts
```diff
@@ -129,7 +129,7 @@ describe('layer2s', () => {
         describe(layer2.display.name, () => {
           type Key = Exclude<
             keyof Layer2Technology,
-            'category' | 'provider' //TODO: Add test for permissions
+            'category' | 'provider' | 'isUnderReview' //TODO: Add test for permissions
```
antooni, line 132: should this permissions test be part of this PR?

packages/config/src/common/ProjectRisk.ts
```diff
@@ -16,7 +16,7 @@ export type ProjectRiskCategory =
   | 'Users can be censored if'
   | 'MEV can be extracted if'
 
-export type Sentiment = 'bad' | 'warning' | 'good' | 'neutral'
+export type Sentiment = 'bad' | 'warning' | 'good' | 'neutral' | 'UnderReview'
```
antooni, line 19: I think that I came up with this name `UnderReview` but the more I see it I think it should be `under-review`. I will create a task to rename it in the future.

michalsidzej: 100%

antooni (changes requested):
There is no way to set permissions as under review

packages/frontend/src/utils/project/isProjectUnderReview.ts
```diff
@@ -0,0 +1,20 @@
+import { Bridge, BridgeRiskView, Layer2, Layer2RiskView } from '@l2beat/config'
+
+export function isProjectUnderReview(project: Layer2 | Bridge): boolean {
```
antooni, line 3: This function needs a comment that in case of adding a new section there is a need to also include it in this function

antooni:
shouldn't whole L2finance row be marked as under review?
image

antooni:
shouldn't here be a banner under review?
image

antooni:
image
why there is a question mark on the rosette in L2finance?

lucadonnoh:
>  why there is a question mark on the rosette in L2finance?

i think it might be ok to have a question mark there but then it's in contrast with the fact that we show the risk analysis

antooni:
> > why there is a question mark on the rosette in L2finance?
> 
> i think it might be ok to have a question mark there but then it's in contrast with the fact that we show the risk analysis

Why we should display question mark there if we have risks info? I dont understand

lucadonnoh:
my understanding is that we are not sure that they're correct

antooni:
I see, so even having all of them in the config, but not finish all the research would mean that the risks are under review? Then the risk section also should be marked as in the review

lucadonnoh:
yeah i think the risk analysis cannot be not complete until the review is finished

antooni (approved):

lucadonnoh (approved):

antooni (changes requested):
image
there is no under review icon in the table and the small rosette does not have question mark

antooni (approved):

antooni (approved):
Approve, the only thing I am not sure is that the rosette is now green and the same color is used for Technology column when a project is a rollup. Maybe we should think about different visual distinction in the future. Because it looks a bit misleading imo, espacially in the Validium case

PR #1628, opened 2023-07-04
Improve reg-viz stability

mateuszradomski:
Resolves L2B-1446

Building on #1443, we now tackle two observed problems:

- ExpandableContainer being flaky, this was fixed adding more text to the container. This takes the test from being on an edge (4 paragraphs ~ 3rem each and button about 4rem and the max height of the container is by default 20rem) to being safely further away from the triggering point.
- All Tooltip tests, there is not real fix for it if we want to keep acting on events. As described in https://github.com/reg-viz/storycap/issues/580 and https://github.com/reg-viz/storycap/issues/586, `storycap` has problems with events such as hover/focus/click. Some improvements were introduced to `storycap` by the flakiness stayed. We fixed this problem by calling the function that would normally be called by an event ourselves.


packages/frontend/src/scripts/configureTooltips.ts
```diff
@@ -102,6 +102,65 @@ export function configureTooltips() {
   }
 }
 
+export function testConfigureTooltipsAndShow() {
```
mateuszradomski, line 105: This is copy-pasta from above but extracting the `show` function would mean that there is some state executed on import. What do you think is better, copy-pasta or state-execution on import? If you have some solution that does not have any of these problems I'm open for suggestions :smiley:

antooni (approved):

PR #1629, opened 2023-07-04
Introduce `BlockNumberProvider` abstraction

mateuszradomski:
Resolves L2B-1645

With this PR we also simplify EtherscanClient class by removing duplication from `shared`/`backend` into a single class in `shared`. The implementation of calling the Etherscan api has been refactored into a class `EtherscanLikeClient` that allows to share the calling code between `EtherscanClient` and `ArbiscanClient` as they have very similar APIs allowing for this abstraction.


packages/backend/src/api/routers/StatusRouter.ts
```diff
@@ -1,8 +1,8 @@
 import Router from '@koa/router'
+import { stringAs } from '@l2beat/shared'
```
antooni, line 2: why not `shared-pure`?

packages/backend/src/config/Config.ts
```diff
@@ -55,6 +55,7 @@ export interface ClockConfig {
 export interface TvlConfig {
   readonly alchemyApiKey: string
   readonly etherscanApiKey: string
+  readonly arbiscanApiKey: string
```
antooni, line 58: This is obviously good for now, but do you have any ideas how it could be handled in the future? I suppose we will have many configurations. This is just an open question and is not a part of this PR

packages/backend/src/config/config.production.ts
```diff
@@ -58,6 +58,7 @@ export function getProductionConfig(): Config {
     tvl: {
       alchemyApiKey: getEnv('ALCHEMY_API_KEY'),
       etherscanApiKey: getEnv('ETHERSCAN_API_KEY'),
+      arbiscanApiKey: getEnv('ARBISCAN_API_KEY'),
```
antooni, line 61: remember you will have to set it on Heroku in advance, otherewise the main Application process will fail

packages/backend/src/config/config.local.ts
```diff
@@ -52,6 +52,7 @@ export function getLocalConfig(): Config {
     tvl: tvlEnabled && {
       alchemyApiKey: getEnv('ALCHEMY_API_KEY'),
       etherscanApiKey: getEnv('ETHERSCAN_API_KEY'),
+      arbiscanApiKey: getEnv('ETHERSCAN_API_KEY'),
```
antooni, line 55:
```suggestion
      arbiscanApiKey: getEnv('ARBISCAN_API_KEY'),
```

packages/shared/src/services/etherscan/EtherscanClient.ts
```diff
@@ -1,83 +1,56 @@
...
 } from './model'
 
-export class EtherscanClient {
-  private readonly rateLimiter = new RateLimiter({
-    callsPerMinute: 120,
-  })
-  private readonly timeoutMs = 10_000
+export class EtherscanError extends Error {}
+
+export class EtherscanClient
```
antooni, line 20: There are two EtherscanClients, this one is used only in tests, what is the purpose of this?

packages/shared/src/services/etherscan/EtherscanClient.ts
```diff
@@ -1,83 +1,56 @@
...
+  extends EtherscanLikeClient
+  implements BlockNumberProvider
+{
+  static API_URL = 'https://api.etherscan.io/api'
```
antooni, line 24: why is it decalred here as a static field?

antooni (changes requested):
First round of review ready, I looked more high level on this one. Please resolve the issues. During next review I will look more closely on the Clients code.

Two open questions:
- how to make config ready for many chains
- how to make folder structure ready to many chains

packages/shared/src/services/etherscan/EtherscanClient.ts
```diff
@@ -1,83 +1,56 @@
...
-    if (response.status !== '1') {
-      throw new Error(
-        `Error response ${response.message} ${JSON.stringify(response.result)}`,
-      )
-    }
-    return ContractSourceResult.parse(response.result)[0]
+    return ContractSourceResult.parse(response)[0]
   }
 
   async getContractDeploymentTx(address: EthereumAddress): Promise<Hash256> {
```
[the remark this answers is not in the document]
mateuszradomski: From what I've seen in the docs, that's correct

packages/shared/src/services/etherscan/EtherscanClient.ts
mateuszradomski: Because it's not going to change, it's shared with every EtherscanClient or ArbiscanClient instance

packages/shared/src/services/etherscan/EtherscanClient.ts
mateuszradomski: I forgot to remove the one from backend, my fault

packages/backend/src/api/routers/StatusRouter.ts
mateuszradomski: good question, I'll move it there since it fits the bill of that package

packages/backend/src/config/config.local.ts
```diff
@@ -52,6 +52,7 @@ export function getLocalConfig(): Config {
     tvl: tvlEnabled && {
       alchemyApiKey: getEnv('ALCHEMY_API_KEY'),
       etherscanApiKey: getEnv('ETHERSCAN_API_KEY'),
+      arbiscanApiKey: getEnv('ARBISCAN_API_KEY'),
```
antooni, line 55: one more important thing, after we merge it people will need to update their .env files for local development, please inform especially Bartek and Luca about this

antooni: you know what? let's make this optional for now, because we will hide all this functionality behind the feature flag anyway

antooni (approved):

PR #1630, opened 2023-07-04
Separate balance fetching into balance provider abstraction

sdlyy:
Resolves L2B-1753

Little code abstraction-away. This change does not contain Arbitrum related changes - however, it will serve as a foundation for such.

### DIFF
--- a/packages/backend/src/core/balances/BalanceProvider.ts
+++ b/packages/backend/src/core/balances/BalanceProvider.ts
@@ -1,0 +1,57 @@
+import {
+  AssetId,
+  ChainId,
+  EthereumAddress,
+  UnixTime,
+} from '@l2beat/shared-pure'
+
+import { BalanceRecord } from '../../peripherals/database/BalanceRepository'
+import { BalanceCall } from '../../peripherals/ethereum/calls/BalanceCall'
+import { MulticallClient } from '../../peripherals/ethereum/MulticallClient'
+
+export interface HeldAsset {
+  holder: EthereumAddress
+  assetId: AssetId
+}
+
+export interface BalanceProvider {
+  fetchBalances: (
+    missingAssets: HeldAsset[],
+    timestamp: UnixTime,
+    blockNumber: number,
+  ) => Promise<BalanceRecord[]>
+  getChainId: () => ChainId
+}
+
+export class EthereumBalanceProvider implements BalanceProvider {
+  constructor(
+    private readonly multiCallClient: MulticallClient,
+    private readonly chainId: ChainId,
+  ) {}
+  public async fetchBalances(
+    missingData: HeldAsset[],
+    timestamp: UnixTime,
+    blockNumber: number,
+  ): Promise<BalanceRecord[]> {
+    const calls = missingData.map((m) =>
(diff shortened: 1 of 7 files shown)


packages/backend/src/core/balances/BalanceProvider.ts
```diff
@@ -0,0 +1,57 @@
...
+export interface BalanceProvider {
+  fetchBalances: (
+    missingAssets: HeldAsset[],
+    timestamp: UnixTime,
+    blockNumber: number,
+  ) => Promise<BalanceRecord[]>
+  getChainId: () => ChainId
+}
+
+export class EthereumBalanceProvider implements BalanceProvider {
```
mateuszradomski, line 26: This seems like it's only going to be using Ethereum mainnet, so taking `chainId` as argument seems useless since it will always be equal to `ChainId.ETHERUM`. If that is not the case then the naming of this class is misleading

packages/backend/src/core/balances/BalanceProvider.ts
```diff
@@ -0,0 +1,57 @@
...
+  constructor(
+    private readonly multiCallClient: MulticallClient,
+    private readonly chainId: ChainId,
+  ) {}
```
mateuszradomski, line 30: It's really hard to read without a space here

mateuszradomski (changes requested):
Some code organization issues only, other than that LGTM

packages/backend/src/core/balances/BalanceProvider.ts
sdlyy: https://github.com/l2beat/l2beat/pull/1630/commits/400987444f2034b3e4d97f61b345f5a55698a1f7

packages/backend/src/core/balances/BalanceProvider.ts
sdlyy: https://github.com/l2beat/l2beat/pull/1630/commits/5a33402f11cc030022fba9b95a0d4e3a450e0f6b

packages/backend/src/core/balances/BalanceProvider.ts
```diff
@@ -0,0 +1,57 @@
...
+import { BalanceRecord } from '../../peripherals/database/BalanceRepository'
+import { BalanceCall } from '../../peripherals/ethereum/calls/BalanceCall'
+import { MulticallClient } from '../../peripherals/ethereum/MulticallClient'
+
+export interface HeldAsset {
+  holder: EthereumAddress
+  assetId: AssetId
+}
+
+export interface BalanceProvider {
```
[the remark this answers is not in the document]
sdlyy: https://github.com/l2beat/l2beat/pull/1630/commits/aaad1067a71a9a4ed7a04dc4bca2912ee2771452

packages/backend/src/core/balances/BalanceProvider.ts
sdlyy: Fair enough, worth mentioning that such thing will shape itself along the way once we start to tackle new 'connectors' to L2

mateuszradomski (approved):

antooni (approved):

PR #1633, opened 2023-07-05
Add `asset_type` to the reports table

antooni:
Resolves L2B-1769


packages/shared-pure/src/types/ValueType.ts
```diff
@@ -0,0 +1,17 @@
...
+export function ValueType(value: string): ValueType {
+  if (!TYPES.includes(value)) {
+    throw new TypeError('Invalid ValueType')
+  }
+  return value as unknown as ValueType
+}
+
+const TYPES = ['TVL', 'CBV', 'EBV', 'NMV']
+
+ValueType.TVL = ValueType(TYPES[0])
```
mateuszradomski, line 14: Maybe leave a comment here that explains what TVL, CBV, EBV and NMV actually stand for?

packages/backend/src/core/reports/createReport.ts
```diff
@@ -29,9 +30,10 @@ export function createReport(
     timestamp: price.timestamp,
     projectId: balance.projectId,
     asset: balance.assetId,
-    balance: balance.balance,
-    balanceUsd,
-    balanceEth,
+    type: balance.type,
+    amount: balance.balance,
+    usdValue: balanceUsd,
+    ethValue: balanceEth,
```
mateuszradomski, line 36: I know that `convertBalance` returns `{ balanceUsd, balanceEth }` but I think it can be changed so that it returns `{ usdValue, ethValue }`. Also if you do so, maybe rename the function to be something like `convertValue`?

mateuszradomski (approved):
LGTM :ok_hand:

PR #1634, opened 2023-07-05
Modify risk description

BergDev22:
In the risk description, users can use 'force exit' to exit. it looks like a shortcoming, and we believe it is necessary to emphasize that this is a design feature. We suggest modifying it as follows.
"However, fortunately they can still exit the system." 

telegram-cloud-photo-size-5-6116393034687035737-y

In addition, we modify a section of text that was not very fluent.


lucadonnoh:
@BergDev22 i made some changes because i didn't like the word "fortunately", it gives a subjective feeling while the presence of the mechanism is not. wdyt?

lucadonnoh (approved):

PR #1637, opened 2023-07-05
Rename TVL tables and classes

antooni:
Resolves L2B-1770

### DIFF
--- a/packages/backend/src/peripherals/database/migrations/061_rename_tvl_tables.ts
+++ b/packages/backend/src/peripherals/database/migrations/061_rename_tvl_tables.ts
@@ -1,0 +1,74 @@
+/*
+                      ====== IMPORTANT NOTICE ======
+
+DO NOT EDIT OR RENAME THIS FILE
+
+This is a migration file. Once created the file should not be renamed or edited,
+because migrations are only run once on the production server.
+
+If you find that something was incorrectly set up in the `up` function you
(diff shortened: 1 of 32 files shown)


packages/backend/src/peripherals/database/migrations/061_rename_tvl_tables.ts
```diff
@@ -0,0 +1,74 @@
...
+    indexes: ['pkey', 'config_hash_index', 'unix_timestamp_index'],
+  },
+  {
+    oldName: 'report_status',
+    newName: 'reports_status',
+    indexes: ['pkey', 'config_hash_index', 'unix_timestamp_index'],
+  },
+  {
+    oldName: 'aggregate_reports',
+    newName: 'reports_aggregated',
```
mateuszradomski, line 34: Do we really want to switch that around? By that I mean, what is better:

- `aggregated_reports`
- `reports_aggregated`

Everywhere else it's the other way around, so the DB would be the only odd one out

mateuszradomski (approved):

packages/backend/src/peripherals/database/migrations/061_rename_tvl_tables.ts
antooni: I agree

sdlyy (approved):

PR #1638, opened 2023-07-05
Add /jobs redirect

antooni:


sz-piotr (changes requested):
Doesn't work
Screenshot 2023-07-05 at 16 45 03

sz-piotr (approved):

PR #1640, opened 2023-07-06
Set up BlockNumberUpdater for Arbitrum

mateuszradomski:
Resolves L2B-1646


sdlyy (approved):
Looks sweet, looking forward to merging this boy, so I can prevent some of the duplicates within work dependencies

packages/backend/src/modules/tvl/ArbitrumTvl.ts
```diff
@@ -0,0 +1,56 @@
...
+  // #endregion
+
+  // #region updaters
+
+  const arbiscanBlockNumberUpdater = new BlockNumberUpdater(
+    arbiscanClient,
+    db.blockNumberRepository,
+    clock,
+    logger,
+    ChainId.ETHEREUM,
```
sdlyy, line 39: Double check if it shouldn't be `ChainId.ARBITRUM`

mateuszradomski: Good catch!

packages/backend/src/modules/tvl/TvlModule.ts
```diff
@@ -40,113 +28,80 @@ export function createTvlModule(
...
 
   // #endregion
 
-  const start = async () => {
-    if (!config.syncEnabled) {
-      logger.info('TVL sync disabled')
-      return
-    }
+  const modules: (ApplicationModule | undefined)[] = [
+    createEthereumTvlModule(db, priceUpdater, config, logger, http, clock),
```
mateuszradomski, line 91: @antooni  Shouldn't we be creating the Arbitrum TVL module here as well? Why skip it?

packages/backend/src/modules/tvl/TvlModule.ts
```diff
@@ -40,113 +29,81 @@ export function createTvlModule(
...
-    aggregatedReportRepository,
-    reportStatusRepository,
-    clock,
-    config.projects,
-    logger,
-  )
 
   // #endregion
+
   // #region api
```
michalsidzej, line 71: What's the deal with those extra spaces? 
```suggestion
  // #endregion
  // #region api
```

packages/backend/src/modules/tvl/TvlModule.ts
```diff
@@ -40,113 +29,81 @@ export function createTvlModule(
...
 
   // #endregion
 
-  const start = async () => {
-    if (!config.syncEnabled) {
-      logger.info('TVL sync disabled')
-      return
-    }
+  const modules: (ApplicationModule | undefined)[] = [
+    createEthereumTvlModule(db, priceUpdater, config, logger, http, clock),
```
michalsidzej, line 92: WYT about renaming this to `Submodule`? It's one step deeper than module, so would be less confusing.
```suggestion
    createEthereumTvlSubmodule(db, priceUpdater, config, logger, http, clock),
```

michalsidzej (commented):
Overall looks good, gj

antooni (approved):

packages/backend/src/modules/tvl/TvlModule.ts
mateuszradomski: I like that idea! I've renamed these functions

PR #1641, opened 2023-07-06
Move aggregate_reports to separate updater

antooni:
Resolves L2B-1771


packages/backend/src/test/mockReports.ts
```diff
@@ -53,39 +103,35 @@ export const BALANCES: BalanceRecord[] = [
...
-    ],
-  },
+const FUTURE_BALANCES = BALANCES.map((balance) => ({
+  ...balance,
+  timestamp: NOW.add(1, 'hours'),
+}))
+
+const REPORTS = createReports(PRICES, BALANCES, PROJECTS)
+const FUTURE_REPORTS = createReports(FUTURE_PRICES, FUTURE_BALANCES, PROJECTS)
+// We need this mock for now, it ensures that the OP token is included in the
```
mateuszradomski, line 113: You'll forget about this comment in a week, mark it `TODO` or something that sticks out

mateuszradomski (approved):

packages/backend/src/core/reports/AggregatedReportUpdater.ts
```diff
@@ -0,0 +1,71 @@
...
+      this.projects,
+      timestamp,
+    )
+
+    await this.aggregatedReportStatusRepository.add({
+      configHash: this.configHash,
+      timestamp,
+    })
+
+    await this.aggregatedReportRepository.addOrUpdateMany(aggregatedReports),
```
michalsidzej, line 68: We should add these 2 in a transaction, or at least, in an opposite order. It's better to sync the same timestamp twice than to display untrue data.

michalsidzej (changes requested):

packages/backend/src/core/reports/AggregatedReportUpdater.ts
antooni: fully agree, nice catch, will update it in other places as well

packages/backend/src/core/reports/AggregatedReportUpdater.ts
```diff
@@ -0,0 +1,72 @@
...
+
+    await this.aggregatedReportRepository.addOrUpdateMany(aggregatedReports)
+
+    await this.aggregatedReportStatusRepository.add({
+      configHash: this.configHash,
+      timestamp,
+    })
+
+    this.logger.info('Report updated', { timestamp: timestamp.toNumber() })
+  }
```
michalsidzej, line 71: Nit pick, but those spaces are not necessary IMO

michalsidzej (approved):
LGTM

PR #1642, opened 2023-07-07
Remove incomplete banner

lucadonnoh:
deleted `isIncomplete`, but what do i need to do to delete this banner too? (solved by #1644)
image
cc @antooni


antooni:
@lucadonnoh I think this is the copy that we have on every bridge page, please contact @luke7211 whether we want to remove it or make it conditional on some pages

lucadonnoh:
i quickly skimmed through the code and it looked conditional, could you please check that this is not the case?

michalsidzej:
> i quickly skimmed through the code and it looked conditional, could you please check that this is not the case?

This is coming from here: https://github.com/l2beat/l2beat/blob/f7692f6a271c095f4d4c950e072ce0e335be33e2/packages/frontend/src/pages/bridges-projects/props/getTechnologyOverview.ts#L39. 

There's a bug (fixed by https://github.com/l2beat/l2beat/pull/1644). After merging this PR, it should work as expected.

michalsidzej (approved):
The bug was fixed, LGTM

PR #1666, opened 2023-07-13
Update TVLController

antooni:
How to include USDC(Arbitrum EBV) correctly in our API? for now we just want to add it to the L1 USDC, so the TVL will match

Probably a good way to go would be to:
- change the ReportRepository.getByTimestamp to include chainId and ValueType
- resolve issues after above change
- concat the responses for CBV, NMV, BV

Resolves L2B-1799


packages/backend/src/api/controllers/tvl/TvlController.ts
```diff
@@ -30,18 +31,38 @@ export class TvlController {
...
+        ChainId.ETHEREUM,
+        ValueType.CBV,
+      ),
     )
-    if (!timestamp) {
+    if (this.isArbitrumEnabled) {
+      timestamps.push(
+        await this.reportStatusRepository.findLatestTimestamp(
+          ChainId.ARBITRUM,
+          ValueType.CBV,
```
mateuszradomski, line 52: EBV?

mateuszradomski (changes requested):

antooni:
I have one question, what would be a better pattern?:
- return deduplicated records from the repository
- deduplicate records in the controller

@mateuszradomski

mateuszradomski:
> I have one question, what would be a better pattern?:
> 
>     * return deduplicated records from the repository
> 
>     * deduplicate records in the controller
> 
> 
> @mateuszradomski

The second option for sure imho

PR #1673, opened 2023-07-17
Add Base (under review)

lucadonnoh:


antooni:
The messaging is not clear here regarding TVL, we show 0$, which probably is not true 
image

lucadonnoh:
LGTM

antooni (approved):

PR #1681, opened 2023-07-18
Fix Base activity URL

antooni:


mateuszradomski (approved):

lucadonnoh (approved):

packages/config/src/layer2s/base.ts
```diff
@@ -70,7 +70,7 @@ export const base: Layer2 = {
     transactionApi: {
       type: 'rpc',
       startBlock: 1,
-      url: 'developer-access-mainnet.base.org',
+      url: 'https://developer-access-mainnet.base.org',
```
sz-piotr, line 73: We should add a check that this starts with https

PR #1706, opened 2023-07-21
Logger better integration with sentry

mateuszradomski:
It simplyfies the intersection between the Logger and TaskQueue, so we actually we the TaskQueue error as if it was thrown by the TaskQueue. This makes for faster understanding of what is happening with each TaskQueue


packages/backend/src/core/discovery/UpdateMonitor.ts
```diff
@@ -72,7 +72,7 @@ export class UpdateMonitor {
       try {
         await this.updateProject(projectConfig, blockNumber, timestamp)
       } catch (error) {
-        this.logger.error(error)
+        this.logger.error({ message: "Failed to update discovered project"}, error)
```
antooni, line 75:
```suggestion
        this.logger.error({ message: `Failed to update project [${projectConfig.name}]``}, error)
```

antooni (approved):

PR #1708, opened 2023-07-21
Update status endpoints

antooni:
Resolves L2B-1922


packages/backend/src/core/assets/ArbitrumEBVUpdater.ts
```diff
@@ -99,6 +99,12 @@ export class ArbitrumEBVUpdater implements AssetUpdater {
         timestamp: timestamp.toNumber(),
         minTimestamp: this.minTimestamp.toNumber(),
       })
+      await this.reportStatusRepository.add({
```
mateuszradomski, line 102: Why are we polluting our DB with useless information?

packages/backend/src/core/assets/CBVUpdater.ts
```diff
@@ -88,6 +88,12 @@ export class CBVUpdater implements AssetUpdater {
         timestamp: timestamp.toNumber(),
         minTimestamp: this.minTimestamp.toNumber(),
       })
+      await this.reportStatusRepository.add({
```
mateuszradomski, line 91: same question

packages/backend/src/core/assets/NMVUpdater.ts
```diff
@@ -90,6 +90,12 @@ export class NMVUpdater implements AssetUpdater {
         timestamp: timestamp.toNumber(),
         minTimestamp: this.minTimestamp.toNumber(),
       })
+      await this.reportStatusRepository.add({
```
mateuszradomski, line 93: same

mateuszradomski (changes requested):

packages/backend/src/core/assets/ArbitrumEBVUpdater.ts
antooni: This way next time backend starts with the same configHash it will not try to update those timestamps, without it we will be skipping every process reload. Maybe it should be handled differently, but I am not sure yet how

sdlyy (approved):
Looks good, @mateuszradomski mentioned all the necessities

mateuszradomski (approved):

mateuszradomski (approved):

PR #1724, opened 2023-07-25
Add bridges to Starknet

lucadonnoh:
Adds wstETH and rETH bridges.

Resolves L2B-1953


antooni (changes requested):
Does starknet keep list of bridges onchain? It would be ideal if those addresses were discovered and not hardcoded into initial ones. How can we be sure that it is indeed a bridge to starknet?

lucadonnoh:
i'll take another look, but i'm not sure that they can be discovered

lucadonnoh:
i don't see anything onchain, just asked them in the tg group with starkware if they have such a registry

antooni:
It depends, let's wait for today, I am just worried we will forget about it, but if you feel like you will not forget then merge it

antooni (approved):

PR #1725, opened 2023-07-25
Introduce chain notion to ConfigReader

antooni:
Resolves L2B-1966

please run the discovery and UpdateMonitor locally to see whether it does not break anything (I have done it twice, but better be safe than sorry)


packages/backend/src/api/controllers/status/discovery/props/getDashboardProjects.ts
```diff
@@ -20,13 +21,15 @@ export async function getDashboardProjects(
   configReader: ConfigReader,
   updateMonitorRepository: UpdateMonitorRepository,
 ): Promise<DashboardProject[]> {
-  const names = (await configReader.readAllConfigs()).map((c) => c.name)
+  const configs = await configReader.readAllConfigs(ChainId.ETHEREUM)
```
michalsidzej, line 24:
```suggestion
  const configs = await configReader.readAllForChain(ChainId.ETHEREUM)
```

michalsidzej (approved):
LGTM

PR #1726, opened 2023-07-25
Aggregate reports over new value types

sdlyy:
Resolves L2B-1916

### DIFF
--- a/packages/backend/src/api/controllers/tvl/TvlController.ts
+++ b/packages/backend/src/api/controllers/tvl/TvlController.ts
@@ -67,9 +67,11 @@ export class TvlController {
       await Promise.all([
         this.aggregatedReportRepository.getHourly(
           getHourlyMinTimestamp(timestamp),
+          ValueType.TVL,
         ),
         this.aggregatedReportRepository.getSixHourly(
           getSixHourlyMinTimestamp(timestamp),
+          ValueType.TVL,
         ),
         this.aggregatedReportRepository.getDaily(),
         this.reportRepository.getByTimestamp(timestamp),
--- a/packages/backend/src/api/controllers/tvl/TvlController.ts
+++ b/packages/backend/src/api/controllers/tvl/TvlController.ts
@@ -73,7 +73,7 @@ export class TvlController {
           getSixHourlyMinTimestamp(timestamp),
           ValueType.TVL,
         ),
-        this.aggregatedReportRepository.getDaily(),
+        this.aggregatedReportRepository.getDaily(ValueType.TVL),
         this.reportRepository.getByTimestamp(timestamp),
       ])
     const tvlApiResponse = generateTvlApiResponse(
--- a/packages/backend/src/core/reports/reportTree.ts
+++ b/packages/backend/src/core/reports/reportTree.ts
@@ -1,5 +1,41 @@
 import { assert } from '@l2beat/shared-pure'
 
+/**
+ * Simple wrapper around two-level nested `Map`.
+ * Useful when aggregating many sub-variants of a value.
+ * For example, when aggregating reports, we have a root value (project) and a branch value (value type from report).
+ * The leaf value may be the aggregated value of i.e all EBV reports
+ *
+ * @example
+ *    const tree = ReportTree.from(
+ *      [ARBITRUM, OPTIMISM],
+ *      [ValueType.CBV, ValueType.EBV, ValueType.NMV, ValueType.TVL],
+ *      () => ({ valueUsd: 0n, valueEth: 0n }),
+ *    )
+ *
+ *    tree.set(ARBITRUM, ValueType.CBV, () => ({ valueUsd: 1n, valueEth: 2n }))
+ *    tree.set(ARBITRUM, ValueType.NMV, () => ({ valueUsd: 10n, valueEth: 20n }))
+ *    tree.set(ARBITRUM, ValueType.TVL, () => ({ valueUsd: 11n, valueEth: 22n }))
+ *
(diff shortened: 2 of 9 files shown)


packages/backend/src/api/controllers/tvl/TvlController.ts
```diff
@@ -67,9 +67,11 @@ export class TvlController {
       await Promise.all([
         this.aggregatedReportRepository.getHourly(
           getHourlyMinTimestamp(timestamp),
+          ValueType.TVL,
         ),
         this.aggregatedReportRepository.getSixHourly(
           getSixHourlyMinTimestamp(timestamp),
+          ValueType.TVL,
         ),
         this.aggregatedReportRepository.getDaily(),
```
antooni, line 76: what about ValueType there?

packages/backend/src/api/controllers/tvl/TvlController.ts
```diff
@@ -67,9 +67,11 @@ export class TvlController {
       await Promise.all([
         this.aggregatedReportRepository.getHourly(
           getHourlyMinTimestamp(timestamp),
+          ValueType.TVL,
         ),
         this.aggregatedReportRepository.getSixHourly(
           getSixHourlyMinTimestamp(timestamp),
+          ValueType.TVL,
```
antooni, line 74: we definitely need a test for this function `getTvlApiResponse()`

packages/backend/src/core/reports/reportTree.ts
antooni, line 1: tbh it is quite hard to grasp this file, can you provide some context in the comment? why ReportTree and how ;p

sdlyy: Yea, so it's a simple wrapper around nested map structure that ease out working with all the data here.

Back then, we had only one level of nesting - simply TVL. It was represented with a simple one-level array.

Now we have a bit more to work on since we introduced `ValueType` per `Project` so now we could still use one-level array, but it would be a pain in the ass to search, filter and finally aggregate it. So instead tree abstracts away such structure:

```ts
const nestedMap = Map<ReportProject, Map<ValueType, ReportRecord[]>>()
```

Why wrapper?

It would be hell of assertion to do each time we read from it - and code looks bit tidier.

How does it look like?

```ts
    const tree = ReportTree.from(
      [ARBITRUM, OPTIMISM, POLYGON],
      [ValueType.CBV, ValueType.EBV, ValueType.NMV],
      () => ({ valueUsd: 0n, valueEth: 0n }),
    )

    console.dir({ tree }, { depth: null })
  ```
  
  Would look like this:
 
 ```ts
   tree: ReportTree {
    tree: Map(3) {
      'arbitrum' => Map(3) {
        'CBV' => { valueUsd: 0n, valueEth: 0n },
        'EBV' => { valueUsd: 0n, valueEth: 0n },
        'NMV' => { valueUsd: 0n, valueEth: 0n }
      },
      'optimism' => Map(3) {
        'CBV' => { valueUsd: 0n, valueEth: 0n },
        'EBV' => { valueUsd: 0n, valueEth: 0n },
        'NMV' => { valueUsd: 0n, valueEth: 0n }
      },
      'polygon' => Map(3) {
        'CBV' => { valueUsd: 0n, valueEth: 0n },
        'EBV' => { valueUsd: 0n, valueEth: 0n },
        'NMV' => { valueUsd: 0n, valueEth: 0n }
      }
    },
    roots: [ 'arbitrum', 'optimism', 'polygon' ]
  }
}
```

So now once we aggregate reports with empty tree already built, we simply check entries accessing them by 

```ts
      reportTree.set(project, valueType, () => ({
        usdValue,
        ethValue,
      }))
```

instead of looping through each time matching `ReportProject` & `ValueType`.

packages/backend/src/api/controllers/tvl/TvlController.ts
sdlyy: Shouldn't be a part of describing the endpoints? For now, we focus on only on aggregates and yet not expose the rest of the logic I presume

antooni:
why when doing `getDaily()` we do not pass value type, but when doing `getSixHourly()` we pass the value type?

sdlyy:
> why when doing `getDaily()` we do not pass value type, but when doing `getSixHourly()` we pass the value type?
@antooni 
https://github.com/l2beat/l2beat/pull/1726/commits/e422a2e60a07d881dd601e20ea6d6a2f56490b47

packages/backend/src/peripherals/database/AggregatedReportRepository.ts
```diff
@@ -19,30 +19,36 @@ export class AggregatedReportRepository extends BaseRepository {
...
     const rows = await knex('aggregated_reports')
-      .where({ value_type: ValueType.TVL.toString() })
+      .where({ value_type: valueType.toString() })
       .andWhereRaw(`EXTRACT(hour FROM unix_timestamp) = 0`)
       .orderBy('unix_timestamp')
     return rows.map(toRecord)
   }
 
-  async getSixHourly(from: UnixTime): Promise<AggregatedReportRecord[]> {
+  async getSixHourly(
```
antooni, line 31: The thing to consider for the future is whether we are using index on those calls

antooni (approved):

antooni (changes requested):
I will test it locally rn

antooni (approved):
works like prod, you can merge now

antooni (changes requested):
change the AGGREGATED_REPORT_LOGIC_VERSION to 1

antooni (approved):

PR #1740, opened 2023-07-28
Add Brine config

lucadonnoh:
Brine is a standard StarkEx Spot Validium.

Resolves L2B-1904.


antooni (approved):
From code perspective it looks good

PR #1742, opened 2023-07-30
Migrate Ethereum TVL to quicknode

sdlyy:
Resolves L2B-1992

## What's changed
Migrated RPC config of ETH TVL module to quicknode instead of current Alchemy setup


antooni (changes requested):
Have you updated the Heroku env? Also, please inform the team on discord about the breaking change

PR #1746, opened 2023-07-31
Add detailed TVL endpoint

sdlyy:
Resolves L2B-2005

## What's changed
- Added new endpoint for stacked charts.
- Response charts data now should include TVL values, along the per Value Type composition of the TVL
- Instead of chart old TVL-in-time now we include TVL-breakdown-in-time - that is:
```ts
const chartDataPoint = [timestamp, usdTvl, ethTvl]

// now's converted to:

const detailedDataPoint = [
  timestamp
  valueUsd,
  cbvUsd,
  ebvUsd,
  nmvUsd,
  valueEth,
  cbvEth,
  ebvEth,
  nmvEth,
]
```

### DIFF
--- a/packages/backend/src/api/controllers/tvl/DetailedTvlController.ts
+++ b/packages/backend/src/api/controllers/tvl/DetailedTvlController.ts
@@ -18,8 +18,8 @@ export class DetailedTvlController {
   constructor(
     private readonly reportStatusRepository: ReportStatusRepository,
     private readonly aggregatedReportRepository: AggregatedReportRepository,
-    private readonly aggregatedReportStatusRepository: AggregatedReportStatusRepository,
     private readonly reportRepository: ReportRepository,
+    private readonly aggregatedReportStatusRepository: AggregatedReportStatusRepository,
     private readonly projects: ReportProject[],
     private readonly logger: Logger,
   ) {
--- a/packages/backend/src/api/controllers/tvl/DetailedTvlController.ts
+++ b/packages/backend/src/api/controllers/tvl/DetailedTvlController.ts
@@ -1,0 +1,94 @@
+import { Logger } from '@l2beat/shared'
+import { DetailedTvlApiResponse, UnixTime } from '@l2beat/shared-pure'
+
+import { ReportProject } from '../../../core/reports/ReportProject'
+import { AggregatedReportRepository } from '../../../peripherals/database/AggregatedReportRepository'
+import { AggregatedReportStatusRepository } from '../../../peripherals/database/AggregatedReportStatusRepository'
+import { ReportRepository } from '../../../peripherals/database/ReportRepository'
+import { ReportStatusRepository } from '../../../peripherals/database/ReportStatusRepository'
+import { getHourlyMinTimestamp } from '../utils/getHourlyMinTimestamp'
+import { getSixHourlyMinTimestamp } from '../utils/getSixHourlyMinTimestamp'
+import {
+  groupByProjectIdAndAsset,
+  groupByProjectIdAndTimestamp,
+} from './detailedTvl'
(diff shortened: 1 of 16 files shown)


packages/backend/src/api/controllers/tvl/DetailedTvlController.ts
```diff
@@ -0,0 +1,94 @@
...
+  ) {
+    this.logger = this.logger.for(this)
+  }
+
+  async getDetailedTvlApiResponse(): Promise<
+    DetailedTvlApiResponse | undefined
+  > {
+    const [latestReportUpdateTime, latestAggregateReportUpdateTime] =
+      await Promise.all([
+        this.reportStatusRepository.findAnyLatestTimestamp(),
```
antooni, line 34: shouldn't we find the minimum here? so we get latest timestamp on which all types are synced?

packages/backend/src/api/controllers/tvl/DetailedTvlController.ts
```diff
@@ -0,0 +1,94 @@
...
+    private readonly aggregatedReportRepository: AggregatedReportRepository,
+    private readonly aggregatedReportStatusRepository: AggregatedReportStatusRepository,
+    private readonly reportRepository: ReportRepository,
+    private readonly projects: ReportProject[],
+    private readonly logger: Logger,
+  ) {
+    this.logger = this.logger.for(this)
+  }
+
+  async getDetailedTvlApiResponse(): Promise<
```
antooni, line 29: I feel like this needs a test at least for the timestamps usage, ideally also for the response creation based on mock DB data

antooni (changes requested):
Please resolve comments

packages/backend/src/api/controllers/tvl/DetailedTvlController.ts
antooni: There are different updaters for EBV, NMV and CBV

antooni: They run in parallel

packages/backend/src/peripherals/database/AggregatedReportRepository.ts
```diff
@@ -28,6 +28,14 @@ export class AggregatedReportRepository extends BaseRepository {
     return rows.map(toRecord)
   }
 
+  async getDailyWithAnyType(): Promise<AggregatedReportRecord[]> {
```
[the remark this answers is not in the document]
sdlyy: Yes, we do (?)

```bash
"Sort  (cost=78863.26..78993.14 rows=51951 width=28) (actual time=163.541..193.689 rows=428596 loops=1)"
"  Sort Key: unix_timestamp"
"  Sort Method: external merge  Disk: 16696kB"
"  ->  Bitmap Heap Scan on aggregated_reports  (cost=581.17..74794.24 rows=51951 width=28) (actual time=18.466..78.509 rows=428596 loops=1)"
"        Recheck Cond: (EXTRACT(hour FROM unix_timestamp) = '0'::numeric)"
"        Heap Blocks: exact=5180"
"        ->  Bitmap Index Scan on aggregated_reports_unix_timestamp_daily_index  (cost=0.00..568.18 rows=51951 width=0) (actual time=17.975..17.975 rows=433048 loops=1)"
"Planning Time: 0.230 ms"
"Execution Time: 211.451 ms"
```

packages/backend/src/api/controllers/tvl/DetailedTvlController.ts
sdlyy: We have to find those in the optimistic way anyway. Also, it would be great to build something to glue together fetching and setting safe sync heights, so we would prevent huge combined traversal query like `amount of chains * amount of supported assets' types on chains`

antooni (approved):

PR #1748, opened 2023-07-31
Add retry to DiscoveryRunner

antooni:
Resolves L2B-2003

### DIFF
--- a/packages/backend/src/core/discovery/DiscoveryRunner.ts
+++ b/packages/backend/src/core/discovery/DiscoveryRunner.ts
@@ -40,13 +40,14 @@ export class DiscoveryRunner {
       ? await this.updateInitialAddresses(projectConfig)
       : projectConfig
 
-    const discovery = await this.discover(config, blockNumber)
+    const discovery: DiscoveryOutput = await this.discoverWithRetry(
+      config,
+      blockNumber,
+    )
 
     if (options.runSanityCheck) {
       await this.sanityCheck(discovery, config, blockNumber)
     }
-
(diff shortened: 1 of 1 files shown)


packages/backend/src/core/discovery/DiscoveryRunner.ts
```diff
@@ -71,6 +72,32 @@ export class DiscoveryRunner {
...
+            discovery = await this.discover(config, blockNumber)
+          } catch {
+            throw error
+          }
+        } else {
+          throw error
+        }
+      } else {
+        throw error
+      }
```
michalsidzej, line 96: Use the negation and early `return` (early `throw` in this case) to avoid nesting

packages/backend/src/core/discovery/DiscoveryRunner.ts
```diff
@@ -40,14 +40,15 @@ export class DiscoveryRunner {
       ? await this.updateInitialAddresses(projectConfig)
       : projectConfig
 
-    const discovery = await this.discover(config, blockNumber)
+    const discovery: DiscoveryOutput = await this.discoverWithRetry(
```
mateuszradomski, line 43: I think that this type declaration (`: DiscoveryOutput`) is unnecessary and can be removed

packages/backend/src/core/discovery/DiscoveryRunner.ts
mateuszradomski, line 1: Test that it actually retries

mateuszradomski (changes requested):

mateuszradomski (approved):

PR #1749, opened 2023-07-31
L2 Assets Frontend: Connect to backend

mateuszradomski:
Before, we were generating random data instead of fetching it from the backend, right now we are fetching using JSON/HTTP from a local server. The only thing that needs to now change is change the backend api that will serve us this data and frontend for L2 Assets will now be connected to backend and showing real time data.


packages/frontend/src/build/buildPages.ts
```diff
@@ -23,7 +24,9 @@ async function main() {
 
   const http = new JsonHttpClient(new HttpClient(), config.backend.skipCache)
 
-  const tvlApiResponse = await fetchTvlApi(config.backend.apiUrl, http)
+  const tvlApiResponse = config.features.detailedTvl
```
antooni, line 27: This one is risky, the endpoint is also behind the feature flag on the backend, so after you enable it here and forget to do so on the backend, the build will fail. I thought that the detailed TVL would be possibly undefined and fetched in a different step, and later passed as a fourth argument to createApi() wdyt?

packages/frontend/src/components/chart/configure/effects/handleEffect.ts
```diff
@@ -54,63 +53,73 @@ function handleFetchDetailedAggregateTvl(
...
+    // eslint-disable-next-line @typescript-eslint/no-unsafe-assignment
+    (data) => ({ type: 'AggregateDetailedTvlLoaded', requestId, data }),
+    () => ({ type: 'AggregateDetailedTvlFailed', requestId }),
+    AggregateDetailedTvlResponse,
+  )
 
-  if (url === 'ESLINT LEAVE ME ALONE') {
-    return
-  }
+  // const timestamps = [
```
antooni, line 66: remove commented out data

packages/frontend/src/components/table/NonZeroDiffedNumberCell.tsx
```diff
@@ -0,0 +1,32 @@
...
+export interface NonZeroDiffedNumberCellProps {
+  value?: string
+  percentChange?: string
+}
+
+// NOTE(radomski): This is not an ASCII minus character, but a UTF-8 "Em Dash"
+// Look here for more: https://www.compart.com/en/unicode/U+2014
+const LONG_HYPHEN = '—'
+
+export function NonZeroDiffedNumberCell(props: NonZeroDiffedNumberCellProps) {
```
antooni, line 15: Can't this just be named ValueCell ?

antooni (approved):

PR #1751, opened 2023-08-01
Add diff tool

lucadonnoh:
**requirements**: [foundry suite](https://book.getfoundry.sh/getting-started/installation), [difft](https://github.com/Wilfred/difftastic)
**usage**: 
- run discovery for selected chains
- `python3 diffc.py zora base`

works only for OP stack chains, try diffing between `zora`, `base` and `publicgoodsnetwork`


packages/backend/scripts/diffc.py
```diff
@@ -0,0 +1,241 @@
...
+
+    directories1, ignored1 = list_directories(args.Folder_Name1)
+    directories2, ignored2 = list_directories(args.Folder_Name2)
+
+    common_directories = directories1 & directories2
+    unique_to_folder1 = directories1 - directories2
+    unique_to_folder2 = directories2 - directories1
+    ignored_directories = ignored1 | ignored2
+
+    if common_directories:
```
mateuszradomski, line 213: What do you think about printing something like this:

```
Common Directories: ....
Directories unique to 1: NONE
Directories unique to 2: ......
Ignored directories: NONE
```

So we would print `NONE` when there are no directories for a given arrangement. That way we always print something about all four of these values and the possible confusion "Where are x?" disappear

packages/backend/scripts/diffc.py
```diff
@@ -0,0 +1,241 @@
...
+                    try:
+                        chunk = os.read(fd, 1024)
+                        if not chunk:  # no more data
+                            break
+                        result += chunk
+                    except OSError:
+                        break
+                result = result.decode()
+                if "No changes." not in result:
+                    # Split result into lines
```
mateuszradomski, line 181: It's obvious what is happening in this place from the code, comments like these IMO add noise instead of help. I'm not going to fight fiercely about it, it can stay here :sweat_smile:

packages/backend/scripts/diffc.py
```diff
@@ -0,0 +1,241 @@
...
+                    lines = result.split('\n')
+                    # Only print lines that are not empty
```
mateuszradomski, line 183: This is my last nitpick about comments, I promise. This comment is redundant with the second one just below. You can understand what is happening by reading either of these comments, so two comments are twice as likely to become outdated when this code changes. Or worse, one will get updated and the second won't and now you have a contradiction in your comments and the reader does not know which one to believe in

packages/backend/scripts/diffc.py
```diff
@@ -0,0 +1,241 @@
...
+            file2 = sol_file.replace(
+                "<placeholder>", dir_path2).replace(folder1, folder2)
+
+            # Run "forge fmt" on each file before diffing
+            # subprocess.run(["forge", "fmt", file1])
+            # subprocess.run(["forge", "fmt", file2])
+
+            pid, fd = pty.fork()
+            if pid == 0:  # child process
+                os.execvp("difft", ["difft", "--skip-unchanged",
```
mateuszradomski, line 167: this assumes that the user has this executable on his computer. There is no check and a friendly error print out when such an executable is not found. Better yet for portability, diffing tools are not that complex, one can implement a simple diff routine in python at most two hounded lines, look at [Levenshtein distance](https://en.wikipedia.org/wiki/Levenshtein_distance) if you're interested

packages/backend/scripts/diffc.py
```diff
@@ -0,0 +1,241 @@
...
+
+        common_sol_files = sol_files1 & sol_files2
+
+        for sol_file in common_sol_files:
+            file1 = sol_file.replace(
+                "<placeholder>", dir_path1).replace(folder1, folder1)
+            file2 = sol_file.replace(
+                "<placeholder>", dir_path2).replace(folder1, folder2)
+
+            # Run "forge fmt" on each file before diffing
```
mateuszradomski, line 161: Is this dead code? If so, can this be deleted?

packages/backend/scripts/diffc.py
```diff
@@ -0,0 +1,241 @@
...
+                f"{folder2} does not contain {directory} with project structure {logical_name2}.")
+            continue
+
+        sol_files1 = [os.path.join(root, f) for root, dirs, files in os.walk(
+            dir_path1) for f in files if f.endswith('.sol')]
+        sol_files2 = [os.path.join(root, f) for root, dirs, files in os.walk(
+            dir_path2) for f in files if f.endswith('.sol')]
+
+        # Map to common format
+        sol_files1 = {f.replace(dir_path1, "<placeholder>")
```
mateuszradomski, line 148: Instead of a comment, you can extract that to a function and the code becomes self documenting. Also, it's way easier to later edit this code. Same applies to `.sol` file filtering above

packages/backend/scripts/diffc.py
```diff
@@ -0,0 +1,241 @@
...
+        sol_files2 = [os.path.join(root, f) for root, dirs, files in os.walk(
+            path2) for f in files if f.endswith('.sol')]
+
+        common_sol_files = set(sol_files1) & set(sol_files2)
+
+        for sol_file in common_sol_files:
+            file1 = sol_file.replace(folder1, folder1)
+            file2 = sol_file.replace(folder1, folder2)
+
+            # Run "forge fmt" on each file before diffing
```
mateuszradomski, line 90: More dead code?

packages/backend/scripts/diffc.py
```diff
@@ -0,0 +1,241 @@
...
+                             "ethereum", ".code", directory, "proxy")
+
+        if not os.path.exists(path1):
+            print(f"'proxy' subdirectory does not exist in {path1}")
+            continue
+        if not os.path.exists(path2):
+            print(f"'proxy' subdirectory does not exist in {path2}")
+            continue
+
+        sol_files1 = [os.path.join(root, f) for root, dirs, files in os.walk(
```
mateuszradomski, line 79: This is duplicate code with the function below. A good rule of thumb for pulling things out into functions is 
> If I repeat myself three times, it has very good grounds to be promoted to a function

packages/backend/scripts/diffc.py
```diff
@@ -0,0 +1,241 @@
...
+
+
+def list_directories(folder_name):
+    base_path = ".."
+    path = os.path.join(base_path, "discovery",
+                        folder_name, "ethereum", ".code")
+
+    try:
+        entries = os.listdir(path)
+        directories = [
```
mateuszradomski, line 42: This is too complex and is unreadable. A better solution would be:

```
def is_accepted_directory(entry):
  return os.path.isdir(...) and not any(keyword in ...)
  
entries.filter(is_accepted_directory)
```

mateuszradomski (changes requested):

packages/backend/scripts/diffc.py
lucadonnoh: done

packages/backend/scripts/diffc.py
lucadonnoh: i particularly want to use difftastic since it is syntax-aware: for example, it has explicit solidity supports and for this reason it shows far better diffs than other tools that we have used in the past (either online or internal tools)

regarding the check, i will add a friendly error

lucadonnoh: done

packages/backend/scripts/diffc.py
lucadonnoh: done

packages/backend/scripts/diffc.py
lucadonnoh: done

mateuszradomski (approved):

PR #1758, opened 2023-08-02
L2 Assets Frontend: Add token modal

mateuszradomski:
Resolves L2B-2032

In this PR a new way to pick tokens is added, instead of a vulcanic eruption of all tokens possible as a list, the most important tokens now get placed inside a table, separated by CBV/EBV/NMV with icons which is a lot more user friendly.


antooni:
window resize refreshes the state

mateuszradomski:
> window resize refreshes the state

It's now fixed, the issue was cause by the fact the when the window is resized the entire previousState get's wiped clean, and my `classList.toggle` class were happening over and over. They still do, but right now I force if classes should be added or removed.

packages/frontend/src/components/chart/Chart.tsx
```diff
@@ -120,10 +121,17 @@ export function Chart({
...
+            {hasTvl && (
+              <div className="flex h-[2rem] items-end">
+                <CurrencyControls />
+                {hasDetailedTvl && <TokenControls tokens={tokens} />}
+              </div>
+            )}
             <ScaleControls />
           </div>
-          {hasTvl && <TokenControls tokens={tokens} />}
+          {hasTvl && !hasDetailedTvl && (
```
antooni, line 132: This flag does not work on staging, the "Regular TVL" shows new modal

mateuszradomski: On production it does not

antooni (approved):

PR #1759, opened 2023-08-02
Update OP message

luke7211:
Update incorrect "based on OP Stack" message with "based on OP Stack or old OVM"


lucadonnoh:
my current understanding is that both Bedrock and OVM chains are OP stack chains, given my private conversation with Kelvin and the many posts by Binji (e.g. https://warpcast.com/binji/0x85045a)

but i also agree that OVM architecture is not really a "stack" and doesn't fit much in the OP stack vision, therefore I'm ok to distinguish OVM from OP stack chains compared to OVM vs Bedrock

lucadonnoh (approved):

PR #1760, opened 2023-08-02
Add Top tokens listed per given project - endpoint data

sdlyy:
Resolves L2B-2006

## What's changed
Added simple types which frontend will require to implement modal with top chain tokens


packages/shared-pure/src/types/api/DetailedTvlApi.ts
```diff
@@ -44,9 +46,19 @@ export const DetailedTvlApiToken = z.object({
...
+  chainId: branded(z.number(), ChainId),
+  valueType: branded(z.string(), ValueType),
+})
+
+export type DetailedTvlApiTopToken = z.infer<typeof DetailedTvlApiTopToken>
+
 export type DetailedTvlApiToken = z.infer<typeof DetailedTvlApiToken>
 
 export const DetailedTvlApiProject = z.object({
+  topTokens: z.array(DetailedTvlApiTopToken),
```
antooni, line 61: let's do Top Tokens per value type, I know right now it is not an issue, but in the future can be helpful (when we have more NMV and EMV tokens in config)

packages/backend/src/api/controllers/tvl/detailedTvl.ts
```diff
@@ -49,27 +49,38 @@ export function nestedGroupBy<T extends ReportRecord | AggregatedReportRecord>(
...
       return {
-        assetId: AssetId(asset),
-        tvl: asNumber(tokenUsdTvl, 2),
+        ...prev,
+        [valueType]: reports
+          .map((report) => ({
+            assetId: report.asset,
+            chainId: report.chainId,
+            valueType: report.type,
+            tvl: asNumber(report.usdValue, 2),
```
antooni, line 78: probably a better name would be `usdValue`

sdlyy:
> have you tested the endpoint locally?

@antooni, yeah - works just fine, even when it comes to the performance

antooni (approved):

PR #1765, opened 2023-08-04
L2 Assets frontend: Feedback changes

mateuszradomski:


antooni (changes requested):
change also the column names to "Canonically Bridged" etc.

antooni (approved):

PR #1767, opened 2023-08-04
Rewrite StarknetClient to use RPC

michalsidzej:
This PR needs adding env variables on heroku: 

`ACTIVITY_STARKNET_CALLS`
`ACTIVITY_STARKNET_URL`

### DIFF
--- a/packages/backend/src/config/config.production.ts
+++ b/packages/backend/src/config/config.production.ts
@@ -113,6 +113,11 @@ export function getProductionConfig(): Config {
           callsPerMinute: getEnv.integer('ACTIVITY_POLYGONZKEVM_CALLS'),
           url: getEnv('ACTIVITY_POLYGONZKEVM_URL'),
         },
+        starknet: {
+          type: 'starknet',
+          callsPerMinute: getEnv.integer('ACTIVITY_STARKNET_CALLS'),
+          url: getEnv('ACTIVITY_STARKNET_URL'),
+        },
       },
     },
     statusEnabled: getEnv.boolean('STATUS_ENABLED', true),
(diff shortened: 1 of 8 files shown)


packages/backend/src/config/config.production.ts
```diff
@@ -113,6 +113,11 @@ export function getProductionConfig(): Config {
           callsPerMinute: getEnv.integer('ACTIVITY_POLYGONZKEVM_CALLS'),
           url: getEnv('ACTIVITY_POLYGONZKEVM_URL'),
         },
+        starknet: {
```
antooni, line 116: update also local config

packages/backend/src/peripherals/starknet/StarknetClient.ts
```diff
@@ -24,8 +24,8 @@ export class StarknetClient {
   }
 
   async getBlockNumber() {
-    const block = await this.getBlock('latest')
-    return block.number
+    const response = await this.getBlock('latest')
```
antooni, line 27: why this cannot be named block?

packages/backend/src/config/config.production.ts
antooni: nvm, I see it below now ;p

packages/config/src/layer2s/starknet.ts
```diff
@@ -166,8 +166,8 @@ export const starknet: Layer2 = {
     ],
     transactionApi: {
       type: 'starknet',
-      callsPerMinute: 60 * 5,
-      url: 'https://alpha-mainnet.starknet.io',
+      url: 'https://starknet-mainnet.g.alchemy.com/v2/demo',
```
antooni, line 169: I am not sure whether is a pattern, but in projects that use env variables on heroku we do not provide url and calls, which signals it is using custom RPC url

antooni (changes requested):
please resolve comments

antooni (approved):

PR #1773, opened 2023-08-08
Display tokens in the detailed TVL chart type

mateuszradomski:
Before, we were connected to the database and displayed stacked parts of each TVL, but we did nothing when a given token was chosen. This PR adds the functionality to display the token value in the detailed TVL style, changing colors and marker with different token sources


antooni:
nice branch name

packages/frontend/src/components/chart/ChartLoader.tsx
```diff
@@ -4,7 +4,7 @@ export function ChartLoader() {
   return (
     <svg
       data-role="chart-loader"
-      className="pointer-events-none relative z-20 mx-auto mt-[62px] hidden select-none opacity-40 sm:mt-[85px]"
+      className="pointer-events-none relative z-40 mx-auto mt-[62px] hidden select-none opacity-40 sm:mt-[85px]"
```
antooni, line 7: comment would be useful, because I see some magic in there ;p

packages/frontend/src/components/chart/TokenControls.tsx
```diff
@@ -58,7 +60,7 @@ function SelectButton() {
 function SelectedTokenButton() {
   return (
     <label
-      className="flex hidden cursor-pointer items-center justify-between gap-1.5 rounded-md px-2 text-base transition-all dark:bg-black"
+      className="flex hidden cursor-pointer items-center justify-between gap-1.5 rounded-md bg-white px-2 text-base transition-all dark:bg-black"
```
antooni, line 63: flex and hidden apply the same property (display)

antooni:
What is this `<p>ARTH</p>` thing?

packages/frontend/src/components/chart/configure/effects/handleEffect.ts
```diff
@@ -80,15 +80,15 @@ function handleFetchAlternativeTvl(
...
+  { url, requestId, token, assetType }: FetchTokenTvlEffect,
   dispatch: (message: Message) => void,
 ) {
   timeoutLoader(requestId, dispatch)
   fetchThenDispatch(
     url,
     dispatch,
     // eslint-disable-next-line @typescript-eslint/no-unsafe-assignment
-    (data) => ({ type: 'TokenTvlLoaded', requestId, token, data }),
+    (data) => ({ type: 'TokenTvlLoaded', requestId, token, assetType, data }),
```
antooni, line 91: wouldn't it break the production token fetching logic? because the detailed tvl is disabled on prod server

packages/frontend/src/components/chart/configure/render/renderHover.ts
```diff
@@ -154,6 +155,16 @@ export function renderHover(
         rows.push(renderCBVRow(point.usdParts.cbv))
         rows.push(renderEBVRow(point.usdParts.ebv))
         rows.push(renderNMVRow(point.usdParts.nmv))
+      } else if (
+        state.view.chart.type === 'TokenDetailedTvlChart' &&
+        'balance' in point
```
antooni, line 160: why do we need this "balance in point"?

antooni (commented):
GJ, have left some comments, mostly because i did not understand some parts of the code

packages/frontend/src/components/chart/ChartLoader.tsx
mateuszradomski: I'm going to leave the reasoning here because it's straight forward I think and putting it as a comment is not really helpful.

`z-20` meant that the Loader element was behind the chart. `z-40` means that the Loader element is now in front of the Chart. Since DetailedTVL is opaque, we need all the things to be in front of the chart.

packages/frontend/src/components/chart/TokenControls.tsx
mateuszradomski: I know, but we are going to remove and add hidden to show/hide this element. If we only specified `hidden` we would have to 

1. Remove `hidden` and add `flex` for showing
2. Add `hidden` and remove `flex` for hiding

Which only adds more complexity

packages/frontend/src/components/chart/configure/render/renderHover.ts
mateuszradomski: Since it's a union, it doesn't know which fields are present in this objects, and if we check that `balance` is present now we know for sure that this object is `TokenDetailedTvlChart`.

packages/frontend/src/components/chart/configure/effects/handleEffect.ts
mateuszradomski: Not really, we just assume that on prod every single token is EBV and off we go :+]

sdlyy (approved):

sdlyy:
Do we have some deprecation/migration plan for classic TVL?

PR #1776, opened 2023-08-08
Refactor detailed TVL data timings

sdlyy:
Resolves L2B-2016

## What's changed

Simplified timestamp logic. No status field, only conditional check beforehand is enough tho.
Optimization at the end of the day, instead of few longer queries we could do the one generic one in the first place since once we got Aggregated Report for given timestamp in place we are also sure plain reports have been generated since aggregated reports are simply basic reports' summary 🤯 

Query benchmark, down below, works as if no changes were made.

Also, we agreed to enable returning 404 if data is in unsycned state. It's optional and controllable via backend feature flag.

Ah, yeah and sync status - get the latest aggregated config hash, count aggregated reports updated with this hash and with different hash (those haven't been synced yet). If the amount of different hashes is not equal to `0` that means data is mid-way through re-indexing process. Plain and simple. Also, if there are no different hashes in case of fresh database i.e. locally no data will be returned since there are no timestamps to query for.

Query written on the knee 😅 - this is production database roa 
image

### DIFF
--- a/packages/backend/README.md
+++ b/packages/backend/README.md
@@ -80,7 +80,7 @@ environment variables. One variable per line `KEY=value`.
 
 ##### Multichain discovery CLI
 
-If you want to use multichain discovery cli, make sure you include the `RPC_URL` and `ETHERSCAN_API_KEY` for desired chains. If you want to use one RPC provider, all of them (unless stated in brackets) are supported by https://www.quicknode.com/ and https://www.ankr.com/. Etherscan API key should be created by creating an account on Etherscan instance for every chain separately (check config.discovery.ts for etherscan links).
+If you want to use multichain discovery cli, make sure you include the `RPC_URL` and `ETHERSCAN_API_KEY` for desired chains. If you want to use one RPC provider, all of them (unless stated in brackets) are supported by <https://www.quicknode.com/> and <https://www.ankr.com/>. Etherscan API key should be created by creating an account on Etherscan instance for every chain separately (check config.discovery.ts for etherscan links).
 
 - `DISCOVERY_ARBITRUM_RPC_URL` (Optional)
 - `DISCOVERY_ARBITRUM_ETHERSCAN_API_KEY` (Optional)
@@ -125,7 +125,10 @@ ETHEREUM_ALCHEMY_API_KEY=
 # ARBITRUM_TVL_ENABLED=
 # ARBISCAN_API_KEY=
 # ARBITRUM_PROVIDER_URL=
-
+## Enable/disable detailed-tvl endpoint exposure - defaults to true
+# DETAILED_TVL_ENABLED=
+## Should detailed-tvl endpoint return 404 if data is not synced yet - defaults to false
+# SKIP_UNSYNCED_DETAILED_TVL=
 # UPDATE_MONITOR_RUN_ON_START=
 # ACTIVITY_PROJECTS_EXCLUDED_FROM_API=project-a project-b
 
--- a/packages/backend/README.md
+++ b/packages/backend/README.md
@@ -125,10 +125,10 @@ ETHEREUM_ALCHEMY_API_KEY=
 # ARBITRUM_TVL_ENABLED=
 # ARBISCAN_API_KEY=
 # ARBITRUM_PROVIDER_URL=
-## Enable/disable detailed-tvl endpoint exposure - defaults to true
-# DETAILED_TVL_ENABLED=
-## Should detailed-tvl endpoint return 404 if data is not synced yet - defaults to false
-# SKIP_UNSYNCED_DETAILED_TVL=
+## Enable/disable detailed-tvl endpoint exposure - defaults to false
+# DETAILED_TVL_ENABLED=true
+## Should detailed-tvl endpoint return 404 if data is not synced yet? - defaults to false
+# SKIP_UNSYNCED_DETAILED_TVL=true
 # UPDATE_MONITOR_RUN_ON_START=
 # ACTIVITY_PROJECTS_EXCLUDED_FROM_API=project-a project-b
 
--- a/packages/backend/README.md
+++ b/packages/backend/README.md
@@ -128,7 +128,7 @@ ETHEREUM_ALCHEMY_API_KEY=
 ## Enable/disable detailed-tvl endpoint exposure - defaults to false
 # DETAILED_TVL_ENABLED=true
 ## Should detailed-tvl endpoint return 404 if data is not synced yet? - defaults to false
-# SKIP_UNSYNCED_DETAILED_TVL=true
+# ERROR_ON_UNSYNCED_DETAILED_TVL=true
 # UPDATE_MONITOR_RUN_ON_START=
 # ACTIVITY_PROJECTS_EXCLUDED_FROM_API=project-a project-b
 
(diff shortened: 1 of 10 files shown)


packages/backend/src/api/controllers/tvl/DetailedTvlController.ts
```diff
@@ -44,25 +45,25 @@ export class DetailedTvlController {
   async getDetailedTvlApiResponse(): Promise<
     DetailedTvlApiResponse | undefined
   > {
-    const minimumTimestamp = await this.getMinimumTimestamp()
+    const dataTimings = await this.getDataTimings()
 
-    if (!minimumTimestamp) {
+    if (!dataTimings.latestTimestamp) {
       return
     }
```
sdlyy, line 52: Do we want to return undefined (404) is data is out sync now or later?

antooni: yeah, let's do it, we will observe how it behaves on staging, but hide it behind the feature flag

packages/backend/src/api/controllers/tvl/DetailedTvlController.ts
```diff
@@ -44,25 +45,25 @@ export class DetailedTvlController {
   async getDetailedTvlApiResponse(): Promise<
     DetailedTvlApiResponse | undefined
   > {
-    const minimumTimestamp = await this.getMinimumTimestamp()
+    const dataTimings = await this.getDataTimings()
```
antooni, line 48: does it really introduce so little overhead to the response time? maybe it would be good to cache it somehow?

antooni (commented):
have left some comments

packages/backend/src/api/controllers/tvl/DetailedTvlController.ts
sdlyy: It is acceptable - actually lower than before.

#### Timing overhead distribution:
- Current (production replica synced up to yesterday): 
  - `findCountsForHash: 19.291166067123413ms`
  - `database query`
image

- Previous method:
  - `getMinimumTimestamp: 47.10783314704895ms`
  - `database query` - skipped since we have quite a few queries there

packages/backend/src/api/controllers/tvl/DetailedTvlController.ts
sdlyy: https://github.com/l2beat/l2beat/pull/1776/commits/96dab8887c3ff7a92dc428953f4701a77e660ce0

packages/backend/README.md
```diff
@@ -125,7 +125,10 @@ ETHEREUM_ALCHEMY_API_KEY=
 # ARBITRUM_TVL_ENABLED=
 # ARBISCAN_API_KEY=
 # ARBITRUM_PROVIDER_URL=
-
+## Enable/disable detailed-tvl endpoint exposure - defaults to true
+# DETAILED_TVL_ENABLED=
+## Should detailed-tvl endpoint return 404 if data is not synced yet - defaults to false
+# SKIP_UNSYNCED_DETAILED_TVL=
```
mateuszradomski, line 131: Better name would be `ERROR_ON_UNSYNCED_DETAILED_TV`

packages/backend/src/api/controllers/tvl/DetailedTvlController.ts
```diff
@@ -25,6 +25,30 @@ import {
 } from './detailedTvl'
 import { generateDetailedTvlApiResponse } from './generateDetailedTvlApiResponse'
 
+interface DetailedControllerBehaviorFlags {
+  skipUnsyncedDetailedTvl: boolean
```
mateuszradomski, line 29: Propagate this new, better name through here

mateuszradomski (approved):
IMO looks good, I'd strongly advise to change that name to indicate that it's going to error out, you can use the one I'm suggesting or anything else if you have something better, but it should be changed either way

packages/backend/README.md
```diff
@@ -125,7 +125,10 @@ ETHEREUM_ALCHEMY_API_KEY=
 # ARBITRUM_TVL_ENABLED=
 # ARBISCAN_API_KEY=
 # ARBITRUM_PROVIDER_URL=
-
+## Enable/disable detailed-tvl endpoint exposure - defaults to true
+# DETAILED_TVL_ENABLED=
```
sz-piotr, line 129: We should not default to true. Best practice is to always default to false and have the user provide non-falsy values.
Also for booleans in this file we could provide `=true` instead of just `=` since that would be the only possible version.

packages/backend/src/api/controllers/tvl/DetailedTvlController.ts
```diff
@@ -34,35 +58,45 @@ export class DetailedTvlController {
     private readonly projects: ReportProject[],
     private readonly tokens: Token[],
     private readonly logger: Logger,
+    private readonly aggregatedConfigHash: Hash256,
+    private readonly flags: DetailedControllerBehaviorFlags,
```
sz-piotr, line 62: Can we just call it options: DetailedTvlControllerOptions?

packages/backend/src/api/routers/TvlRouter.ts
```diff
@@ -57,14 +57,22 @@ export function createTvlRouter(
 
   if (features.detailedTvlEnabled) {
     router.get('/api/detailed-tvl', async (ctx) => {
-      const data = await detailedTvlController.getDetailedTvlApiResponse()
+      const detailedTvlResult =
+        await detailedTvlController.getDetailedTvlApiResponse()
+
+      if (detailedTvlResult.result === 'error') {
```
sz-piotr, line 63: I think I'd prefer a more holistic approach to error handling and codes in the controller. This can happen as another PR

sz-piotr (changes requested):

packages/backend/src/api/routers/TvlRouter.ts
sdlyy: We need to standardize it either way. Can't estimate how many endpoints' backend will grow to. Now we have 'some' convention which relies on partly handled cases and exceptions I suppose. Another PR it is 😅

packages/backend/README.md
sdlyy: https://github.com/l2beat/l2beat/pull/1776/commits/c2e508e7976519fcbc105e390e3688848bf87989

packages/backend/src/api/controllers/tvl/DetailedTvlController.ts
sdlyy: https://github.com/l2beat/l2beat/pull/1776/commits/c2e508e7976519fcbc105e390e3688848bf87989

packages/backend/README.md
sdlyy: Fair catch https://github.com/l2beat/l2beat/pull/1776/commits/36dc613c2e0e1fbd2fa98cf8d781b84e5a79ca49

packages/backend/src/api/controllers/tvl/DetailedTvlController.ts
sdlyy: Yess we can
https://github.com/l2beat/l2beat/pull/1776/commits/e5a48d7e0833e68dce0fde495a2cbe44590fd7e2

sz-piotr (approved):

PR #1780, opened 2023-08-09
Lz discovery

michalsidzej:


packages/discovery/src/discovery/handlers/user/ArrayHandler.ts
```diff
@@ -60,8 +60,9 @@ export class ArrayHandler implements Handler {
     const resolved = resolveDependencies(this.definition, previousResults)
 
     const value: ContractValue[] = []
+    const startIndex = resolved.startIndex
     const maxLength = Math.min(resolved.maxLength, resolved.length ?? Infinity)
-    for (let i = resolved.startIndex; i < maxLength; i++) {
+    for (let i = startIndex; i < maxLength + startIndex; i++) {
```
mateuszradomski, line 65: `maxLength + startIndex` feels wrong, maybe it's worth explaining why it's even valid?

mateuszradomski (approved):

PR #1781, opened 2023-08-10
L2 Assets Frontend: Mobile designs

mateuszradomski:
All the work up to this point was only on Desktop, this PR makes the frontend mobile friendly.


mateuszradomski:
reg-viz triggers for some components because I've unified the margin between all elements in the summary, I'm aware of this change.

sdlyy (changes requested):
Overall GJ, will do second round later on.

Also nitpick: can we do the sliding animation a bit slower? Just a bit - I mean, it is still 'to be discussed' but now I feel like I'm in a F1 Race car

packages/frontend/src/components/SlideCard.tsx
```diff
@@ -0,0 +1,73 @@
...
+            <div className="from sticky inset-x-0 top-0 h-4 bg-gradient-to-b from-gray-100 via-gray-100/60 to-transparent dark:from-zinc-800 dark:via-zinc-800/60" />
+            {props.children}
+            <div className="sticky inset-x-0 bottom-0 h-8 bg-gradient-to-t from-gray-100 via-gray-100/60 to-transparent dark:from-zinc-800 dark:via-zinc-800/60" />
+          </div>
+        </div>
+      </div>
+      <div className="flex h-[12%] w-full items-center justify-center bg-gray-200 dark:bg-[#373B41] ">
+        <div className="flex h-[6%] items-center justify-center rounded-md border border-pink-900 p-[6%] px-4">
+          <SlideCardToggle type="close"> Close the overlay </SlideCardToggle>
+        </div>
```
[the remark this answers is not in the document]
sdlyy: Also, maybe we could try something else than `Close the overlay` - `Close tokens`? I don't know tbh, something more related to the actual slide card

packages/frontend/src/scripts/configureSlideCard.ts
```diff
@@ -0,0 +1,33 @@
...
+  if (slideCards.length === 0) {
+    return
+  }
+
+  for (const slideCard of slideCards) {
+    const { $ } = makeQuery(slideCard)
+    const openButton = $('.SlideCard-Toggle[data-role="open"]')
+    const closeButton = $('.SlideCard-Toggle[data-role="close"]')
+
+    const content = $('.SlideCard-Content')
```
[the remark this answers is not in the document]
mateuszradomski: That's actually Tomek's suggestion, as he said that he would prefer to do this that way, it was just a `querySelector` before. I'm going to resolve this conversation but feel free to re-open

packages/frontend/src/components/header/TvlSummary.tsx
```diff
@@ -35,14 +35,16 @@ export function TvlSummary(props: TvlSummaryProps) {
     : []
 
   return props.detailedTvlEnabled ? (
-    <div className="w-[30%] p-4 md:rounded-lg md:bg-gray-100 md:px-6 md:py-5 md:dark:bg-zinc-800">
-      <div className="flex h-1/2 flex-col gap-2">
-        <div className="text-xs text-gray-500 dark:text-gray-600">
+    <div className="w-full bg-gray-100 p-4 dark:bg-zinc-800 md:w-[30%] md:rounded-lg md:px-6 md:py-5">
+      <div className="md:justify-normal mb-0 flex w-full justify-between md:mb-4 md:flex-col md:gap-2">
+        <span className="text-lg font-bold text-white text-gray-500 dark:text-gray-600 md:text-xs md:font-normal">
```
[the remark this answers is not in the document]
mateuszradomski: Text is white without this color definition and it needs to be gray

packages/frontend/src/components/SlideCard.tsx
```diff
@@ -0,0 +1,80 @@
...
+  closeButtonText?: string
+  children: ReactNode
+}
+
+function SlideCardContent(props: SlideCardContentProps) {
+  return (
+    <div className="SlideCard-Content transition-300 fixed left-0 top-0 bottom-0 z-999 flex h-[100dvh] w-full translate-y-full flex-col whitespace-normal bg-black/50 transition-transform ease-out">
+      <div className="h-[10%]" />
+      <div className="flex h-[90%] flex-col rounded-t-2xl bg-gray-100 dark:bg-zinc-800">
+        <h2 className="m-4 mb-0 text-3xl font-bold">{props.title}</h2>
```
sdlyy, line 64: This guy's beefy, I feel like maybe we could get rid of that title? Wdyt?

image

mateuszradomski: I don't really mind it, otherwise the visual noise it too great I think with all the different coin logos and also without the list just kinda starts weirdly

packages/frontend/src/components/SlideCard.tsx
```diff
@@ -0,0 +1,82 @@
...
+        <div className="relative m-4 flex grow flex-wrap">
+          <div className="absolute inset-0 overflow-auto scroll-smooth">
+            <div className="from sticky inset-x-0 top-0 h-4 bg-gradient-to-b from-gray-100 via-gray-100/60 to-transparent dark:from-zinc-800 dark:via-zinc-800/60" />
+            {props.children}
+            <div className="sticky inset-x-0 bottom-0 h-8 bg-gradient-to-t from-gray-100 via-gray-100/60 to-transparent dark:from-zinc-800 dark:via-zinc-800/60" />
+          </div>
+        </div>
+      </div>
+      <div className="flex h-[12%] w-full items-center justify-center bg-gray-200 dark:bg-[#373B41] ">
+        <div className="flex h-[6%] items-center justify-center rounded-md bg-pink-900 p-[6%] px-4">
```
sdlyy, line 74:
```suggestion
      <div className="flex w-full items-center justify-center bg-gray-100 pb-4 dark:bg-zinc-800">
        <div className="flex items-center justify-center rounded-md bg-pink-900 px-6 py-3 text-center text-base font-medium text-white">
```

Last thing to check out how it looks like. That's all from me for today 🚀

sdlyy (approved):

sdlyy:
Double check mobile behavior on different scaling factors, please.
Also, it would be great to let other ppl take a look and see what's up with their opinions since my design taste - well - If it was up to me, I would be fine with plain HTML and only a sprinkle of CSS 😆

packages/frontend/src/components/SlideCard.tsx
mateuszradomski: The colors do not really fit, and the padding is a good idea, I've added that

PR #1786, opened 2023-08-14
Update discovery

lucadonnoh:
- **base**: changed ProxyAdmin structure, now it's a 2/2 multisig composed of the OptimismMultisig and the BaseMultisig
- **beamer**: added whitelisted liquidity providers to permission section
- **cbridge**: updated pausers and governors, LEFT TO DO: fix handler to get the pausers based on the enum (which is really a uint8!!) and see how to join them with the other ones (there is a FIXME left in the code, look for it)
- **deversifi**: big implementation upgrade, i'll leave it for another PR, just ignored a value for now
- **linea**: implementation upgrade, i'll leave it for another PR
- **orbit**: owners update
- **polygon-plasma**: validator set update
- **polynetwork**: added a lock proxy
- **stargate**: chain paths length update
- **starknet**: max total balance update

please fix **cbridge**!


packages/config/src/layer2s/rhinofi.ts
```diff
@@ -93,12 +93,7 @@ export const rhinofi: Layer2 = {
     dataAvailability: {
       ...RISK_VIEW.DATA_EXTERNAL_DAC,
       sources: [
-        {
-          contract: 'StarkExchange',
-          references: [
-            'https://etherscan.io/address/0x1c3A4EfF75a287Fe6249CAb49606FA25659929A2#code#F34#L183',
-          ],
-        },
+        // TODO: add reference to new StarkExchange implementation!
```
michalsidzej, line 96: Should this stay here?

lucadonnoh: yes, as i said in the PR description rhino (= deversifi) upgraded the contracts and i will address it in another PR, that reference has been removed otherwise it breaks the build and has been replaced with a TODO

michalsidzej (approved):
LGTM

PR #1795, opened 2023-08-16
Fix non-unique OVM gradientId

mateuszradomski:
Because the `gradientId` was non-unique, on different pages then the main - like Archived/Upcoming - the icon became invisible on light mode


maciekzygmunt (approved):

mateuszradomski:
Fixed in #1796

PR #1802, opened 2023-08-17
Assess OP stack transaction count

michalsidzej:
Resolves L2B-2124

### DIFF
--- a/packages/config/src/layer2s/common/assessCount.ts
+++ b/packages/config/src/layer2s/common/assessCount.ts
@@ -2,5 +2,9 @@ import { AssessCount } from '../types'
 
 export const subtractOne: AssessCount = (count) => count - 1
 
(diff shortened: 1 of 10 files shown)


packages/config/src/layer2s/common/assessCount.ts
```diff
@@ -0,0 +1,6 @@
+import { AssessCount } from '../types'
+
+export const subtractOne: AssessCount = (count) => count - 1
+
+export const subtractOneAfterBlock: (blockNumber: number) => AssessCount = () =>
```
lucadonnoh, line 5: can you explain me this one? 😅

lucadonnoh (approved):

packages/config/src/layer2s/common/assessCount.ts
michalsidzej: This is a helper function, returning an AssessCount function. I'll rewrite this to be a bit more readable.

PR #1803, opened 2023-08-17
Fix missing Ethereum among tokens to query for

sdlyy:
Resolves L2B-2143

## What's changed
- Do not skip missing addresses
- Handle ETH native coin within chart's token list
- Added deep fallback if token has no address (temp fix since if address is there but trustwallet has no icon for such token we still display default HTML no-image fallback)


michalsidzej (commented):
Please mark the unmergable PRs as draft

sdlyy:
Fallbacks would look like that:
- Desktop
  image
- Mobile
mobile tokens - no address

michalsidzej (approved):
Other than the fact, that external USDC has no icon (I guess it's a known issue), LGTM
Screenshot 2023-08-22 at 14 50 59

PR #1805, opened 2023-08-18
Fix tooltips on mobile

maciekzygmunt:
Resolves L2B-2131

This PR change that tooltips now show without the user being redirected to the project page.


torztomasz:
I think it's a correct solution however we still have something like this in code that was used to fix this problem.
```tsx
    {
      name: 'Risks',
      tooltip: 'Risks associated with this project.',
      minimalWidth: true,
      noHrefMobile: true,
      alignCenter: true,
      getValue: (project) => <RosetteCell riskValues={project.riskValues} />,
    },
```

We should probably get rid of noHrefMobile field.

packages/frontend/src/scripts/configureTooltips.ts
```diff
@@ -95,6 +95,12 @@ export function configureTooltips() {
 
     element.addEventListener('click', (e) => {
       e.stopPropagation()
+      const isDisabledOnMobile = element.getAttribute(
+        'data-tooltip-mobile-disabled',
+      )
```
torztomasz, line 100: Extract it outside event listener. Best would be 85 line I guess.

Also the best would be to pass isDisabledOnMobile to `show` function.

torztomasz (changes requested):

torztomasz (approved):

PR #1808, opened 2023-08-21
Add flagTrueValues/flagFalseValues to ArrayFromOneEvent Handler

mateuszradomski:
This PR adds two new configuration fields to the ArrayFromOneEventHandler, they allow you to specify which values are treated as true/false. The rest of the behavior of this handler is the same, only the way we determine true/false is changed but the old boolean true/false is still supported. One might use these fields in the case where there are more then two states represented by a boolean, something like an enum flag that is represented by a numeral value. For more information look at the usage inside cBridge


packages/backend/discovery/cbridge/ethereum/config.jsonc
```diff
@@ -72,12 +72,21 @@
...
         },
-        "pausers": {
-          "type": "arrayFromOneEventWithArg",
+        "pausersPauseOnly": {
+          "type": "arrayFromOneEventReinterpreted",
+          "event": "event PauserUpdated(address account, uint8 role)",
+          "valueKey": "account",
+          "arg": "role",
+          "argsTrue": ["1"],
+          "argsFalse": ["0", "2"]
```
sz-piotr, line 81: Can we add options to the old handler instead?

sz-piotr (approved):

PR #1815, opened 2023-08-22
Add iconUrl to token defintion

mateuszradomski:
Resolves L2B-2053

This PR introduces a new optional field into the `Token` type `iconUrl`. It's value is automatically fetched using Coingecko's API for canonical tokens (currently, external and native tokens require manual addition). We decided to use Coingecko as a source of token icons since we already store a Coingecko ID for each tokens and resources like [trustwallet](https://github.com/trustwallet/assets) while great are very often incomplete. This allows us to make sure that we are always serving as high quality icons as Coingecko.
The entire `tokenList.json` has been refetched from scratch to add new fields and populate the `iconUrl` field.


packages/shared/src/services/coingecko/CoingeckoClient.ts
```diff
@@ -49,6 +50,20 @@ export class CoingeckoClient {
...
+      localization: 'false',
+      tickers: 'false',
+      market_data: 'false',
+      community_data: 'false',
+      developer_data: 'false',
+      sparkline: 'false',
+    })
+
+    const parsed = CoinMetadata.parse(data)
+    return parsed.image.large
```
antooni, line 64: I remember we had a conversation that we want "all the icon sizes", what has changed since then?

mateuszradomski: It's trivial to go from large -> small/thumb with a simple `.replace()`.

packages/shared-pure/src/types/Token.ts
```diff
@@ -17,4 +17,5 @@ export interface Token {
   chainId: ChainId
   type: ValueType
   premintHolderAddresses?: EthereumAddress[]
+  iconUrl?: string
```
antooni, line 20: One thing we need to keep in mind, is that conigecko likes to change CoingeckoId of a token from time to time, so probably in such event developer (when updating CoingeckoId) should also check whether the iconUrl still works after the change. Do you have an idea how can we enforce it?

antooni (approved):
Approved, please answer my comments

packages/shared-pure/src/types/Token.ts
mateuszradomski: We could try to write a test that will go through these iconUrls and try to fetch them. If it's `200` that's good and fail otherwise, but what if Coingecko replaces that url with a placeholder image, then the tests should fail but won't. Ideally we would fetch test image, hash the response and compare the hashes but then we have to store somewhere those hashes. Something to keep in mind for later I suppose.

PR #1818, opened 2023-08-23
Allow for key-less Coingecko tokens fetching

mateuszradomski:
Because the Coingecko free API Tier does not have a key, we need to support having this key not set. Otherwise everybody would need to share the same production key which is not safe or easy to manage

### DIFF
--- a/packages/config/scripts/tokens/add/index.ts
+++ b/packages/config/scripts/tokens/add/index.ts
@@ -10,6 +10,8 @@ import { getCanonicalTokens } from '../../../src'
 import { getEnv } from '../../checkVerifiedContracts/utils'
 import { getTokenInfo } from './getTokenInfo'
 
+const API_KEY_ENV_FALLBACK = '_FALLBACK'
+
 async function main() {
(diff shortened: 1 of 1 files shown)


packages/config/scripts/tokens/add/index.ts
```diff
@@ -10,14 +10,18 @@ import { getCanonicalTokens } from '../../../src'
...
+
 async function main() {
   const [address, category] = handleCLIParameters()
   if (!address) {
     return
   }
 
   const http = new HttpClient()
-  const coingeckoClient = new CoingeckoClient(http, getEnv('COINGECKO_API_KEY'))
+  const apiKeyEnv = getEnv('COINGECKO_API_KEY', API_KEY_ENV_FALLBACK)
```
antooni, line 22: wouldn't it be better to just do `process.env.COINGECKO_API_KEY`? it can resolve to undefined

antooni (changes requested):

antooni (approved):

PR #1822, opened 2023-08-23
Update tvl tooltip text

maciekzygmunt:
Resolves L2B-2216


packages/frontend/src/utils/project/getProjectTvlTooltipText.ts
```diff
@@ -0,0 +1,48 @@
...
+  const hasExternal = config.tokenList?.some(
+    (token) => token.type === ValueType.EBV,
+  )
+  const hasNative = config.tokenList?.some(
+    (token) => token.type === ValueType.NMV,
+  )
+  if (
+    (!hasCanonical && !hasExternal && !hasNative) ||
+    (hasCanonical && !hasExternal && !hasNative)
+  ) {
```
michalsidzej, line 15:
```suggestion
  if (!hasExternal && !hasNative) {
```
The hasCanonical can be either true or false so we can skip it

packages/frontend/src/utils/project/getProjectTvlTooltipText.ts
```diff
@@ -0,0 +1,48 @@
...
+    if (hasNative) {
+      sentence += ' and '
+    } else if (hasCanonical) {
+      sentence += ' '
+    }
+  }
+
+  if (hasNative) {
+    sentence += 'natively minted '
+  }
```
michalsidzej, line 43: Instead of this logic try: create an array or types, e.g. ['canonically bridged', 'natively minted'], then use `languageJoin` function (join with `, `, but `and` for last join).

michalsidzej (changes requested):
GJ. We can improve the logic a tiny bit in `getProjectTvlTooltipText`

packages/frontend/src/utils/project/getProjectTvlTooltipText.ts
maciekzygmunt: d2d382d

packages/frontend/src/utils/utils.ts
```diff
@@ -30,3 +30,16 @@ export function renderInlineMarkdown(markdown: string) {
 
   return markdownProcessor.renderInline(markdown)
 }
+
+export function languageJoin(items?: string[]) {
```
michalsidzej, line 34: Please use this function in `getContractSections.ts` as well

PR #1823, opened 2023-08-23
Rename optimistic chain to optimium

torztomasz:
Resolves L2B-2232

Instead of hiding this behind feature flag - let's merge as soon as we want it to be live. Hiding behind feature flag will just consume more time and add useless checks in code which we would have to remove after all changes related to the updates for the L2 Definition will be released.


maciekzygmunt (approved):

lucadonnoh (approved):
all the references seems to be updated

PR #1827, opened 2023-08-25
Refactor to use discovery-types package

michalsidzej:
Resolves L2B-2174

Context: we want to move `@l2beat/discoverry` package to https://github.com/l2beat/tools repo. This requires extracting discovery types so they can be used in all the relevant packages. This PR is the last step before moving discovery package.


lucadonnoh:
@michalsidzej could you give a small description for the reason behind this change? just curious :p

michalsidzej:
> @michalsidzej could you give a small description for the reason behind this change? just curious :p

@lucadonnoh provided more context in PR description

packages/discovery/src/discovery/config/RawDiscoveryConfig.ts
```diff
@@ -12,7 +8,7 @@ export type DiscoveryContract = z.infer<typeof DiscoveryContract>
 export const DiscoveryContract = z.object({
   ignoreDiscovery: z.optional(z.boolean()),
   proxyType: z.optional(ManualProxyType),
-  ignoreInWatchMode: z.optional(z.array(z.string())),
+  ignoreInWatchMode: z.z.optional(z.array(z.string())),
```
sdlyy, line 11: Does it provide some form of stronger validation - the double z? 😄

sdlyy (approved):
Looks fine, I'll do the second run later on

packages/discovery/src/discovery/config/RawDiscoveryConfig.ts
michalsidzej: I wonder why does TS not shout about it

michalsidzej: 7948981e

PR #1831, opened 2023-08-25
Start refactoring L2 Assets backend to better fit requirements

mateuszradomski:
Resolves L2B-2146

Beforehand we assumed that each external token is going to be calculated as `(totalSupply - balanceOfPremintAccounts) * price` and all native tokens are going to be calculated in their own circulating supply fashion. This assumption got thrown out of the window when we decided to classify USDC on Arbitrum as a native token because it's of Omnichain origin. Now the requirement is to calculate token TVL with three formulas 

- locked formula
- total supply formula
- circulating supply formula

And the updaters that handle each of these formulas shouldn't be parametrized on a given asset type (like external or native). In this PR, change the ArbitrumEBVUpdater to TotalSupplyUpdater, it now handles all tokens with TVL calculated as `totalSupply * price` - both native and external. The work done here is not the final look of the system, but it's a small and calculated step in the right direction that allows us to add more tokens to chains like Arbitrum and Optimism before L2 Assets release.


packages/backend/src/core/reports/getTotalSupplyFormulaConfigHash.ts
```diff
@@ -1,19 +1,18 @@
...
+const LOGIC_VERSION = 0
 
-export function getEBVConfigHash(tokens: Token[]) {
-  return hashJson([extractEBVTokens(tokens), EBV_LOGIC_VERSION])
+export function getTotalSupplyFormulaConfigHash(tokens: Token[]) {
+  return hashJson([extractTokens(tokens), LOGIC_VERSION])
 }
 
-function extractEBVTokens(tokens: Token[]) {
+function extractTokens(tokens: Token[]) {
```
antooni, line 9: maybe we should add an assert here to be sure that there are only totalSupply tokens?

packages/backend/src/core/assets/CirculatingSupplyFormulaUpdater.ts
```diff
@@ -64,9 +75,8 @@ export class NMVUpdater implements AssetUpdater {
 
   async start() {
     const known = await this.reportStatusRepository.getByConfigHash(
-      this.configHash,
-      this.getChainId(),
-      'NMV',
+      this.getConfigHash(),
+      this.chainId,
```
antooni, line 79: can we also get rid of the chainId?

packages/backend/src/core/reports/createTotalSupplyFormulaReports.ts
```diff
@@ -1,15 +1,14 @@
...
 
-import { BalanceRecord } from '../../peripherals/database/BalanceRepository'
+import { CirculatingSupplyRecord } from '../../peripherals/database/CirculatingSupplyRepository'
 import { PriceRecord } from '../../peripherals/database/PriceRepository'
 import { ReportRecord } from '../../peripherals/database/ReportRepository'
 import { TotalSupplyRecord } from '../../peripherals/database/TotalSupplyRepository'
 import { BalancePerProject, createReport } from './createReport'
 
-export function createEBVReports(
+export function createTotalSupplyFormulaReports(
```
antooni, line 9: This one should be named differently, it is used in a formula agnostic way right now

packages/backend/src/core/assets/CirculatingSupplyFormulaUpdater.ts
```diff
@@ -127,10 +142,26 @@ export class NMVUpdater implements AssetUpdater {
...
+    )
+
+    const native = await this.reportRepository.getByTimestampAndPreciseAsset(
       timestamp,
       this.getChainId(),
       'NMV',
     )
+
+    const all = canonical.concat(external).concat(native)
+    return all.filter((t) => this.tokens.some((m) => m.id === t.asset))
```
antooni, line 165: It is good for now, please create a TODO to refactor it

packages/backend/src/core/reports/getTotalSupplyFormulaConfigHash.ts
```diff
@@ -1,19 +1,18 @@
 import { hashJson, Token } from '@l2beat/shared-pure'
 
-const EBV_LOGIC_VERSION = 0
+const LOGIC_VERSION = 0
 
-export function getEBVConfigHash(tokens: Token[]) {
-  return hashJson([extractEBVTokens(tokens), EBV_LOGIC_VERSION])
+export function getTotalSupplyFormulaConfigHash(tokens: Token[]) {
```
antooni, line 5: rename this function so it reads more generic

packages/frontend/src/build/config/config.local.ts
```diff
@@ -8,7 +8,7 @@ export function getLocalConfig(): Config {
       ...common.features,
       // The local backend doesn't support activity
       activity: false,
-      detailedTvl: false,
+      detailedTvl: true,
```
antooni, line 11: should it be enabled?

packages/frontend/src/build/api/sanityCheck.ts
```diff
@@ -77,6 +77,7 @@ export function checkIfDelayedTvl(
   now: UnixTime,
 ) {
   const delayedProjects = allProjects
+
```
antooni, line 80: not needed space

mateuszradomski:
Moved to #1851 and #1854

PR #1836, opened 2023-08-25
Update Axelr / Satellite multichain discovery

luke7211:
This PR attempts to:
- update multichain discovery for some AxelR chains (not all) A list of all chains and gateways: https://docs.axelar.dev/resources/mainnet
- pulls token lists (where possible) from gateways. In some cases discovery does not work (block range too wide when querying events)

Tokens are fetched to recognize omnichain tokens handled by AxelR/Satellite/Squid


michalsidzej (commented):
Please provide list of chains that do not wor k (because of block range too wide when querying events). This should be fixable by changing RPC endpoints.

lucadonnoh:
now the discovery fetches token addresses as well: 
image

lucadonnoh (approved):

lucadonnoh:
@luke7211 @sz-piotr what do we do with this PR?

PR #1838, opened 2023-08-27
Prune zero-value reports

sdlyy:
Resolves L2B-2088

## What's changed

- No more zeroed values across repots
- Detailed controller is aware of potential data gaps due to lack of zeroed reports and fills the timeframes beforehand
- Data migration to get rid of zeroed values across reports
- insert filtering for zeroed reports

## How to test it?
Run staging dump locally and start local backend and frontend local variant. Data should be slim, charts will work as if no changes were made


packages/backend/src/core/reports/AggregatedReportUpdater.ts
```diff
@@ -70,7 +71,9 @@ export class AggregatedReportUpdater {
       timestamp,
     )
 
-    await this.aggregatedReportRepository.addOrUpdateMany(aggregatedReports)
+    const nonZeroReports = aggregatedReports.filter((x) => x.usdValue > 0n)
+
+    await this.aggregatedReportRepository.addOrUpdateMany(nonZeroReports)
```
sz-piotr, line 76: What if the value was != 0 but changed to zero?

packages/backend/src/core/assets/CBVUpdater.ts
```diff
@@ -105,7 +105,9 @@ export class CBVUpdater implements AssetUpdater {
     // TODO(radomski): This really needs to be refactored
     reports = filterOutNVMReports(reports)
 
-    await this.reportRepository.addOrUpdateMany(reports)
+    await this.reportRepository.addOrUpdateMany(
+      reports.filter((r) => r.amount > 0),
+    )
```
sz-piotr, line 110: Same concern: was != 0 but changed to zero.

packages/backend/src/core/assets/ArbitrumEBVUpdater.ts
```diff
@@ -126,7 +126,10 @@ export class ArbitrumEBVUpdater implements AssetUpdater {
       this.getProjectId(),
       this.getChainId(),
     )
-    await this.reportRepository.addOrUpdateMany(reports)
+
+    await this.reportRepository.addOrUpdateMany(
+      reports.filter((r) => r.amount > 0),
```
sz-piotr, line 131: Same concern: was != 0 but changed to zero.

sz-piotr (changes requested):
Maybe we should delete reports from a specific time and only then insert new ones (all in a transaction) instead of a simple adding of filtered ones?

packages/backend/src/peripherals/database/ReportRepository.ts
```diff
@@ -63,25 +63,29 @@ export class ReportRepository extends BaseRepository {
     return rows.map(toRecord)
   }
 
-  async addOrUpdateMany(reports: ReportRecord[]) {
+  async addOrUpdateMany(
+    reports: ReportRecord[],
+    where: { timestamp: UnixTime; chainId: ChainId },
+  ) {
+    if (reports.length === 0) {
+      return 0
```
sz-piotr, line 71: we still need to remove records from the database in this case

packages/backend/src/peripherals/database/AggregatedReportRepository.ts
```diff
@@ -112,23 +112,33 @@ export class AggregatedReportRepository extends BaseRepository {
     return row ? toRecord(row) : undefined
   }
 
-  async addOrUpdateMany(reports: AggregatedReportRecord[]) {
+  async addOrUpdateMany(
+    reports: AggregatedReportRecord[],
+    where: { timestamp: UnixTime },
+  ) {
+    if (reports.length === 0) {
+      return 0
```
sz-piotr, line 120: we still need to remove records from the database in this case

sz-piotr (changes requested):

packages/backend/src/peripherals/database/ReportRepository.ts
sdlyy: https://github.com/l2beat/l2beat/pull/1838/commits/65b872dba0346c18c980447d4e516c3ebbd9f1ed

packages/backend/src/peripherals/database/AggregatedReportRepository.ts
sdlyy: https://github.com/l2beat/l2beat/pull/1838/commits/65b872dba0346c18c980447d4e516c3ebbd9f1ed

sdlyy:
Works locally - ignore unsynced data, please (it more or less proves that we actually fill the gaps ;p)


image

image

image

sz-piotr (approved):

antooni (changes requested):
wait

sdlyy:
Merging postponed for some time due to l2assets things

PR #1842, opened 2023-08-28
Change USDC -> USDC.e for canonically bridged on arbitrum

maciekzygmunt:
Resolves L2B-2142


mateuszradomski (approved):
Leave a message as a comment for any future code readers that this is a hack and a better solution should be implemented.

PR #1844, opened 2023-09-01
L2Warsaw is over!

mateuszradomski:
Since L2Warsaw and Gitcoin is over, disable the navbar.

### DIFF
--- a/packages/frontend/src/build/config/common.ts
+++ b/packages/frontend/src/build/config/common.ts
@@ -28,7 +28,7 @@ export const common: Omit<Config, 'backend'> = {
     stages: true,
     buildAllProjectPages: false,
     upcomingRollups: true,
-    multisigReport: true,
+    multisigReport: false,
   },
   layer2s,
   bridges,


packages/frontend/src/build/config/common.ts
```diff
@@ -28,7 +28,7 @@ export const common: Omit<Config, 'backend'> = {
     stages: true,
     buildAllProjectPages: false,
     upcomingRollups: true,
-    multisigReport: true,
+    multisigReport: false,
```
michalsidzej, line 31: This does not look like disabling the topbar

mateuszradomski: Check out the preview. If I don't understand something we'll talk :+]

michalsidzej (approved):

michalsidzej (commented):
Wait, but don't we want to keep the multisigReport topbar?

mateuszradomski:
> Wait, but don't we want to keep the multisigReport topbar?

The code is removed already :(((

PR #1855, opened 2023-09-05
Add Manta Pacific (upcoming)

lucadonnoh:
Resolves L2B-2323

# Summary
Manta Pacific is an OP stack chain using Celestia as DA (= Optimium), built with zk applications in mind. The documentation is very confusing since they claim to be an L2 when they are not.

## Open question
- Should we point out anywhere that the documentation is misleading? Or our the "Optimium" designation is enough? The alternative is to add a small phrase in the description to warn users.


sz-piotr (approved):
> Should we point out anywhere that the documentation is misleading? Or our the "Optimium" designation is enough? The alternative is to add a small phrase in the description to warn users.

I think that since this is upcoming we don't need to do anything. The description doesn't really claim anything about L2s and the type in the code is something we should refactor anyway but as part of a different PR

lucadonnoh:
the description doesn't claim anything about L2s because i've changed it :p

DanielZhangReal:
@lucadonnoh

We have released the Manta Pacific Mainnet. How to modify the status of Manta Pacific on L2Beat to Active?
Do you need any information from us? Thanks

lucadonnoh:
i'll put the project under review and once reviewed we will show full info

PR #1860, opened 2023-09-06
Bring back the pretty token modal

mateuszradomski:
Resolves L2B-2294

Bring back the beautiful top 15 token view with icons and full names. The expanding of that to everything we have is going to be addressed in stage 2 of L2 Assets.


sdlyy (approved):
ACs met, good to have that thingy back.

Nitpick: Maybe that link to google forms for requesting new tokens can be tweaked a bit? Some more margin/padding - whatever. It merges itself with the actual tokens list in some way.

PR #1873, opened 2023-09-07
Fix L2Assets config

antooni:
There was an issue with `sinceTimestamp` in some of the tokens config. This PR aims to solve it and add context in the comment of Token.ts


packages/backend/src/core/totalSupply/CirculatingSupplyUpdater.ts
```diff
@@ -70,7 +71,9 @@ export class CirculatingSupplyUpdater {
   }
 
   async update() {
-    const from = this.clock.getFirstHour()
+    const from = this.clock.getFirstHour().gt(this.minTimestamp)
```
adamiak, line 74: I guess a test would be useful, but approving.

sdlyy (approved):

adamiak (approved):

PR #1888, opened 2023-09-13
Update node to 18

michalsidzej:
Updating node version to 18 as 16 reached EOL. Tested locally both backend with all modules enabled and frontend. Works like a charm.

After this PR is merged, this codebase will not work with Node 16.
Make sure you install new version of node: 

```nvm install 18
nvm use 18
```
or if using `fnm` (you should check it out, it's **rusty**)
```fnm install 18
fnm use 18
fnm default 18
```

The storybook deployment does not work right now, as the `Node` version set in vercel project is set to 16.x. In `l2beat-production` and `l2beat-staging` this version is overridden, so it works. In storybook the root directory is set to `packages/frontend`, so it does not know about the `engine` version. 
Screenshot 2023-09-13 at 11 07 30

I will bump the `Node` version in `Vercel - @l2beat/storybook` after this PR is approved.


sz-piotr (approved):

sz-piotr:
Maybe you could specify the engine field in frontend too? For storybook

PR #1891, opened 2023-09-13
Refactor multicall and multicall users

sz-piotr:


packages/backend/src/core/balances/BalanceProvider.ts
```diff
@@ -1,22 +1,216 @@
...
+      native.map((balanceQuery) =>
+        this.ethereumClient.getBalance(balanceQuery.holder, blockNumber),
+      ),
+    )
+
+    let nonNativeIndex = 0
+    let nativeIndex = 0
+    return balanceQueries.map(({ holder, assetId }): BalanceRecord => {
+      if (this.isNativeCoin(assetId)) {
+        const response = nativeResponses[nativeIndex++]
```
antooni, line 107: This value can be undefined

antooni: If from 84 line should be used there

packages/backend/src/core/balances/BalanceProvider.test.ts
```diff
@@ -7,34 +7,37 @@ import {
...
-import { MulticallClient } from '../../../peripherals/ethereum/MulticallClient'
-import { EthereumBalanceProvider } from './EthereumBalanceProvider'
+import { EthereumClient } from '../../peripherals/ethereum/EthereumClient'
+import { MulticallClient } from '../../peripherals/ethereum/MulticallClient'
+import { BalanceProvider } from './BalanceProvider'
 
-describe(EthereumBalanceProvider.name, () => {
-  describe(EthereumBalanceProvider.prototype.fetchBalances.name, () => {
+describe(BalanceProvider.name, () => {
+  describe(BalanceProvider.prototype.fetchBalances.name, () => {
```
antooni, line 15: add the test for native and nonNative split

packages/backend/src/core/balances/BalanceProvider.test.ts
```diff
@@ -45,13 +48,16 @@ describe(EthereumBalanceProvider.name, () => {
...
+          assetId: AssetId.USDC,
           holderAddress: holderB,
           balance: 420n,
           timestamp,
           chainId: ChainId.ETHEREUM,
         },
       ])
     })
+
+    it.skip('handles native assets with multicall')
```
antooni, line 60: should it be there?

packages/backend/src/peripherals/ethereum/MulticallClient.ts
```diff
@@ -1,30 +1,16 @@
...
+  MulticallConfigEntry,
+  MulticallRequest,
+  MulticallResponse,
+} from './types'
 
 export class MulticallClient {
-  constructor(private readonly ethereumClient: EthereumClient) {}
+  constructor(
+    private readonly ethereumClient: EthereumClient,
+    private readonly config = ETHEREUM_MULTICALL_CONFIG,
```
antooni, line 12: I do not like this default param, maybe a static method?

packages/backend/src/peripherals/ethereum/MulticallClient.ts
```diff
@@ -1,30 +1,16 @@
...
+import {
+  MulticallConfigEntry,
+  MulticallRequest,
+  MulticallResponse,
+} from './types'
 
 export class MulticallClient {
-  constructor(private readonly ethereumClient: EthereumClient) {}
+  constructor(
+    private readonly ethereumClient: EthereumClient,
```
antooni, line 11: This should not be a scope of this PR, but I think that EthereumClient should have a notion of its ChainId, the same is with the multicall config, so we can run the assert that the chainIds match

packages/backend/src/peripherals/ethereum/MulticallConfig.ts
antooni, line 1: maybe all the files connected with multicall should be moved to multicall folder? because right now "ethereum" is a bit misleading because the MulticallConfig.ts includes configs for multiple chains

packages/backend/src/core/balances/BalanceProvider.ts
```diff
@@ -1,22 +1,219 @@
...
+  encode: encodeGetEthBalance,
+  decode: decodeGetEthBalance,
+}
+
+export class BalanceProvider {
+  constructor(
+    private readonly ethereumClient: EthereumClient,
+    private readonly multicallClient: MulticallClient,
+    private readonly chainId: ChainId,
+    private readonly nativeBalanceEncoding: NativeBalanceEncoding | undefined,
```
antooni, line 66: maybe this param can be stored in Multicall?

antooni: nativeBalanceEncoding

sz-piotr: It's actually only partially related. While it's true that the multicall contract is the vehicle by which we query for eth balance it doesn't feel right to me that the MulticallClient is responsible for this encoding and decoding.

I thought about it but came to the conclusion that ultimately it is the responsibility of the caller to encode and decode calls and so calls to multicall itself should be no different

packages/backend/src/peripherals/ethereum/MulticallConfig.ts
sz-piotr: Maybe just renaming ethereum to blockchain or evm would suffice. In any case I think it is out of scope of this refactoring.

packages/backend/src/peripherals/ethereum/MulticallClient.ts
sz-piotr: Might be a nice improvement for the future

packages/backend/src/peripherals/ethereum/MulticallClient.ts
sz-piotr: Removed

antooni (approved):

PR #1902, opened 2023-09-14
Support markdown on project page

sz-piotr:
Resolves L2B-2444

Because I wanted the markdown text to look at least somewhat presentable I wanted to increase it's line height. Line heights are in general broken on our frontend so I just selectively applied `leading-snug` whenever it was super bad. So this PR does change how some of the text looks on our page.


antooni:
Is it a desired behaviour that after setting part of the description to bold it also gets yellow color?
image

antooni:
I like the change, although I am missing some things:
- basic Storybook story showing all the markdown functionality
- what is the planned workflow? does it needs to be a responsibility of a component to display this markdown? maybe it can be enforced in some way? currently it is hard to know whether markdown is added to every important places & especially when filling a config one cannot be sure that this field supports markdown
- please provide the description for the PR, some parts are not clear to me: why you are adding this class "leading-snug" everywhere? how did you come up with which component to update? how can I test that components looks the same with new markdown logic?

PR #1903, opened 2023-09-14
Remove redundant feature flags

torztomasz:
Resolves L2B-2449


maciekzygmunt (approved):

maciekzygmunt (approved):

packages/frontend/src/build/config/common.ts
```diff
@@ -20,17 +20,13 @@ export const common: Omit<Config, 'backend'> = {
     banner: new Date() <= GITCOIN_BETA_ROUND_END,
     gitcoinOption: false,
     hiring: true,
-    hiringBadge: false,
+    hiringBadge: true,
```
sz-piotr, line 23: Why this change?

sz-piotr (changes requested):

packages/frontend/src/build/config/common.ts
torztomasz: Oh, my mistake, I thought I discarded it.

sz-piotr (approved):

PR #1908, opened 2023-09-15
Add Base tvl module

maciekzygmunt:
Resolves L2B-2391

This PR:

- implements TVL module for Base
- adds USDC to the token list


packages/backend/src/peripherals/ethereum/multicall/MulticallConfig.ts
```diff
@@ -53,6 +49,16 @@ export const OPTIMISM_MULTICALL_CONFIG: MulticallConfigEntry[] = [
   },
 ]
 
+export const BASE_MULTICALL_CONFIG: MulticallConfigEntry[] = [
+  {
+    sinceBlock: 5022,
+    batchSize: 150,
+    address: EthereumAddress('0xcA11bde05977b3631167028862bE2a173976CA11'),
+    encodeBatch: encodeMulticallV2,
+    decodeBatch: decodeMulticallV2,
```
sz-piotr, line 58: Btw since this is multicall3 we could add support for it, but not needed in this PR

packages/shared-pure/src/types/AssetId.ts
```diff
@@ -15,6 +15,7 @@ AssetId.USDT = AssetId('usdt-tether-usd')
 AssetId.USDC = AssetId('usdc-usd-coin')
 AssetId.USDC_ON_ARBITRUM = AssetId('arbitrum:usdc-usd-coin')
 AssetId.USDC_ON_OPTIMISM = AssetId('optimism:usdc-usd-coin')
+AssetId.USDC_ON_BASE = AssetId('base:usdc-usd-coin')
```
sz-piotr, line 18: Honestly not a fan of those. We aren't really reading them anywhere

sz-piotr (approved):

PR #1931, opened 2023-09-19
Update Chart's scale

antooni:
Currently the scale on most of the charts does not start from zero, which is a bit misleading, this PR aims to improve it.

Before:
image

After:
image


sz-piotr (changes requested):
I think this needs more discussion. Change is too drastric

PR #1934, opened 2023-09-19
Update cbridge discovery

adamiak:
Cbridge removed one pauser


adamiak:
It was just a test. Closing.

PR #1919, opened 2023-09-18
Update discovery

lucadonnoh:
- **arbitrum**: 
  - the canceller role in the L1ArbitrumTimelock contract has been changed from the Security Council to the UpgradeExecutor contract, which is owned by both the Security Council and the L1ArbitrumTimelock itself.
  - the security council has switched to a new address, what? [contract](https://etherscan.io/address/0x22ec545357162c342f643bddb2ed4c3fb6b42eb0#code) (check `perform` function), [tx](https://app.tokenflow.live/anytx/mainnet/0x7165ad22cc2bbca36019963ba1adff21c1d48444a3cfecf980f752aaa9594d94)
- **base**: multisig change
- **cbridge**: pausers list change
- **debridge**: implementation change, will open another PR
- **lzomnichain**: configuration change
- **nova**: similar change as arbitrum

PR #1937, opened 2023-09-19
update config with new arbitrum l1 security council; run discovery

DZGoldman:
With the [execution](https://etherscan.io/tx/0x7165ad22cc2bbca36019963ba1adff21c1d48444a3cfecf980f752aaa9594d94) of [AIP 6](https://www.tally.xyz/gov/arbitrum/proposal/108413626736577087081818577238162267924459697981830202052718122463860611528602) (activating security council elections), the L1 security council is a new contract.


lucadonnoh:
thanks for the PR! the change was already in progress here: #1919 

@DZGoldman why is the canceller now the UpgradeExecutor and not the SC directly? what's the advantage? it looks to me that the DAO still cannot cancel pending proposals, right?

DZGoldman:
@lucadonnoh 

> thanks for the PR! the change was already in progress here: https://github.com/l2beat/l2beat/pull/1919

ah, didn't see that, ty! feel to close this

> why is the canceller now the UpgradeExecutor and not the SC directly? what's the advantage? it looks to me that the DAO still cannot cancel pending proposals, right?

Correct, the functionality of giving the role to SC would be materially the same; we give it to the UpgradeExecutor for consistency and simplicity. I.e., (now) wherever a contract wants to give an affordance to the SC, it does so via granting it to UpgradeExecutor, making the UpgradeExecutor the single source of truth for upgrade rights.

lucadonnoh:
solved in the other PR

PR #1942, opened 2023-09-20
add: chainport

odedNea:
Hey L2beat,

We'd like to add chainport (https://www.chainport.io/) to L2Beat platform.
Few things I'm not sure about:
1. Do we need to specify our contract addresses under 'escrows' or under 'technology' -> 'contracts' -> 'addresses'?
2. Can we add contracts across different networks? as we enable bridging not only from eth, but from many other networks.
3. Could we add Cardano contracts as a source chain for TLV checks? as most of our TVL comes from this network.

Thanks in advanced!


odedNea:
Thanks @lucadonnoh 
Just to clarify:
TVL that is locked in non-Ethereum chains won't be counted in L2Beat?

lucadonnoh:
yes that's correct. we recently added general EVM support, meaning that in the close future we will also start tracking contracts on other EVM chains, but currently we don't plan to support other chains like Cardano

odedNea:
@lucadonnoh Thanks for the efforts you've put in editing this PR.
I've noticed a small diff error, fixed it and pushed the fix.
Can you please check and approve?

lucadonnoh:
we're fixing some issue with the CI and then we are ready to merge this. Also, I found the escrows from Vault 1 to 6 but I'm missing number 4, do you know where i can find it, if it exists? @odedNea

lucadonnoh (approved):

odedNea:
@lucadonnoh Here are 2 more vault addresses we're using:

- 0x894107B7b5051409f279E8300774B2f62Febe057
- 0xfE83475880d3592833249BAAacfEC5eD51E29D82

By the way - This info is available on our website at: https://www.chainport.io/transparency

PR #1960, opened 2023-09-25
Add Kinto (upcoming)

VSBDev:
Adding the upcoming project kinto.xyz


lucadonnoh:
moved to #1967 since i cannot push

PR #1973, opened 2023-09-26
Rename Zora

lucadonnoh:
Resolves L2B-2614


Huberacht (approved):
Great, now we can contact them regarding L2Days

PR #1984, opened 2023-09-28
Fix tablet view

torztomasz:
Resolves L2B-2481, L2B-2478


maciekzygmunt (changes requested):
### There is still an issue with this part of the page. Doesn't look good imo

image

antooni:
In the description please provide the screenshots of changes (before & after)

torztomasz:
@antooni @maciekzygmunt Good catch - although we decided with Adi to go with that. There is no other way to make it better than it is implemented currently.

maciekzygmunt (approved):

PR #1985, opened 2023-09-29
Update Kroma discovery with the dynamic array handler

mateuszradomski:
Resolves L2B-2646

This PR update the discovery package to 0.5.0 and uses the newly added dynamic array handler to fetch the validators in Kroma. The discovery has been performed on the same block number as it was done before (18213384) so only the fact that we now track validators are the thing that has changed.


antooni (commented):
Maybe we should add validators to the webpage?

lucadonnoh:
> Maybe we should add validators to the webpage?

it changes too often so it doesn't make sense atm, would have been cool in general

lucadonnoh (approved):

PR #1994, opened 2023-10-03
Add support for descriptions in diffHistory.md

mateuszradomski:
Resolves L2B-2567

Right now, you can add a description field to the diffHistory in the following way.

```
## Diff at Tue, 03 Oct 2023 08:28:18 GMT:

- author: Mateusz Radomski (<radomski.main@protonmail.com>)
- comparing to: master@03894908622bf9e8ff40d5177fd0b65776e38dc3

## Description

Hello, this is a test. I'm testing a single line.
Now I'm testing multiple lines.
This also is a friend of that line.

And now I should also be present in the description above and I will be.

```diff
....
```

Description will be preserved between `yarn discover` runs.


packages/backend/src/tools/updateDiffHistory.ts
```diff
@@ -45,14 +45,25 @@ async function updateDiffHistoryFile() {
   )
 
   if (diff.length > 0) {
-    const newHistoryEntry = generateDiffHistoryMarkdown(diff, mainBranchHash)
     const diffHistoryPath = `${discoveryFolder}/diffHistory.md`
+    const diskDiffHistory = readFileSync(diffHistoryPath, 'utf-8')
```
adamiak, line 49: Program breaks here if there is no existing `diffHistory.md` file. Please cover this scenario.

adamiak (changes requested):
I think this functionality would benefit from following additions:
- Automatically generate an empty "Description" section with text "Provide description of changes. This section will be preserved."
- Fix section hierarchy (via the amount of hash symbols). It should be:
  - `# Diff at` <-- one hash
  - `## Description` <-- two hashes
  - `## Watched changes` <-- let's add it for the diff section
  - update already existing diffHistory files to follow this schema.
- To find description to preserve, find text between `## Description` and the next `##`.

adamiak (approved):

PR #1996, opened 2023-10-03
Update morphism.ts and change category

jommi9:


lucadonnoh:
once the project will be live on mainnet we will be able to actually verify what's going on and properly decide the correct designation. My understanding is that the design is similar to Kroma's (https://l2beat.com/scaling/projects/kroma), which is currently listed as an optimistic rollup.

PR #2007, opened 2023-10-05
Diff source code of contracts during discovery

adamiak:
Resolves L2B-2516

When `yarn discover` is run, it downloads source code (for blockNumber from `main` branch) to `.code@<block_number>` folder. It removes any other such code folders, so there always will be only two: one `.code` and one `.code@<main_branch_block>`. 

Once downloaded, it compares two folders using `git diff` and if any difference is discovered, puts the summary to the `diffHistory.md` file.

Due to bump of discovery version, there are also some related changes in some files.


packages/backend/src/config/config.production.ts
```diff
@@ -176,6 +176,9 @@ export function getProductionConfig(env: Env): Config {
         {
           chainId: ChainId.ETHEREUM,
           rpcUrl: env.string('DISCOVERY_ETHEREUM_RPC_URL'),
+          rpcGetLogsMaxRange: env.optionalInteger(
+            'DISCOVERY_ETHEREUM_RPC_GETLOGS_MAX_RANGE',
```
mateuszradomski, line 180: We should add this env var to the read in this the section under `#-----DISCOVERY-----`

packages/backend/src/tools/updateDiffHistory.ts
```diff
@@ -27,6 +27,7 @@ async function updateDiffHistoryFile() {
   }
   const chainId = ChainId.fromName(chainName)
 
+  // Get discovered.json from main branch and compare to current
```
mateuszradomski, line 30: Since this script is getting more and more useful and the amount of features that it supports is only going up not down I'd extract the code that this comment is referencing and put it in a function called something like `compareDiscoveredWithMaster()`, this way it's easier to see what's going on. The same with the newly added code.

packages/backend/src/tools/updateDiffHistory.ts
```diff
@@ -36,14 +37,42 @@ async function updateDiffHistoryFile() {
...
   const diff = diffDiscovery(
-    discoveryFromMainBranch.contracts,
+    discoveryFromMainBranch?.contracts ?? [],
     curDiscovery.contracts,
     config,
   )
 
+  // To check for changes to source code,
+  // download sources for block number from main branch
+  let codeDiff = ''
```
mateuszradomski, line 50: This doesn't need to be defined here. Also it does not need to be a `let`, can easily be `const`.

packages/backend/src/tools/updateDiffHistory.ts
```diff
@@ -36,14 +37,42 @@ async function updateDiffHistoryFile() {
...
+  if (discoveryFromMainBranch) {
```
mateuszradomski, line 51: What if the `blockNumberFromMainBranch` is the same as the already downloaded one? Should we check if it already exists or always remove and download? One is faster the other is more robust. Discovery is already slow as it is.

packages/backend/src/tools/updateDiffHistory.ts
```diff
@@ -176,11 +230,21 @@ function generateDiffHistoryMarkdown(
     )
     result.push('')
   }
+
   result.push('## Watched changes')
   result.push('')
   result.push(discoveryDiffToMarkdown(diffs))
   result.push('')
 
+  if (codeDiff !== '') {
```
mateuszradomski, line 239: I don't like this compare against an empty string, assumes that we know what is going on outside of this function. Maybe it can be optional just like the `description`?

packages/backend/src/tools/updateDiffHistory.ts
```diff
@@ -36,14 +37,42 @@ async function updateDiffHistoryFile() {
...
+  // To check for changes to source code,
+  // download sources for block number from main branch
+  let codeDiff = ''
+  if (discoveryFromMainBranch) {
+    const root = `discovery/${projectName}/${chainName}`
+    // Remove any old sources we fetched before, so that their count doesn't grow
+    await rimraf(`${root}/.code@*`, { glob: true })
+
+    const blockNumberFromMainBranch = discoveryFromMainBranch.blockNumber
+    const cli = [
```
mateuszradomski, line 57: This feels bad but robust at the same time. WDYT about exposing and using the source code fetching API from discovery?

mateuszradomski (changes requested):

packages/backend/src/tools/updateDiffHistory.ts
adamiak: For simplicity, I'll keep current solution. We can refactor when we decide that this functionality is useful.

mateuszradomski (approved):
gj

PR #2018, opened 2023-10-09
Add BigQuery client

maciekzygmunt:
Resolves L2B-2688
Adds BigQuery Client


packages/shared-pure/src/utils/getMethodQuery.ts
antooni, line 1: add test

packages/shared-pure/src/utils/getTransferQuery.ts
antooni, line 1: add test

antooni (approved):

antooni (approved):

PR #2031, opened 2023-10-11
Fix /status/discovery returning Internal Server Error

mateuszradomski:
Resolves L2B-2780


adamiak (approved):

lucadonnoh (changes requested):
i would fix the issue before merging

lucadonnoh (approved):

PR #2036, opened 2023-10-12
Enforce/Migrate db timestamp columns to without timezone

aminlatifi:
Added migration to change timestamp with time zone column type to tim…stamp without time zone

Added a local eslint plugin to complain each time a table column of timestamp with timezone type is created

Resolve L2B-2779


aminlatifi:
@antooni Supported `activity` schema also.

packages/backend/.eslint/rules/db_ts_no_tz.js
```diff
@@ -0,0 +1,42 @@
...
+          if (
+            args.length !== 2 ||
+            !args[1].properties.some(
+              (prop) => prop.key.name === 'useTz' && prop.value.value === false,
+            )
+          ) {
+            context.report({
+              node,
+              message:
+                "The 'table.dateTime' method should be called with ('unix_timestamp', { useTz: false })",
```
antooni, line 33:
```suggestion
                "The 'dateTime' column should be without timezone, call it with '{ useTz: false }'",
```

antooni (approved):

PR #2038, opened 2023-10-12
add network: opbnb

huihzhao:
opBNB is a high-performance layer 2 solution for BNB Smart Chain, which is a blockchain network that supports smart contracts and decentralized applications. opBNB uses the OP Stack. opBNB aims to provide a scalable, secure, and low-cost platform for various use cases, such as gaming, decentralized exchanges, digital collectibles, and more.


lucadonnoh:
we currently only consider rollups on ethereum at the moment, but i'll leave the PR here to discuss it better with the team

huihzhao:
> we currently only consider rollups on ethereum at the moment, but i'll leave the PR here to discuss it better with the team

Thank you. I think opBNB is a good topic and has a good potential to expand the l2beat scope. And we are engineering for the lower cost and with big user and developer community.

PR #2039, opened 2023-10-12
eslint prettier support

aminlatifi:
Eslint error on prettier issues


aminlatifi:
By vscode format on save feature, it's not highly necessary

PR #2056, opened 2023-10-16
Add Scroll (full)

lucadonnoh:
Resolves L2B-2786

# Left to do
- add RPC for Activity
- contracts diagram (will leave this to Bartek to force him to take a look at Scroll as well 😛)

# How to review
- Read the page and see if you're able to understand what's going on
- Check whether all claims are backed by references
- Check whether the sections contain all the contracts from `discovered.json`
  - "[...]Fallback" contracts have been omitted on purpose since they are dummy contracts with the same address as L2 counterparts and are there only to recover funds sent incorrectly to them. Not displayed because they are a lot and would pollute the interface a lot, and also not informative.

I also added the state derivation section but left the node software requirement as 'UnderReview' in stages because i want to discuss with Bartek the details of it.


packages/backend/discovery/scroll/ethereum/diffHistory.md
```diff
@@ -1,3 +1,224 @@
+# Diff at Thu, 19 Oct 2023 08:22:39 GMT:
+
+- author: Luca Donno (<donnoh99@gmail.com>)
+- comparing to: master@0beeab609744b4e4247dca817daaa8caed24ab12
+
+## Description
+
+Provide description of changes. This section will be preserved.
```
antooni, line 8: Please update description

packages/config/src/layer2s/scroll.ts
antooni, line 1: please provide a more precise description
image

antooni (approved):

PR #2057, opened 2023-10-16
Adding Applicative Escape hatches

hydrogenbond007:


hydrogenbond007:
Starknet uses Applicative hatches as a security measure in situation of prover failure. 
https://community.starknet.io/t/starknet-escape-hatch-research/1108

lucadonnoh:
could you provide an example of an application on starknet that makes use of this mechanism? i’ve always been skeptical about its practicality

hydrogenbond007:
Yo sorry just saw this, I talked to the starknet team and they said they haven't worked on it much lately but this might be useful for [code context](https://github.com/Kelvyne/starknet-storage-proof-solidity/blob/shamir_trick/contracts/InterleavedPedersenHashZZ.sol)

And as for applicative, it means it depends on the bridge since mainnet addresses don't match L2 addresses in starknet and zk rollups, so it depends on the bridge contract and how they handle it.

lucadonnoh:
if a bridge that uses this mechanism ever pops out this is something that we can consider, but until then i believe the mechanism to be highly impractical

PR #2070, opened 2023-10-17
OP Mainnet state derivation section

lucadonnoh:
Reference: https://www.notion.so/l2beat/Source-available-nodes-review-c2f07682a07b4dda93e6a674cf0c329b

Resolves L2B-2818


packages/config/src/layer2s/common/stages/getStage.ts
```diff
@@ -39,7 +39,7 @@ const getBlueprint = (opts?: GetStageOptions) =>
           positive:
             'A source-available node exists that can recreate the state from L1 data.' +
             (opts?.rollupNodeLink
-              ? ` [View code](${opts.rollupNodeLink})`
+              ? ` [View code](${opts.rollupNodeLink}).`
```
antooni, line 42: I am not sure whether we need this dot at the end
image

antooni (changes requested):

packages/config/src/layer2s/common/stages/getStage.ts
lucadonnoh: yeah looks a bit weird, but there is a test failing so maybe we should fix that

antooni (approved):

PR #2072, opened 2023-10-18
Added svg support to hover render icon

aminlatifi:
Resovle L2B-2778
Resolve L2B-2794


packages/frontend/src/components/icons/symbols/CanonicalIcon.tsx
```diff
@@ -3,6 +3,10 @@ import React, { SVGAttributes } from 'react'
 
 import { Icon } from '../Icon'
 
+/*  IMPORTANT
+  If you change this file you need to update the following file too:
+  * packages/frontend/src/scripts/charts/styles.ts
```
torztomasz, line 8: Provide more context on which icon should be changed as `styles.ts` file does not tell which one should change too

torztomasz (approved):

PR #2078, opened 2023-10-18
Fix daily message in update monitor not taking version into account

michalsidzej:
Resolves L2B-2846

NOTE: I was unable to run it locally due to an outdated database. @antooni can you please take a look? Otherwise I'll download the DB during the night and test it tommorow.


adamiak (changes requested):
I'm not sure if this is the right approach, since it would cause constant re-runs of discovery of many projects on "master" when discovery starts returning new version. That's because old files on master would use old version until someone manually re-discovers them, which can take a long time.

A temporary solution is to re-run discovery (with --dev) on all projects when such a big change happens and commit their updated version in one big PR. I'll add a script to do it in one go.

Ultimately we must find a proper solution, but I'm not sure that this PR is the right approach.

michalsidzej:
Resolved by #2098

PR #2079, opened 2023-10-19
Add L2BEAT Socials to README

davidcardenasus:
Adds Twitter, Discord, and YouTube hyperlinks to README.

The Discord shield will not function correctly unless the widget setting is enabled withing the L2BEAT Discord server settings. https://shields.io/badges/discord


mateuszradomski:
Thank you for your contribution but we don't need these entries. I hate that your work is going to waste, so I'd suggest to create a issue before hand to see if the change you've created is needed.

PR #2080, opened 2023-10-19
First step into op stack templates

mateuszradomski:
Resolves L2B-2702


antooni:
Have you tested that the production preview has the data aligned with what is currently on master?

antooni (commented):
Code looks good, step in the right direction. What is left IMO is the one last check whether the output of the build works as expected (probably manually)

adamiak (approved):
Please create issues to:
* add unit tests for ProjectDiscovery.ts
* don't read config.jsonc (which is not necessary here - I'm aware it requires changing how invert works in tools)

lucadonnoh (approved):

PR #2082, opened 2023-10-19
Migrate timestamps column 

aminlatifi:
Whitelisted tables whose timestamp columns type must be migrated to without timezone.

Resolves L2B-2779


packages/backend/src/peripherals/database/migrations/080_timestamp_column_tz.ts
```diff
@@ -0,0 +1,62 @@
...
+      'column_name',
+      'data_type',
+    )
+    .where({
+      data_type: 'timestamp with time zone',
+      table_schema: 'public',
+    })
+    .whereIn('table_name', tableWhitelist)) as ColumnInfo[]
+
+  for (const column of columns) {
```
antooni, line 47: iterate over tableWhitelist instead of dynamically getting the columns, they are named the same, there is no need to be implicit there

aminlatifi: Done

packages/backend/src/peripherals/database/migrations/080_timestamp_column_tz.ts
```diff
@@ -0,0 +1,59 @@
...
+  'circulating_supplies',
+  'total_supplies',
+  'total_supplies_status',
+  'update_notifier',
+]
+
+export async function up(knex: Knex) {
+  for (const table_name of tableWhitelist) {
+    // get all columns that are of type timestamp with time zone
+    const columns = (await knex('information_schema.columns')
```
antooni, line 35: This line introduced bug previously, although now we have whitelist I would like the code to be even more explicit. Please change the `tableWhitelist` to `tablesToUpdate` and just iterate over them and on each step run the raw query to change the column type

packages/backend/src/peripherals/database/migrations/080_timestamp_column_tz.ts
```diff
@@ -0,0 +1,59 @@
...
+      // get list of indexes table is involved in
+      await knex.raw(
+        `ALTER TABLE "${table_schema}"."${table_name}" ALTER COLUMN ${column_name}
+        SET DATA TYPE timestamp without time zone USING ${column_name}::timestamp without time zone`,
+      )
+    }
+  }
+}
+
+export async function down(_: Knex) {}
```
antooni, line 59: update the down function, to migrate the columns back to the type with timezone

antooni (changes requested):

packages/backend/src/peripherals/database/migrations/080_timestamp_column_tz.ts
```diff
@@ -0,0 +1,49 @@
...
+}
+
+export async function up(knex: Knex) {
+  for (const [table_name, columns] of Object.entries(tableColumnToUpdate)) {
+    for (const column of columns) {
+      // https://www.postgresql.org/docs/14/sql-altertable.html
+      // Indexes and simple table constraints involving the column will be automatically converted
+      // to use the new column type by reparsing the originally supplied expression.
+
+      // get list of indexes table is involved in
```
antooni, line 31: is this and other comments needed?

PR #2086, opened 2023-10-20
Add Optimistic Vision Knowledge Nugget

L2BeatSinkas:
First pull request!


torztomasz (dismissed):
These two nuggets redirect user to the same URL, also the PNGs are the same

antooni (approved):

PR #2090, opened 2023-10-22
zkSync Lite state derivation

aminlatifi:
Added state derivation for zksync lite

Ref L2B-2798


packages/config/src/layer2s/zksynclite.ts
```diff
@@ -335,6 +335,13 @@ export const zksynclite: Layer2 = {
     ],
     risks: [CONTRACTS.UPGRADE_NO_DELAY_RISK],
   },
+  stateDerivation: {
+    nodeSoftware: `The node software is open-sourced and the source can be found [here](https://github.com/matter-labs/zksync).`,
+    compressionScheme: 'Transactions saved off-chain are not compressed.',
```
lucadonnoh, line 340: the compression scheme entry is about data posted onchain, so in this case we care whether state diffs are compressed or not and not about tx data saved offchain

lucadonnoh (changes requested):

lucadonnoh (approved):

PR #2092, opened 2023-10-22
dydx state derviation

aminlatifi:
Ref L2B-2798


adamiak (changes requested):
This file mixes both dydx v3 and dydx v4, which are completely different. We should only write about v3 (and v4 is not live yet).

V3 is a proper rollup on Ethereum, and we should probably refer to our own StarkEx explorer as `nodeSoftware`

V4 is a non-Ethereum construct running on Cosmos, and STARKs from Starkware are not used here.

Please reach to me for more info.

PR #2098, opened 2023-10-23
Update all discovery files to v3. Fix script bug.

adamiak:


packages/backend/scripts/rediscoverRawDevAll.sh
mateuszradomski, line 1: There is no mention anywhere about how one would use that script

mateuszradomski (changes requested):

michalsidzej (approved):

packages/backend/scripts/rediscoverRawDevAll.sh
adamiak: Added

mateuszradomski (approved):

PR #2112, opened 2023-10-26
Rerun discoveries

adamiak:
Since our discovery supports multicall, function ordering has been fixed to be always alphabetical (and some other issues been fixed, see PRs in tools). This PR is just rerunning discovery with the most recent discovery version on existing projects, and the same block number, to update the output. A tiny fix was needed for metis (irrelevant for update monitoring).


michalsidzej (approved):
I looked at all of them very carefully (except for stargate, but sampled randomly few contracts) and looks good.

PR #2126, opened 2023-10-29
discover method instead of execute process

aminlatifi:
Used discover method from package instead of execute process

Ref L2B-2774


aminlatifi:
Blocked by https://github.com/l2beat/tools/pull/90

mateuszradomski (approved):
overall looks good, just remember to update the discovery version in package.json to 0.23 since it's already bumped on master

PR #2139, opened 2023-10-31
Improve TaskQueue retry strategy

antooni:
Context:
Our current TaskQueue retry strategy is to try retry task 10 times and after that halt the queue. This can be problematic for the 3rd party API downtimes longer than this 10 retries exponential backoff (~ 30min time span). 

Solution:
- update the default strategy to never halt (stop) the queue
- add param "notifyAfterAttempts" that allows to set arbitrary number of retires after which error will be sent to Sentry

Most important things to check:
- the TaskQueue behaviour should be the same as before, only extended by this `notifyAfterAttempts`
- is the interface for exponential backoff clear?

Resolves L2B-2731


packages/backend/src/core/SequenceProcessor.ts
```diff
@@ -219,4 +218,10 @@ export class SequenceProcessor extends EventEmitter {
     activityLatest.labels({ project: this.id }).set(state.latest)
     this.state = state
   }
+
+  // WARNING: this method should be used only in tests
+  stopQueue(): void {
```
sz-piotr, line 223: Use `/**` for doc comments. This comment won't show up on hover. Also you could make it obvious from the method name e.g. `_TEST_ONLY_stopQueue()`

packages/backend/src/core/queue/TaskQueue.ts
```diff
@@ -202,15 +211,13 @@ export class TaskQueue<T> {
...
-      this.halted = false
-    }
+  // WARNING: this method should be used only in tests
+  stop() {
+    this.stopped = true
+  }
+
+  // WARNING: this method should be used only in tests
+  clear() {
+    this.queue.splice(0, this.queue.length)
```
sz-piotr, line 221: Use `/**` for doc comments. This comment won't show up on hover. Also you could make it obvious from the method name e.g. `_TEST_ONLY_stop()` and `_TEST_ONLY_clear()`

sz-piotr (changes requested):

packages/backend/src/core/SequenceProcessor.ts
```diff
@@ -219,7 +219,9 @@ export class SequenceProcessor extends EventEmitter {
     this.state = state
   }
 
-  // WARNING: this method should be used only in tests
+  /**
+  WARNING: this method should be used only in tests
+  **/
```
sz-piotr, line 224: With `/**` comments the end should be `*/`

sz-piotr (approved):

PR #2145, opened 2023-10-31
Add LivenessApiResponse type

antooni:
Context:
Before proceeding with frontend work (Liveness) we need to define the schema of the endpoint response.

Solution:
Add new type `LivenessApiResponse` to share-pure package

Resolves L2B-2682

### DIFF
--- a/packages/shared-pure/src/types/api/LivenessApi.ts
+++ b/packages/shared-pure/src/types/api/LivenessApi.ts
@@ -1,0 +1,77 @@
+import z from 'zod'
+
+import { branded } from '../branded'
+import { UnixTime } from '../UnixTime'
+
+const DataPoint = z.object({
+  averageInSeconds: z.number().positive(),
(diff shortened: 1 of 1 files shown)


packages/shared-pure/src/types/api/LivenessApi.ts
```diff
@@ -0,0 +1,77 @@
...
+  }),
+  stateUpdates: z.object({
+    last30Days: DataPoint,
+    last90Days: DataPoint,
+    max: DataPoint,
+  }),
+  anomalies: z.array(
+    z.object({
+      timestamp: branded(z.number(), (n) => new UnixTime(n)),
+      durationInSeconds: z.number().positive(),
```
torztomasz, line 25: make sure it is integer

torztomasz (approved):

maciekzygmunt (approved):

PR #2146, opened 2023-10-31
Fill projects liveness configs

maciekzygmunt:
Resolves L2B-2959


antooni (changes requested):
Check sinceTimestamp:
- transfers - get Timestamp of the first transfer from to
- functionCalls :
 - try to get first call of this method (Etherscan internal calls API?)
 - if you cannot, then check all the timestamps added whether they are all reasonably old

packages/config/src/layer2s/metis.ts
```diff
@@ -65,6 +65,31 @@ export const metis: Layer2 = {
       url: 'https://andromeda.metis.io/',
       startBlock: 1,
     },
+    liveness: {
+      batchSubmissions: [
+        {
```
antooni, line 70: delete it

packages/config/src/layer2s/nova.ts
```diff
@@ -103,6 +103,32 @@ export const nova: Layer2 = {
       assessCount: subtractOne,
       startBlock: 1,
     },
+    liveness: {
+      batchSubmissions: [
+        {
+          formula: 'functionCall',
```
antooni, line 109: should we add 2 more methods to batchSubmissions

packages/config/src/layer2s/linea.ts
```diff
@@ -116,6 +116,21 @@ export const linea: Layer2 = {
       type: 'rpc',
       startBlock: 1,
     },
+    liveness: {
+      batchSubmissions: [],
```
antooni, line 120: is it posting txs?

packages/config/src/layer2s/polygonzkevm.ts
```diff
@@ -104,6 +104,32 @@ export const polygonzkevm: Layer2 = {
       type: 'rpc',
       startBlock: 1,
     },
+    liveness: {
```
antooni, line 107: double check

packages/config/src/layer2s/arbitrum.ts
```diff
@@ -327,6 +327,32 @@ export const arbitrum: Layer2 = {
       assessCount: subtractOneAfterBlockInclusive(22207818),
       startBlock: 1,
     },
+    liveness: {
+      batchSubmissions: [
```
antooni, line 331: should we add 2 more?

torztomasz (approved):

PR #2163, opened 2023-11-03
Update lzomnichain discovery and project page

michalsidzej:
Resolves L2B-2910


lucadonnoh:
image
this part of the description is a bit misleading. While it's true that there is only one contract that is not verified, it doesn't specify which.

michalsidzej:
> 2d6183b

[2d6183b](https://github.com/l2beat/l2beat/pull/2163/commits/2d6183b9b9025bde683660a8f159b4ea11a78b31) WDYT?

michalsidzej:
> image this part of the description is a bit misleading. While it's true that there is only one contract that is not verified, it doesn't specify which.

[2d6183b](https://github.com/l2beat/l2beat/pull/2163/commits/2d6183b9b9025bde683660a8f159b4ea11a78b31) WDYT?

lucadonnoh (approved):

PR #2180, opened 2023-11-08
Update BaseRepository convention

antooni:
This PR will:
- add convention for `update` methods
- add convention for `set` method
- fix the issue with passing the arguments in `wrapAddMany`


packages/backend/src/peripherals/database/shared/BaseRepository.ts
```diff
@@ -188,12 +230,18 @@ export abstract class BaseRepository {
     )
   }
 
+  protected wrapSet<T extends SetMethod>(method: T): T {
+    return this.wrap(method, (count) =>
+      this.logger.info({ method: method.name, count }),
```
torztomasz, line 235: why does set method have count argument?

torztomasz (changes requested):

packages/backend/src/peripherals/database/shared/BaseRepository.ts
antooni: This returns how many row were updated as a result of this function call. Maybe the result there should be `success: boolean` instead of this count. WDYT?

torztomasz: I guess it's fine, approved.

PR #2209, opened 2023-11-16
Different TVL header description for bridges

aminlatifi:
Resolves L2B-2873


packages/frontend/src/components/header/TvlHeader.tsx
```diff
@@ -2,7 +2,7 @@ import React from 'react'
 
 import { HorizontalSeparator } from '../HorizontalSeparator'
 
-export function TvlHeader() {
+export function TvlHeader({ isBridge = false }) {
```
torztomasz, line 5: no need to provide default value, change it to this:
```tsx
interface TvlHeaderProps {
  isBridge: boolean
}
...
export function TvlHeader({ isBridge }: TvlHeaderProps) {
```

torztomasz (changes requested):

torztomasz (approved):

PR #2218, opened 2023-11-20
Add liveness status endpoint

maciekzygmunt:
Resolves L2B-3093

This PR:
- adds Liveness Status endpoint
- fix small bug in `LivenessConfigurationIdentifier` and add tests for future changes
- adds migration to fix database


packages/backend/src/peripherals/database/migrations/086_fix_liveness_params.ts
```diff
@@ -0,0 +1,60 @@
...
+        params: JSON.stringify({
+          from: config.from.toString(),
+          to: config.to.toString(),
+        }),
+      })
+      .where('identifier', '=', identifier)
+  }
+}
+
+export async function down() {}
```
antooni, line 60: here restore the previous "faulty" state, this will ensure our migrations are atomic

packages/backend/src/peripherals/database/migrations/086_fix_liveness_params.ts
```diff
@@ -0,0 +1,60 @@
...
+        }
+      })
+      .filter(notUndefined)
+    const stateUpdates = l2.config.liveness?.stateUpdates
+      .map((c) => {
+        if (c.formula === 'transfer') {
+          return {
+            ...c,
+            projectId: l2.id,
+            type: LivenessType('DA'),
```
antooni, line 39: STATE?

antooni (approved):

PR #2226, opened 2023-11-21
Update Linea

mateuszradomski:
Resolves L2B-3103

## Description

AdminMultisig safe was extended with a role oriented module.

### AdminMultisig

Added a new module called Roles. The owner of this module (the AdminMultisig itself) can grant or remove access to given targets (whole contracts) or just a certain function in a given contract per role. The module checks if the message sender has a given role and checks if the role is allowed to execute a given operation. This whole module seems like it solves the same issue as AccessControl but with more granularity just like the extended AccessControl in Scroll.

```diff
    contract AdminMultisig (0x892bb7EeD71efB060ab90140e7825d8127991DD3) {
      upgradeability.modules[0]:
+        "0xF24f1DC519d88246809B660eb56D94048575d083"
    }
```

## Created contracts

```diff
+   Status: CREATED
    contract MultiSend (0xA238CBeb142c10Ef7Ad8442C6D1f9E89e07e7761) {
    }
```

```diff
+   Status: CREATED
    contract Roles (0xF24f1DC519d88246809B660eb56D94048575d083) {
    }
```

### DIFF
--- a/packages/backend/discovery/linea/ethereum/config.jsonc
+++ b/packages/backend/discovery/linea/ethereum/config.jsonc
@@ -16,7 +16,8 @@
     "0x051F1D88f0aF5763fB888eC4378b4D8B29ea3319": "ERC20Bridge",
     "0x5B0bb17755FBa06028530682E2FD5bc373931768": "ERC20BridgeProxyAdmin",
     "0xC84832f69bFFbC1A94E44a157A342766E959ED27": "PlonkVerifierFull2",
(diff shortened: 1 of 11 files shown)


luke7211:
Current LINEA setup:

image

lucadonnoh:
new contracts need to be added into the contracts section

luke7211 (changes requested):
We need to extract roles/permissions from the Module. Given that this module is written using Zodiac standard, we also need to extract Modules from it

luke7211 (approved):

PR #2240, opened 2023-11-22
Optimise liveness API response

antooni:
Introduction of Promise.all will paralelize the calls to the DB, resulting in reducing this bottleneck. But still, the introduction of indexes to the table is needed. That will be done in the next PR


maciekzygmunt (approved):

mateuszradomski (approved):
🔥

PR #2243, opened 2023-11-22
Add logic for duplicated data in liveness

maciekzygmunt:
Resolves L2B-3133 L2B-3095

This PR:
- adds logic to handle situations when batch submission and state updates are posted in the same tx. Now in those cases we need to fill only one config, and specify in new `duplicatedData` property which we want to copy:
image
So in this case we will copy stateUpdates data into batchSubmission
- fills config for Linea

### DIFF
--- a/packages/backend/src/api/controllers/liveness/LivenessController.ts
+++ b/packages/backend/src/api/controllers/liveness/LivenessController.ts
@@ -1,4 +1,5 @@
-import { LivenessApiResponse } from '@l2beat/shared-pure'
+import { layer2s } from '@l2beat/config'
+import { LivenessApiResponse, LivenessType } from '@l2beat/shared-pure'
 
 import { Project } from '../../../model'
 import { LivenessRepository } from '../../../peripherals/database/LivenessRepository'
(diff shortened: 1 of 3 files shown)


packages/backend/src/api/controllers/liveness/LivenessController.ts
```diff
@@ -39,6 +40,34 @@ export class LivenessController {
         const withAnomalies = calculateAnomaliesPerProject(intervals)
         console.timeEnd(`withAnomalies ${project.projectId.toString()}`)
 
+        if (withAnomalies) {
+          const l2project = layer2s.find((l) => l.id === project.projectId)
```
antooni, line 44: get rid of this find, add duplicateData optional field to backend liveness config

packages/backend/src/api/controllers/liveness/LivenessController.ts
```diff
@@ -39,6 +40,34 @@ export class LivenessController {
...
+          const anomalies = withAnomalies.anomalies
+          if (l2project?.config.liveness?.duplicatedData?.batchSubmissions) {
+            withAnomalies.batchSubmissions = { ...withAnomalies.stateUpdates }
+            if (anomalies) {
```
antooni, line 48: I am not sure what we should do with the anomalies in this case, maybe let's skip them for now, and keep only the "source" ones

antooni (approved):

antooni (approved):

PR #2247, opened 2023-11-22
Add "Config related changes" and Improve Loopring discovery

luke7211:
Resolves L2B-3156


packages/backend/src/tools/updateDiffHistory.ts
```diff
@@ -305,14 +342,11 @@ function findDescription(
     return undefined
   }
 
-  const lastIndex = lines.findIndex((l) => l.startsWith('## Watched changes'))
+  const followingLines = lines.slice(index + 1)
```
mateuszradomski, line 345: I don't like this solution because it does not work if you want to use markdown inside the description. For example in [this PR](https://github.com/l2beat/l2beat/pull/2350) I've added headers to contracts that have changed using the `#` character.

To fix this I'd find the line that contains `## Watched changes` or `## Config related changes` to allow for headers inside the description header

mateuszradomski (changes requested):
@adamiak 

Looks fine but the issue in the comment described below is big for me

packages/backend/src/tools/updateDiffHistory.ts
adamiak: Very good point, my solution would break on subsections.

But hardcoding exact names of following same-level sections to find the end of description is very error-prone. Even now there's a different section name than what you mentioned, namely "## Initial discovery'. 

What I did, is to look for the next line that begins with  '##<space' - this should achieve exactly what we want - description ends when the new section at the same level begins.

mateuszradomski (approved):

PR #2254, opened 2023-11-23
Refactor BigQuery queries

sz-piotr:


antooni (approved):

antooni (approved):
I have tested it locally, it behaves the same as the previous version of code

PR #2282, opened 2023-11-28
Add raw liveness data endpoint

maciekzygmunt:
Resolves L2B-3197


packages/backend/src/peripherals/database/LivenessRepository.ts
```diff
@@ -71,6 +71,26 @@ export class LivenessRepository extends BaseRepository {
     return rows.map(toRecordWithProjectIdAndType)
   }
 
+  async getByProjectIdAndType(
```
antooni, line 74: Add a test for that

antooni: + maybe we also want to add `.distinct(timestamp)` ?

antooni (changes requested):

antooni (approved):

PR #2284, opened 2023-11-28
Add projects to liveness

maciekzygmunt:
Resolves L2B-3192 L2B-3193 L2B-3191

This PR adds liveness config for:
- Scroll
- Manta Pacific
- Paradex


antooni (approved):
I assume those were tested locally

PR #2287, opened 2023-11-28
Remove TXs in the same block for Liveness

maciekzygmunt:
Resolves L2B-3200

Now we do not count transactions that happened in the same block.


packages/backend/src/peripherals/database/LivenessRepository.test.ts
```diff
@@ -151,4 +151,62 @@ describe(LivenessRepository.name, () => {
...
+            projectId: LIVENESS_CONFIGS[0].projectId,
+            type: LIVENESS_CONFIGS[0].type,
+          },
+        ]
+
+        expect(result).toEqual(expected)
+      })
+
+      it('filters out transactions with the same timestamp', async () => {
+        const NEW_DATA = [
```
antooni, line 174: this test will be more readable if you delete all records in the beginning and add all them here

antooni (approved):

PR #2315, opened 2023-12-03
Add pheasant network

arkeysun:


lucadonnoh:
i see that most of the logic is on L2, and we're going to add L2 contracts support soon so i think the best is to wait a bit for it and add proper discover. in the current form the page is not very useful.

mateuszradomski:
Closing, tracked by L2B-3304.

lucadonnoh:
added to the backlog

PR #2321, opened 2023-12-04
Allow for l3s to be added

maciekzygmunt:
Resolves L2B-3235 L2B-3250


packages/frontend/src/components/table/props/getScalingTableColumnsConfig.tsx
```diff
@@ -302,6 +296,48 @@ export function getArchivedScalingTvlColumnsConfig() {
...
+    },
+    {
+      name: 'Type',
+      tooltip: 'The type stack used.',
+      shortName: 'Type',
+      getValue: (project) => project.provider,
+    },
+    {
+      name: 'Host Chain',
+      tooltip: 'Name of host chain.',
```
radinat, line 328:
```suggestion
      tooltip: 'The chain, on top of which the L3 is built.',
```

radinat (changes requested):

radinat (approved):

PR #2347, opened 2023-12-09
Refactor API data fetching and caching

torztomasz:
This pull request refactors the API data fetching and caching logic.


packages/frontend/src/build/buildPages.ts
```diff
@@ -38,31 +37,37 @@ async function main() {
 
   const http = new JsonHttpClient(httpClient, config.backend.skipCache)
 
+  console.time('Fetching TVL data')
```
antooni, line 40: What about such pattern?

console.log('[TVL] fetching data ...')
console.time('[TVL] fetched')

packages/frontend/src/build/caching/JsonHttpClient.ts
```diff
@@ -33,7 +33,7 @@ export class JsonHttpClient {
   }
 }
 
-const TEN_MINUTES_IN_MS = 10 * 60 * 1000
+const THIRTY_MINUTES_IN_MS = 30 * 60 * 1000
```
antooni, line 36: tbh, all the data that we currently have refreshes on hourly granularity, maybe make it to invalidate cache when hour is different ?

torztomasz: yup probably better, will change

antooni (approved):

PR #2350, opened 2023-12-11
Update zkSync Era

mateuszradomski:
Resolves L2B-3257

## Description

The big picture of these changes are the following:

- Migrated Solidity from 0.8.13 to 0.8.20.
- Moved from working on blocks to working on batches.
- Any mention to Security Council have been removed, contracts have been simplified basing on that fact
- Refactor and updates of the verifier algorithm

Here follows a description of changes to each particular contract.

### LibMap

A new library described as "Library for storage of packed unsigned integers". Taken from open-source solady library package (https://github.com/Vectorized/solady). It's just an gas optimized uint32 array. It seems like they just extracted what they needed from it, because the solady version supports more than just uint32s.

### ValidatorTimelock

Uses LibMap to replace a mapping of `uint256 => uint256`, it mapped from the block number to the block timestamp. Right now using LibMap they essentially have `uint64 => uint32`, it maps from batch number to the block timestamp. The timestamp is truncated to `uint32`, this will wrap in the year 2106. They have introduced a getter for the timestamp of the block `getCommittedBatchTimestamp`, but it's not used by any of the code. In `executeBatches` the assertion changed from `require(block.timestamp > commitBlockTimestamp + delay)` to `require(block.timestamp >= commitBatchTimestamp + delay)`, now it's inclusive.

### L1ERC20Bridge

Moved to batches, simple renames.

### Diamond

Better check for facet without code, instead of doing `require(_facet != address(0))` they now do `require(_facet.code.length > 0)`. Reverts in `_initializeDiamondCut()` will now revert with decoded calldata if it's possible, if not it will just revert with the previous error code.

### Verifier

Extracted an interface into IVerifier. Verifier now a single file contract, where before it included `PairingsBn254.sol`, `TranscriptLib.sol`, `Plonk4VerifierWithAccessToDNext.sol`. It was flattened and refactored for readability, but I'm unable to verify that this is the same verification algorithm.

### Storage

`UpgradeStorage upgrades` is now deprecated. A new role, called `admin` and `pendingAdmin`.

### Base

Removed `onlySecurityCouncil` modifier and added `onlyGovernorOrAdmin`.

### AllowList

New event

```
event UpdateDepositLimit(address indexed l1Token, bool depositLimitation, uint256 depositCap);
```

that is emitted in the call to this function - that is in AllowList.sol:

```
function setDepositLimit(address _l1Token, bool _depositLimitation, uint256 _depositCap) external onlyOwner
```

### zkSync

Facets:

- changed from DiamondCut => Admin (0x409560DE546e057ce5bD5dB487EdF2bB5E785baB)
- same => Getters (0xF3ACF6a03ea4a914B78Ec788624B25ceC37c14A4)
- same => Mailbox (0x63b5EC36B09384fFA7106A80Ec7cfdFCa521fD08)
- same => Executor (0x9e3Fa34a10619fEDd7aE40A3fb86FA515fcfd269)

The 5th facet was Governance facet, it was removed so we now have 4 facets instead of 5. The first facet (DiamondCut) has been replaced with Admin.

#### Admin

Admin is an implementation of IAdmin interface. That interface is a combination of IGovernance and IDiamondCut. IAdmin implements the entire IGovernance interface. It also implements parts of the IDiamondCut interface, like `executeUpgrade()`, `freezeDiamond()` and `unfreezeDiamond()`. Everything relating to proposal and security council has been removed. The new functions that were in neither of these two interfaces are `setPendingAdmin()` and `acceptAdmin()`. Implementation of Admin is a combination of implementations for Governance and DiamondCut facets from before.

#### Getters

The default IGetters interface has moved to working on batches instead of blocks. Additionally getters for functions referencing Security Council have been removed (since they returned zero addresses either way). The legacy interface is still supported, but it just returns the values in batches.

#### Mailbox

Variable renaming to move to batches instead of blocks

#### Executor

Batch commitment uses three inputs: pass through from batches, meta params and auxiliary output from the batch. The first two, stay the same after the update. But the third input (auxiliary output from the batch) changed most of the input that it takes. Right now, the prover verifies both the main and the recursive proof in the same step. Refactored the way logs from L2 indicate what they are, instead of different senders use `logKeys` and `logValues`.


packages/config/src/discovery/values/projects/zksync2/ZKSYNC_2.test.ts
lucadonnoh, line 1: please remove this test as not needed anymore

lucadonnoh (changes requested):

antooni (changes requested):
Please update the liveness config

antooni (approved):

lucadonnoh (approved):

PR #2358, opened 2023-12-12
Make project icons stick to the left in the tables

torztomasz:


packages/frontend/tailwind.config.js
```diff
@@ -1,5 +1,8 @@
 /** @type import("tailwindcss/types").Config */
 module.exports = {
+  future: {
+    hoverOnlyWhenSupported: true,
```
antooni, line 4: please add a comment with explaination, I assume it was added to prevent hover events from happening on mobile?

packages/frontend/src/components/table/props/getRowType.ts
```diff
@@ -0,0 +1,30 @@
...
+  if (entry.showProjectUnderReview) {
+    return 'under-review'
+  }
+  if (entry.isUpcoming) {
+    return 'upcoming'
+  }
+}
+
+export function getRowTypeClassNames(): string {
+  return classNames(
```
antooni, line 23: this one makes inspecting HTML really hard

packages/frontend/tailwind.config.js
torztomasz: yup, basically its going to be default in tailwind v4

antooni (approved):

PR #2395, opened 2023-12-19
Add BOB network

nud3l:
- Adds the BOB network (an OP Stack rollup)


lucadonnoh:
we currently only supports network using Ethereum consensus (so Ethereum rollups or Ethereum validiums). My understanding from the docs is that the project didn't decide yet what to use (ETH vs BTC vs Celestia), right?

nud3l:
We are using Ethereum consensus with an unmodified OP Stack. We may use Celestia or Bitcoin for DA but settlement will happen on Ethereum.

lucadonnoh:
please fix tests

danielsimao:
> please fix tests

Hey @lucadonnoh. I have added what was missing in the PR. I ran the linter and it is fine but the tests fail.

image

This does not seem to be related to our PR. Let me know if there is something else that we need do. Cheers

lucadonnoh (approved):

PR #2398, opened 2023-12-19
Add discovered.json hashes database and a test

mateuszradomski:
Resolves L2B-3253


packages/backend/src/tools/updateDiffHistory.ts
```diff
@@ -316,3 +324,14 @@ function findDescription(
...
+
+async function updateHashes(projectName: string, chainName: string) {
+  const databasePath = 'discovery/discoveredHashes.json'
+  const chainId = ChainId.fromName(chainName)
+  const shaSum = await getDiscoveryHash(projectName, chainId)
+  const database = getHashesDatabase(databasePath)
+
+  database[getHashesDatabaseKey(projectName, chainId)] = shaSum
+
+  writeFileSync(databasePath, JSON.stringify(database, null, 2))
```
adamiak, line 336: This is an unexpected place to write hashes to the file, since reading is implemented in `hashDatabase.ts`. I think that functionality of this whole function should be encapsulated in `hashDatabase.ts`, so that here you simple call one function e.g. `updateHash(projectName, hashName)`.

packages/backend/src/core/discovery/utils/hashDatabase.ts
```diff
@@ -0,0 +1,31 @@
...
+  hasher.update(JSON.stringify(curDiscovery))
+  return `0x${hasher.digest('hex')}`
+}
+
+export function getHashesDatabase(path: string): Record<string, string> {
+  let result = undefined
+
+  try {
+    const hashContent = readFileSync(path, 'utf-8')
+    result = JSON.parse(hashContent) as Record<string, string>
```
adamiak, line 18: In most places we use zod to parse JSON for additional checks, and I think it would also make sense here.

packages/backend/src/core/discovery/utils/hashDatabase.ts
```diff
@@ -0,0 +1,31 @@
...
+  } catch (_) {
+    result = {}
```
adamiak, line 20: I suggest removing the whole try/catch block. Why do we want to silently fail?

mateuszradomski: To be honest, it's one of those _I'll do something clever in here_. The idea behind it is that if the file does not exists - `readFileSync` throws - or the content of the file is somehow corrupt and parsing it as JSON fails - also throws - just start with an empty "database". I don't really mind this being like this, lmk what you think

adamiak: From the point of view of the code reviewer it's unclear (or seemingly incorrect) of what happens here and why. It either needs an extensive comment, or simply to implement the intention directly - rather than try/catch, simply check if file exists and return {} if not. I prefer the "direct implementation" approach.

As for other unexpected errors during file reading, I see no reason of silently suppressing them. I prefer to fail early - this way you don't let any potential error to propagate further and pollute the rest of the system.

adamiak (approved):

PR #2401, opened 2023-12-19
Add discovery for every single day in backend

mateuszradomski:
Resolves L2B-3374

### DIFF
--- a/packages/backend/src/config/config.local.ts
+++ b/packages/backend/src/config/config.local.ts
@@ -22,6 +22,7 @@ export function getLocalConfig(env: Env): Config {
   )
   const livenessEnabled = env.boolean('LIVENESS_ENABLED', false)
   const updateMonitorEnabled = env.boolean('WATCHMODE_ENABLED', false)
+  const diffHistoryEnabled = env.boolean('DIFF_HISTORY_ENABLED', false)
   const discordToken = env.optionalString('DISCORD_TOKEN')
   const internalDiscordChannelId = env.optionalString(
     'INTERNAL_DISCORD_CHANNEL_ID',
@@ -288,6 +289,16 @@ export function getLocalConfig(env: Env): Config {
           ...getChainConfig(ChainId.POLYGON_ZKEVM),
           reorgSafeDepth: env.optionalInteger(
             'DISCOVERY_POLYGON_ZKEVM_REORG_SAFE_DEPTH',
+          ),
+        },
+      ],
+    },
+    diffHistory: diffHistoryEnabled && {
+      chains: [
+        {
+          ...getChainConfig(ChainId.ETHEREUM),
+          reorgSafeDepth: env.optionalInteger(
+            'DISCOVERY_ETHEREUM_REORG_SAFE_DEPTH',
           ),
         },
       ],
--- a/packages/backend/src/core/Clock.ts
+++ b/packages/backend/src/core/Clock.ts
@@ -5,25 +5,44 @@ export class Clock {
     private readonly minTimestamp: UnixTime,
     private readonly delayInSeconds: number,
     private readonly refreshIntervalMs = 1000,
-  ) {
-    if (!this.minTimestamp.isFull('hour')) {
-      this.minTimestamp = this.minTimestamp.toNext('hour')
-    }
-    if (this.minTimestamp.gt(this.getLastHour())) {
-      throw new Error('minTimestamp must be in the past')
-    }
-  }
+  ) {}
 
   getFirstHour(): UnixTime {
-    return this.minTimestamp
+    let result = this.minTimestamp
+
+    if (!result.isFull('hour')) {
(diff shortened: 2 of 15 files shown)


packages/backend/src/modules/diff-history/createDiffHistoryModule.ts
```diff
@@ -0,0 +1,102 @@
...
+    logger.info('DiffHistory module disabled')
+    return
+  }
+
+  const configReader = new ConfigReader()
+
+  const discoveryLogger =
+    config.name === 'Backend/Local'
+      ? DiscoveryLogger.SILENT
+      : DiscoveryLogger.SILENT
```
adamiak, line 35: Why both are SILENT?

packages/backend/src/config/config.local.ts
```diff
@@ -279,5 +280,15 @@ export function getLocalConfig(env: Env): Config {
         },
       ],
     },
+    diffHistory: diffHistoryEnabled && {
+      chains: [
+        {
+          ...getChainConfig(ChainId.ETHEREUM),
+          reorgSafeDepth: env.optionalInteger(
```
adamiak, line 287: Don't we need to the max-logs-range related stuff?

packages/backend/src/config/config.local.ts
```diff
@@ -279,5 +280,15 @@ export function getLocalConfig(env: Env): Config {
...
+            'DISCOVERY_ETHEREUM_REORG_SAFE_DEPTH',
```
adamiak, line 288: Don't we need the max-logs-range related stuff?

packages/backend/src/core/Clock.ts
```diff
@@ -5,25 +5,44 @@ export class Clock {
...
     return UnixTime.now().add(-this.delayInSeconds, 'seconds').toStartOf('hour')
   }
 
+  getLastDay(): UnixTime {
+    return UnixTime.now().add(-this.delayInSeconds, 'seconds').toStartOf('day')
+  }
+
   onEveryHour(callback: (timestamp: UnixTime) => void) {
-    let next = this.minTimestamp
+    let next = this.getFirstHour()
```
adamiak, line 45: A lot of stuff (TVL, Activitity, ?) rely on this loop. How does changing this initial value impact those?

packages/backend/src/peripherals/database/migrations/091_update_monitor_pkey.ts
```diff
@@ -0,0 +1,28 @@
...
+should create a new migration file that fixes the issue.
+
+*/
+
+import { Knex } from 'knex'
+
+export async function up(knex: Knex) {
+  await knex.schema.alterTable('update_monitor', (table) => {
+    table.dropPrimary()
+    table.primary(['project_name', 'chain_id', 'unix_timestamp'])
```
adamiak, line 19: Previously we only had one entry per project in this table. Update monitor functionality relied on this fact. Won't switching to multiple entries per project break update monitor?

packages/backend/src/core/discovery/ProjectDiscoverer.ts
```diff
@@ -0,0 +1,108 @@
...
+  async start() {
+    this.logger.info('Started')
+
+    const known = await this.repository.getTimestamps(
+      this.projectName,
+      this.chainId,
+    )
+
+    for (const timestamp of known) {
+      this.knownSet.add(timestamp.toNumber())
```
adamiak, line 45: knowSet should keep days, not times up to the second.

adamiak: e.g. use .toStartOf('day')

packages/backend/src/peripherals/database/migrations/091_add_discovery_history.ts
```diff
@@ -14,15 +14,19 @@ should create a new migration file that fixes the issue.
 import { Knex } from 'knex'
 
 export async function up(knex: Knex) {
-  await knex.schema.alterTable('update_monitor', (table) => {
-    table.dropPrimary()
+  await knex.schema.createTable('discovery_history', function (table) {
```
adamiak, line 17: How about calling the table "daily_discovery"

packages/backend/src/peripherals/database/discovery/DiscoveryHistoryRepository.ts
```diff
@@ -0,0 +1,105 @@
...
+    name: string,
+    chainId: ChainId,
+  ): Promise<DiscoveryHistoryRecord | undefined> {
+    const knex = await this.knex()
+    const row = await knex('discovery_history')
+      .where({
+        project_name: name,
+        chain_id: +chainId,
+      })
+      .first()
```
adamiak, line 36: Missing sorting

adamiak (approved):

PR #2415, opened 2023-12-20
Remove wUSDM

radinat:
The token is causing issues because its historical data is not returned from Coingecko (it has been added recently).
Will be added later once there is some data.


lucadonnoh (approved):

michalsidzej (approved):

DanielZhangReal:
@radinat  Do you know how long this will take?

PR #2425, opened 2023-12-21
fix: update of discord links

krauspt:
Hello again,
today I'm updating outdated discord invites, all links are taken from project webpages. Note that Obscuro is branded as Ten. Also, note that Stargate has a special flow for Discord invites - they don't use a general invite link, but for each user who passes the captcha check is generated a specific time-limited invite.


packages/config/src/bridges/stargate.ts
```diff
@@ -20,7 +20,7 @@ export const stargate: Bridge = {
         'https://github.com/LayerZero-Labs/LayerZero',
       ],
       socialMedia: [
-        'https://discord.gg/agEXvDB',
```
mateuszradomski, line 23: We required the discord link to be in the format `https://discord.gg/<code>` or `https://discord.com/invite/<code>`. The `https://stargate.finance/discord` is not going to work.

mateuszradomski (changes requested):

packages/config/src/bridges/stargate.ts
krauspt: @mateuszradomski I went with the first option (keep the old outdated link) and now the test passes. The other 3 updated links are working without problem.

mateuszradomski (approved):

PR #2441, opened 2023-12-29
Add Lisk as an upcoming OP Stack L2

Huberacht:


packages/frontend/src/static/icons/lisk.png
lucadonnoh, line 1: icon should be 256x256

lucadonnoh (changes requested):

lucadonnoh (approved):

PR #2444, opened 2023-12-29
Add Ancient8 as an upcoming OP Stack L2

Huberacht:


packages/frontend/src/static/icons/ancient.png
lucadonnoh, line 1: icon should be 256x256

lucadonnoh (changes requested):

PR #2451, opened 2024-01-02
Add STONE to linea

luke7211:


packages/config/src/layer2s/linea.ts
```diff
@@ -75,6 +79,27 @@ const withdrawalLimitString = `Currently, there is a general limit of ${utils.fo
...
+  {
+    id: AssetId('lyra:stone-stakestone-ether'),
+    name: 'StakeStone Ether',
+    symbol: 'STONE',
+    decimals: 18,
+    iconUrl:
+      'https://assets.coingecko.com/coins/images/33103/large/200_200.png?1702602672',
+    address: EthereumAddress('0x93F4d0ab6a8B4271f4a28Db399b5E30612D21116'),
+    coingeckoId: CoingeckoId('stakestone-ether'),
+    sinceTimestamp: new UnixTime(1695507314),
```
mateuszradomski, line 92: Use `1699781729` instead. It's the timestamp when the first price on Coingecko appeared.

mateuszradomski (approved):
LGTM, but please change the two things I've mentioned below

PR #2454, opened 2024-01-02
Relist wUSDM

DanielZhangReal:
Can we relist wUSDM on L2Beat now? It's been two weeks. Thanks

https://github.com/l2beat/l2beat/pull/2415
https://github.com/l2beat/l2beat/pull/2410

### DIFF
--- a/packages/config/src/tokens/tokenList.json
+++ b/packages/config/src/tokens/tokenList.json
@@ -3011,6 +3011,20 @@
       "formula": "locked"
     },
     {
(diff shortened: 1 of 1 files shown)


mateuszradomski:
Hi,

We're using Coingecko to determine the price of tokens. As you can see for yourself, Coingecko currently does provide historical data for this token (https://www.coingecko.com/en/coins/wrapped-usdm) but it's very sparse and limited. As of this moment we do not have mechanism to support tokens with this many holes in its price chart.

PR #2460, opened 2024-01-03
Add wUSDM

mateuszradomski:
Closes #2454


DanielZhangReal:
cool, thanks @mateuszradomski 

I think using `mountain-protocol-usdm` for the price of token wUSDM is a good solution

antooni (approved):

PR #2465, opened 2024-01-03
Update Aevo

radinat:
PGNMultiSig owners change.

Resolves L2B-3419

@lucadonnoh Is it correct for this multisig to be called PGNMultiSig? I see it's reused across projects, but for example for Lyra it's named LyraMultisig.


lucadonnoh:
please fix name as discussed

lucadonnoh (approved):

PR #2474, opened 2024-01-04
Add ZKFair DAC

lucadonnoh:
Resolves L2B-3428


adamiak (dismissed):
There is diffHistory file missing, looks like an error in our tool. I'll look into it.

luke7211 (approved):

PR #2475, opened 2024-01-04
feat: add token ASM

DanielZhangReal:
One question: where did you get Circulating Supply?  CoinGecko?


luke7211:
Tokens needs to have circulating supply available on Coingecko and - as a rule of thumb - needs to have significant trading volume and be in Top300 Coingecko list. See more details in https://medium.com/l2beat/redefining-total-value-locked-for-l2s-756160602747

PR #2476, opened 2024-01-04
Add logic to remove historical data from TVL module

antooni:
Context:
There is an ongoing issue with our DB size, specifically reports and balances tables:
- our postgres instance will soon reach the capacity
- per-token requests are timing out

Currently reports table has ~300M records which translates to around 120GB disk space (with indexes)

Solution:
We store data with hourly granularity for all the historical timestamps (from 2019). If we remove hourly data for timestamp older than 7D and six hourly data for timestamp older than 90D we should reduce the size to `5%` of what we currently have. This should be step in the right direction, and the next should be to get rid of zeroes (which currently constitutes 92% of reports table)

Needed changes:
- change the Clock tick behaviour
- add method to remove historical data if it goes out of range (7D, 90D) - every repository & updater keeping TVL data should have it
- update hardcoded config hash, so the reports will recalculate 

Resolves L2B-2296

### DIFF
--- a/packages/backend/src/core/Clock.test.ts
+++ b/packages/backend/src/core/Clock.test.ts
@@ -19,7 +19,9 @@ describe(Clock.name, () => {
     UnixTime.fromDate(new Date(`2022-06-29T${hhmmss}.000Z`))
 
   function setTime(hhmmss: string) {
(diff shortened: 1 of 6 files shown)


sdlyy:
👀

PR #2478, opened 2024-01-04
Hover same address on hover

maciekzygmunt:
image
image


lucadonnoh:
looks great! let's make it a bit subtler like in etherscan tho
image

packages/frontend/src/components/project/ContractEntry.tsx
```diff
@@ -99,11 +99,12 @@ export function ContractEntry({
             ))}
             {contract.links.map((x, i) => (
               <Link
+                data-role="etherscan-link"
                 key={i}
                 className={cx(
                   verificationStatus.contracts[x.address] === false &&
                     !x.isAdmin
```
torztomasz, line 106: `!x.isAdmin && 'text-red-300'`

torztomasz (approved):

lucadonnoh (changes requested):
is this checking the full address or just the text displayed on the website (first few chars and last few chars)? because if that's the case, it could happen that two addresses looks similar but are different and they should not be highlighted together

maciekzygmunt:
> is this checking the full address or just the text displayed on the website (first few chars and last few chars)? because if that's the case, it could happen that two addresses looks similar but are different and they should not be highlighted together

It is checking the full address

lucadonnoh (approved):

PR #2480, opened 2024-01-04
Add shortName property to project configs

torztomasz:
This pull request adds a new `shortName` property to the project configs. The `shortName` property will be used as a display name in some places on the website e.g. in the table.


lucadonnoh:
Metis, Boba and Honeypot can be shortened too

antooni (approved):

PR #2485, opened 2024-01-05
Remove get included projects

maciekzygmunt:
This PR removes getIncludedProjects


antooni (changes requested):
There is a change in `getScalingL2SummaryEntry`needed

antooni (approved):

PR #2518, opened 2024-01-09
feat: add WAGMI omnichain token

michalsidzej:
Found a token that will add ~20M USD to lzomnichain bridge page tvl.


packages/config/src/tokens/tokenList.json
```diff
@@ -2912,6 +2912,20 @@
...
       "formula": "locked"
     },
+    {
+      "id": "wagmi-wagmi",
+      "name": "Wagmi",
+      "coingeckoId": "wagmi-2",
+      "address": "0x92CC36D66e9d739D50673d1f27929a371FB83a67",
+      "symbol": "WAGMI",
+      "decimals": 18,
+      "sinceTimestamp": 1688483015,
```
maciekzygmunt, line 2922: sinceTimestamp for this token should be changed to 1695376800 because this is the time coingecko started to tracking its price

maciekzygmunt (changes requested):

packages/config/src/tokens/tokenList.json
michalsidzej: Ok, but how do you know that?

maciekzygmunt (approved):

PR #2519, opened 2024-01-10
add more details about parallel

0x8f701:


lucadonnoh:
most info is incorrect, we'll open another PR

PR #2529, opened 2024-01-10
Adding Ola

PhilipEriksson:
Adding Ola L2


lucadonnoh (changes requested):
refactor to use the new upcoming template, it should be pretty straightforward, use this as a reference https://github.com/l2beat/l2beat/blob/master/packages/config/src/layer2s/ancient.ts

PhilipEriksson:
> refactor to use the new upcoming template, it should be pretty straightforward, use this as a reference https://github.com/l2beat/l2beat/blob/master/packages/config/src/layer2s/ancient.ts

Refactored to match new template @lucadonnoh

PhilipEriksson:
> > refactor to use the new upcoming template, it should be pretty straightforward, use this as a reference https://github.com/l2beat/l2beat/blob/master/packages/config/src/layer2s/ancient.ts
> 
> Refactored to match new template @lucadonnoh

Seems like the "purpose" field was changed to "purposes" after I had opened the example you sent. I'll change that to match as well.

PhilipEriksson:
Tests were failing on a formatting issue even though links were valid. I refactored it, tests should pass now I believe, not sure if it needs your approval to trigger the CI/CD pipeline again @lucadonnoh 

Cheers mate, hope it's all good now.

lucadonnoh (approved):

PR #2532, opened 2024-01-11
Fix coingecko since timestamp

maciekzygmunt:
Resolves L2B-3525

Previously, when adding a new token with the sinceTimestamp parameter, we used the contract creation date, which does not correspond to the date when CoinGecko began tracking the token's price. Now, we will use the date of the first recorded price on CoinGecko. I have also implemented a test to check for any gaps in the price data, which can occur with CoinGecko.

I have also moved CoingeckoQueryService to shared package


packages/config/scripts/tokens/add/getTokenInfo.ts
```diff
@@ -124,7 +135,45 @@ async function getSinceTimestamp(
...
+  )
+
+  const coingeckoPriceHistoryData =
+    await coingeckoClient.getCoinMarketChartRange(
+      coingeckoId,
+      'usd',
+      new UnixTime(0),
+      UnixTime.now(),
+    )
+  const firstCoingeckoPriceTimestamp =
```
antooni, line 154: Assert that length is longer than 1

packages/config/scripts/tokens/add/getTokenInfo.ts
```diff
@@ -124,7 +135,45 @@ async function getSinceTimestamp(
...
+    coingeckoPriceHistoryData.prices[0].date.getTime() / 1000
+
+  //We call it to check if there is a gap in price history
+  const coingeckoQueryService = new CoingeckoQueryService(coingeckoClient)
+  await coingeckoQueryService.getUsdPriceHistory(
+    coingeckoId,
+    new UnixTime(firstCoingeckoPriceTimestamp),
+    UnixTime.now(),
+  )
+
```
antooni, line 164: add a comment, that if code reaches there price data is okay

antooni (approved):

PR #2535, opened 2024-01-11
Add get-lz-coingecko script

michalsidzej:
Resolves L2B-3463

### Context

New omnichain tokens that use layer zero as a bridge are popping out often. It's important to find them before it's too late. 
This script is using a set of heuristics to compile a list of native omnichain tokens. This script is not perfect as it has some false negatives (ex. it does not show agEUR) but it fills in the gap of the old script.
The current game plan is to merge this script to l2beat, but in a near future we'll delete it from here after we add this logic to the [token-list repository](https://github.com/l2beat/token-list), see [L2B-3552](https://linear.app/l2beat/issue/L2B-3552/integrate-handwritten-scripts-to-information-sources-of-tokenlist)


packages/config/scripts/omnichain/coingecko/EtherescanClient.ts
mateuszradomski, line 1: Can't we just re-use the `EtherscanClient` from the `shared` package and extend it if does not meet the required specifications? Plus, there is a typo in the name of the file, `Ether[e]scanClient`

packages/config/scripts/omnichain/coingecko/index.ts
```diff
@@ -0,0 +1,202 @@
...
+
+const pricesSchema = z.record(z.string(), z.object({ usd: z.number() }))
+
+async function main() {
+  const env = getEnv()
+  const etherscanApiKey = env.string('ETHERSCAN_API_KEY')
+  const coingeckoApiKey = env.string('COINGECKO_API_KEY')
+  const provider = new providers.AlchemyProvider(
+    'homestead',
+    env.string('CONFIG_ALCHEMY_API_KEY'),
```
mateuszradomski, line 29: Do we have to use Alchemy here? Why not just a generic provider?

packages/config/scripts/omnichain/coingecko/index.ts
```diff
@@ -0,0 +1,202 @@
...
+function filterTokenAddresses(
+  coinList: {
+    symbol: string
+    name: string
+    platforms: Record<string, string | null>
+    id?: CoingeckoId
+  }[],
+  addresses: string[],
+) {
+  const tokenAddresses = coinList
```
mateuszradomski, line 170: This variable name is deceiving, it's more than just addresses. This naming goes back to `index.ts` as well

packages/config/scripts/omnichain/coingecko/index.ts
```diff
@@ -0,0 +1,202 @@
...
+  bridgedValueUsd: bigint
+}
+
+async function compareWithLive(
+  bridgedValues: BridgedValue[],
+  totalBridgedValueUsd: bigint,
+) {
+  // TODO: agEUR is not included in the total value as the token that is bridged is a wrapped version of it
+  // so coingecko doesn't have the price for it. This
+  const THRESHOLD_USD = 10_000_000n
```
mateuszradomski, line 83: This would be nice to have as a parameter to the script

packages/config/scripts/omnichain/coingecko/index.ts
```diff
@@ -0,0 +1,202 @@
...
+        dataToSave,
+        (_, value) =>
+          // eslint-disable-next-line @typescript-eslint/no-unsafe-return
+          typeof value === 'bigint' ? value.toString() : value,
+        2,
+      ),
+    )
+  } else {
+    console.log(
+      `Total value is ${diff} greater than live value. This is within the ${THRESHOLD_USD} threshold. No action required.`,
```
mateuszradomski, line 122: The first sentence is repeated from above. Total value is less actually in here

mateuszradomski (changes requested):

PR #2547, opened 2024-01-12
Exclude l3s from the main tvl

maciekzygmunt:
Resolves L2B-3568


antooni (changes requested):
Looks good, can you provide at least one escrow example? So I will test it locally

PR #2556, opened 2024-01-15
feat: add zkLink Nexus

zkTyral:
Adding zkLink Nexus (mainnet) to the layer3 project list.

Note that zkLink Nexus is a multi-chain zkRollup architecture, in addition to Linea, it currently supports zkSyncEra and will support more Layer2s in the future.

zkLink Nexus architecture documentation and mainnet contract address information:
[https://docs.zk.link/protocol/settlement-layer/multi-chain-state-synchronization/in-detail-nexus-multi-chain-state-synchronization](https://docs.zk.link/protocol/settlement-layer/multi-chain-state-synchronization/in-detail-nexus-multi-chain-state-synchronization)


packages/config/src/layer3s/zklinknexus.ts
```diff
@@ -0,0 +1,42 @@
...
+import { CONTRACTS, TECHNOLOGY, UPCOMING_RISK_VIEW } from '../common'
+import { Layer3 } from './types'
+
+export const zklinknexus: Layer3 = {
+  type: 'layer3',
+  isUpcoming: false,
+  isUnderReview: true,
+  id: ProjectId('zklinknexus'),
+  // TODO(Linea, zkSync Era): zkLink Nexus is a multi-chain zkRollup architecture. In addition to Linea, it currently supports zkSync Era and will support more Layer2s in the future.
+  hostChain: ProjectId('linea'),
```
lucadonnoh, line 12:
```suggestion
  hostChain: 'Multiple',
```
@maciekzygmunt can you add support for something like this as a short-term solution? we will refine it once we fully review the project

lucadonnoh (changes requested):

PR #2559, opened 2024-01-15
Add finality for OPStack

maciekzygmunt:
Resolves L2B-3558 L2B-3538

### DIFF
--- a/packages/backend/src/api/controllers/finality/calcAvgsPerProject.ts
+++ b/packages/backend/src/api/controllers/finality/calcAvgsPerProject.ts
@@ -1,30 +1,14 @@
 import { LivenessDataPoint } from '@l2beat/shared-pure'
 
+import { LivenessRecordWithType } from '../../../peripherals/database/LivenessRepository'
 import {
-  LivenessRecordWithProjectIdAndType,
-  LivenessRecordWithType,
-} from '../../../peripherals/database/LivenessRepository'
-import { calculateDetailsFor } from '../liveness/calculateIntervalWithAverages'
-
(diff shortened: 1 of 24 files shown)


packages/backend/src/config/Config.ts
```diff
@@ -19,6 +19,7 @@ export interface Config {
   readonly health: HealthConfig
   readonly tvl: TvlConfig
   readonly liveness: LivenessConfig | false
+  readonly finality: FinalityConfig
```
antooni, line 22: this can be just a boolean

packages/config/src/layer2s/optimism.ts
```diff
@@ -142,6 +142,10 @@ export const optimism: Layer2 = opStack({
   ),
   inboxAddress: EthereumAddress('0xFF00000000000000000000000000000000000010'),
   genesisTimestamp: new UnixTime(1686074603),
+  finality: {
```
antooni, line 145: we only set it for optimism? what about other projects?

packages/backend/src/api/controllers/finality/calcAvgsPerProject.ts
```diff
@@ -0,0 +1,30 @@
...
+}
+
+export function calcAvgsPerProject(
+  records: LivenessRecordWithType[],
+): LivenessDataPoint | undefined {
+  calculateIntervals(records)
+  return calculateDetailsFor(records, '30d')
+}
+
+export function calculateIntervals(
```
antooni, line 23: why you cannot reuse function from liveness?

antooni (approved):

PR #2586, opened 2024-01-16
fix: outdated links in Boba network

krauspt:
Hello,
I've found some outdated links in the Boba network page. This PR fixes the issue.

Hope it helps.


lucadonnoh (approved):
thanks!

PR #2609, opened 2024-01-18
Add warning bars to risk values

torztomasz:
This pull request adds warning bars to the risk values in the code. The warning bars will be displayed when there is a warning associated with a risk value.


lucadonnoh (changes requested):
add sentiment to warnings

lucadonnoh (approved):

PR #2611, opened 2024-01-18
Update Across-V2

mateuszradomski:
Resolves L2B-3262

OptimisticGovernor got discovered as a GnosisSafe module.
Smaller values for bonds in the OptimisticGovernor.
New owner added to the EmergencyProposalExecutor.


lucadonnoh:
it's weird that we don't have that contract listed and those owners shown on the project page :/

PR #2613, opened 2024-01-18
Update getFaqItems.tsx

edwardtay:
DeFi Pulse is an old reference, perhaps use growthepie instead


mateuszradomski:
Hi,

You're right, DeFi Pulse is an old reference but if we change it to growthepie the question is different and the answer becomes weird since growthepie tracks more than just the TVL. It's not really about DeFi Pulse, it's about the answer that explains that L2BEAT is more than just TVL tracking.

PR #2615, opened 2024-01-18
Add Manta Pacific milestone

edwardtay:


mateuszradomski:
Hi,

We've decided not to include as a milestone since the wording and content of it is not objective. One can assume that if we included this into our milestones we are promoting or somehow being positive this program. We try to stay as neutral as possible so I hope you understand our decision here.

edwardtay:
Totally understand, just felt would be helpful to explain what caused the activity to go parabolic

PR #2641, opened 2024-01-22
Update Chainport

mateuszradomski:
Resolves L2B-3546

Executed two proposalas 46 and 47.
The first one according to the description submitted by the Chainport Congress aims to: "Deploy Arbitrage, Upgrade Main and Side Bridge and add new Maintainer".
While the second one wants to: "Refund HODL tokens".

In the process the implementation of the Vault changed, since this contract is unverfied it's impossible to analyze what has really happened.


lucadonnoh (changes requested):
tests should fail but don't because of a cache hit which should have been a cache miss.

luke7211 (approved):

lucadonnoh (approved):

PR #2646, opened 2024-01-22
Rename devId to chain

sz-piotr:
Additionally this PR:
- enforces all chain names to be a-z0-9 and <20 characters
- removes chainsByDevId and chainsByChainId maps


michalsidzej (approved):
LGTM

PR #2659, opened 2024-01-23
Finally remove list of chain ids

sz-piotr:
Resolves L2B-3648


packages/backend/src/modules/tvl/TvlModule.ts
```diff
@@ -104,20 +105,20 @@ export function createTvlModule(
       logger,
     )
 
-  const submodules: (TvlSubmodule | undefined)[] = [
+  const submodules: TvlSubmodule[] = [
```
antooni, line 108: Let's for now write a simple assert that will ensure all of them are defined when the env is production or staging

packages/backend/src/api/controllers/status/StatusController.ts
antooni, line 1: methods that can be deleted:
- getPricesStatus
- getBalancesStatus
- getReportsStatus
- getAggregatedReportsStatus
- getTotalSuppliesStatus

those are no longer actively being used and can be removed, please decide whether this should be in the scope of this PR

antooni (changes requested):
please resolve my comments

packages/backend/src/modules/tvl/TvlModule.ts
sz-piotr: Why would this be in the scope of this PR? I haven't changed the behaviour of this code.

packages/backend/src/api/controllers/status/StatusController.ts
sz-piotr: I will make a new PR removing those

antooni (approved):

PR #2661, opened 2024-01-23
Rename Routescan to Blockscout

antooni:
Long time ago via a mistake the Blockscout was misnamed as Routescan. This PR aims to fix this issue. 
It also slightly refactors the way that tvl config is being created


packages/config/src/chains/avalanche.ts
```diff
@@ -6,10 +6,10 @@ export const avalanche: ChainConfig = {
   name: 'avalanche',
   chainId: 43114,
   explorerUrl: 'https://snowtrace.io',
-  explorerApi: {
-    url: 'https://api.snowtrace.io/api',
-    type: 'routescan',
-  },
+  // explorerApi: {
```
mateuszradomski, line 9: Why comment this out?

sz-piotr: This could simply be removed. Right now we're not calling this anywhere

packages/shared/src/services/blockscout/BlockscoutLikeClient.test.ts
```diff
@@ -19,7 +19,7 @@ describe(RoutescanLikeClient.name, () => {
         },
       })
 
-      const etherscanClient = new RoutescanLikeClient(
+      const etherscanClient = new BlockscoutLikeClient(
```
sz-piotr, line 22: Why is the variable named `etherscanClient`? Also several other instances below

packages/shared/src/services/blockscout/BlockscoutLikeClient.test.ts
```diff
@@ -131,7 +131,7 @@ describe(RoutescanLikeClient.name, () => {
         },
       })
 
-      const arbiscanClient = new RoutescanLikeClient(
+      const arbiscanClient = new BlockscoutLikeClient(
```
sz-piotr, line 134: `arbiscanClient`?

packages/shared/src/services/blockscout/UniversalBlockscoutClient.ts
```diff
@@ -3,10 +3,10 @@ import { ChainId, UnixTime } from '@l2beat/shared-pure'
 
 import { HttpClient } from '../HttpClient'
 import { BlockNumberProvider } from '../providers/BlockNumberProvider'
-import { RoutescanLikeClient } from '../routescan/RoutescanLikeClient'
+import { BlockscoutLikeClient } from './BlockscoutLikeClient'
 
-export class UniversalRoutescanClient
-  extends RoutescanLikeClient
+export class UniversalBlockscoutClient
```
sz-piotr, line 8: I really dislike having both `UniversalBlockscoutClient` and `BlockscoutLikeClient`. Same for Etherscan. IMO passing chainId to them was simply a mistake. There should only be one class - the universal one.

mateuszradomski (approved):

PR #2665, opened 2024-01-23
Configure backend through feature flags

sz-piotr:
Example use in the .env
```
FEATURE_FLAGS=tvl,!tvl.arbitrum,activity
```
Will enable things like:
- tvl
- tvl.ethereum
- tvl.optimism
- activity
- activity.ethereum

But won't enable
- tvl.arbitrum
- updateMonitor (for example)


packages/backend/src/config/FeatureFlags.ts
```diff
@@ -0,0 +1,85 @@
...
+    for (const item of items) {
+      if (item.startsWith('!')) {
+        this.disabled.add(item.substring(1))
+      } else {
+        this.enabled.add(item)
+      }
+    }
+  }
+
+  with(input: string): FeatureFlags {
```
mateuszradomski, line 29: this function name is too generic for me. I need to go to the implementation to understand what it's doing. Maybe `add`, `withFlag`, `append`, `push`. In context 

```typescript
const flags = getFlags()

// Current
flags.with('diffHistory')

// Proposed
flags.add('diffHistory')
flags.withFlag('diffHistory')
flags.append('diffHistory')
flags.push('diffHistory')
```

mateuszradomski (approved):
One gripe, maybe the proposed names would be better? Otherwise looks good!

PR #2672, opened 2024-01-24
Add JSON and JSONC files to test inputs

torztomasz:


antooni (changes requested):
I would also add it to build

antooni (approved):

PR #2675, opened 2024-01-24
Remove Sass

torztomasz:
This pull request removes the Sass dependency and updates the code to use TailwindCSS only


sz-piotr:
The more we wait on it the more work it's going to be to merge main to this branch. This change should be one and done

antooni (approved):

PR #2678, opened 2024-01-24
Add POL token

radinat:
Resolves L2B-3566


packages/config/src/tokens/tokens.jsonc
```diff
@@ -584,6 +584,10 @@
       "symbol": "PNT",
       "address": "0x89Ab32156e46F46D02ade3FEcbe5Fc4243B9AAeD"
     },
+    {
```
sz-piotr, line 587: @antooni we need a test that informs the developer that they need to run a script!

sz-piotr (approved):

PR #2688, opened 2024-01-25
Convert Lyra into Optimium

lucadonnoh:
Resolves L2B-3733


radinat (commented):
Maybe we should update the description and in the future handle cases where the OP Stack template is used and saying "L2" is not correct

radinat (approved):

PR #2701, opened 2024-01-28
Fix broken links and add milestone update in `metis` docs

pr0x2b:
### Project name: `Metis Andromeda`
- [x] typo-fix in Risk analysis
- [x] Fix broken social media links
- [x] add a new milestone

### DIFF
--- a/packages/config/src/layer2s/metis.ts
+++ b/packages/config/src/layer2s/metis.ts
@@ -40,11 +40,11 @@ export const metis: Layer2 = {
       ],
       repositories: ['https://github.com/MetisProtocol'],
       socialMedia: [
-        'https://medium.com/@MetisDAO',
+        'https://metisl2.medium.com/',
         'https://twitter.com/MetisL2',
-        'https://discord.gg/RqfEJZXnxd',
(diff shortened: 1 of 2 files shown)


packages/config/src/layer2s/metis.ts
```diff
@@ -297,6 +297,13 @@ export const metis: Layer2 = {
       link: 'https://metisdao.medium.com/decentralized-storage-goes-live-da876dc6eb70',
       date: '2022-04-12T00:00:00Z',
       description: 'Update moving data to an off-chain committee.',
+    },    
+    {
```
sz-piotr, line 301: In our understanding this is not a milestone as nothing has changed on the deployed chain. This is about the testnet and so it doesn't qualify. Please remove.

sz-piotr (changes requested):

packages/config/src/layer2s/metis.ts
pr0x2b: sure. I have reverted it now.

sz-piotr (approved):

PR #2718, opened 2024-01-30
Add Tailwind plugin to ESLint

torztomasz:
This pull request adds the Tailwind plugin to ESLint, which will enable linting for Tailwind CSS classes in the codebase.


packages/frontend/src/components/ExpandableContainer.tsx
```diff
@@ -11,28 +11,39 @@ interface ExpandableContainerProps {
 
 export function ExpandableContainer(props: ExpandableContainerProps) {
   return (
-    <div className="ExpandableContainer">
```
antooni, line 14: I thought we got rid of those classes in previous PR?

antooni:
So the only thing you have done was to add this plugin and later run it, and it resulted in this diff right?

packages/frontend/src/components/ExpandableContainer.tsx
torztomasz: Nope, these are the classes responsible for being queryable in scripts. Previously we deleted Sass classes

torztomasz:
> So the only thing you have done was to add this plugin and later run it, and it resulted in this diff right?

Yup, also had to manually fix some stuff

antooni (approved):

PR #2723, opened 2024-01-30
Add tinify-logos CLI script

torztomasz:
This pull request adds a new cli script called `tinify-logos` that resizes and optimizes logos. The script uses the Tinify API and requires the `TINIFY_API_KEY` environment variable to be set. To use the script, run `yarn tinify-logos`, it will tinify all logos that were not tinified.


packages/frontend/src/test/icons.test.ts
```diff
@@ -158,4 +36,30 @@ describe('icons', () => {
...
+  })
 })
+
+function getTinifiedLogos() {
+  const tinifiedLogosFile = path.join(
+    __dirname,
+    '..',
+    '..',
+    'cli',
+    'tinifiedLogos.json',
```
sz-piotr, line 60:
```suggestion
    '../../cli/tinifiedLogos.json',
```

packages/frontend/src/static/icons/beamerbridgev2.png
sz-piotr, line 1: Note that we're upscaling here. Doesn't look bad in this case, but maybe we should prevent that in the general case?

packages/frontend/cli/tinifiedLogos.json
```diff
@@ -0,0 +1,132 @@
+[
```
sz-piotr, line 1: maybe add file hashes here to prevent weird situations. Record<filename, hash>

packages/frontend/cli/tinifyLogos.ts
```diff
@@ -0,0 +1,89 @@
...
+    __dirname,
+    '..',
+    'src',
+    'static',
+    'icons',
+    fileName,
+  )
+
+  const source = tinify.fromFile(pathToLogo)
+  const resized = source.resize({
```
sz-piotr, line 59: we only want to downscale!

sz-piotr (changes requested):

packages/frontend/src/test/icons.test.ts
```diff
@@ -151,11 +31,28 @@ describe('icons', () => {
...
+        expect(tinifiedLogos[`${project.display.slug}.png`]).toEqual(hash)
       })
     }
   })
+
+  describe('every icon has been tinified by script', () => {
+    for (const project of projects) {
+      const id = project.id.toString()
+
+      it(id, () => {})
```
sz-piotr, line 48: Missing?

sz-piotr (changes requested):
Fix the unimplemented test

sz-piotr (approved):

PR #2750, opened 2024-02-02
Refactor stack template discovery to support backlinks

adamiak:
Resolves L2B-3039

I have confirmed that this PR renders *exactly* the same output for all .ts config files.


packages/config/src/discovery/ProjectDiscovery.ts
```diff
@@ -190,90 +188,58 @@ export class ProjectDiscovery {
...
+        if (description !== '') {
+          return {
+            name: contract.name,
+            accounts: [this.formatPermissionedAccount(contract.address)],
+            description,
+            chain: this.chain,
+          }
+        }
+      })
+      .filter((x) => !!x) as ScalingProjectPermission[]
```
mateuszradomski, line 242: I don't like this operator, it's to broad. Wouldn't it be better to return undefined above in the else clause and later filter undefined? In addition, I feel like this operator makes the code harder to read/reason about the possibilities than can happen.

We have this util function (`packages/shared-pure/src/utils/notUndefined.ts`), wdyt about using it?
```typescript
export function notUndefined<T>(value: T | undefined): value is T {
  return value !== undefined
}
```

mateuszradomski (approved):
small suggestion, overall gj!

packages/config/src/discovery/ProjectDiscovery.ts
adamiak: nice, I'll use it.

PR #2764, opened 2024-02-05
Improve frontend build logs

antooni:
Log error message
Additionally I have removed /api/blocks which is no longer being used


mateuszradomski (approved):
lgtm

PR #2782, opened 2024-02-07
Add finality indexer

maciekzygmunt:
Resolves L2B-3687 L2B-3684


packages/backend/src/core/finality/FinalityIndexer.ts
```diff
@@ -0,0 +1,169 @@
...
+  override async start(): Promise<void> {
+    this.logger.info('Starting...')
+    await this.initialize()
+    await super.start()
+  }
+
+  override async update(from: number, to: number): Promise<number> {
+    const [configurationsToSync, adjustedTo] = await this.getConfiguration(to)
+    if (configurationsToSync.length === 0) {
+      this.logger.info('Update skipped', { from, to: adjustedTo })
```
antooni, line 35: change to debug, also in liveness

packages/backend/src/core/finality/FinalityIndexer.ts
```diff
@@ -0,0 +1,169 @@
...
+        )
+      if (projectFinalityData === undefined) {
+        return
+      }
+      return {
+        projectId: config.projectId,
+        ...projectFinalityData,
+      }
+    })
+    const data = await Promise.all(finalityPromises)
```
antooni, line 100: change to for

packages/backend/src/core/finality/FinalityIndexer.ts
```diff
@@ -0,0 +1,169 @@
...
+import { ChildIndexer } from '@l2beat/uif'
+import { Knex } from 'knex'
+
+import { FinalityRepository } from '../../peripherals/database/FinalityRepository'
+import { IndexerStateRepository } from '../../peripherals/database/IndexerStateRepository'
+import { LivenessIndexer } from '../liveness/LivenessIndexer'
+import { FinalityConfig } from './types/FinalityConfig'
+import { findConfigurationsToSync } from './utils/findConfigurationsToSync'
+
+export class FinalityIndexer extends ChildIndexer {
```
antooni, line 12: add comment with explaination how it is being synced

packages/backend/src/config/makeConfig.ts
```diff
@@ -117,7 +117,13 @@ export function makeConfig(
...
       minTimestamp: UnixTime.fromDate(new Date('2023-05-01T00:00:00Z')),
     },
-    finality: flags.isEnabled('finality'),
+    finality: flags.isEnabled('finality') && {
+      ethereumProviderUrl: env.string('FINALITY_ETHEREUM_PROVIDER_URL'),
+      ethereumProviderCallsPerMinute: env.integer(
+        'FINALITY_ETHEREUM_PROVIDER_CALLS_PER_MINUTE',
+        600,
+      ),
+    },
```
michalsidzej, line 126: Please add those .env variables to readme.

packages/backend/src/modules/finality/analyzers/LineaFinalityAnalyzer.ts
```diff
@@ -29,12 +29,12 @@ export class LineaFinalityAnalyzer {
     to: UnixTime,
     granularity: number,
   ) {
-    const interval = (from.toNumber() - to.toNumber()) / granularity
+    const interval = (to.toNumber() - from.toNumber()) / granularity
```
michalsidzej, line 32: Should we assert here that interval > 0?

packages/backend/src/modules/finality/FinalityIndexer.ts
```diff
@@ -0,0 +1,184 @@
...
+    }
+
+    return data
+  }
+
+  private async initialize() {
+    await this.initializeIndexerState(this.minTimestamp.toNumber())
+  }
+
+  async initializeIndexerState(safeHeight: number) {
```
michalsidzej, line 134: This could be one function with no parameter (using this.minTimestamp)

packages/backend/src/modules/finality/FinalityIndexer.ts
```diff
@@ -0,0 +1,184 @@
...
+      .toStartOf('day')
+      .add(1, 'days')
+      .toNumber()
+
+    if (targetTo < this.minTimestamp.toNumber()) {
+      this.logger.debug(
+        'Update skipped: target earlier than minimumTimestamp',
+        { from, targetTo },
+      )
+      return targetTo
```
michalsidzej, line 53: You can just return Math.min(min, this.minTimestamp). This will result in reaching the `this.minTimestamp` in the least possible amount of steps.

packages/backend/src/modules/finality/FinalityIndexer.ts
```diff
@@ -0,0 +1,184 @@
...
+    }
+
+    if (targetTo > UnixTime.now().toNumber()) {
```
michalsidzej, line 56: This conditional will be unnecessary if you apply my previous comment

packages/backend/src/modules/finality/FinalityIndexer.ts
```diff
@@ -0,0 +1,184 @@
...
+    const targetTimestamp = new UnixTime(targetTo)
+
+    const configurationsToSync = await this.getSyncStatus(targetTimestamp)
+
+    if (configurationsToSync.length === 0) {
+      this.logger.debug('Update skipped: no configurations to sync', {
+        from,
+        to: targetTo,
+      })
+      return targetTimestamp.toNumber()
```
michalsidzej, line 69: Watch out here. This can result in `targetTimestamp > to` being returned. This will never happen in our current setup, but a change in a parentIndexer for this one can become a problem.

packages/backend/src/Application.ts
```diff
@@ -51,8 +54,14 @@ export class Application {
...
-      createLivenessModule(config, logger, database, clock),
-      createFinalityModule(config, logger, database, clock),
+      livenessModule,
+      createFinalityModule(
+        config,
+        logger,
+        database,
+        clock,
+        livenessModule?.indexer as LivenessIndexer,
+      ),
```
michalsidzej, line 64: I'm not a big fan of this approach here. The whole point of creating those modules in an array is some separation of concerns. The problem here is that the finality module depends on the liveness module, which does not fit this style. What about creating a liveness and finality module as one?

michalsidzej (approved):
Please disable it on prod.

PR #2787, opened 2024-02-07
Discover sequencerInbox for all opStack chains

michalsidzej:
Resolves L2B-3782


lucadonnoh (changes requested):
nice job!!! can you add the value in the template directly since it works for every chain?

lucadonnoh (approved):

PR #2797, opened 2024-02-08
Rename to "Time to inclusion"

maciekzygmunt:
Resolves L2B-3906


lucadonnoh:
tooltip should be updated too

torztomasz (approved):
Shouldn't we change the tooltip content for this column??

lucadonnoh (changes requested):
🙄

lucadonnoh (approved):

PR #2806, opened 2024-02-09
Liveness caching

antooni:
This is the first step into the process of fixing liveness endpoint performance. Two high level steps we need to take:

1) Stop request timeouts when querying this endpoint
2) Change liveness logic to offload work from Controller to Indexer

Description of those two steps:

1) "Ugly solution" - hide our problems by caching the response: I have decreased the memory usage of the `.getLiveness` to make it possible to cache it every 10 minutes in our Controller's memory. Now, the endpoint response is almost instant, but still we have deep underlying performance issues that need revisiting

2) Offload work to Indexer - liveness statistics such as min, max, avg, anomalies could be calculated in the Indexer that will be dependent on LivenessIndexer. When LivenessIndexer syncs to certain timestamp the results of liveness calculations will be stored to the database and then the Controller will only perform one simple and lightweight query to the DB to expose the Liveness data for frontend

Resolves L2B-3903

### DIFF
--- a/packages/backend/src/modules/liveness/api/LivenessController.ts
+++ b/packages/backend/src/modules/liveness/api/LivenessController.ts
@@ -72,35 +72,31 @@ export class LivenessController {
 
     const projects: LivenessApiResponse['projects'] = {}
 
-    await Promise.all(
-      this.projects
-        .filter((p) => !p.isArchived)
-        .map(async (project) => {
-          if (project.livenessConfig === undefined) {
-            return
+    for (const project of this.projects.filter((p) => !p.isArchived)) {
+      if (project.livenessConfig === undefined) {
+        continue
+      }
+      const records =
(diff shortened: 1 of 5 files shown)


packages/backend/src/modules/liveness/api/LivenessController.ts
```diff
@@ -72,35 +100,31 @@ export class LivenessController {
...
-        .filter((p) => !p.isArchived)
-        .map(async (project) => {
-          if (project.livenessConfig === undefined) {
-            return
-          }
-          const records =
-            await this.livenessRepository.getWithTypeDistinctTimestamp(
-              project.projectId,
-            )
+    for (const project of this.projects.filter((p) => !p.isArchived)) {
```
torztomasz, line 103: extract filter to separate variable

torztomasz (approved):

michalsidzej (approved):

PR #2811, opened 2024-02-12
Add Hypr token

radinat:
Resolves L2B-3934


lucadonnoh (changes requested):
Hypr ts file should be updated to consider the token as associated, see Manta Pacific as an example

lucadonnoh (approved):

PR #2824, opened 2024-02-13
Add zkSyncEraFinalityAnalyzer

maciekzygmunt:
Resolves L2B-3753

### DIFF
--- a/packages/backend/src/modules/finality/FinalityIndexer.test.ts
+++ b/packages/backend/src/modules/finality/FinalityIndexer.test.ts
@@ -41,12 +41,6 @@ describe(FinalityIndexer.name, () => {
     })
 
     it('correctly syncs not synced projects', async () => {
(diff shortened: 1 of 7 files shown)


packages/backend/src/modules/finality/FinalityModule.ts
```diff
@@ -56,13 +57,23 @@ export function createFinalityModule(
...
+  )
+
   const runtimeConfigurations = config.projects
     .map((p) => {
       if (p.finalityConfig?.type === 'Linea') {
         return {
           projectId: p.projectId,
           analyzer: lineaAnalyzer,
         }
+      } else if (p.finalityConfig?.type === 'zkSyncEra') {
```
torztomasz, line 72: ELSE IF:
image

torztomasz (approved):

PR #2839, opened 2024-02-15
Add discovery to near rainbow bridge

vincfurc:
Resolves L2B-3986


luke7211 (changes requested):
Contracts are not upgradable. Admin permissions require more detailed description

luke7211 (approved):

PR #2844, opened 2024-02-15
Fix address in etherscan for PolygonZkEvmBridge.sol

omnifient:
Previous address is for PolygonZkEVMUpgraded.sol, not the bridge

### DIFF
--- a/packages/config/src/layer2s/polygonzkevm.ts
+++ b/packages/config/src/layer2s/polygonzkevm.ts
@@ -408,7 +408,7 @@ export const polygonzkevm: Layer2 = {
         references: [
           {
             text: 'PolygonZkEvmBridge.sol - Etherscan source code, claimAsset function',
-            href: 'https://etherscan.io/address/0xb1585916487AcEdD99952086f2950763D253b923',
+            href: 'https://etherscan.io/address/0x2a3DD3EB832aF982ec71669E178424b10Dca2EDe',
           },
         ],
       },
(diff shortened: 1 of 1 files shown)


packages/config/src/layer2s/polygonzkevm.ts
```diff
@@ -408,7 +408,7 @@ export const polygonzkevm: Layer2 = {
         references: [
           {
             text: 'PolygonZkEvmBridge.sol - Etherscan source code, claimAsset function',
-            href: 'https://etherscan.io/address/0xb1585916487AcEdD99952086f2950763D253b923',
+            href: 'https://etherscan.io/address/0x2a3DD3EB832aF982ec71669E178424b10Dca2EDe',
```
lucadonnoh, line 411:
```suggestion
            href: 'https://etherscan.io/address/0x0feb850b183c57534b56b7d56520133c8f9bdb65',
```

packages/config/src/layer2s/polygonzkevm.ts
```diff
@@ -408,7 +408,7 @@ export const polygonzkevm: Layer2 = {
         references: [
           {
             text: 'PolygonZkEvmBridge.sol - Etherscan source code, claimAsset function',
```
lucadonnoh, line 410:
```suggestion
            text: 'PolygonZkEvmBridgeV2.sol - Etherscan source code, claimAsset function',
```

lucadonnoh (changes requested):

packages/config/src/layer2s/polygonzkevm.ts
omnifient: huh, bridging/claiming is done through the proxy (0x2a3DD3EB832aF982ec71669E178424b10Dca2EDe), not the implementation (0x0feb850b183c57534b56b7d56520133c8f9bdb65), so why point to the impl?

lucadonnoh: because we want to show the line of code that implements the functionality

PR #2874, opened 2024-02-21
Update L2 definition on FAQ page

john-light:
- Made the definition more precise and inclusive of state channel protocols
- Removed section calling out Polygon PoS specifically (could add it back but if so imo should be a separate FAQ item from the one about the L2 definition)
- Unrelated: fixed formatting on one of the FAQ items to make it consistent with the rest of the page


lucadonnoh:
we will have a bigger PR including a glossary to update the definitions, but thanks for the input

PR #2880, opened 2024-02-22
Generate TVL aggregates in controller

antooni:
Resolves L2B-4090


packages/backend/src/modules/tvl/api/generateTvlApiResponse.ts
```diff
@@ -10,7 +11,139 @@ import {
...
+  )
+
+  assert(
+    aggregate.slice(startingIndexInAggregates).length === project.length,
+    'Programmer error: Expected arrays to be aligned to the same untilTimestamp',
+  )
+
+  let projectIndex = 0
+
+  for (let i = startingIndexInAggregates; i < aggregate.length; i++) {
```
sz-piotr, line 128: Start i at 0, only increment i.

packages/backend/src/modules/tvl/api/generateTvlApiResponse.ts
```diff
@@ -10,7 +11,139 @@ import {
...
+
+  return aggregate
+}
+
+// Function to sum two chart points with the same timestamp
+function sumChartPoints(
+  a: TvlApiChartPoint,
+  b: TvlApiChartPoint,
+): TvlApiChartPoint {
+  return a.map((value, index) =>
```
sz-piotr, line 141: Instead of mapping and a new object do a for loop starting at index 1

sz-piotr (changes requested):

PR #2891, opened 2024-02-23
Review Scroll upgrade

lucadonnoh:
Resolves L2B-4112


vincfurc:
Updated implementations referenced in risk section. 
EnforcedTxGateway is set but still not enabled, contract is paused.

PR #2893, opened 2024-02-23
Add Ancient8 (full)

radinat:
Resolves L2B-4120

OP Stack chain, no difference compared to Mode (both created through Conduit)


luke7211 (changes requested):
Ancient appears to be an Optimium posting data to Celestia

PR #2910, opened 2024-02-28
Fix government review 6 thumbnail

torztomasz:
This pull request fixes the issue with the government review 6 thumbnail. The thumbnail was not displaying correctly, and this PR resolves that problem.


maciekzygmunt (approved):

antooni (changes requested):
image
#4 is wrong

PR #2931, opened 2024-03-04
Create SECURITY.md

philliprossii:
Security Policy


mateuszradomski:
Hi,

Sorry but I don't see how this file relates in any way to this repository. We're not a library, so we don't have "versions".

PR #2936, opened 2024-03-04
Create tracked txs config table & repo

maciekzygmunt:
Resolves L2B-4207 L2B-4208


packages/backend/src/peripherals/database/types.ts
```diff
@@ -171,6 +171,17 @@ declare module 'knex/types/tables' {
     average_time_to_inclusion: number
   }
 
+  interface TrackedTxConfigRow {
```
torztomasz, line 174: This type should be renamed to `TrackedTxsConfigRow`

torztomasz (changes requested):

torztomasz (changes requested):

torztomasz (approved):

PR #2949, opened 2024-03-06
Delete unifra     'https://scroll.unifra.xyz/' because it is unavailable

IronLu233:
image

We (Unifra) remove the old domain, and https://scroll.l2scan.co/ is already in the list.


mateuszradomski (approved):
Thanks for the contribution!

PR #2966, opened 2024-03-07
Add the ability to configure a TVL warning

mateuszradomski:
Resolves L2B-4239


packages/frontend/src/components/table/props/getScalingTableColumnsConfig.tsx
```diff
@@ -88,21 +89,21 @@ export function getActiveScalingSummaryColumnsConfig() {
       noPaddingRight: true,
       headClassName: '-translate-x-[72px]',
       getValue: (project) =>
-        project.slug === 'polygonzkevm' ? (
+        project.tvlWarning !== undefined ? (
```
torztomasz, line 92: lets extract it to`TotalCell` component or smth

packages/config/src/layer2s/types/Layer2.ts
```diff
@@ -88,3 +90,10 @@ export interface Layer2Config extends ScalingProjectConfig {
   /** Configuration for getting finality data */
   finality?: Layer2FinalityConfig
 }
+
+export interface Layer2TVLWarning {
+  /** Content of the warning */
+  content: string
+  /** Color with which the warning should be displayed */
+  severity: "yellow" | "red"
```
torztomasz, line 98: lets use `sentiment: Pick<Sentiment, 'yellow' | 'red'>`

torztomasz (changes requested):

packages/config/src/layer2s/types/Layer2.ts
```diff
@@ -95,5 +95,5 @@ export interface Layer2TVLWarning {
   /** Content of the warning */
   content: string
   /** Color with which the warning should be displayed */
-  severity: "yellow" | "red"
+  sentiment: 'bad' | 'warning'
```
torztomasz: lets pick it from Sentiment type: `Pick<Sentiment, 'bad' | 'warning'>`

torztomasz (approved):

packages/config/src/layer2s/types/Layer2.ts
mateuszradomski: You can't pick from a union type only from an interface/class

torztomasz: Extract was the correct utility type, sorry 👍

PR #2974, opened 2024-03-07
Add TVL fallback for projects without data

torztomasz:
Resolves L2B-4263, L2B-4253

This pull request adds a fallback mechanism for projects that do not have TVL data available. It displays a "Coming soon" badge for these projects instead of showing a blank value. Additionally, it includes a fix for sorting projects by TVL to handle cases where TVL data is missing.

### DIFF
--- a/packages/frontend/src/utils/orderByTvl.ts
+++ b/packages/frontend/src/utils/orderByTvl.ts
@@ -17,14 +17,10 @@ export function orderByTvl<
     const tvl =
       tvlApiResponse.projects[project.id.toString()]?.charts.hourly.data.at(
         -1,
-      )?.[1] ?? 0
+      )?.[1]
 
     if (tvl) {
-      return (
-        tvlApiResponse.projects[project.id.toString()]?.charts.hourly.data.at(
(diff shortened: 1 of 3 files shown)


packages/frontend/src/utils/orderByTvl.ts
```diff
@@ -9,10 +13,25 @@ export function orderByTvl<
   const archived = projects.filter((project) => project.isArchived)
   const upcoming = projects.filter((project) => project.isUpcoming)
 
-  const getTvl = (project: T) =>
-    tvlApiResponse.projects[project.id.toString()]?.charts.hourly.data.at(
-      -1,
-    )?.[1] ?? 0
+  const getTvl = (project: T) => {
```
sz-piotr, line 16: Will this work with the sorting on the table?

sz-piotr (approved):

packages/frontend/src/utils/orderByTvl.ts
torztomasz: nope but I thought it should not

PR #2980, opened 2024-03-08
Make isNodeAvailable not required for Optimiums

mateuszradomski:
Resolves L2B-3922

### DIFF
--- a/packages/config/src/layer2s/templates/opStack.ts
+++ b/packages/config/src/layer2s/templates/opStack.ts
@@ -1,5 +1,10 @@
 import { ContractParameters } from '@l2beat/discovery-types'
-import { EthereumAddress, ProjectId, UnixTime } from '@l2beat/shared-pure'
+import {
+  assert,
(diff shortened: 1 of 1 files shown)


packages/config/src/layer2s/templates/opStack.ts
```diff
@@ -91,6 +96,14 @@ export function opStack(templateVars: OpStackConfig): Layer2 {
   const daProvider =
     templateVars.daProvider ??
     (postsToCelestia ? CELESTIA_DA_PROVIDER : undefined)
+
+  if (daProvider === undefined) {
+    assert(
```
lucadonnoh, line 101: can it be done at the type level?

mateuszradomski: No, because we are now fetching things from discovery. So only at run time you'll know for sure if a chain is an optimum or not.

lucadonnoh (approved):

PR #2986, opened 2024-03-09
Add ezETH and add ezETH to mode network

agileurbanite:


antooni:
Hello @agileurbanite,

I have looked at your PR but there is a problem with explorerAPI, our backend currently does not support your custom route for Blockscout (the one where you specify chainId https://modescan.io/documentation/etherscan-compatibility/blocks).

Currently we have a lot of things to do, so probably ETA to add your token will be next week. But if you want to add it faster you can try modifying our backend, I will give you some hints:

- take a look at this route "Get block number by timestamp" https://modescan.io/documentation/etherscan-compatibility/blocks, it is being used by our BlockNumber updater, to fetch paring timestamp-blocknumber from mode network
- modify the BlockNumber provider, pass a flag to Blockscout client stating that this will use custom route https://modescan.io/documentation/etherscan-compatibility/blocks
- modify BlockNumber.call to handle your custom route https://github.com/l2beat/l2beat/blob/main/packages/shared/src/services/blockscout/BlockscoutClient.ts#L64

After you make those changes to test it locally set this env variables and run `yarn start:dev` from backend folder
```
TVL_MODE_PROVIDER_URL=https://1rpc.io/mode	
FEATURES=tvl,!tvl.*,tvl.mode
```

mateuszradomski (approved):

mateuszradomski:
Fixed the configuration, tested it locally and everything is working.

PR #3012, opened 2024-03-12
New milestone updates to Arbitrum and Fuel

ltodd22:
Both projects have had a few noteworthy milestones recently, this provides context on them.


lucadonnoh:
we usually just track upgrades and fuel refers to the old project so the milestone don't apply

PR #3014, opened 2024-03-13
Implement tracked transactions

torztomasz:


packages/backend/src/peripherals/database/types.ts
```diff
@@ -150,7 +150,7 @@ declare module 'knex/types/tables' {
     timestamp: Date
     block_number: number
     tx_hash: string
-    liveness_id: string
+    tracked_tx_id: string
   }
 
   interface LivenessConfigurationRow {
```
antooni, line 156: this type is no longer needed

packages/backend/src/Application.ts
```diff
@@ -47,6 +47,13 @@ export class Application {
     )
 
     const livenessModule = createLivenessModule(config, logger, database, clock)
```
antooni, line 49: why do we need this module? can it be moved to trackedTxsModule?

packages/config/src/layer2s/polygonzkevm.ts
```diff
@@ -192,7 +186,10 @@ export const polygonzkevm: Layer2 = {
           sinceTimestamp: new UnixTime(1679653163),
           untilTimestamp: new UnixTime(1707822059),
         },
-        {
+      },
+      {
+        uses: [{ type: 'liveness', subtype: 'batchSubmissions' }],
```
antooni, line 191: Just a random thought, isn't this property more like `usedBy` than `uses` ?

packages/frontend/src/pages/scaling/liveness/props/getScalingLivenessView.tsx
```diff
@@ -130,11 +130,11 @@ function typeToDisplayType(
   anomaly: NonNullable<LivenessApiProject['anomalies']>[0],
 ) {
   switch (anomaly.type) {
-    case 'DA':
+    case 'batchSubmissions':
```
antooni, line 133: I like this change, acronyms are nice in theory, later it is hard to use them while discussing code

packages/backend/src/config/Config.ts
```diff
@@ -29,7 +29,7 @@ export interface Config {
   readonly health: HealthConfig
   readonly tvl: TvlConfig
   readonly tvl2: Tvl2Config | false
-  readonly liveness: LivenessConfig | false
+  readonly trackedTxsConfig: TrackedTxsConfig | false
```
antooni, line 32: Another example that we should get rid of liveness module

packages/backend/src/modules/tracked-txs/repositories/TrackedTxsConfigsRepository.ts
```diff
@@ -0,0 +1,170 @@
...
+    const insertedRows = await knex('tracked_txs_configs')
+      .insert(records.map(toRow))
+      .returning('id')
+
+    return insertedRows
+      .map((row) => (row.id ? TrackedTxId.unsafe(row.id) : undefined))
+      .filter(notUndefined)
+  }
+
+  // TODO: (tracked_txs_status) function useful for implementing tracked txs status
```
antooni, line 61: we can try leaving it there, but I feel it would be better to describe it in the task in Linear, with maybe hash to the code version that had this code in liveness repository

packages/backend/src/modules/tracked-txs/TrackedTxsIndexer.ts
```diff
@@ -38,88 +42,100 @@ export class LivenessIndexer extends ChildIndexer {
   }
 
   override async update(from: number, to: number): Promise<number> {
-    const [configurations, adjustedTo] = await this.getConfiguration(from, to)
+    const unixFrom = new UnixTime(from)
```
antooni, line 45: this is not needed, it is just used in one place

packages/backend/src/modules/tracked-txs/TrackedTxsModule.ts
```diff
@@ -0,0 +1,81 @@
...
+
+export function createTrackedTxsModule(
+  config: Config,
+  logger: Logger,
+  database: Database,
+  clock: Clock,
+  livenessUpdater: LivenessUpdater | undefined,
+): ApplicationModuleWithIndexer<TrackedTxsIndexer> | undefined {
+  if (!config.trackedTxsConfig || !livenessUpdater) {
+    logger.info('Tracked transactions module disabled')
```
antooni, line 24: TrackedTxsModule module disabled

packages/backend/src/Application.ts
torztomasz: Well, we are creating the updater and router inside

packages/backend/src/modules/tracked-txs/utils/findConfigurationsToSync.ts
```diff
@@ -0,0 +1,54 @@
+import { assert } from '@l2beat/backend-tools'
+import { notUndefined, UnixTime } from '@l2beat/shared-pure'
+
+import { TrackedTxsConfigRecord } from '../repositories/TrackedTxsConfigsRepository'
+import { TrackedTxConfigEntry } from '../types/TrackedTxsConfig'
+import { isTimestampInRange } from './isTimestampInRange'
+
+export function findConfigurationsToSync(
```
antooni, line 8: Please provide a comment with the simple A,B,C example of until timestamp being used

packages/backend/src/modules/tracked-txs/TrackedTxsIndexer.ts
```diff
@@ -38,88 +42,98 @@ export class LivenessIndexer extends ChildIndexer {
...
-    return adjustedTo.toNumber()
+    return syncTo.toNumber()
   }
 
   async getConfiguration(
     from: number,
     to: number,
-  ): Promise<[LivenessConfigEntry[], UnixTime]> {
+  ): Promise<[TrackedTxConfigEntry[], UnixTime]> {
     const adjustedTo = adjustToForBigqueryCall(from, to)
```
antooni, line 83: .getConfiguration does too many things, I think we can move this line to .update

packages/backend/src/modules/tracked-txs/TrackedTxsIndexer.ts
```diff
@@ -38,88 +42,98 @@ export class LivenessIndexer extends ChildIndexer {
...
+      adjustedTo: syncTo,
       usedConfigurations: configurations.length,
-      fetchedDataPoints: data.length,
+      fetchedTxsCount: txs.length,
     })
-    return adjustedTo.toNumber()
+    return syncTo.toNumber()
   }
 
   async getConfiguration(
```
antooni, line 79: getConfigurationsToSync

packages/backend/src/modules/tracked-txs/TrackedTxsIndexer.ts
```diff
@@ -38,88 +42,98 @@ export class LivenessIndexer extends ChildIndexer {
...
-    return [configurationsToSync, adjustedTo]
+    return [configurationsToSync, syncTo]
   }
 
-  private async initialize() {
-    const databaseEntries = await this.configurationRepository.getAll()
-    const { toAdd, toRemove, toTrim } = diffLivenessConfigurations(
-      this.runtimeConfigurations,
+  private async initialize(): Promise<void> {
+    this.logger.info('Initializing...')
```
antooni, line 98: Move this log to the end and say Initialized

antooni: Do the same with "Starting..."

packages/backend/src/modules/tracked-txs/TrackedTxsIndexer.ts
```diff
@@ -38,88 +42,98 @@ export class LivenessIndexer extends ChildIndexer {
...
+    const databaseEntries = await this.configRepository.getAll()
+    const { toAdd, toRemove, toTrim } = diffTrackedTxConfigurations(
+      this.configs,
       databaseEntries,
     )
 
+    this.logger.info('Modifying tracked txs configs', {
```
antooni, line 105: This log should be conditional

packages/backend/src/modules/tracked-txs/TrackedTxsIndexer.ts
```diff
@@ -38,88 +42,98 @@ export class LivenessIndexer extends ChildIndexer {
...
     const safeHeight = getSafeHeight(databaseEntries, toAdd, this.minTimestamp)
 
     await this.stateRepository.runInTransaction(async (trx) => {
-      await this.configurationRepository.addMany(toAdd, trx)
-      // this will also delete records from "liveness" using CASCADE constraint
-      await this.configurationRepository.deleteMany(toRemove, trx)
+      await this.configRepository.addMany(toAdd, trx)
+      await this.configRepository.deleteMany(toRemove, trx)
       await this.trimConfigurations(toTrim, trx)
       await this.initializeIndexerState(safeHeight, trx)
```
antooni, line 117: Maybe inside this function add a log that it was initialiezed for the first time

packages/backend/src/modules/tracked-txs/TrackedTxsModule.ts
```diff
@@ -0,0 +1,98 @@
...
+    trackedTxsConfigsRepository,
+    runtimeConfigurations,
+    config.trackedTxsConfig.minTimestamp,
+  )
+
+  const start = async () => {
+    logger = logger.for('TrackedTxsModule')
+    logger.info('Starting...')
+
+    const areUpdatersEmpty = Object.values(updaters).every((u) => !u)
```
antooni, line 76: IMO this code should not be there, in such case TrackedTxIndexer will have empty configurations and there is a code to handle that already. This line of code adds a false notion that TrackedTxsIndexer relies on updaters - but it  rather relies on configurations

packages/backend/src/modules/tracked-txs/TrackedTxsIndexer.ts
```diff
@@ -0,0 +1,220 @@
...
+  override async start(): Promise<void> {
+    await this.initialize()
+    await super.start()
+    this.logger.info('Started')
+  }
+
+  override async update(from: number, to: number): Promise<number> {
+    const { from: unixFrom, to: unixTo } = adjustRangeForBigQueryCall(from, to)
+
+    const [configurations, syncTo] = await this.getConfigurationToSync(
```
antooni, line 57:
```suggestion
    const [configurations, syncTo] = await this.getConfigurationsToSync(
```

packages/backend/src/modules/tracked-txs/TrackedTxsIndexer.ts
```diff
@@ -0,0 +1,220 @@
...
+    })
+    return syncTo.toNumber()
+  }
+
+  async getConfigurationToSync(
+    from: UnixTime,
+    to: UnixTime,
+  ): Promise<[TrackedTxConfigEntry[], UnixTime]> {
+    const databaseEntries = await this.configRepository.getAll()
+    const updaterTypes = Object.keys(this.updaters)
```
antooni, line 99:
```suggestion
    const availableUpdaters = Object.keys(this.updaters)
```

packages/backend/src/modules/tracked-txs/utils/findConfigurationsToSync.ts
```diff
@@ -0,0 +1,89 @@
...
+- {configurationsToSync: [config1, config2, config4], syncTo: 60}
+
+Next range from 60 to 100 will be:
+
+- {configurationsToSync: [config1, config2], syncTo: 100}
+
+*/
+
+export function findConfigurationsToSync(
+  updaterTypes: string[],
```
antooni, line 39:
```suggestion
  availableUpdaters: string[],
```

packages/backend/src/modules/tracked-txs/utils/findConfigurationsToSync.test.ts
```diff
@@ -0,0 +1,247 @@
...
+      FROM.add(1, 'hours'),
+    )
+
+    expect(result).toEqual({
+      configurationsToSync: [],
+      syncTo: FROM.add(1, 'hours'),
+    })
+  })
+
+  it('should skip configuration uses that are not enabled', () => {
```
antooni, line 170:
```suggestion
  it('should skip configurations which updater is not available', () => {
```

packages/backend/src/modules/tracked-txs/TrackedTxsIndexer.ts
antooni: or maybe enabled

packages/backend/src/modules/tracked-txs/utils/findConfigurationsToSync.ts
antooni: or maybe enabled

packages/backend/src/modules/tracked-txs/types/TrackedTxsConfig.ts
```diff
@@ -0,0 +1,48 @@
...
+  | TrackedTxTransferConfig
+  | TrackedTxSharpSubmissionConfig
+
+export interface TrackedTxUseWithId extends Layer2TrackedTxUse {
+  id: TrackedTxId
+}
+
+interface TrackedTxConfigBase {
+  projectId: ProjectId
+  uses: TrackedTxUseWithId[]
```
antooni, line 26: I still feel like this should be named "usedBy"

packages/backend/src/modules/tracked-txs/TrackedTxsIndexer.ts
```diff
@@ -0,0 +1,220 @@
...
+      from,
+      to,
+    )
+
+    return [configurationsToSync, syncTo]
+  }
+
+  private async initialize(): Promise<void> {
+    const databaseEntries = await this.configRepository.getAll()
+    const { toAdd, toRemove, toTrim } = diffTrackedTxConfigurations(
```
antooni, line 114: we also need to add this "enabled updaters" context to this diffTrackedTxConfigurations functions, we do not want to add a configuration if updater for it is not enabled

packages/backend/src/modules/tracked-txs/types/TrackedTxsConfig.ts
torztomasz: I will keep it for now, later we can change it

packages/backend/src/modules/tracked-txs/TrackedTxsIndexer.ts
torztomasz: Why not? We should add it but do not sync.

packages/backend/src/modules/tracked-txs/TrackedTxsModule.ts
torztomasz: good catch

antooni (approved):
GG WP

PR #3042, opened 2024-03-14
Metis updates: One multisig signer is replaced

sekuba:
Resolves L2B-4411.


lucadonnoh (changes requested):
multisig owners change too often and we have no idea who those people are. please change the severity to low and do the same to other multisigs. the only important multisigs that we want to track closer are security councils, which should be medium or even high severity since they affect the stages designation

lucadonnoh:
also: DAC have high severity because it affects DA risk assessment

lucadonnoh:
also: please don't use merge when not necessary because it messes up git blame / history and creates confusion, the PR when merged will take care of this itself. But if you really want to do it, use rebase instead:
image

lucadonnoh (approved):

PR #3043, opened 2024-03-14
Immutable zkEVM: add flowrate buckets description to .ts

sekuba:
This is a cleaned PR of the one i made yesterday. It adds descriptions of the 'flowrate buckets' mechanism in the bridge.


lucadonnoh (approved):
LGTM !

PR #3044, opened 2024-03-14
Convert Astar to Validium

lucadonnoh:
as requested via telegram

### DIFF
--- a/packages/config/src/layer2s/astarzkevm.ts
+++ b/packages/config/src/layer2s/astarzkevm.ts
@@ -7,7 +7,7 @@ export const astarzkevm: Layer2 = upcoming({
     name: 'Astar zkEVM',
     slug: 'astarzkevm',
     description:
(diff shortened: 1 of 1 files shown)


packages/config/src/layer2s/astarzkevm.ts
```diff
@@ -9,7 +9,7 @@ export const astarzkevm: Layer2 = upcoming({
     description:
       "Astar zkEVM is an Ethereum Layer 2 solution that leverages Polygon's CDK and zero-knowledge cryptography to enable off-chain transactions while maintaining EVM equivalence.",
```
radinat, line 10:
```suggestion
      "Astar zkEVM is a Validium that leverages Polygon's CDK and zero-knowledge cryptography to enable off-chain transactions while maintaining EVM equivalence.",
```

radinat (approved):

PR #3057, opened 2024-03-14
Decode Arbtrium governance scheduled transactions

adamiak:
Resolves L2B-4028


adamiak:
This will ultimately be done as a handler in discovery, not a separate script. See https://github.com/l2beat/tools/pull/158

PR #3065, opened 2024-03-15
Implement PriceIndexer

antooni:
Resolves L2B-4282

### DIFF
--- a/packages/backend/src/modules/tvl2/PriceIndexer.ts
+++ b/packages/backend/src/modules/tvl2/PriceIndexer.ts
@@ -35,8 +35,9 @@ export class PriceIndexer extends ChildIndexer {
 
   override async update(_from: number, _to: number): Promise<number> {
     const from = this.syncService.getTimestampToSync(new UnixTime(_from))
-    const to = this.syncService.getTimestampToSync(from)
+    const to = this.syncService.getTimestampToSync(from.add(1, 'hours'))
 
+    //todo figure out the flow with tvl cleaner, should it write more to the db which will be cleaned up later?
     const prices = await this.coingeckoQueryService.getUsdPriceHistory(
       this.priceConfigEntry.coingeckoId,
       from,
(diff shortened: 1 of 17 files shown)


packages/config/src/tokens/types.ts
```diff
@@ -18,6 +18,7 @@ export const GeneratedToken = z.object({
   symbol: z.string(),
   decimals: z.number(),
   deploymentTimestamp: numberAs((n) => new UnixTime(n)),
+  // This is either coingecko listing timestamp or min timestamp of ethereum
```
sz-piotr, line 21: Why?

packages/backend/src/modules/tvl2/PriceIndexer.ts
```diff
@@ -0,0 +1,142 @@
...
+export class PriceIndexer extends ChildIndexer {
+  indexerId: string
+
+  constructor(
+    logger: Logger,
+    parentIndexer: HourlyIndexer,
+    private readonly coingeckoQueryService: CoingeckoQueryService,
+    private readonly stateRepository: IndexerStateRepository,
+    private readonly pricesRepository: PricesRepository,
+    private readonly token: PriceConfigEntry,
```
sz-piotr, line 21: I would actually see this done slightly differently. The PriceIndexer would receive a single coingeckoId and multiple PriceConfigEntries with this coingeckoId set. Then it will start downloading data from the min of those entries timestamps and then save to the entries that need it. That way we're not redownloading data for the same coingeckoId.

sz-piotr (changes requested):

sz-piotr (approved):

packages/backend/src/modules/tvl2/Tvl2Module.ts
```diff
@@ -1,24 +1,75 @@
...
 import { createTvl2StatusRouter } from './api/Tvl2StatusRouter'
+import { PriceIndexer } from './PriceIndexer'
+import { PriceRepository } from './repositories/PriceRepository'
+import { SyncOptimizer } from './SyncOptimizer'
 
 export function createTvl2Module(
   config: Config,
   logger: Logger,
+  http: HttpClient,
+  database: Database,
```
sz-piotr, line 24: Please use the new Peripherals class

sz-piotr (changes requested):

sz-piotr (approved):

PR #3092, opened 2024-03-18
Add stack to L3s in review

sekuba:
Resolves L2B-4370


lucadonnoh (changes requested):
the logo is not correct (not even centered), use the one on stack.so that has gray outline and black inner filling

lucadonnoh (approved):

PR #3099, opened 2024-03-19
Severity of program hash change is now HIGH for starknet

sekuba:


packages/backend/discovery/starknet/ethereum/meta.json
```diff
@@ -690,7 +690,7 @@
         },
         "programHash": {
           "description": "The hash changes when the L2 cairo state machine logic changes.",
```
lucadonnoh, line 692: use the description that you provided in paradex, i really liked that

lucadonnoh (changes requested):

PR #3140, opened 2024-03-21
Rhino.fi disc update: 6 new signers added to gov Multisig

sekuba:
Resolves L2B-4509


lucadonnoh (changes requested):
add this multisig in the permissions section

sekuba:
> add this multisig in the permissions section

was only listed as a governor, is now listed additionally as a multisig

lucadonnoh (approved):

PR #3157, opened 2024-03-21
Track Scroll time to inclusion

sdlyy:
Resolves L2B-4515


packages/backend/src/modules/finality/analyzers/ScrollFinalityAnalyzer.ts
```diff
@@ -0,0 +1,98 @@
...
+      version: decodedInput._version,
+      parentBatchHeader: decodedInput._parentBatchHeader,
+      chunks: decodedInput._chunks,
+      skippedL1MessageBitmap: decodedInput._skippedL1MessageBitmap,
+      /* eslint-enable @typescript-eslint/no-unsafe-assignment */
+    })
+  }
+
+  /**
+   * @see https://github.com/scroll-tech/scroll/blob/develop/contracts/src/libraries/codec/ChunkCodec.sol#L13
```
michalsidzej, line 64: Invalid link. There's V0 and V1 now.

packages/backend/src/modules/finality/analyzers/ScrollFinalityAnalyzer.ts
```diff
@@ -0,0 +1,98 @@
...
+    const decodedTransactionData = this.decodeTransaction(tx.data)
+
+    const blockNumbers = decodedTransactionData.chunks
+      .map((chunk) => this.decodeChunkBlockNumbers(chunk))
+      .flat()
+
+    const firstBlockInData = Math.max(...blockNumbers)
+    const lastBlockInData = Math.min(...blockNumbers)
+
+    const l2Blocks = await Promise.all([
```
michalsidzej, line 38: No need to call l2Rpc here. You can decode the timestamp from the block comtext in the batch: https://github.com/scroll-tech/scroll/blob/develop/contracts/src/libraries/codec/ChunkCodecV1.sol#L22-L23

michalsidzej: Also, this does not change across the versions: https://github.com/scroll-tech/scroll/blob/develop/contracts/src/libraries/codec/ChunkCodecV0.sol#L20-L21

sdlyy: L2 calls dropped

michalsidzej (approved):
Please disable scroll finality on production before merging that. Let's see if there's no problems on staging first.

PR #3163, opened 2024-03-22
Add the zk-newcrypto risk to the zkRollups that were missing it

sekuba:
Resolves L2B-4537


lucadonnoh (changes requested):
i'll make some changes

lucadonnoh (approved):

PR #2883, opened 2024-02-23
Add Upgrades & Governance section for Starknet

mzoettl:
starknet-upgrade-governance

PR #3170, opened 2024-03-22
Add Starknet Upgrades/Gov section

sekuba:
Part of L2B-4521

This is an external contribution by @mzoettl 
I reviewed, then revised it slightly and added values from discovery where possible. Original PR was #2883


packages/config/src/layer2s/starknet.ts
```diff
@@ -589,6 +589,18 @@ export const starknet: Layer2 = {
...
+The Upgrading mechanism of Starknet follows a similar scheme for all of their smart contracts. A contract initializes with the creator of the contract as a Governor, who can then nominate or remove other Governors allowing them to call restricted governor functions.
+
+The Starknet core contract is upgradable by 2 appointed \`Starknet Proxy Governors\`: A Proxy multisig with a ${discovery.getMultisigStats('ProxyMultisig')} threshold and an EOA. Implementations can be upgraded ${starknetDelaySeconds === 0 ? 'without delay, thus users are not provided with an exit window in case of unwanted upgrades.' : 'with a delay of ' + formatSeconds(starknetDelaySeconds) + '.'}
+
+\`Starknet Implementation Governors\` have the authority to execute governed functions that modify contract parameters without delay. These actions encompass registering/removing Operators, specifying the program and config hash, or setting the Message Cancellation Delay between L1 and L2. Currently it is governed by a Multisig with a ${discovery.getMultisigStats('ImplementationMultisig')} threshold and an EOA. The verifier address is set upon initialization of the Starknet Implementation contract.
```
lucadonnoh, line 597: here there should be an assert that makes sure that indeed there is a multisig and a EOA

lucadonnoh (changes requested):

lucadonnoh (approved):

PR #2816, opened 2024-02-12
feat(frontend): add Upgrades & Governance to Polygon zkEVM

alexhooketh:
Added "Upgrades & Governance" section to Polygon zkEVM, diagram with source included

Onchain contracts & data match what's written it the [docs](https://docs.polygon.technology/zkEVM/), but I didn't find anything related to escrows there. USDC escrow seems to still be active (others have 0x00 admins) so I added it to the diagram

PR #3174, opened 2024-03-23
Add Upgrades and Gov section to Polygon zkEvm

sekuba:
This is an external contribution by @alexhooketh
I reviewed and revised their PR #2816 and the included diagram.

Part of L2B-4521


lucadonnoh (changes requested):
this is not enough, try to understand how the `updateRollup` function works

sekuba:
- do you mean that i should further detail how an upgrade would go through the timelock? (schedule...execution)
- or describe how parts can be updated too? (like verifier, root, bridge, PolygonZkEVMEtrog..)? rolluptypeid explanation?
all of it, sth else? :D

sekuba:
In RollupManager, the `_UPDATE_ROLLUP_ROLE` needed to call 'updateRollup()' is initialized to the timelock.
The AdminMultisig was granted the role tonight though: [in this transaction (through timelock emergency)](https://ethtx.info/mainnet/0xb1f0e00b6642adb65f10521ba7fa349b6aeb0a0e09697057b964eff6324f44c6/), so in the current state i think you are correct (the AdminMultisig can directly update verifier and / or implementation).

But i'm assuming this is only an emergency measure and not the usual path as the emergency state is needed. For this to be correct the `_UPDATE_ROLLUP_ROLE` should be renounced soon by the AdminMultisig though, and i do not understand why the update did not just go through the emergency path of the timelock tonight.

As i understand it i could either describe the state as it was until yesterday, or as it is at the moment (timelock bypassable by admin multisig for upgrades).

lucadonnoh (approved):

PR #3184, opened 2024-03-25
Handle Polygon CDK shared contracts

vincfurc:
Resolves L2B-4138 L2B-3731


packages/config/src/layer2s/templates/polygonCDKStack.ts
```diff
@@ -0,0 +1,548 @@
...
+              'https://etherscan.io/address/0x5132A183E9F3CB7C848b0AAC5Ae0c4f0491B7aB2',
+            ],
+          },
+        ],
+      },
+      dataAvailability: {
+        ...riskViewDA(daProvider),
+        description:
+          riskViewDA(daProvider).description +
+          ' Unlike most ZK rollups transactions are posted instead of state diffs.',
```
vincfurc, line 313: To remove as it's not true for all cdk chains, i.e. validiums post only hash of transactions

packages/config/src/layer2s/polygonzkevm.ts
```diff
@@ -500,6 +211,7 @@ export const polygonzkevm: Layer2 = {
       },
     ],
   },
+  isForcedBatchDisallowed,
```
vincfurc, line 214: I don't think this does the job here

packages/config/src/layer2s/templates/polygonCDKStack.ts
```diff
@@ -0,0 +1,577 @@
...
+        ],
+        references: [],
+      },
+      forceTransactions: templateVars.nonTemplateTechnology
+        ?.forceTransactions ?? {
+        ...FORCE_TRANSACTIONS.SEQUENCER_NO_MECHANISM,
+        description:
+          'The mechanism for allowing users to submit their own transactions is currently disabled.',
+      },
+      exitMechanisms: templateVars.nonTemplateTechnology?.exitMechanisms ?? [],
```
vincfurc, line 445: rather than empty this can be default as they share the same bridge / claimAsset function:
      {
        ...EXITS.REGULAR('zk', 'merkle proof'),
        references: [
          {
            text: 'PolygonZkEvmBridgeV2.sol - Etherscan source code, claimAsset function',
            href: 'https://etherscan.io/address/0x0feb850b183c57534b56b7d56520133c8f9bdb65',
          },
        ],
      },

packages/config/src/layer2s/templates/polygonCDKStack.ts
```diff
@@ -482,6 +493,7 @@ export function polygonCDKStack(templateVars: PolygonCDKStackConfig): Layer2 {
           templateVars.rollupModuleContract.name,
           {
             description: `The main contract of the ${templateVars.display.name}. Contains sequenced transaction batch hashes and forced transaction logic.`,
```
vincfurc, line 495: I suppose this is taken from Astar but it does not apply to polygon zkEVM or cdk rollups. For Polygon it was "Contains sequencing and forced transaction logic.", we can have this for cdk rollups and  "Contains sequenced transaction batch hashes and forced transaction logic." for validiums as defaults

vincfurc: or we can do with "Contains sequencing" for both and be less specific

lucadonnoh: i think it's ok as it is

lucadonnoh (changes requested):
image

lucadonnoh:
the rest LGTM

lucadonnoh (approved):

PR #3196, opened 2024-03-26
Add Syndicate Frame L3 in review

sekuba:
Resolves L2B-4586


packages/config/src/layer3s/syndicateframe.ts
```diff
@@ -0,0 +1,31 @@
...
+  id: 'syndicateframe',
+  hostChain: ProjectId('base'),
+  display: {
+    name: 'Syndicate Frame Chain',
+    slug: 'syndicateframe',
+    category: 'Optimium',
+    description:
+      'Syndicate Frame Chain is an OP Stack L3 on Base with Celestia underneath unlocking high throughput and extremely low gas costs for Frame developers.',
+    purposes: ['Social', 'NFT'],
+    provider: 'Arbitrum Orbit',
```
radinat, line 16: Shouldn't the provider be OP Stack? According to the description

sekuba: yes, sorry!

radinat (approved):

PR #3198, opened 2024-03-27
Add zklighter as upcoming project

sekuba:
Resolves L2B-4588


packages/config/src/layer2s/zklighter.ts
```diff
@@ -0,0 +1,23 @@
+import { upcoming } from './templates/upcoming'
+import { Layer2 } from './types'
+
+export const zklighter: Layer2 = upcoming({
+  id: 'zklighter',
+  display: {
+    name: 'zkLighter',
+    slug: 'zklighter',
+    description:
+      'zkLighter is an efficient order book L2 on Ethereum - low cost, low latency, verifiable matching.',
```
radinat, line 10: We aim to not use "L2" for Validiums as it doesn't fit our definition for an L2

sekuba: okay!

radinat (approved):

PR #3204, opened 2024-03-27
Implement AmountRepository

antooni:
Resolves L2B-4594


packages/backend/src/modules/tvl2/repositories/AmountConfigurationRepository.ts
```diff
@@ -0,0 +1,141 @@
...
+  address: EthereumAddress | 'native'
+  escrowAddress?: EthereumAddress
+  origin: 'canonical' | 'external' | 'native'
+  type: 'totalSupply' | 'circulatingSupply' | 'escrow'
+  includeInTotal: boolean
+  sinceTimestampInclusive: UnixTime
+  untilTimestampExclusive?: UnixTime
+}
+
+export class AmountConfigurationRepository extends BaseRepository {
```
sz-piotr, line 39: I think it might be good to actually have this be in AmountRepository. They are very tightly connected and if you are reading one you are likely reading the other.

sz-piotr (changes requested):

sz-piotr (approved):

PR #3205, opened 2024-03-27
Add STONE on Mode Network

dananderson3999:
added STONE (StakeStone ETH) as a token on Mode Network, using LayerZero bridge


dananderson3999:
sorry, duplicate!

PR #3207, opened 2024-03-27
Add blobstream discovery

michalsidzej:
Resolves L2B-4635
Resolves L2B-4604


lucadonnoh (changes requested):
image

this verifiers need better discovery, have a look at the FunctionRegistry contract

lucadonnoh (approved):

PR #3224, opened 2024-03-28
feat: update RSS3 logo and name

HenryQW:
RSS3 had a new branding, see: https://rss3.io/blog/en/the-rss3-branding-evolution


packages/config/src/layer2s/rss3.ts
```diff
@@ -20,7 +20,7 @@ const FINALIZATION_PERIOD_SECONDS: number = discovery.getContractValue<number>(
 export const rss3: Layer2 = opStack({
   discovery,
   display: {
-    name: 'RSS3',
+    name: 'RSS3 Value Sublayer',
```
lucadonnoh, line 23: short name should be added

lucadonnoh (changes requested):

PR #3231, opened 2024-03-29
Add dcamonster L3 upcoming

sekuba:


packages/frontend/cli/tinifyLogos.ts
```diff
@@ -19,7 +19,7 @@ async function main() {
     If the limit is reached (500 contributions), create a new one for yourself at https://tinypng.com/developers.
   */
   const apiKey =
-    process.env.TINIFY_API_KEY ?? 'kqzJMkBtL5wmJk1JVR7l6BNbhvZBf4wS'
+    process.env.TINIFY_API_KEY ?? 'yVNSp71vjB7Hn0zmCpRVTl2Z1dznT1Gw'
```
radinat, line 22: Should this one be updated?

sekuba:
dca.monster seems to be testnet-only so far

PR #3236, opened 2024-03-29
Calculate finality for all transactions

michalsidzej:
Resolves L2B-4681

## Context

Too many calls to db crush our backend. With those performance improvements we can easily decrease db calls by 144x (1 per day instead on 24 * 6).

Own implementation of rlp decoding is more than 50x faster than the one from ethers.js (thanks @sz-piotr). With the blobClient being stable now, we are only network bounded and we can calculate finality for all the transactions (including the out of order transactions!)


sdlyy (approved):
Looks good, and I like it. I'll proceed with the second round of review tomorrow. My only concern is regarding the number of calls within the specified timeframe. With this approach, we're actively transitioning from making an excessive number of database calls to making RPC calls 'n+x' times. Please double-check the number of RPC calls issued because, from my understanding, the granularity might have limited the number of transactions we considered. In other words, these queries might have overlooked some transactions. I might be mistaken, but I'll verify this tomorrow.

PR #3241, opened 2024-04-02
Add Degen L3 in review

sekuba:
Resolves L2B-4657


packages/config/src/layer3s/syndicatedegen.ts
```diff
@@ -0,0 +1,29 @@
+import { ProjectId } from '@l2beat/shared-pure'
+
+import { underReviewL3 } from '../layer2s/templates/underReview'
+import { Layer3 } from './types'
+
+export const syndicateframe: Layer3 = underReviewL3({
```
radinat, line 6:
```suggestion
export const syndicatedegen: Layer3 = underReviewL3({
```
Actually I think we can just use `degen` or `degenchain` (here and the rest of the files) and not include infra provider names, unless there is a reason about it?

sekuba: I used it for consistency because syndicate launched another chain called frame shortly before this and that one was called syndicateframe because 1) they branded it [Syndicate Frame Chain](https://syndicate.io/blog/syndicate-frame-chain) 2) We already have [Frame](https://l2beat.com/scaling/projects/frame)

Syndicate is kind of the governing body and higher lvl infra provider in one i guess, they use Conduit etc. for lowlvl infra and provide a 'transaction cloud' for devs to do something similar to account abstraction with semi-custody.

I am fine with either leaving this as is and call all syndicate chains accordingly in the code or changing only degen to leave out syndicate. What do you think?

lucadonnoh: i vote for `degen`, it makes sense to disambiguate only when it's needed

sekuba: renamed

lucadonnoh (approved):

PR #3258, opened 2024-04-03
Track starknet finality

sdlyy:
Resolves L2B-4133

### DIFF
--- a/packages/backend/src/modules/finality/analyzers/LineaFinalityAnalyzer.ts
+++ b/packages/backend/src/modules/finality/analyzers/LineaFinalityAnalyzer.ts
@@ -33,9 +33,9 @@ const iface = new utils.Interface([calldataFn, blobFn])
 
 export class LineaFinalityAnalyzer extends BaseAnalyzer {
   constructor(
-    protected override readonly provider: RpcClient,
-    protected override readonly livenessRepository: LivenessRepository,
-    protected override readonly projectId: ProjectId,
+    provider: RpcClient,
+    livenessRepository: LivenessRepository,
+    projectId: ProjectId,
     private readonly l2Provider: RpcClient,
   ) {
     super(provider, livenessRepository, projectId)
--- a/packages/backend/src/modules/finality/analyzers/LineaFinalityAnalyzer.ts
+++ b/packages/backend/src/modules/finality/analyzers/LineaFinalityAnalyzer.ts
@@ -1,6 +1,12 @@
-import { assert, TrackedTxsConfigSubtype, UnixTime } from '@l2beat/shared-pure'
+import {
+  ProjectId,
+  TrackedTxsConfigSubtype,
+  UnixTime,
+} from '@l2beat/shared-pure'
 import { utils } from 'ethers'
 
+import { RpcClient } from '../../../peripherals/rpcclient/RpcClient'
+import { LivenessRepository } from '../../tracked-txs/modules/liveness/repositories/LivenessRepository'
 import { BaseAnalyzer } from './types/BaseAnalyzer'
 
 type Decoded = [[string, string, string, number, number]]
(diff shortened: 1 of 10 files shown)


packages/backend/src/modules/finality/analyzers/LineaFinalityAnalyzer.ts
```diff
@@ -26,6 +32,15 @@ const blobFn = `function submitBlobData(
 const iface = new utils.Interface([calldataFn, blobFn])
 
 export class LineaFinalityAnalyzer extends BaseAnalyzer {
+  constructor(
+    protected override readonly provider: RpcClient,
+    protected override readonly livenessRepository: LivenessRepository,
+    protected override readonly projectId: ProjectId,
```
michalsidzej, line 38:
```suggestion
    provider: RpcClient,
    livenessRepository: LivenessRepository,
    projectId: ProjectId,
```
No need to override

packages/backend/src/modules/finality/analyzers/StarknetFinalityAnalyzer.ts
```diff
@@ -0,0 +1,80 @@
...
+  const iface = new utils.Interface([`function ${signature}`])
+
+  const decodedInput = iface.decodeFunctionData(signature, data)
+
+  return StarknetStateUpdate.parse({
+    /* eslint-disable @typescript-eslint/no-unsafe-assignment */
+    programOutput: decodedInput.programOutput,
+    kzgProof: decodedInput.kzgProof,
+    /* eslint-enable @typescript-eslint/no-unsafe-assignment */
+  })
```
michalsidzej, line 66: Can't you just parse the decoded?
```suggestion
  return StarknetStateUpdate.parse(decodedInput)
```

packages/backend/src/modules/finality/analyzers/StarknetFinalityAnalyzer.ts
```diff
@@ -0,0 +1,80 @@
...
+} from '@l2beat/shared-pure'
+import { BigNumber, utils } from 'ethers'
+import { z } from 'zod'
+
+import { RpcClient } from '../../../peripherals/rpcclient/RpcClient'
+import { StarknetClient } from '../../../peripherals/starknet/StarknetClient'
+import { LivenessRepository } from '../../tracked-txs/modules/liveness/repositories/LivenessRepository'
+import { BaseAnalyzer } from './types/BaseAnalyzer'
+
+const ZBigNumber = z.instanceof(BigNumber).transform(Number)
```
michalsidzej, line 15: Please use Bigint and transform to number safely inside `extractBlockNumber` function
```suggestion
const ZBigNumber = z.instanceof(BigNumber).transform(Number)
```

michalsidzej (changes requested):

packages/backend/src/modules/finality/analyzers/StarknetFinalityAnalyzer.ts
sdlyy: Ethers when parsing is doing some weird mumbo jumbo and instead of named properties also provides you with something like this if you are using precise iface:
```ts
[
  "foo",
  "bar",
  "foo": "foo",
  "foo": "foo",
]
```

packages/backend/src/modules/finality/analyzers/StarknetFinalityAnalyzer.ts
sdlyy: https://github.com/l2beat/l2beat/pull/3258/commits/3d5d874d1694462980e715896a10871ec9171e23

michalsidzej (approved):

packages/backend/src/modules/finality/analyzers/StarknetFinalityAnalyzer.ts
```diff
@@ -0,0 +1,80 @@
...
+  programOutput: z.array(ZBigNumber),
+  kzgProof: z.string(),
+})
+type StarknetStateUpdate = z.infer<typeof StarknetStateUpdate>
+
+export class StarknetFinalityAnalyzer extends BaseAnalyzer {
+  constructor(
+    protected override readonly provider: RpcClient,
+    protected override readonly livenessRepository: LivenessRepository,
+    protected override readonly projectId: ProjectId,
```
michalsidzej, line 27: No need to override
```suggestion
    provider: RpcClient,
    livenessRepository: LivenessRepository,
    projectId: ProjectId,
```

PR #3264, opened 2024-04-03
Add arbitrum canonically bridged tokens

sekuba:
Part of L2B-4625


sekuba:
These tokens all get routed to the vanilla `L1ERC20Gateway` by arbitrums `L1GatewayRouter`. They are eth-native tokens, thus added as CBV on ethereum.

lucadonnoh (approved):

PR #3297, opened 2024-04-05
Track Paradex finality

sdlyy:
Resolves L2B-4595

TODOs:
- [ ] Populate envs with Paradex RPC


packages/config/src/layer2s/paradex.ts
michalsidzej, line 1: I think that we should add just this file after we confirm that StarknetAnalyzer works as intended on staging.

michalsidzej (approved):

PR #3298, opened 2024-04-05
Add Degen L3 on Base (full)

sekuba:
Resolves L2B-4700

Degen is a baseL2-settled orbitstack L3 with anytrust DA.

Contracts are the same as Deri/Rari except for ERC20Bridge.sol and ERC20Inbox.sol which altogether yields a matt-factor of .78.
Changes in the two contracts are limited to 2 new libraries (Address, ERC20), the deletion of L1 fork-specific functions from the Inbox and the adding of custom native L3 token support (in form of the DEGEN ERC-20)

The L3 has no L1GatewayRouter (it was deployed but is not used and was initialized to `0x000000000000000000000000000000000000dEaD`). Instead, all deposits that are not DEGEN are routed/swapped using UTB.sol which is an integration by Decent(LayerZero app) and allows bridging and/or swapping in from other chains and tokens than Base and DEGEN. The UTB contract is not Degen L3 specific (and governed by Decent) but currently almost exclusively used for Degen routing/bridging.
- Should UTB be inside the smart contracts section of Degen L3?

Ready for review with the above q.


lucadonnoh:
yes i think UTB should be included as it is a core contract

sekuba:
To clarify, the UTB currently sits at a position equivalent to the gateways in this diagram:

image

It would e.g. be possible to deposit/withdraw DEGEN without it and the UTB uses the Inbox's public deposit (`createretryableticket`) function.
The current PR has the UTB included btw

lucadonnoh (approved):

PR #3305, opened 2024-04-08
Expose `syncedUntil` per tracked tx subtype

imxeno:


packages/backend/src/modules/tracked-txs/modules/liveness/api/LivenessController.ts
```diff
@@ -146,9 +149,26 @@ export class LivenessController {
...
+            configurations.filter((c) => c.subtype === subtype),
+          )
+          if (!syncedUntil) return obj
+          obj[subtype] = {
+            ...subtypeData[subtype],
+            syncedUntil,
+          }
+          return obj
+        },
+        {} as LivenessApiProject,
```
maciekzygmunt, line 166:
```suggestion
      const withSyncedUntil = TrackedTxsConfigSubtypeValues.reduce<LivenessApiProject>(
        (obj, subtype) => {
          const syncedUntil = getSyncedUntil(
            configurations.filter((c) => c.subtype === subtype),
          )
          if (!syncedUntil) return obj
          obj[subtype] = {
            ...subtypeData[subtype],
            syncedUntil,
          }
          return obj
        },
        {},
```

PR #3337, opened 2024-04-10
Fix Rollups only breaking activity chart

maciekzygmunt:
Resolves L2B-4797


packages/backend/src/modules/tvl/reports/ReportProject.ts
antooni, line 1: why this change?

antooni (changes requested):

antooni (approved):

PR #3338, opened 2024-04-10
Add AmountId for AmountService

michalsidzej:
Resolves L2B-4717


packages/backend/src/modules/tvl2/utils/amountId.ts
antooni, line 1: rename to createAmountId and add tests

antooni (approved):

PR #3347, opened 2024-04-10
Add EigenDA discovery

vincfurc:
Resolves L2B-4790


luke7211 (changes requested):
Shouldn't EigenDA be part of EigenLayer discovery ? For the most parts it discovers the same contracts as in EigenLayer. Additionally there are many more AVSs available now that are not discovered. 

We could see AVSs similarly to Gateways for Rollups. There is an AVSRegistry that allows us to discover all of them. We should consider having a discovery service for all in EigenLayer. 

The downside will be that EigenLayer discovery will be huge as it will discover all changes in all AVSs

luke7211 (approved):
After the discussion we leave it as it is - EigenDA will have a separate discovery. In the future we might want to make EigenLayer contracts a shared module

PR #3358, opened 2024-04-10
Track PolygonZkEvm finality

sdlyy:
Resolves L2B-4003

Timings match, zkevm posts data every 30 minutes - thus maximum will be around ~30 minutes and the minimum will depend on how closely a given transaction has been posted to the L1 posting time.

And yes, it is disabled on production.

image


packages/backend/src/modules/finality/analyzers/polygon-zkevm/PolygonZkevmFinalityAnalyzer.ts
```diff
@@ -0,0 +1,112 @@
...
+    const l1Timestamp = transaction.timestamp
+
+    const hashes = extractTransactionData(tx.data)
+      .map(byteArrFromHexStr)
+      // Might be memory intensive - if so, consider batching/sequential processing
+      .flatMap(decodeBatch)
+      .flatMap((block) => block.transactions)
+      .map(toTransactionHash)
+
+    //TODO: poor stuff given the batch can contain ~250 txs, either make sure its rate limited or pick at random a few txs
```
michalsidzej, line 45: Aren't the transactions sorted? Then it's enough to know the first and last to know the min and max. It should be also enough for avg. This call looks a lot too big.

michalsidzej: Especially that we only download the max block. Please check how are the txs usually sorted.

lucadonnoh: min and max are enough for the avg only if the distribution is uniform and i'm not sure this is an assumption that approximates good enough

michalsidzej (dismissed):
Please add a warning that "we assume an uniform block distribution for calculating averages", both on the FE and in the code of the anazyzer. Otherwise LGTM, good job.

michalsidzej (approved):

PR #3364, opened 2024-04-11
Display publish date in governance articles

imxeno:
Resolves L2B-4556.


packages/frontend/src/pages/governance/index/props/getGovernancePublicationEntry.ts
```diff
@@ -31,6 +32,7 @@ export function getGovernancePublicationEntry(
     description: post.data.description,
     excerpt: post.excerpt,
     readTimeInMinutes: post.readTimeInMinutes,
+    publishedOn: post.data.publishedOn,
```
torztomasz, line 35: format the value here

torztomasz (approved):

PR #3371, opened 2024-04-11
Configure Starknet costs split

maciekzygmunt:
Resolves L2B-4822

### DIFF
--- a/packages/config/src/layer2s/starknet.ts
+++ b/packages/config/src/layer2s/starknet.ts
@@ -399,7 +399,7 @@ export const starknet: Layer2 = {
             '1865367024509426979036104162713508294334262484507712987283009063059134893433',
           ],
         },
-        _hackCostMultiplier: 0.85,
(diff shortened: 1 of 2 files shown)


packages/config/src/layer2s/starknet.ts
```diff
@@ -429,7 +429,7 @@ export const starknet: Layer2 = {
             '2479841346739966073527450029179698923866252973805981504232089731754042431018',
           ],
         },
-        _hackCostMultiplier: 0.85,
+        _hackCostMultiplier: 0.65,
```
radinat, line 432: Shouldn't it be 0.7 here? Since the untilTimestamp is 12.03?

maciekzygmunt: [ca2413b](https://github.com/l2beat/l2beat/pull/3371/commits/ca2413b3f51f028c0208b72dcbe31c14a0225f44)

radinat (approved):

PR #3373, opened 2024-04-11
Add Proof of Play Apex (full)

sekuba:
Resolves L2B-4698

Apex is an orbit stack L3 with AnyTrust DA settling on Arbitrum. They deployed with ArbOS 10 and are using the default escrows with ETH only.
Deploy tx: [0x2b3afe91774ac66b00b40909ee92830313aea5687731aa8ba10f0dfc47af7046](https://dashboard.tenderly.co/tx/arbitrum/0x2b3afe91774ac66b00b40909ee92830313aea5687731aa8ba10f0dfc47af7046)


sekuba:
hello github y no merge

lucadonnoh:
btw we're missing info related to the DAC

sekuba:
oh i see ok

lucadonnoh:
but i see it's the same for all the projects, we should remember that tho

lucadonnoh:
we should add the DAC in the permission section but i think it's not trivial

lucadonnoh (approved):

PR #3374, opened 2024-04-11
Rhinofi update: Blocklist implemented

sekuba:
Resolves L2B-4634


lucadonnoh (changes requested):
add in the withdrawal section this info and a risk warning

packages/config/src/common/exits.ts
```diff
@@ -2,6 +2,7 @@ import { formatSeconds } from '@l2beat/shared-pure'
 
 import { ScalingProjectRisk } from './ScalingProjectRisk'
 import { ScalingProjectTechnologyChoice } from './ScalingProjectTechnologyChoice'
+import { text } from 'stream/consumers'
```
lucadonnoh: ?

lucadonnoh (approved):

lucadonnoh (changes requested):
image
let's also add a phrase here that users can be censored explicitely

sekuba:
oke

lucadonnoh (approved):

PR #3379, opened 2024-04-12
Fix `getConfigurationsInRange` performance

michalsidzej:
Before: 
Screenshot 2024-04-12 at 11 00 50
After: 
Screenshot 2024-04-12 at 11 09 05


mateuszradomski (approved):
DS&A be like

PR #3383, opened 2024-04-12
Fix font preload

imxeno:


torztomasz (changes requested):
image

Hmm, seems like anonymous and blank string should be the same

imxeno:
Yeah, of course it's the same. 🤦‍♂️ It seems that this preload issue was somehow fixed in the meantime.

PR #3394, opened 2024-04-12
OP Mainnet: upgrade 6 review

lucadonnoh:
Resolves L2B-4836


vincfurc (approved):
May be time to add [L1ERC721Bridge](https://github.com/ethereum-optimism/optimism/blob/op-contracts/v1.3.0/packages/contracts-bedrock/src/L1/L1ERC721Bridge.sol) to the OP template, creating a task

PR #3396, opened 2024-04-12
Add kakarot zkEVM to upcoming L2s 

danilowhk:
This PR adds Kakarot zkEVM as an upcoming L2 list.


danilowhk:
Added kakarot to layers2 index.ts and added the logo

lucadonnoh (approved):

PR #3408, opened 2024-04-15
Remove minTimestamp asserts

antooni:


michalsidzej (approved):
This probably should be added in the future https://linear.app/l2beat/issue/L2B-4944/make-sure-that-indexer-configs-are-addedchanged-properly

PR #3427, opened 2024-04-18
Migrate to biome formatting from prettier

mateuszradomski:
Prettier has not been removed at the top-level dependency because we sometimes want to format the tailwind classes. To be decided if we can remove it.


mateuszradomski:
This is waiting for L2B-4969.

imxeno (approved):

PR #3431, opened 2024-04-19
Update boba network info

boyuan-chen:
Boba Network has completed an Anchorage upgrade at block 1149019. Boba Network is running the Op stack using Erigon as the sequencer node.


lucadonnoh:
Resolved here https://github.com/l2beat/l2beat/pull/3438

PR #3438, opened 2024-04-21
Boba Bedrock (anchorage) upgrade

sekuba:
Ready for review

Highest similarity: .83 for karak, but latest version of the op stack contracts.
1) ~discovery error:~

2) ~`genesisTimestamp` in .ts should be boba genesis or boba bedrock genesis?~
3) ~upgradeability TODO~


packages/backend/discovery/bobanetwork/ethereum/diffHistory.md
```diff
@@ -1,3 +1,500 @@
...
+
+# Diff at Sun, 21 Apr 2024 14:24:12 GMT:
+
+- author: sekuba (<sekuba@users.noreply.githum.com>)
+- comparing to: main@262f9e3e98ac8a85b09235e0b440b48e826f1f9f block: 19531460
+- current block number: 19704353
+
+## Description
+
+Provide description of changes. This section will be preserved.
```
lucadonnoh, line 11: !!

lucadonnoh (changes requested):

packages/config/src/layer2s/bobanetwork.ts
```diff
@@ -1,57 +1,26 @@
...
 const discovery = new ProjectDiscovery('bobanetwork')
-
-const upgradesProxy = {
-  upgradableBy: ['Owner'],
-  upgradeDelay: 'No delay',
-}
-
-const upgradesAddressManager = {
-  upgradableBy: ['Owner'],
+const upgradeability = {
```
lucadonnoh, line 9: given this PR, this is not needed anymore: https://github.com/l2beat/l2beat/pull/3421

please check if there is anything else that can be simplified

lucadonnoh (changes requested):

packages/config/src/layer2s/bobanetwork.ts
sekuba: nice, only little problem is that the L1ERC721Bridge does not show the upgradability then..

sekuba: ideal would be if it could also be added to the template (as optional if possible)

lucadonnoh (changes requested):
image
taken from Optimism: we should subtract 1 only after upgrade block and not override past blocks

sekuba:
> taken from Optimism: we should subtract 1 only after upgrade block and not override past blocks

cool, and genesisTimestamp stays at the boba network genesis correct?

packages/config/src/layer2s/bobanetwork.ts
```diff
@@ -68,379 +33,54 @@ export const bobanetwork: Layer2 = {
...
+        'Owner of the ProxyAdmin. It can upgrade the rollup system and the bridge implementation, potentially gaining access to all funds.',
+      )
+    })(),
+  ],
+  rpcUrl: 'https://mainnet.boba.network/',
+  transactionApi: {
+    type: 'rpc',
+    defaultUrl: 'https://mainnet.boba.network/',
+    startBlock: 1,
+    assessCount: subtractOneAfterBlockInclusive(1713297000), // anchorage upgrade
```
lucadonnoh, line 72: should be a block number, not a timestamp (read function name)

lucadonnoh (changes requested):

lucadonnoh:
> > taken from Optimism: we should subtract 1 only after upgrade block and not override past blocks
> 
> cool, and genesisTimestamp stays at the boba network genesis correct?

image
no it should be the genesis timestamp of the upgrade, see opStack template code for the reason. Also, we are missing a way to add nonTemplate tracked tx, pls go to matt

sekuba:
nontemplateTxs for opstack is merged and now used here. I also added the vanilla opstack trackedTxs in the .ts since they are not additive in the template. 
timestamp of upgrade: 1713303530
genesisTimestamp: 1635393439
boba l2 bedrock block: 1149019

PR #3446, opened 2024-04-22
Create op stack discovery script

michalsidzej:


packages/backend/scripts/createOpStack.ts
```diff
@@ -0,0 +1,214 @@
...
+    ],
+    fields: {
+      deletedOutputs: {
+        type: 'stateFromEvent' as const,
+        event: 'OutputsDeleted',
+        returnParams: ['prevNextOutputIndex', 'newNextOutputIndex'],
+      },
+    },
+  },
+  LyraMultisig: {
```
lucadonnoh, line 172: remove

lucadonnoh (changes requested):

packages/backend/scripts/createOpStack.ts
```diff
@@ -0,0 +1,214 @@
...
+}
+
+function printUsage(): void {
+  console.log(
+    'Usage: yarn add-opstack-project project initialAddress1 initialAddress2',
+  )
+}
+
+const names = {
+  '0xBB08cf90DEb93492b463f1Ee5DA9453e51643586': 'SystemConfig',
```
lucadonnoh, line 136: what are these addresses?

packages/backend/scripts/createOpStack.ts
```diff
@@ -0,0 +1,214 @@
...
+const execPromise = util.promisify(exec)
+
+interface HelpMode {
+  mode: 'help'
+}
+
+interface OpStackMode {
+  mode: 'opStack'
+  project: string
+  initialAddresses: [EthereumAddress, EthereumAddress]
```
mateuszradomski, line 18: This type is weird, why not `EthereumAddress[]`?

packages/backend/scripts/createOpStack.ts
```diff
@@ -0,0 +1,214 @@
...
+      await createOpStackProject(config.project, config.initialAddresses)
+      break
+  }
+}
+
+async function createOpStackProject(
+  project: string,
+  initialAddresses: EthereumAddress[],
+): Promise<void> {
+  const dir = path.join(__dirname, '..', 'discovery', project, 'ethereum')
```
mateuszradomski, line 44: There should be a small check that the resulting path is the one you'd expect. If I move this script or the discovery directory everything is going to suck big time

packages/backend/package.json
mateuszradomski, line 1: I'd rename the entire script to be like `init-stack-project` or something more generic because it will probably grow out to support orbit stacks etc.

packages/backend/scripts/createOpStack.ts
```diff
@@ -0,0 +1,214 @@
...
+      acc[contract.address] = contract.name
+    }
+    return acc
+  }, {} as Record<string, string>)
+
+  assert(typeof oldConfig === 'object', 'oldConfig must be an object')
+  return configTemplate(oldConfig.name, oldConfig.initialAddresses, names)
+}
+
+function parseCliParameters(): Config {
```
mateuszradomski, line 110: This should take the name of the stack as a subcommand, so we call:

- `yarn init-project opstack 0xabcd`
- `yarn init-project orbitstack 0xabcd`

packages/backend/scripts/createOpStack.ts
```diff
@@ -0,0 +1,214 @@
...
+    names,
+  )
+
+  console.log('Running initial discovery for', project)
+  await writeFile(
+    path.join(dir, 'config.jsonc'),
+    JSON.stringify(projectConfig, null, 2),
+  )
+  console.log('Initial discovery...')
+  await execPromise('yarn discover ethereum ' + project, { cwd: __dirname })
```
mateuszradomski, line 59: You can execute a discovery directly through a call to `import { discover } from '@l2beat/discovery'`

packages/backend/scripts/createOpStack.ts
```diff
@@ -0,0 +1,214 @@
...
+import util from 'util'
+
+const execPromise = util.promisify(exec)
+
+interface HelpMode {
+  mode: 'help'
+}
+
+interface OpStackMode {
+  mode: 'opStack'
```
mateuszradomski, line 16: What about L3s? We should be able to specify the name of the chain.

packages/backend/scripts/createOpStack.ts
mateuszradomski, line 1: Bonus tip, you can use [cmd-ts](https://www.npmjs.com/package/cmd-ts) as we are going to be migrating all our scripts to this package anyway..

mateuszradomski (changes requested):

PR #3450, opened 2024-04-22
Add mBTC tvl to Mode

fede2442:


radinat (changes requested):
Please run those commands in the config:
- yarn tokens
- yarn format:fix

Also, since it's externally bridged, can you also specify the `bridge`? The `slug` is not needed unless we do not have the bridge listed already.

sekuba:
mesonFi/freeTech does not offer any information about their bridging method for m-btc.
Usually they bridge assets like a liquidity network: https://docs.meson.fi/protocol/meson/process
But m-btc gets locked on merlin [here](https://scan.merlinchain.io/address/0xbb913acCDaBF1F952AAE3dFe78E18a0893B395fF) and then natively minted on mode (see below) so the bridging process must be different...

L2 contracts on Mode:
minter EOA: https://explorer.mode.network/address/0x0014Eb4Ac6Dd1473b258d088E6EF214b2BCdc53C
receiver Multisig: https://explorer.mode.network/address/0x1f478F5f6F83b26a2E4D3a605d430d9DC9c23bf3?tab=token_transfers
mint TX (see the signatures as inputs): https://dashboard.tenderly.co/tx/mode/0x91b2371c2a599372fb8c9a052aaa1a9bb4e1f9bb6860340699967c24f641a2a2

sekuba (approved):

sekuba (approved):

PR #3456, opened 2024-04-22
Fix issue with indexer_configurations table size

antooni:
Resolves L2B-4952


michalsidzej (approved):
Approving as it does not affect prod

PR #3460, opened 2024-04-22
Mode and Zora: upgrade 6

lucadonnoh:
Closes, L2B-4933, L2B-4935


packages/config/src/layer2s/mode.ts
```diff
@@ -65,14 +65,47 @@ export const mode: Layer2 = opStackL2({
...
+      'FoundationMultisig_2',
+      'This address is designated as a Guardian of the OptimismPortal, meaning it can halt withdrawals.',
     ),
     ...discovery.getMultisigPermission(
       'ChallengerMultisig',
       'This address is the permissioned challenger of the system. It can delete non finalized roots without going through the fault proof process. It is also designated as the Guardian.',
     ),
   ],
+  nonTemplateContracts: [
+    discovery.getContractDetails('SuperchainConfig', {
```
sekuba, line 103: this should have `SuperchainProxyAdmin` as upgradeability but currently has the proxyadmin

sekuba (changes requested):

packages/config/src/layer2s/mode.ts
sekuba: haha it seems i can't quote but i do mean the SuperchainConfig's upgradeability

sekuba (approved):

PR #3471, opened 2024-04-23
Fix toFixed error

torztomasz:
This pull request fixes the toFixed error in the getFirstTwoNonZeroPrecision function. The error occurred when the precision parameter was set to a value less than 0 or greater than 100. The fix adds validation for the precision parameter and throws an error if it is outside the valid range.


maciekop-l2b (approved):
LGTM

PR #3491, opened 2024-04-24
Cbridge Update: Sentinel unrelaxed

sekuba:
Closes L2B-5000

### DIFF
--- a/packages/backend/discovery/cbridge/ethereum/diffHistory.md
+++ b/packages/backend/discovery/cbridge/ethereum/diffHistory.md
@@ -1,3 +1,33 @@
+Generated with discovered.json: 0x5bfc86fa944292c5527f9ce286fa22c496536f90
+
+# Diff at Wed, 24 Apr 2024 08:48:18 GMT:
+
+- author: sekuba (<sekuba@users.noreply.githum.com>)
+- comparing to: main@64cf04758a7da5aaed34239f1c6d1672c16260ad block: 19704406
+- current block number: 19724145
+
+## Description
+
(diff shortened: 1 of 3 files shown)


packages/backend/discovery/cbridge/ethereum/diffHistory.md
```diff
@@ -1,3 +1,33 @@
...
+```diff
+    contract Sentinel (0xF140024969F6c76494a78518D9a99c8776B55f70) {
+    +++ description: The Sentinel is itself a Governor and Pauser in the bridge contracts. It allows additional Sentinel-Governors to make changes in its name that can be restricted by Guards changing the Sentinel's relaxed state.
++++ description: Number of relaxed guards in the Sentinel.
+      values.numRelaxedGuards:
+-        2
++        0
++++ description: An unrelaxed Sentinel allows only parameter changes in the Bridge that would make it more secure (decrease limits, increase delay period etc.)
++++ severity: MEDIUM
+      values.relaxed:
```
lucadonnoh, line 25: so being relaxed means to have broader permissions, right?

lucadonnoh, line 25: so being relaxed means to have broader permissions, right?

sekuba: the Sentry being relaxed means all Governors behind the Sentry (Sentry-Governors in my description) inherit the full permissions and functions of the Sentry, who itself is a Governor and full pauser in the bridges (but there are also other govs in the bridges that are without sentry)

sekuba: Plus there is SimpleGovernance for upgrades, which is unrelated
A diagram would probably be interesting here

lucadonnoh (approved):

PR #3493, opened 2024-04-24
Fix liveness coping logic

maciekzygmunt:
Resolves L2B-5008

### DIFF
--- a/packages/backend/src/modules/tracked-txs/modules/liveness/api/LivenessController.ts
+++ b/packages/backend/src/modules/tracked-txs/modules/liveness/api/LivenessController.ts
@@ -142,13 +142,6 @@ export class LivenessController {
 
       const withAnomalies = calculateAnomalies(intervals)
 
-      if (project.livenessConfig) {
-        const { from, to } = project.livenessConfig.duplicateData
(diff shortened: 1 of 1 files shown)


packages/backend/src/modules/tracked-txs/modules/liveness/api/LivenessController.ts
```diff
@@ -167,6 +160,17 @@ export class LivenessController {
...
+      if (project.livenessConfig) {
+        const { from, to } = project.livenessConfig.duplicateData
+
+        withSyncedUntil[to] = {
+          ...withSyncedUntil[from],
+        } as LivenessApiProject[
+          | 'batchSubmissions'
+          | 'proofSubmissions'
+          | 'stateUpdates']
+      }
```
imxeno, line 172: This looks a bit better than the 'as'
```suggestion
     if (project.livenessConfig) {
        const { from, to } = project.livenessConfig.duplicateData
        const data = withSyncedUntil[from]
        assert(data, 'From data must exist')
        withSyncedUntil[to] = { ...data }
      }
```

imxeno (approved):

PR #3496, opened 2024-04-24
Introduce tagging to indexers

antooni:
- [x] refactor createAmountId
- [x]  write migration that clears the tables accordingly (indexers ids are being changed)

Resolves L2B-5010

By introducing tags it will be easy to build dynamic grafana dashboards. Also this PR introduces change from "id" to "name" in ManagedIndexers interfaces + name is no longer being injected but rather defined inside the class

Example response from /metrics
image


packages/backend/src/modules/tvl2/indexers/BlockTimestampIndexer.ts
```diff
@@ -10,20 +10,19 @@ import { SyncOptimizer } from '../utils/SyncOptimizer'
...
 
 export class BlockTimestampIndexer extends ManagedChildIndexer {
-  indexerId: string
-
   constructor(private readonly $: BlockTimestampIndexerDeps) {
-    super($)
-    this.$.logger = this.$.logger.for(this)
-    this.indexerId = $.id
+    const logger = $.logger.tag($.chain)
+    const name = 'block_timestamp_indexer'
```
michalsidzej, line 24: You could do the following in ManagedMultiIndexer (or UIF directly). Could be generic and less things to remember when creating indexers.
```suggestion
    // gets name of the parent class and converts `PascalCase` to `snake_case`
    const name = this.constructor.name.split(/\.?(?=[A-Z])/).join('_').toLowerCase()
```

torztomasz (approved):

PR #3498, opened 2024-04-24
Add zkSync Era native tokens

radinat:
Resolves L2B-4585, L2B-4691

Requested through the form


packages/config/src/tokens/tokens.jsonc
```diff
@@ -1865,6 +1865,25 @@
...
+    },
+    {
+      "symbol": "ZORRO",
+      "address": "0x244c238325fc1bdf6eded726ee1b47d55895d944",
+      "type": "NMV",
+      "formula": "circulatingSupply"
+    },
+    {
+      "symbol": "SLR",
+      "address": "0xe027d939f7de6f521675907cf086f59e4d75b876",
```
sekuba, line 1884: This one will not be tracked correctly if added like this because it is on three chains (ethereum, energyweb, zksync era) and coingecko just tracks all circulating. (which is ~80% energyweb)

We could add the [ethereum-native SLR](https://etherscan.io/token/0x4e9e4ab99cfc14b852f552f5fb3aa68617825b6c) instead (as EBV) because it is [canonically bridged](https://etherscan.io/token/0x4e9e4ab99cfc14b852f552f5fb3aa68617825b6c#balances). TVL is only ~ USD 80k though.

sekuba (changes requested):

packages/config/src/tokens/tokens.jsonc
radinat: Removed

sekuba (approved):

PR #3512, opened 2024-04-25
Add `/tvl2/token` endpoint

michalsidzej:
Resolves L2B-4553

*Blazingly fast*
Screenshot 2024-04-25 at 14 27 09

Only daily data for now.

### DIFF
--- a/packages/backend/src/modules/tvl2/repositories/AmountRepository.ts
+++ b/packages/backend/src/modules/tvl2/repositories/AmountRepository.ts
@@ -32,8 +32,19 @@ export class AmountRepository extends BaseRepository {
     const rows = await knex('amounts')
       .whereIn('configuration_id', configIds)
       .where('timestamp', timestamp.toDate())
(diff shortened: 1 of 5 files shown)


packages/backend/src/modules/tvl2/repositories/AmountRepository.ts
```diff
@@ -32,8 +32,19 @@ export class AmountRepository extends BaseRepository {
     const rows = await knex('amounts')
       .whereIn('configuration_id', configIds)
       .where('timestamp', timestamp.toDate())
+      .andWhereRaw(`extract(hour from "timestamp") % 24 = 0`)
```
antooni, line 35: this change will modify the /api/tvl behaviour - is it intended?

michalsidzej: I think this will modify the `/api/tvl2`. We do not use it anywhere, so it does not matter. This will fix a bug there.

antooni (approved):

PR #3513, opened 2024-04-25
Eigenlayer discovery: Ignore AVS list in watch mode

sekuba:
Closes L2B-5002

A new AVS called _DODOchain MACH (Powered by AltLayer)_ on Eigenbeat with the same implementation contract as the existing Altlayer MACH (0 diff) is added to the Registry.

Moved the AVS list into ignoreInWatchMode to prevent future spam but keep it accessible.


vincfurc (approved):
added a linear task for dodochain

PR #3514, opened 2024-04-25
Enable escrow configuration for different chains (L3 tvl)

antooni:
- [x] double check if ETH on those chains will not be counted as an external/native token in TVL v1

Resolves L2B-4844


maciekop-l2b (approved):
LGTM

packages/config/src/projects/layer2s/arbitrum.ts
```diff
@@ -295,12 +295,16 @@ export const arbitrum: Layer2 = {
         batchSize: 150,
         sinceBlock: 7654707,
         version: '3',
+        // TODO: fix this
+        isNativeBalanceSupported: false,
```
michalsidzej, line 299: This is dependant only on the multicall version. V3 does support native balance https://arbiscan.io/address/0xcA11bde05977b3631167028862bE2a173976CA11#code#F1#L199. What's the problem here?

packages/backend/src/modules/tvl/modules/ChainTvlModule.ts
```diff
@@ -108,9 +109,12 @@ export function chainTvlModule(
       logger,
     )
 
-  const circulatingSupplyTokens = tokens.filter(
-    (t) => t.chainId === config.chainId && t.formula === 'circulatingSupply',
-  )
+  const circulatingSupplyTokens = tokens
+    // temporary solution - will be removed once tvl2 migration is complete
```
michalsidzej, line 113: Please create a task for that

packages/backend/src/model/Project.ts
```diff
@@ -117,6 +129,47 @@ function toBackendTrackedTxsConfig(
...
+                .map((tokenSymbol) => {
+                  const token = tokensOnChain.find(
+                    (t) => t.symbol === tokenSymbol,
+                  )
+                  assert(
+                    token,
+                    `${layer3.id}: token with symbol ${tokenSymbol} not found on ${chain}`,
+                  )
+                  return token
+                }),
```
michalsidzej, line 166: Please make those calculations higher in the tree. It's very hard to see what's going on with this little space in the line left.

packages/backend/src/modules/tvl/modules/ChainTvlModule.ts
antooni: This will be removed once we remove tvl v1

michalsidzej (approved):

PR #3515, opened 2024-04-25
Update metis risk review after the upgrade

luke7211:
Upgrade of April 2024 introduces decentralised sequencer pool which is managed offchain. Hashes of transaction batches are now posted to an EOA address (like in OP-Stack chains) and old CTC is used only for L1->L2 transactions. There is an EOA address with upgrade power of some critical components


packages/config/src/layer2s/metis.ts
```diff
@@ -133,11 +132,14 @@ export const metis: Layer2 = {
       ],
     },
     operator: {
-      ...OPERATOR.CENTRALIZED_SEQUENCER,
+      name: 'The system has a decentralized sequencer set',
+      description:
+        'As o April 2024 Metis uses a permissioned sequencer pool running a Tendermint consensus. Once consensus is reached on a block, an MPC address is used to submit a block hash to Ethereum. The infrastracture to manage the MPC is offchain needs to be trusted with Ethereum not verifying the validity of MPC address.',
```
lucadonnoh, line 137:
```suggestion
        'As of April 2024 Metis uses a permissioned sequencer pool running a Tendermint consensus. Once consensus is reached on a block, an MPC address is used to submit a block hash to Ethereum. The infrastracture to manage the MPC is offchain needs to be trusted with Ethereum not verifying the validity of MPC address.',
```

lucadonnoh (dismissed):

packages/config/src/layer2s/metis.ts
```diff
@@ -17,17 +17,19 @@ const discovery = new ProjectDiscovery('metis')
...
-  isUnderReview: true,
+  isUnderReview: false,
   type: 'layer2',
   id: ProjectId('metis'),
   display: {
     name: 'Metis Andromeda',
     shortName: 'Metis',
     slug: 'metis',
+    redWarning:
+      'Critcal contracts can be upgraded by an EOA which could result with the loss of all funds.',
```
lucadonnoh, line 28:
```suggestion
      'Critical contracts can be upgraded by an EOA which could result with the loss of all funds.',
```

lucadonnoh (changes requested):

packages/config/src/layer2s/metis.ts
```diff
@@ -239,13 +245,20 @@ export const metis: Layer2 = {
...
+        'This contract implements a voting scheme with which the majority of Verifiers can challenge malicious state roots proposed. There are no whitelisted verifiers, hence this contract is not used in practice.',
+      ),
+
       discovery.getContractDetails(
         'CanonicalTransactionChain',
-        'The Canonical Transaction Chain (CTC) contract is an append-only log of transactions which must be applied to the OVM state. It defines the ordering of transactions by writing them to the CTC:batches instance of the Chain Storage Container. CTC batches can only be submitted by OVM_Sequencer. The CTC also allows any account to enqueue() a transaction, which the Sequencer must eventually append to the rollup state.',
+        'The Canonical Transaction Chain (CTC) contract is an append-only log of transactions which must be applied to the OVM state. Given that transactions batch hashes are sent to an EOA address, it allows any account to enqueue() a transaction, which the Sequencer must eventually append to the rollup state.',
       ),
       discovery.getContractDetails(
         'StateCommitmentChain',
```
sekuba, line 264: We should edit this description similarly to the CTC one, because 1:1 correspondence with CTC does not get checked anymore [(see commented-out part in the SC)](https://etherscan.io/address/0x6b1cd90796dc97c859734dda4ff7816f703ddb48#code#F1#L244).

sekuba (changes requested):

packages/config/src/layer2s/metis.ts
sekuba: (should i do it?)

sekuba (approved):

PR #3520, opened 2024-04-26
Socket update: New plugs, new vaults

sekuba:
Closes L2B-5055

The socket-specific plug and vault discovery script now also detects the new plugs/vaults that have slightly different view functions.

All discovered vaults with locked value are added to socket.ts.
Permissions and descriptions for the project page added.


vincfurc (commented):
this is the reason for the tooltip not showing, should be a list of chains

packages/config/src/projects/bridges/socket.ts
vincfurc, line 1: this is the reason for the tooltip not showing, should be a list of chains

sekuba:
'every escrow is unique'-test fails here because newly added reya has the same escrow listed.
(`0xdFf78A949E47c1e90f3Dd6dd7Fe2Fa72B42a75f7`)

the governance of the reya vault (owner) is the socket admin EOA (as listed in socket permissions section) but the vault has only one connector attached, which is restricted to bridging to Reya L2. 
If we have to remove it from one .ts, it would be reya IMO.

luke7211 (approved):

PR #3521, opened 2024-04-26
Build the site half past instead of full hour

michalsidzej:


michalsidzej:
Not needed in the end

PR #3530, opened 2024-04-26
Use local JSON schemas

adamiak:
Referening local (instead of online http://) JSON schemas helps with discovery development. Otherwise we get IDE errors until merge to master.

Currently schemas will live in `discovery` package because that's where ConfigReader lives.


packages/backend/discovery/zora/ethereum/config.jsonc
```diff
@@ -28,6 +28,7 @@
       "ignoreInWatchMode": ["params"]
     },
     "L2OutputOracle": {
+      "extends": "opstack/L2OutputOracle",
```
mateuszradomski, line 31: Something snuck it's way into this PR :p

mateuszradomski (changes requested):

packages/backend/discovery/zora/ethereum/config.jsonc
adamiak: XD, good catch!

adamiak:
Closing due to adding too many unrelated changes to this PR.

PR #3534, opened 2024-04-29
Refactor logger to allow for multiple backends

maciekop-l2b:
- Refactor logger to allow for multiple backends and formatters
- Push logs to Elastic Search if enabled by env configuration

part of L2B-4977

### DIFF
--- a/packages/backend/discovery/metis/ethereum/config.jsonc
+++ b/packages/backend/discovery/metis/ethereum/config.jsonc
@@ -22,7 +22,7 @@
       "ignoreDiscovery": true
     },
     "LockingInfo": {
-      "ignoreInWatchMode": ["totalLocked"]
+      "ignoreInWatchMode": ["totalLocked", "totalRewardsLiquidated"] // 'liquidated' here = withdrawn
     },
     "LockingPool": {
       "fields": {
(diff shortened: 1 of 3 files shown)


packages/backend-tools/src/logger/Logger.test.ts
```diff
@@ -48,9 +63,13 @@ describe(Logger.name, () => {
   it('supports bigint values in pretty output', () => {
     const backend = createTestBackend()
     const logger = new Logger({
-      backend,
+      backends: [
+        {
+          backend,
+          formatter: new LogFormatterPretty(false, true),
```
torztomasz, line 69: let's put that (false,true) to the object of logger options

packages/backend-tools/src/logger/interfaces.ts
```diff
@@ -0,0 +1,39 @@
+import { LogLevel } from './LogLevel'
```
torztomasz, line 1: usually we name that type of files `types.ts`

packages/backend-tools/src/helpers/formatDate.ts
```diff
@@ -0,0 +1,4 @@
+export function formatDate(date: Date): string {
```
torztomasz, line 1: usually we name that type of directories `utils` instead of `helpers`

torztomasz (changes requested):
got some more stuff, lets talk about it

torztomasz (approved):

PR #3553, opened 2024-04-30
Use local JSON schemas

adamiak:
Referening local (instead of online http://) JSON schemas helps with discovery development. Otherwise we get IDE errors until merge to master.

Currently schemas will live in `discovery` package because that's where ConfigReader lives.


mateuszradomski (approved):
LGTM

PR #3556, opened 2024-04-30
Add KOI and HOLD

radinat:
Requested by zkSync
Resolves L2B-4867, L2B-5131
https://www.coingecko.com/en/coins/holdstation
https://www.coingecko.com/en/coins/koi-3

### DIFF
--- a/packages/config/src/tokens/tokens.jsonc
+++ b/packages/config/src/tokens/tokens.jsonc
@@ -497,6 +497,10 @@
       "address": "0xdeFA4e8a7bcBA345F687a2f1456F5Edd9CE97202"
     },
     {
+      "symbol": "KOI",
+      "address": "0x9d14bce1daddf408d77295bb1be9b343814f44de"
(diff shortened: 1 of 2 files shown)


packages/config/src/tokens/tokens.jsonc
```diff
@@ -496,6 +496,10 @@
       "symbol": "KNC",
       "address": "0xdeFA4e8a7bcBA345F687a2f1456F5Edd9CE97202"
     },
+    {
+      "symbol": "KOI",
+      "address": "0x9d14bce1daddf408d77295bb1be9b343814f44de"
```
sekuba, line 501: do you think we could make an exception here and count circulating only? The token is -although bridged- 100% on zksync and coingecko counts circulating at 50% of the value in the zksync bridge escrow.

I'm also fine with just going with this version if we want to respect our usual practice

sekuba (approved):

PR #3560, opened 2024-05-01
Aztecconnect update: Rollup ownership renounced

sekuba:
Closes L2B-5144

### DIFF
--- a/packages/config/src/layer2s/aztecconnect.ts
+++ b/packages/config/src/layer2s/aztecconnect.ts
@@ -1,8 +1,4 @@
-import {
-  EthereumAddress,
-  ProjectId,
-  UnixTime,
-} from '@l2beat/shared-pure'
+import { EthereumAddress, ProjectId, UnixTime } from '@l2beat/shared-pure'
 
 import {
   addSentimentToDataAvailability,
@@ -101,17 +97,20 @@ export const aztecconnect: Layer2 = {
       ],
     },
     proposerFailure: {
-      description: 'The system is EOL and the user must self-propose to withdraw.',
+      description:
+        'The system is EOL and the user must self-propose to withdraw.',
       sentiment: 'neutral',
-      value: 'No rollup proposer'
+      value: 'No rollup proposer',
     },
     sourceUpgradeability: {
-      description: 'The ownership of the rollup contract (ProxyAdmin) has been irrevocably renounced, which makes it immutable.',
+      description:
+        'The ownership of the rollup contract (ProxyAdmin) has been irrevocably renounced, which makes it immutable.',
       sentiment: 'good',
-      value: 'Immutable'
+      value: 'Immutable',
     },
     dataAvailability: {
-      description: 'Since EOL the data is not automatically published and the users must self-publish to L1.',
+      description:
+        'Since EOL the data is not automatically published and the users must self-publish to L1.',
       sentiment: 'bad',
       value: 'No automatic data publishing',
       sources: [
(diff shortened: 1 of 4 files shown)


lucadonnoh (changes requested):
is the exit window still zero?

sekuba:
No, the rollup contract is fully immutable and withdraw-only now.
I hesitated to update more of the risks though because the project is archived and only usable for ppl that run the rollup infra locally AND are deposited from a year ago.

For fully correct risk would have to update Operator, Proposer fail, Exit window on top of this PR
shall i do eet?

sekuba:
description updated to current immutable state :)

packages/config/src/layer2s/aztecconnect.ts
```diff
@@ -180,40 +134,28 @@ export const aztecconnect: Layer2 = {
...
-  }),
+  },
   stage: getStage(
     {
       stage0: {
         callsItselfRollup: true,
-        stateRootsPostedToL1: true,
-        dataAvailabilityOnL1: true,
+        stateRootsPostedToL1: false,
+        dataAvailabilityOnL1: false,
```
lucadonnoh, line 145: why?

packages/config/src/layer2s/aztecconnect.ts
```diff
@@ -180,40 +134,28 @@ export const aztecconnect: Layer2 = {
...
-        usersCanExitWithoutCooperation: false,
+        usersHave7DaysToExit: true,
+        usersCanExitWithoutCooperation: true,
         securityCouncilProperlySetUp: null,
       },
       stage2: {
         proofSystemOverriddenOnlyInCaseOfABug: null,
-        fraudProofSystemIsPermissionless: null,
-        delayWith30DExitWindow: false,
+        fraudProofSystemIsPermissionless: true,
```
lucadonnoh, line 157: null, there is no fraud proof system

packages/config/src/layer2s/aztecconnect.ts
sekuba: before: the central operator posted data and state on L1
now: you have to post data on L1 yourself (but all data to advance state is available)
you can basically only advance because it is only one user at a time with their own data that processes the rollup

so i keep DA as true?

lucadonnoh: both of them are true

packages/config/src/layer2s/aztecconnect.ts
```diff
@@ -135,10 +74,10 @@ export const aztecconnect: Layer2 = {
     bridge: { type: 'Enshrined' },
     mode: 'State diffs',
   }),
-  riskView: makeBridgeCompatible({
+  riskView: {
     stateValidation: {
       ...RISK_VIEW.STATE_ZKP_SN,
-
+      ...RISK_VIEW.DATA_ON_CHAIN,
```
sekuba, line 80: just lost in the diff :p

packages/config/src/layer2s/aztecconnect.ts
```diff
@@ -157,8 +96,23 @@ export const aztecconnect: Layer2 = {
...
+      value: 'No rollup proposer',
+    },
+    sourceUpgradeability: {
+      description:
+        'The ownership of the rollup contract (ProxyAdmin) has been irrevocably renounced, which makes it immutable.',
+      sentiment: 'good',
+      value: 'Immutable',
+    },
     dataAvailability: {
-      ...RISK_VIEW.DATA_ON_CHAIN,
```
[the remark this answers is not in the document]
sekuba: see below diff didn't catch it

packages/config/src/layer2s/aztecconnect.ts
sekuba: ooooh and in the wrong  place, sorry fixing...

sekuba: fixed

lucadonnoh (approved):

PR #3575, opened 2024-05-02
BlobstreamX on Base and Arbitrum update: new Prover, tests added

sekuba:
Closes L2B-5122


vincfurc (changes requested):
we need to add for provers the same test we have for verifiers on functionId changes

sekuba:
they are a watched value, we should get a notif like for all values

vincfurc (approved):

PR #3580, opened 2024-05-03
Add reya full

sekuba:
Orbit stack L2 with 100% code similarity to popapex. (Specializing on in-house DEX,Perps,liquidity efficiency etc.)
Red warning because of Admin EOA.
'DAC' comprising of one member.

The LGE they are currently running accepts USDC-only (bridged via socket, escrow added)
ETH is bridged canonically with standard orbit `Bridge` contract and gelato frontend.
Gelato also provides their AA implementation, which is currently used for all users AFAIK (all blocks on the explorer have a batcher tx and nothing else)

- Activity is disabled because of AA
- TrackedTxs commented out due to tests failing with 'duplicate' error (do we have automatic trackedTxs in the orbit template?)


packages/config/src/layer2s/reya.ts
```diff
@@ -0,0 +1,119 @@
...
+  // adding trackedTxs yields Error: Assertion Error: Duplicate function call config in reya
+  // trackedTxs: [
+  //   {
+  //     uses: [
+  //       { type: 'liveness', subtype: 'batchSubmissions' },
+  //       { type: 'l2costs', subtype: 'batchSubmissions' },
+  //     ],
+  //     query: {
+  //       formula: 'functionCall',
+  //       address: EthereumAddress('0xF4Ef823D57819AC7202a081A5B49376BD28E7b3a'),
```
lucadonnoh, line 45: this address is kinto

lucadonnoh (changes requested):

lucadonnoh (approved):

PR #3581, opened 2024-05-04
Add swell pre-launch discovery

sekuba:
do we need the discovery in .ts (and move to inreview) to have update-monitor catch it?

### DIFF
--- a/packages/backend/discovery/swell/ethereum/config.jsonc
+++ b/packages/backend/discovery/swell/ethereum/config.jsonc
@@ -1,0 +1,19 @@
+{
+    "$schema": "https://raw.githubusercontent.com/l2beat/tools/main/schemas/config.schema.json",
+    "chain": "ethereum",
+    "name": "swell",
+    "initialAddresses": [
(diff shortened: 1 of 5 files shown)


lucadonnoh:
> do we need the discovery in .ts (and move to inreview) to have update-monitor catch it?

no, it works for eigen without a ts :p

lucadonnoh (changes requested):
i don't see anything that would tell us here that they launched. Maybe we need to monitor `rescueERC20` calls?

sekuba:
i think eigen might only work because of `packages/config/src/onChainProjects.ts` 
> anything that would tell us here that they launched

oh yeah true, but at least we would def get the multisig nonce since they have to push a timelock tx to do anything with the escrow funds (like bridging) idk if that would be enough

lucadonnoh:
> oh yeah true, but at least we would def get the multisig nonce since they have to push a timelock tx to do anything with the escrow funds (like bridging) idk if that would be enough

then add this in meta.json so we don't forget

lucadonnoh (approved):

PR #3609, opened 2024-05-07
Auto-detect contract templates

adamiak:
Resolves L2B-5085

This is the initial attempt to add contract "recognition" logic to discovery. The ultimate goal is for discovery to be smart enough to recognize most of the contracts based on their ABI/source-code and automatically apply matching contract templates.

Current implementation works as follows. If you want given contract to be recognized during discovery, you need to create a `shape/` folder inside that contract template folder and add flattened solidity file(s). During discovery, flattened sources are compared to those solidity files and they're considered a match if similarity is >0.55 (configurable). If there are any matches, discovery will generate a `config.hints.jsonc` file with details.

This PR is actually quite simple, but because of many functions moved to a common file and an added example it seems large.

What's happening in this PR:

* a lot of "code similarity" functions have been moved from `compare-flat` script to a `flatten/utils.ts` file
* two solidity files were added to L2OutputOracle contract template - just as an example of a contract shape that can be automatically recognized
* there's a TemplateService class that can compare currently analyzed sources with template shapes
* AddressAnalyzer returns additional field in the result: `.matchingTemplates: Record<template, similarity>`


packages/discovery/src/discovery/output/saveDiscoveryResult.ts
```diff
@@ -27,6 +28,51 @@ export interface SaveDiscoveryResultOptions {
...
+  skipHints?: boolean
+}
+
+export async function saveConfigHints(
+  rootPath: string,
+  results: Analysis[],
+  options: SaveDiscoveryResultOptions,
+): Promise<void> {
+  const filePath = path.join(rootPath, 'config.hints.jsonc')
+  await rimraf(filePath)
```
mateuszradomski, line 40: do want to remove it when we're not going to overwrite it?

packages/discovery/src/discovery/analysis/TemplateService.ts
```diff
@@ -0,0 +1,84 @@
...
+        const templateId = path.substring(TEMPLATES_PATH.length + 1)
+        result[templateId] = maxSimilarity
+      }
+    })
+
+    return result
+  }
+}
+
+function iterateFoldersRecursively(
```
mateuszradomski, line 73: I don't like this pattern, it's hard to read. What do you think about a call to "iterateFoldersRecursivley" just returning an array of paths - no callbacks. And a the top we just iterate over each path.

packages/discovery/src/discovery/output/saveDiscoveryResult.ts
```diff
@@ -27,6 +28,51 @@ export interface SaveDiscoveryResultOptions {
...
+
+  const contractsWithMatchedTemplates = results.filter(
+    (r) => r.type === 'Contract' && !isEmpty(r.matchingTemplates),
+  ) as AnalyzedContract[]
+
+  if (contractsWithMatchedTemplates.length === 0) {
+    return
+  }
+
+  const content = [
```
mateuszradomski, line 53: Maybe it's worth to do it nicer? Either use something like [comment-json](https://www.npmjs.com/package/comment-json) or just `JSON.stringify()` the object and append comments at the top.

mateuszradomski (changes requested):

packages/discovery/src/discovery/output/saveDiscoveryResult.ts
adamiak: Yes, if there are no more hints then we want the hints from the previous run to disappear - probably someone has just added the hints to the config file.

packages/discovery/src/discovery/analysis/TemplateService.ts
adamiak: Done

mateuszradomski (approved):

packages/discovery/src/discovery/output/saveDiscoveryResult.ts
adamiak: I like the idea of using comment-json here but the implementation I'm getting is even more complex and it's difficult to make it work nicely with TypeScript.

I've agreed with @mateuszradomski that we will keep current implementation as is and refactor it if we decide to keep using config.hint.jsonc file in the furure (see: L2B-5274)

PR #3628, opened 2024-05-09
Remove mock warning 

sdlyy:
mb, sry :((


maciekzygmunt (approved):

maciekzygmunt (commented):
This was the previous warning, I guess we should bring it back:
                "It's assumed that transaction data batches are submitted sequentially.",

PR #3637, opened 2024-05-09
Connext bridge diagram and discovery update

vincfurc:
Resolves L2B-5288 and L2B-5133


sekuba (approved):
vincenzo says lgtm and me too

PR #3642, opened 2024-05-10
[L2B-5075] Use aggregated data in cost controller

maciekop-l2b:
Resolves L2B-5075


torztomasz (approved):
LGTM

packages/backend/src/modules/tracked-txs/modules/utils/addToMap.ts
michalsidzej, line 1: I know this file has been here already, but it kinda doesn't make sense that it's outside of `l2-costs` folder when it's only used there. But that's a nitpick.

michalsidzej (approved):
Overall LGTM

PR #3653, opened 2024-05-11
Template support in meta.json

adamiak:
Resolves L2B-5036

If config.jsonc uses template for some contract, meta.json will automatically use the same template. The way it works is the same as for config.jsonc files, namely there's a new keyword `extends` that specifies the template to be used - this keyword is added/removed automatically for a contract if someone uses `extends` on that contract in config.jsonc. 

While working on this PR I decided to switch from saving this structure for every single contract value:

```
{
  desciption: null,
  severity: null,
  type: null
}
```

into simply writing `{}`. It makes meta.json files enormously shorter and easier to navigate. 

Therefore quite a bit of code in this PR is for parsing/saving this new structure.


packages/discovery/src/discovery/runDiscovery.ts
```diff
@@ -56,6 +52,12 @@ export async function runDiscovery(
     blockNumber,
     config.chain.rpcGetLogsMaxRange,
   )
+
+  const projectMeta = await configReader.readMeta(
+    config.project,
+    config.chain.name,
+    true,
+  )
```
mateuszradomski, line 60: Why do we skip the inlining of templates here? There is nothing to indicate the reason for it. The `boolean` as argument nature of this code feels hacky and is hard to understand. Because now I don't know if it should be also set to `true` in other places that call `readMeta()`. If we really need to do this, I'd argue that it's better to create two functions - `readMetaWithoutInlining()` and just `readMeta()`

packages/discovery/src/discovery/config/ConfigReader.ts
```diff
@@ -3,19 +3,26 @@ import path, { posix } from 'path'
...
 
 export const TEMPLATES_PATH = path.join('discovery', '_templates')
 
 export class ConfigReader {
+  public templateService: TemplateService
+
+  constructor() {
+    this.templateService = new TemplateService()
+  }
+
```
mateuszradomski, line 25: Do we really need an entire object? I'm starting to think that `TemplateService` should really be just a couple of functions. The connection between `ConfigReader` and `TemplateService` is really hard to understand.

- `ConfigReader` still exports `TEMPLATES_PATH`, that should probably be in the `TemplateService`
- But `TEMPLATES_PATH` is only used to read the correct files, reading now is split between two classes - `ConfigReader` and `TemplateService` - some files are read by one and others are read by the other

I think that this is just a big mess. I'm currently trying to refactor the project to use `ConfigReader` exclusively and it's really hard because it's not entirely contained in itself. The fact that it's split like that makes this code _very_ rigid, it just **has** to be used from a given directory and any attempt to make it support reading from a different place is a nightmare.

It feels like `TemplateService` has no well defined scope of things it's responsible for without taking some responsibility from `ConfigReader`. My suggestion is to make `ConfigReader` a base class (not in the inheritance kind of way) that does only reading. Any class/function that wants to read some files from the `discovery/project` subdirectories **has** to go through `ConfigReader`.

mateuszradomski (changes requested):

mateuszradomski (approved):

adamiak:
I have agreed with @mateuszradomski that this PR will go in without refactoring suggested in the comments because an upcoming PR will remove meta.json files completely and move data defined there into config.jsonc, making any refactoring of meta.json obsolete.

PR #3654, opened 2024-05-12
Add templates of some core OP stack contracts

adamiak:
Added a few templates of core OP stack contracts and used them in zora config.jsonc


mateuszradomski (approved):
LGTM

PR #3682, opened 2024-05-14
Update build command in gulpfile.js to use bunx

torztomasz:
The build command in the gulpfile.js has been updated to use the "bunx" command instead of "node -r esbuild-register".


sz-piotr:
We probably don't want to do this as we're migrating to next

torztomasz:
Closing because of the Next.js migration

PR #3707, opened 2024-05-15
Make finality warnings about our assumptions greyed out

imxeno:


packages/config/src/projects/layer2s/types/Layer2.ts
```diff
@@ -81,11 +81,16 @@ export interface Layer2LivenessDisplay {
   }
 }
 
+export interface Layer2FinalityDisplayWarning {
+  type: 'project' | 'l2beat'
```
torztomasz, line 85: I would say it should use `Sentiment` type. `project` and `l2beat` are really abstracted type names. Using `neutral` Sentiment should be good there.

torztomasz (changes requested):
Isn't that PR about finality? Title claims its about liveness

imxeno:
> Isn't that PR about finality? Title claims its about liveness

Oops, yes, it's about finality warnings. Sorry!

packages/config/src/projects/layer2s/types/Layer2.ts
imxeno: I've changed the type of the warning fields to - already present - `WarningWithSentiment`.

torztomasz (approved):

lucadonnoh (changes requested):
polygon zkevm should be grey too

lucadonnoh (approved):

PR #3724, opened 2024-05-16
Configjson updates

sekuba:
Closes L2B-5456.

First test of using templates, started with orbitstack configs for SequencerInbox and RollupProxy that were 100% identical among projects. No suggestions based on shape yet though ;)

### DIFF
--- a/packages/backend/discovery/arbitrum/ethereum/config.jsonc
+++ b/packages/backend/discovery/arbitrum/ethereum/config.jsonc
@@ -64,47 +64,7 @@
       }
     },
     "RollupProxy": {
-      "ignoreMethods": [
-        // Execution reverted
-        "proxiableUUID"
-      ],
-      "ignoreInWatchMode": [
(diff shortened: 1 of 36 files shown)


packages/backend/discovery/arbitrum/ethereum/config.jsonc
```diff
@@ -64,47 +64,7 @@
...
-        "validators": {
-          "handler": {
-            "type": "arbitrumActors",
-            "actorType": "validator"
-          }
-        }
-      }
+      "extends": "orbitstack/RollupProxy"
     },
     "SequencerInbox": {
```
adamiak, line 69: Can you use the opstack/SequencerInbox template here?

sekuba: no it isn't identical, but i made a note to look at the orbitstacks that are different ;) (i think it's minor things)

adamiak (approved):

PR #3736, opened 2024-05-17
Implement asset risk design changes

maciekzygmunt:
Resolves L2B-5430 L2B-5429 L2B-5431 L2B-5433


imxeno (commented):
- Something's wrong with the footer padding
- We probably need some rule to autoformat `"use client"` at the top (either with a double newline or not)
- I suggest that we use `svgr` for icons, but we can do this change while merging this project with frontend2
- Imo we lack a button on the main page to proceed with the entered address, I'll leave a comment for Jan

imxeno (approved):

PR #3738, opened 2024-05-17
Remove mergeMetaIntoConfig script

adamiak:
This was a one-time-use script but it imports types type like ConfigReader and RawDiscoveryConfig, so if we don't remove it, we'll start getting build errors when these types change.


mateuszradomski (changes requested):
Remove it from the `package.json` scripts section in `packages/backend`

adamiak:
Good spot! Fixed.

mateuszradomski (approved):

PR #3746, opened 2024-05-20
Remove reorged L3s activity data

torztomasz:
This pull request includes a migration that removes reorged L3s activity. The migration deletes records from the 'activity.block' and 'sequence_processor' tables where the project_id or id is 'degen' or 'popapex'.

### DIFF
--- a/packages/backend/src/peripherals/database/migrations/114_resync_degen_popapex.ts
+++ b/packages/backend/src/peripherals/database/migrations/114_resync_degen_popapex.ts
@@ -1,0 +1,34 @@
+/*
+                      ====== IMPORTANT NOTICE ======
+DO NOT EDIT OR RENAME THIS FILE
+This is a migration file. Once created the file should not be renamed or edited,
+because migrations are only run once on the production server.
+If you find that something was incorrectly set up in the `up` function you
+should create a new migration file that fixes the issue.
+*/
+
+import { Knex } from 'knex'
+
+export async function up(knex: Knex) {
+  await knex('activity.block')
+    .delete()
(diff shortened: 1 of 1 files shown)


packages/backend/src/peripherals/database/migrations/114_resync_degen_popapex.ts
```diff
@@ -0,0 +1,34 @@
...
+export async function down(knex: Knex) {
+  await knex('activity.block')
+    .delete()
+    .where('project_id', '=', 'degen')
+    .orWhere('project_id', '=', 'popapex')
+
+  await knex('sequence_processor')
+    .delete()
+    .where('id', '=', 'degen')
+    .orWhere('id', '=', 'popapex')
```
michalsidzej, line 33: the `down` function can be empty.

packages/backend/src/peripherals/database/migrations/114_resync_degen_popapex.ts
```diff
@@ -0,0 +1,34 @@
...
+DO NOT EDIT OR RENAME THIS FILE
+This is a migration file. Once created the file should not be renamed or edited,
+because migrations are only run once on the production server.
+If you find that something was incorrectly set up in the `up` function you
+should create a new migration file that fixes the issue.
+*/
+
+import { Knex } from 'knex'
+
+export async function up(knex: Knex) {
```
michalsidzej, line 12: Please provide a comment that this is done because of a reorg.

michalsidzej (approved):

PR #3758, opened 2024-05-21
Add token breakdown TVL summary

torztomasz:


lucadonnoh (changes requested):
@sekuba can you review each project and add the associated tokens in this PR? i see that we are missing some like Mantle token and Matic/POL

maciekzygmunt (approved):

lucadonnoh (approved):

PR #3765, opened 2024-05-21
Use db transactions in multiIndexer

michalsidzej:
Resolves L2B-4843, L2B-5475


packages/backend/src/modules/tvl2/repositories/AmountRepository.ts
```diff
@@ -61,9 +62,9 @@ export class AmountRepository extends BaseRepository {
     return rows.map(toRecord)
   }
 
-  async addMany(records: AmountRecord[]) {
+  async addMany(records: AmountRecord[], tx?: Knex.Transaction) {
     const rows: AmountRow[] = records.map(toRow)
-    const knex = await this.knex()
+    const knex = tx ?? (await this.knex())
```
antooni, line 67: We use this pattern in other parts of code `const knex = await this.knex(trx)`

packages/backend/src/tools/uif/KnexMiddleware.ts
```diff
@@ -0,0 +1,24 @@
+import { Knex } from 'knex'
+import { DbTransaction } from './multi/types'
+
+export class KnexTx implements DbTransaction {
```
antooni, line 4: Maybe let's use `KnexTrx` and in other parts of the code also call it `trx` because the naming is not consistent

packages/backend/src/tools/uif/multi/MultiIndexer.ts
```diff
@@ -137,6 +141,7 @@ export abstract class MultiIndexer<T> extends ChildIndexer {
     const oldSafeHeight = (await this.getSafeHeight()) ?? safeHeight
 
     this.saved = new Map(toSave.map((c) => [c.id, c]))
+    // TODO: should we remove data in a transaction?
```
antooni, line 144: Here the SIGTERM will not cause the bug - if the configurations are not saved (line 148) then removal will be retriggered. But the more I think about I think we should run it in transaction just for code clarity + maintainability

packages/backend/src/tools/uif/KnexMiddleware.ts
antooni, line 1: Filename and class have different names

packages/backend/src/tools/uif/KnexMiddleware.test.ts
antooni, line 1: rename the file and the function, this is not multiUpdate but rather a function that calls callbacks in the transaction, maybe _TEST_ONLY_invokeInTransaction?

packages/backend/src/tools/uif/KnexMiddleware.ts
```diff
@@ -0,0 +1,30 @@
+import { Knex } from 'knex'
+import { DbTransaction } from './multi/types'
+
+export class KnexTrx implements DbTransaction {
```
antooni, line 4: tbh I am not a big fan of this interface, lets get rid of it for now

packages/backend/src/tools/uif/multi/ManagedMultiIndexer.ts
```diff
@@ -16,13 +15,20 @@ export interface ManagedMultiIndexerOptions<T> extends IndexerOptions {
   decode: (blob: string) => T
   logger: Logger
   updateRetryStrategy?: RetryStrategy
+  getDbTrx: () => Promise<DbTransaction | undefined>
```
antooni, line 18: use consistent naming in Multi and ManagedMultiIndexer

packages/backend/src/tools/uif/multi/types.ts
```diff
@@ -31,3 +31,8 @@ export interface ConfigurationRange<T> {
   to: number
   configurations: Configuration<T>[]
 }
+
+export interface DatabaseMiddleware {
```
maciekop-l2b, line 35: Let's hide implementation details and alias the type of database transaction. Basically all classes using `DatabaseMiddleware` should refer to `DatabaseTransaction` rather than `Knex.Transaction`

```
export type DatabaseTransaction = Knex.Transaction

export interface DatabaseMiddleware {
  add: (cb: (trx?: DatabaseTransaction) => Promise<void>) => void
  execute: () => Promise<void>
}
```

I would also move `DatabaseMiddleware` and `KnexMiddleware` to some more generic place like `src/peripherals/database`

maciekop-l2b (changes requested):

maciekop-l2b (changes requested):
Overall good approach but some minor things need to be fixed

packages/backend/src/tools/uif/KnexMiddleware.ts
```diff
@@ -0,0 +1,30 @@
...
+        await cb(tx)
+      }
+      await tx.commit()
+    } catch (e) {
+      await tx.rollback()
+      throw e
+    }
+  }
+
+  async _TEST_ONLY_execute() {
```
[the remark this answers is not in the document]
michalsidzej: [084c0b0](https://github.com/l2beat/l2beat/pull/3765/commits/084c0b07ecbbfd9613834269c55b723736d36232)

packages/backend/src/tools/uif/multi/ManagedMultiIndexer.ts
```diff
@@ -16,13 +15,20 @@ export interface ManagedMultiIndexerOptions<T> extends IndexerOptions {
   decode: (blob: string) => T
   logger: Logger
   updateRetryStrategy?: RetryStrategy
+  newDatabaseMiddleware: () => Promise<DatabaseMiddleware>
```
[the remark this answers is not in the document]
michalsidzej: [273f734](https://github.com/l2beat/l2beat/pull/3765/commits/273f73492e3359dfdbb76dea93b0a9f0924481f4)

maciekop-l2b (approved):

PR #3787, opened 2024-05-22
Add molten l3 full

sekuba:
This is weird.
They announced an op stack L2 and launched an orbit stack L3. And all their websites are currently down (but weren't yesterday)
We will not be able to add them like this.

edit: wow, website updated, replaced op stack with orbit, still claiming CelestiaDA while having AnyTrust 1/1
next suggested steps: replace L2 with L3 and remove the Celestia claim


packages/config/src/projects/layer3s/molten.ts
```diff
@@ -0,0 +1,35 @@
...
+      repositories: [],
+      socialMedia: [
+        'https://twitter.com/UniDexFinance',
+        'https://discord.gg/unidex',
+        'https://mirror.xyz/unidexexchange.eth',
+        'https://t.me/unidexfinance',
+      ],
+    },
+  },
+  rpcUrl: '',
```
lucadonnoh, line 31: ?

lucadonnoh (changes requested):

lucadonnoh (dismissed):
not added to L3s' index.ts

radinat (approved):

PR #3790, opened 2024-05-22
Kinto discovery updates

sekuba:
Closes L2B-5547


lucadonnoh (changes requested):
please update discovery here

lucadonnoh (approved):

PR #3794, opened 2024-05-22
Next frontend

imxeno:
Resolves L2B-5506, L2B-5507, L2B-5568.


packages/frontend2/next.config.js
```diff
@@ -0,0 +1,28 @@
+import { withPlausibleProxy } from 'next-plausible'
+import { env } from './src/env.js'
+
+/** @type {import('next').NextConfig} */
+const nextConfig = {
+  // biome-ignore lint/suspicious/useAwait: rewrites must be
```
torztomasz, line 6: must be what?

packages/frontend2/src/app/layout.tsx
```diff
@@ -0,0 +1,42 @@
...
+}>) {
+  return (
+    <html lang="en">
+      <body className={roboto.className}>
+        <PlausibleProvider
+          domain={env.NEXT_PUBLIC_PLAUSIBLE_DOMAIN}
+          enabled={env.NEXT_PUBLIC_PLAUSIBLE_ENABLED}
+        >
+          <TRPCReactProvider>{children}</TRPCReactProvider>
+        </PlausibleProvider>
```
torztomasz, line 38: not sure if its correct moment, but we should create some `withProviders` stuff

torztomasz (approved):

packages/frontend2/src/app/layout.tsx
imxeno: I guess there won't be that many of them, so I guess it's not really needed...? Anyway we can add it later if we feel like it's worth.

packages/frontend2/next.config.js
imxeno: async, fixed!

PR #3808, opened 2024-05-22
New socket script, vaults updated

sekuba:
A new socket discovery script that adds new vaults semi-automatically from the list of plugs.

It does not directly write results to the configs as that was the most annoying part of the last script (it never wrote the correct things) Instead it suggests copypasta for config.jsonc and socket.ts in a .txt file and outputs a structured json file with all data, including 0 TVL vaults and non-standard connectors, in a separate file.

Thanks to this we have new vaults added and 0-TVL vaults discarded from the escrows in this PR.
Will add upgradability in socket.ts too in the near future as the script already outputs it.

Except for a simple test and hint to run the script, I am keeping it independent of the other code so that we can find a real dev (discovery?) solution in the long term and use this script as a band-aid in the short term, but without it socket vaults are hardly manageable rn.


sekuba:
tests fail because i have axios and we don't, oke

lucadonnoh (approved):

PR #3818, opened 2024-05-23
Display out of sync activity chart

torztomasz:


packages/frontend/src/scripts/charts/renderer/getPointGroups.test.ts
```diff
@@ -0,0 +1,57 @@
+import { expect } from 'earl'
+import { Point } from './ChartRenderer'
+import { getPointGroups } from './getPointGroups'
+
+describe(getPointGroups.name, () => {
```
sz-piotr, line 5: You could make the assumption that dashed is only at the end and simplify the code.
Also each line can be dashed/not separately.
Also, how does this work for stacked chart?

sz-piotr (changes requested):

packages/frontend/src/scripts/charts/view-controller/charts/getActivityRenderParams.ts
```diff
@@ -6,27 +7,39 @@ import { getEntriesByDays } from '../getEntriesByDays'
...
     const points = dataInRange.map(([timestamp, txs, ethTxs]) => {
       const tps = getTps(txs)
       const ethTps = getTps(ethTxs)
+      assert(state.data?.type === 'activity', 'Invalid data type')
+      // const isOutOfSync =
+      //   !!state.data.values.estimatedImpact &&
+      //   !!state.data.values.estimatedSince &&
+      //   state.data.values.estimatedImpact > ESTIMATED_IMPACT_THRESHOLD &&
+      //   timestamp > state.data.values.estimatedSince
+      const isOutOfSync = timestamp > 1708683180
```
sz-piotr, line 32: Is this supposed to be commented out?

sz-piotr (approved):

packages/frontend/src/scripts/charts/view-controller/charts/getActivityRenderParams.ts
torztomasz: I left it just for preview

PR #3837, opened 2024-05-23
Allow templates to suggest target's template

adamiak:


mateuszradomski (approved):
It's a little convoluted but it looks good

PR #3882, opened 2024-05-27
Rewrite donate page

torztomasz:


packages/frontend2/src/types/PropsFrom.ts
```diff
@@ -0,0 +1,3 @@
+export type PropsFrom<TComponent> = TComponent extends React.FC<infer Props>
+  ? Props
+  : never
```
torztomasz, line 3: I believe there was a built-in type, but I can not find it anywhere. Any idea?

torztomasz: yup

packages/frontend/src/pages/donate/view/DonatePage.tsx
```diff
@@ -50,7 +50,7 @@ function Header(props: HeaderProps) {
             transparency to the blockchain space.
           </p>
           <p className="mt-6">
-            We’re committed to delivering accurate and reliable information. We
+            We’re committed to delivering ac\rate and reliable information. We
```
imxeno, line 53: A little oopsie happened here 😉

packages/frontend2/src/app/utils/getMetadata.ts
```diff
@@ -0,0 +1,27 @@
+import { type Metadata } from 'next'
+
+interface MetadataParams extends Metadata {
+  title: string
+  image: string
+  description?: string
+}
+
+export function getMetadata(params: MetadataParams): Metadata {
```
imxeno, line 9: Metadata is automatically merged in the app router, there's no need for this kind of stuff.

packages/frontend/src/pages/donate/view/DonatePage.tsx
torztomasz: good catch

packages/frontend2/src/app/utils/getMetadata.ts
torztomasz: oh, forgot about that

imxeno (approved):

PR #3889, opened 2024-05-27
Return all charts in `/api/tvl2/token` endpoint

michalsidzej:
Needed for https://github.com/l2beat/l2beat/pull/3840

### DIFF
--- a/packages/backend/src/modules/tvl2/api/Tvl2Controller.ts
+++ b/packages/backend/src/modules/tvl2/api/Tvl2Controller.ts
@@ -9,6 +9,7 @@ import {
   NativeAssetBreakdownData,
   ProjectAssetsBreakdownApiResponse,
   ProjectId,
+  TokenTvlApiCharts,
   TvlApiChart,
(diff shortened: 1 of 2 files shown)


packages/backend/src/modules/tvl2/api/Tvl2Controller.ts
```diff
@@ -682,51 +683,110 @@ export class Tvl2Controller {
...
     )
     const pricesMap = new Map(
       prices.map((x) => [x.timestamp.toNumber(), x.priceUsd]),
     )
-
     const amountsByTimestamp = groupBy(amounts, 'timestamp')
 
-    const values: [number, number][] = []
+    const values = new Map<number, { amount: number; value: number }>()
     for (const [timestamp, amounts] of Object.entries(amountsByTimestamp)) {
```
antooni, line 717: Here instead of iterating over entries iterate over timestamps (use getTimestampsForProject)

antooni (approved):

PR #3902, opened 2024-05-28
Detect change in `source` and `includeInTotal`

antooni:
Resolves L2B-5239


packages/backend/src/modules/tvl2/utils/getValuesConfigHash.ts
```diff
@@ -0,0 +1,30 @@
...
+  CoingeckoPriceConfigEntry,
+} from '@l2beat/shared-pure'
+import { createAmountId } from './createAmountId'
+import { createValueId } from './createValueId'
+
+export function getValuesConfigHash(
+  amountConfigs: AmountConfigEntry[],
+  priceConfigs: CoingeckoPriceConfigEntry[],
+): string {
+  const input: string[] = []
```
michalsidzej, line 14: I'd make it more explicit
```suggestion
  const valueIds: string[] = []
```

michalsidzej (approved):
LGTM

PR #3904, opened 2024-05-28
[L2B-5587] Add persistance for verifier status

maciekop-l2b:
Resolves L2B-5587

### DIFF
--- a/packages/backend/src/modules/verifiers/repositories/VerifierStatusRepository.ts
+++ b/packages/backend/src/modules/verifiers/repositories/VerifierStatusRepository.ts
@@ -1,0 +1,77 @@
+import { Logger } from '@l2beat/backend-tools'
+import { ChainId, UnixTime } from '@l2beat/shared-pure'
+import { VerifierStatusRow } from 'knex/types/tables'
+import {
+  BaseRepository,
+  CheckConvention,
+} from '../../../peripherals/database/BaseRepository'
+import { Database } from '../../../peripherals/database/Database'
+
+export interface VerifierStatusRecord {
+  address: string
+  chainId: ChainId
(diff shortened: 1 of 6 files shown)


packages/backend/src/peripherals/database/types.ts
```diff
@@ -261,12 +268,14 @@ declare module 'knex/types/tables' {
     finality: FinalityRow
     tracked_txs_configs: TrackedTxsConfigRow
     l2_costs: L2CostsRow
+    l2_costs_prices: L2CostsPricesRow
```
antooni, line 271: why this change?

michalsidzej (approved):
LGTM

packages/backend/src/peripherals/database/types.ts
maciekop-l2b: I missed this before when the the table was being added

packages/backend/src/modules/verifiers/repositories/VerifierStatusRepository.ts
```diff
@@ -0,0 +1,77 @@
+import { Logger } from '@l2beat/backend-tools'
+import { ChainId, UnixTime } from '@l2beat/shared-pure'
+import { VerifierStatusRow } from 'knex/types/tables'
```
antooni, line 3: IMO it is easier to read if this type is also in the Repository .ts file

maciekop-l2b: On the other hand Row types are heavily coupled with knex table definitions so I would prefer to keep it here.  What's more I would like to prevent referencing modules here as it creates circular dependency: modules => peripherals => modules

PR #3910, opened 2024-05-28
Add hychain full

sekuba:
Closes L2B-5551


sekuba:
oopsie this depends on another one

PR #3911, opened 2024-05-28
Add ignoreInWatchMode to discovered

adamiak:
Resolves L2B-5659

After introduction of template suggestions (`field.target.template`), the main config.jsonc will not always define explicitly which templates will be used, so it's no longer possible to directly read `ignoreInWatchMode` from it for every discovered contract. In the future, this value might be potentially overwritten/extended, meaning it must be gathered from multiple places / template files. That's why `ignoreInWatchMode` is added to discovered.json in its final form.

This PR also follows the direction of having all relevant data in discovered.json, removing requirement of reading config and template files for any subsequent processing of discovery results.


packages/backend/src/modules/implementation-change-report/api/ImplementationChangeController.ts
mateuszradomski, line 1: Higher up from this place we read all the configs which is no longer necessary, we use the config only for `config.name` which could be replaced with `project`. This allows us to not read the config at all in this file.

packages/backend/src/modules/update-monitor/UpdateMonitor.ts
mateuszradomski, line 1: What about counting severities? Maybe that should also go to `discovered.json`? Plus, here we can stop reading the configs and just read the `discovered.json` with `readAllProjectsForChain()`.

packages/backend/src/modules/update-monitor/api/props/utils/getDiff.ts
mateuszradomski, line 1: This function does not need to take the config.

mateuszradomski (changes requested):

packages/backend/src/modules/implementation-change-report/api/ImplementationChangeController.ts
adamiak: Done

packages/backend/src/modules/update-monitor/UpdateMonitor.ts
adamiak: They should go to `discovered.json`, as you suggest. It's going to be done as part of L2B-5690.

mateuszradomski (approved):

PR #3925, opened 2024-05-28
Calls and logs batching

sz-piotr:
Resolves L2B-5888

This PR aims to reimplement discovery provider to make it simpler to do the following:

- query everything, because there would be no need to pass `blockNumber` around
- switch chains, because you'd just call `provider.switchChain('arbitrum', l2BlockNumber)`
- use multicall, because `Promise.all` will put all `provider.call` calls into a single multicall, like `viem`
- batch log calls, because `Promise.all` will batch possible log calls to a single request
- drop support for batch log querying via quicknode

Stuff to do
- [x] Tests for existing functionality
- [x] Discover more issues by running discovery for other projects (currently I only ran against mode)
- [x] Fix issues
- [x] Add `callUnbatched`


packages/backend/src/config/features/updateMonitor.ts
adamiak, line 1: `rpcGetLogsMaxRange` can be removed here and in all other places

adamiak (approved):

PR #3929, opened 2024-05-29
Add vercel toolbar to localhost

torztomasz:


packages/frontend2/next.config.js
```diff
@@ -75,6 +76,7 @@ function createNextConfig() {
   const plugins = [
     createPlausibleProxyPlugin(),
     createNextIntlPlugin('./src/i18n.tsx'),
+    withVercelToolbar(),
```
imxeno, line 79: Here I'd stick with the naming I proposed before, so:
```ts
import createVercelToolbarPlugin from '@vercel/toolbar/plugins/next'
// ...
const plugins = [
  // ...
  createVercelToolbarPlugin(),
]
```
mainly because `withVercelToolbar` suggests a config wrapper, which is not the case here, as the default export is a function that creates a wrapper.

I'd reserve directly using `with*` names for things that don't need to be called before using in the plugins array.

imxeno (approved):

PR #3943, opened 2024-05-29
Merge asset risks to frontend2

maciekzygmunt:
Resolves L2B-5576


packages/frontend2/src/app/(new)/(other)/asset-risks/_components/address-form.tsx
```diff
@@ -25,7 +25,7 @@ export function AddressForm() {
 
   useModal({
     onConnect: ({ address }) => {
-      router.push(`/wallet/${address}`)
+      router.push(`asset-risks/${address}`)
```
imxeno, line 28: It's better to use absolute paths in case we move it somewhere.

packages/frontend2/src/env.js
```diff
@@ -45,6 +46,9 @@ export const env = createEnv({
   client: {
     NEXT_PUBLIC_PLAUSIBLE_DOMAIN: z.string().default('localhost'),
     NEXT_PUBLIC_PLAUSIBLE_ENABLED: coerceBoolean.optional(),
+    NEXT_PUBLIC_WALLETCONNECT_PROJECT_ID: z
+      .string()
+      .default('00000000000000000000000000000000'),
```
imxeno, line 51: Let's remove the default here and bailout if it's missing in the Asset Risks' layout, at least until asset-risks become production-ready.

packages/frontend2/svgr.d.ts
```diff
@@ -5,7 +5,6 @@ declare module '*.svg' {
 }
 
 declare module '*.svg?url' {
-  // eslint-disable-next-line @typescript-eslint/no-explicit-any
-  const content: any
+  const content: string
```
imxeno, line 8: I checked it recently and it's not a string. Svgr also recommends typing it as any.
```suggestion
  // eslint-disable-next-line @typescript-eslint/no-explicit-any
  const content: any
```

imxeno (approved):

PR #4038, opened 2024-06-05
twitter fix & update

IperGiove:
Hi from the Routescan team,
I've noticed that the Twitter link for Avalanche is incorrect. I've updated it and also added some new social media links.


lucadonnoh (approved):

IperGiove:
@lucadonnoh now the test passes

PR #4040, opened 2024-06-06
Hide sections in TVL breakdown if we don't have the tokens

maciekzygmunt:
Resolves L2B-5718


packages/frontend/src/components/tvl-breakdown/ExternallyBridgedTable.tsx
```diff
@@ -14,16 +14,14 @@ export function ExternallyBridgedTable(props: ExternallyBridgedTableProps) {
     return acc + Number(token.usdValue)
   }, 0)
 
-  return (
+  return props.tokens.length === 0 ? null : (
```
torztomasz, line 17: lets go with `props.tokens.length !== 0 ? component : null` easier to read

torztomasz (approved):

PR #4062, opened 2024-06-07
Enable tailwind class sorting in Biome

mateuszradomski:


packages/backend/src/modules/tvl/api/status/TvlStatusPage.tsx
```diff
@@ -31,9 +31,7 @@ export function TvlStatusPage({
   return (
     <Page title="TVL module status (24h)">
       <div
-        className={`card ${
-          aggregatedStatus === 'not synced' ? 'warn' : 'hint'
-        }`}
+        className={`card${aggregatedStatus === 'not synced' ? 'warn' : 'hint'}`}
```
torztomasz, line 34: wont it result in `cardwarn` instead of `card warn`?

packages/backend/src/modules/tvl/api/status/TvlStatusPageDetailed.tsx
```diff
@@ -32,7 +32,7 @@ export function TvlStatusPageDetailed({
   return (
     <Page title="Detailed updater status">
       <div
-        className={`card ${syncStatus === 'not synced' ? 'warn' : 'hint'}`}
+        className={`card${syncStatus === 'not synced' ? 'warn' : 'hint'}`}
```
torztomasz, line 35: same

torztomasz (changes requested):

PR #4089, opened 2024-06-10
Update the way chain is configured for escrows

maciekzygmunt:
Resolves L2B-5256


packages/config/src/discovery/ProjectDiscovery.ts
```diff
@@ -147,7 +149,8 @@ export class ProjectDiscovery {
       tokens,
       contract,
       isUpcoming,
-      isLayer3,
+      chain,
+      includeInTotal,
```
antooni, line 153: if chain is ethereum and include in total is undefined then set it to true

antooni (changes requested):
modify .getEscrowDetails() a little bit

antooni (approved):

PR #4099, opened 2024-06-10
Adjust market share column hide breakpoint

imxeno:


J-Mick (approved):
The share column breakdown appears on 1440px & larger sizes – works as expected.

PR #4121, opened 2024-06-11
getEscrowDetails improvements

maciekzygmunt:
Resolves L2B-5836


antooni (approved):
LGTM, please run it locally (can be read only) to see whether the config for tvl is created correctly

PR #4135, opened 2024-06-11
New milestones for re.al chain

chasebrownn:
Added 8 new milestones on behalf of re.al
Also added all native tokens


chasebrownn:
Will need an admin to run `yarn tokens` I was having issues with this command to generate the generated objects for the new tokens

sekuba:
Hellooo
- removed wreETH (would be double-count as reETH is already counted and the escrow is on L2)
- added lz-bridge where applicable (all NMV totalSupply, full OFT)
- pruned milestones (they should be about the L2 or closely related, and have a launch blog post or similar as the link)

sekuba (approved):

PR #4140, opened 2024-06-12
Update arbitrum links

hyperchris123:
Update arbitrum links and remove archive repo


mateuszradomski:
If you want to farm commits at least don't do it from multiple accounts. #4122

PR #4142, opened 2024-06-12
[L2B-5863] Move logger init closer to app root

maciekop-l2b:
Resolves L2B-5863


packages/backend/src/index.ts
```diff
@@ -18,22 +28,25 @@ async function main() {
...
 
+  const logger = createLogger(environment)
+
   try {
     const config = getConfig()
-    const app = new Application(config)
+    const app = new Application(config, logger)
     await app.start()
   } catch (e) {
     console.error(e)
```
torztomasz, line 38: we can remove this then

torztomasz (approved):

packages/backend/src/index.ts
maciekop-l2b: Removed

PR #4160, opened 2024-06-12
Implement DA main page

torztomasz:


packages/frontend2/types.d.ts
```diff
@@ -0,0 +1,7 @@
+import '@tanstack/react-table' //or vue, svelte, solid, qwik, etc.
```
imxeno, line 1: remove comment

packages/frontend2/src/app/_components/table/table-faceted-filter.tsx
```diff
@@ -0,0 +1,135 @@
...
+          <div className="flex size-3 items-center justify-center rounded-sm bg-black dark:bg-white">
+            <CloseIcon className="size-2.5 fill-white dark:fill-black dark:group-hover:fill-gray-950" />
+          </div>
+        </div>
+      </button>
+    )
+  }
+
+  const showDrawer = breakpoint === 'tablet' || breakpoint === 'mobile'
+  return showDrawer ? (
```
imxeno, line 60: It's very hard to see where the alternative case of this ternary operator starts. Let's maybe do the same as above, so:
```ts
 if(showDrawer) {
   return ...
 }
 
 return ...
 ```
 It's way more readable.

File: packages/frontend2/src/app/(new)/data-availability/summary/_components/table/da-bridge-cell.tsx [hunk 3]
```diff
@@ -0,0 +1,24 @@
+import React from 'react'
+import { type DataAvailabilityProvider } from './columns'
+
+interface Props {
+  daBridge: DataAvailabilityProvider['daBridge']
+}
+export function DaBridgeCell({ daBridge }: Props) {
+  if (!daBridge) return 'No bridge'
```
imxeno, line 8: Should this really return plain string?

imxeno (approved):

packages/frontend2/src/app/(new)/data-availability/summary/_components/table/da-bridge-cell.tsx
torztomasz: yup, no need to wrap it in anything

PR #4162, opened 2024-06-12
Add gathered meta to discovered.json

adamiak:
Resolves L2B-5663

After this change the discovery process will gather every target meta specified in templates (e.g. `target.role`, `target.category`, `target.permission`), create a union of everything gathered per contract (e.g. some template might say that 'target.role' is 'Challanger' while a different contract can say that 'target.role' is 'Guardian' for the same contract) and puts all gathered meta into contract data in discovered.json.

Currently I'm ignoring all those values when diffing to not trigger update-monitor. Once we're happy with the structure, I'll update discovered.json files in bulk.


mateuszradomski (approved):
Looks nice

PR #4163, opened 2024-06-12
DA template improvements 

sdlyy:
Resolves L2B-5878

### DIFF
--- a/packages/config/src/projects/other/da-beat/types/DaLayer.ts
+++ b/packages/config/src/projects/other/da-beat/types/DaLayer.ts
@@ -38,6 +38,12 @@ export type BlockchainDaLayer = CommonDaLayer & {
    * @unit seconds
    */
   consensusFinality: number
+
+  /**
+   * Duration of time for unbonding in seconds
+   * @unit seconds
+   */
+  unbondingPeriod: number
 }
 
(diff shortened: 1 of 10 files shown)


packages/config/src/projects/other/da-beat/types/DaLayer.ts
```diff
@@ -32,6 +32,12 @@ export type BlockchainDaLayer = CommonDaLayer & {
    * @unit seconds
    */
   storageDuration: number
+
+  /**
+   * Consensus finality time.
+   * @unit seconds
+   */
+  consensusFinality: number
```
vincfurc, line 40: unbondingPeriod is usually an hardcoded parameter that would require an hardfork to change, so we can specify it manually here too (e.g., for Celestia it can be taken from here https://celestiaorg.github.io/celestia-app/specs/params.html - consensus.evidence.MaxAgeDuration)

sdlyy: be my guest
[feat: add unbonding period to blockchain dalayer definition](https://github.com/l2beat/l2beat/pull/4163/commits/0d8d4f4e7c97501c3ba490c649af36cb711c0425)

vincfurc (approved):

PR #4189, opened 2024-06-14
Allow to configure canonical L2 tokens

antooni:
Resolves L2B-5521
Resolves L2B-4555


maciekop-l2b (approved):

maciekop-l2b:
@antooni Approved, but:
- did we consult with researchers the rename from `CBV`, `EBV`, `NMV` to `canonical`, `external`, `native`?
- it would be a bit cleaner doing it in two steps 1. Rename 2. Add new supply type (but let's leave it as it is)

PR #4192, opened 2024-06-17
Update ZK Catalog

maciekzygmunt:
Resolves L2B-5939

### DIFF
--- a/packages/frontend/src/pages/zk-catalog/view/ZkCatalogView.tsx
+++ b/packages/frontend/src/pages/zk-catalog/view/ZkCatalogView.tsx
@@ -75,7 +75,7 @@ export function ZkCatalogView(props: ZkCatalogViewProps) {
               tooltip="Shows if a trusted setup is used anywhere in the proving stack."
               className="flex-row items-baseline justify-between md:flex-col md:items-start md:justify-start"
             >
-              {item.hasTrustedSetup ? 'Yes' : 'No'}
+              {item.hasTrustedSetup}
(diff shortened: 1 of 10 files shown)


packages/frontend/src/pages/zk-catalog/view/ZkCatalogView.tsx
```diff
@@ -75,7 +75,7 @@ export function ZkCatalogView(props: ZkCatalogViewProps) {
               tooltip="Shows if a trusted setup is used anywhere in the proving stack."
               className="flex-row items-baseline justify-between md:flex-col md:items-start md:justify-start"
             >
-              {item.hasTrustedSetup ? 'Yes' : 'No'}
+              {item.hasTrustedSetup}
```
torztomasz, line 78: if its string then lets just name it `trustedSetup`

torztomasz (changes requested):

packages/frontend/src/utils/zk-catalog/trustedSetup.ts
```diff
@@ -0,0 +1,20 @@
+import { ZkCatalogOnchainVerifier } from './types'
+
+export function trustedSetup(verifiers: ZkCatalogOnchainVerifier[]) {
```
torztomasz, line 3: well, this could be named `getTrustedSetup` or smth. We do not have this kind of pattern in our codebase

torztomasz (approved):

PR #4235, opened 2024-06-20
Fix integration tests

sdlyy:
Resolves L2B-6033
- Whitelisted envs via turbo spec
- Down-side migration regenerated from pre-117 table declaration SQL
- These do not have to be super accurate since we will be squashing them as a baseline once migration to Prisma is in place

and it actually works 😬 
image
image
5 pending from BigQuery client

### DIFF
--- a/turbo.json
+++ b/turbo.json
@@ -1,53 +1,25 @@
 {
   "$schema": "https://turbo.build/schema.json",
-  "globalDependencies": [
-    "tsconfig.json"
-  ],
+  "globalEnv": ["TEST_DB_URL", "CI"],
+  "globalDependencies": ["tsconfig.json"],
   "tasks": {
     "build": {
-      "dependsOn": [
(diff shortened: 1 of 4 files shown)


turbo.json
```diff
@@ -1,53 +1,25 @@
 {
   "$schema": "https://turbo.build/schema.json",
-  "globalDependencies": [
-    "tsconfig.json"
-  ],
+  "globalEnv": ["TEST_DB_URL", "CI"],
```
torztomasz, line 3: aren't those only related to backend?

sdlyy: yessssir, this should do the trick unless there is a better option to specify these specially for backend/test
https://github.com/l2beat/l2beat/pull/4235/commits/bc6a5ca58418635d896aa2bf64ae707c3ecfeeff

imxeno (approved):

PR #4241, opened 2024-06-20
fix: correct the hyperlinks related to JSON-RPC

aganisgash:
you are welcome. : ]


mateuszradomski:
Those files are not written by humans, they're auto generated from source code found on chain. We're using them to find similarities between other files, changing them is the complete opposite of what we want to do.

PR #4243, opened 2024-06-21
Add ZRO token

radinat:
Resolves L2B-6008
https://www.coingecko.com/en/coins/layerzero


sekuba:
ZRO is full OFT (burn-mint everywhere) so i will add it to the other chains first, otherwise we would find 0 TVL

PR #4260, opened 2024-06-21
Port scaling summary

torztomasz:
Please do review CAREFULLY as it is a foundation for the future page migration

- [x] Hovers in table

Co-authored-by: Piotr Adamczyk <xeno@xeno.yt>


packages/frontend2/src/app/_components/callout.tsx
```diff
@@ -0,0 +1,57 @@
...
+    background = 'bg-red-600/20'
+  } else if (color === 'yellow') {
+    background = 'bg-yellow-700/20'
+  } else if (color === 'blue') {
+    background = 'bg-blue-700/20'
+  } else if (color === 'gray') {
+    background = 'bg-gray-500/20'
+  } else {
+    background = ''
+  }
```
imxeno, line 38: Can we use variants here? I saw you were doing a refactor like this somewhere else.

packages/frontend2/src/app/_components/callout.tsx
```diff
@@ -0,0 +1,57 @@
...
+        'flex',
+        'rounded-lg',
+        background,
+        className,
+        small ? 'gap-2 rounded text-sm' : 'gap-3',
+      )}
+    >
+      <span>{icon}</span>
+      <div className="flex flex-col gap-2">
+        {message && message}
```
imxeno, line 52: Just `{message}` is fine 😉

packages/frontend2/src/app/_components/table/cells/index-cell.tsx
```diff
@@ -0,0 +1,20 @@
...
+
+export function IndexCell({ index, className }: IndexCellProps) {
+  return (
+    <span
+      className={cn(
+        'ml-auto tabular-nums font-medium text-gray-50 text-xs dark:font-normal dark:text-gray-600',
+        className,
+      )}
+    >
+      {index + 1}
```
imxeno, line 17: Can we do sth about this? I don't really get the idea of supplying an index just to display `index + 1` here if the component is named IndexCell.

packages/frontend2/src/app/_components/table/cells/project-name-cell.tsx
```diff
@@ -0,0 +1,119 @@
...
+        <NotSyncedBadge
+          className="relative top-[-3px] ml-2"
+          syncedUntil={project.data?.syncStatus.syncedUntil}
+        />
+      )}
+    </div>
+  )
+}
+
+// TODO: Move it somewhere else later
```
imxeno, line 104: Has the time come yet, or we can live with this there?

packages/frontend2/src/app/_components/table/cells/stage-cell.tsx
```diff
@@ -0,0 +1,144 @@
...
+  stageConfig: StageConfig
+}
+
+export function StageCell({ stageConfig }: StageCellProps) {
+  if (stageConfig.stage === 'NotApplicable') {
+    return <StageBadge stage={stageConfig.stage} oneSize />
+  }
+
+  return (
+    // TODO: disabledOnMobile
```
imxeno, line 23: 👀

packages/frontend2/src/app/_components/table/cells/stage-cell.tsx
```diff
@@ -0,0 +1,144 @@
...
+      return 'Stage under review'
+    case 'Stage 0':
+      return 'Full training wheels'
+    case 'Stage 1':
+      return 'Limited training wheels'
+    case 'Stage 2':
+      return 'No training wheels'
+    default:
+      return ''
+    // assertUnreachable(stage)
```
imxeno, line 131: Shouldn't this comment be removed?

packages/frontend2/src/app/_components/table/sorting/functions/sentiment-sorting.ts
```diff
@@ -0,0 +1,36 @@
...
+  }
+  const sentimentOrderValueB = getSentimentOrderValue(rowB)
+  if (sentimentOrderValueB === undefined) {
+    return 1
+  }
+
+  return sentimentOrderValueA > sentimentOrderValueB ? 1 : -1
+}
+
+export function getSentimentOrderValue<T extends { sentiment: Sentiment }>(
```
imxeno, line 19: Can't we add sth like `const sentimentOrder = ['bad', 'warning', 'neutral', 'good']` so we can do `sentimentOrder.indexOf('bad') inside`? I mean I get why this is a function, as you can call it with any object with sentiment, but this switch and if can be replaced with a oneliner. There will never be a case where we have equal sentiments, because it doesn't make sense.

packages/frontend2/src/styles/globals.css
```diff
@@ -28,6 +28,12 @@ body {
   mask-image: linear-gradient(90deg, transparent, black 30px calc(100% - 30px), transparent) !important;
 }
 
+@layer utilities {
+  .transition-duration-50 {
+    transition-duration: 50ms !important;
+  }
+}
```
imxeno, line 35: Can't we extend the theme in tailwind?

```ts
    extend: {
     // ...
      transitionDuration: {
        '50': '50ms',
      }
```

imxeno (changes requested):

packages/frontend2/src/app/_components/table/cells/project-name-cell.tsx
torztomasz: Actually I would leave it as it is now

packages/frontend2/src/app/_components/table/cells/stage-cell.tsx
torztomasz: yup, copy paseted it

packages/frontend2/src/app/_components/table/sorting/functions/sentiment-sorting.ts
torztomasz: Sure, nice change

packages/frontend2/src/hooks/use-on-click-outside.tsx
```diff
@@ -0,0 +1,42 @@
...
+  handler: (event: MouseEvent | TouchEvent | FocusEvent) => void,
+  eventType: EventType = 'mousedown',
+  eventListenerOptions: AddEventListenerOptions = {},
+): void {
+  useEventListener(
+    eventType,
+    (event) => {
+      const target = event.target as Node
+
+      // Do nothing if the target is not connected element with document
```
imxeno, line 24: ?

imxeno (approved):
LGTM, but make sure to rename the page directory before merging. Maybe `/scaling/summary-next` with meta robots: `{ index: false }`?

PR #4261, opened 2024-06-21
Remove "we're hiring" badge

imxeno:


sdlyy (changes requested):
Still visible on next production preview
image

sdlyy (approved):

PR #4267, opened 2024-06-24
zklink nova: Reclassify tokens as externally bridged (apart from linea)

sekuba:


packages/config/src/projects/layer3s/zklinknova.ts
```diff
@@ -108,62 +132,98 @@ export const zklinknova: Layer3 = underReviewL3({
...
     },
     {
       chain: 'arbitrum',
       includeInTotal: false,
       address: EthereumAddress('0xfB0Ad0B3C2605A7CA33d6badd0C685E11b8F5585'),
       sinceTimestamp: new UnixTime(1709296973),
       tokens: '*',
+      source: 'external',
+      bridge: {
+        name: 'zkLink Nova Bridge',
```
lucadonnoh, line 181: this should be arbitrum bridge etc

lucadonnoh (changes requested):

packages/config/src/projects/layer3s/zklinknova.ts
sekuba: like this?

PR #4276, opened 2024-06-24
Refactor EtherscanClient

antooni:
There was a code duplication in EtherscanClient and BlockscoutClient, this PR merges them into one class and parametrizes it


packages/backend/src/modules/tvl/modules/ChainModule.ts
```diff
@@ -86,26 +86,32 @@ function createChainModule(
...
+  const options =
+    chainConfig.config.blockTimestampClientConfig === undefined
+      ? undefined
+      : chainConfig.config.blockTimestampClientConfig.type === 'etherscan'
+        ? {
+            type: 'Etherscan' as const,
+            apiKey:
+              chainConfig.config.blockTimestampClientConfig.etherscanApiKey,
+            url: chainConfig.config.blockTimestampClientConfig.etherscanApiUrl,
+            maximumCallsForBlockTimestamp: 3,
```
maciekzygmunt, line 103: SHould we use `env` for this?

antooni: in theory we can but in practice I not see this one as being changed frequently, we already have a lot of envs, I am afraid this will introduce not needed noise

maciekop-l2b (approved):
LGTM

maciekzygmunt (approved):

PR #4277, opened 2024-06-24
Add Orderly escrows, description and archi diagram

sekuba:


packages/config/src/projects/layer2s/orderly.ts
```diff
@@ -39,6 +41,63 @@ export const orderly: Layer2 = opStackL2({
...
+      sinceTimestamp: new UnixTime(1705831672),
+      tokens: ['USDC'],
+      source: 'external',
+      bridge: {
+        name: 'Mantle escrow -> LayerZero AMB',
+      },
+    },
+    {
+      chain: 'ethereum',
+      includeInTotal: false,
```
lucadonnoh, line 91: why is this false here? is there any double counting with an ethereum escrow? or are there other reasons

lucadonnoh (approved):

PR #4279, opened 2024-06-25
Fix DA mode misalignment & transaction data typo

imxeno:


packages/frontend/src/pages/scaling/finality/props/getScalingFinalityColumnsConfig.tsx
```diff
@@ -28,7 +28,8 @@ export function getScalingFinalityColumnsConfig() {
     {
       name: 'DA MODE',
       getValue: (project) =>
-        project.dataAvailabilityMode ?? <span>&mdash;</span>,
+        // Replace the first space with a non-breaking space
+        project.dataAvailabilityMode?.replace(' ', ' ') ?? <span>&mdash;</span>,
       className: 'whitespace-nowrap md:whitespace-normal',
```
torztomasz, line 33: can't we just use whitespace-pre and keep everything in one line?

torztomasz (approved):

packages/frontend/src/pages/scaling/finality/props/getScalingFinalityColumnsConfig.tsx
imxeno: We still want the `(compressed)` part to wrap. It saves some space on smaller screens.

PR #4291, opened 2024-06-25
Getting TVL for DA providers

sdlyy:
Resolves L2B-5657


packages/database/src/value/repository.ts
```diff
@@ -106,4 +106,39 @@ export class ValueRepository {
...
+            projectIds.map((id) => id.toString()),
+          ),
+      )
+      .selectFrom('latest_values')
+      .select(selectValue)
+      .where('combination_number', '=', 1)
+      .execute()
+
+    return rows.map(toRecord)
+  }
```
torztomasz, line 143: we do not have tests in the db package?

packages/frontend2/src/server/features/data-availability/utils/get-da-projects-tvl.ts
```diff
@@ -0,0 +1,56 @@
...
+        return acc
+      },
+      { canonical: 0n, external: 0n, native: 0n },
+    )
+
+    const tvl = canonical + external + native
+
+    return {
+      projectId: ProjectId(projectId),
+      tvl: Number(tvl),
```
torztomasz, line 35: is it safe to cast it to a number here?

torztomasz (changes requested):

packages/frontend2/src/server/features/data-availability/utils/get-da-projects-tvl.ts
sdlyy: Yea I would say so given:
- tvl is denominated in fiat/USD cents:
- max safe int here is around 9 quadrillion

packages/database/src/value/repository.ts
sdlyy: I'll try to set up the suite on the database package side so we can have at least some basic foundation for later use

.github/workflows/ci.yml
```diff
@@ -93,7 +93,7 @@ jobs:
         uses: actions/setup-node@v3
         with:
           node-version: 18.x
-          cache: "yarn"
+          cache: 'yarn'
```
torztomasz, line 96: why is that?

torztomasz (changes requested):

.github/workflows/ci.yml
sdlyy: https://github.com/l2beat/l2beat/pull/4291/commits/42981c6bc52b2a00d5ce9331af589bbb29cd2cce

Leftover - was debugging CI

torztomasz (approved):

PR #4295, opened 2024-06-26
Allow to specify an address in the call handler

mateuszradomski:
Resolves L2B-6054


packages/discovery/src/discovery/handlers/user/CallHandler.ts
```diff
@@ -18,6 +18,7 @@ export const CallHandlerDefinition = z.strictObject({
   ignoreRelative: z.optional(z.boolean()),
   pickFields: z.optional(z.array(z.string())),
   expectRevert: z.optional(z.boolean()),
+  inAddress: z.optional(z.string()),
```
adamiak, line 21: I checked with AIs and `atAddress` would be more correct here.

adamiak (approved):

PR #4305, opened 2024-06-26
Get Orbit Sequencers and Validators straight from Discovery (by Role)

adamiak:
Resolves L2B-5757

Fixes a long-standing issue of having only one sequencer and validator in many Orbit chains due to limitations of legacy stack templates.

Changes in this PR:
* all Orbit chain configs and templates have `Sequencer` and `Validator` roles assigned in `target.role`. (I didn't try to templatize configs as this must be done properly for all orbit stacks in a separate effort)
* `orbitStack.ts` calls a new method `getPermissionsByRole(role)` to fetch sequencers and validators.

Important comment: I ran full discovery (without `--dev`) for most orbit chains, so there might be some updates in their `discovered.json` but never anything significant that would go to `diffHistory.json`

(I have also rendered all projects locally and made sure that the only change that happens is addition of validators and sequencers).


packages/config/src/discovery/ProjectDiscovery.test.ts
```diff
@@ -13,7 +13,7 @@ import {
 } from '../test/stubs/discoveredJson'
 import { ProjectDiscovery } from './ProjectDiscovery'
 
-describe(ProjectDiscovery.name, () => {
+describe.only(ProjectDiscovery.name, () => {
```
mateuszradomski, line 16: `.only` 😱

mateuszradomski (approved):

PR #4334, opened 2024-07-01
Trivial discovery updates 07/01

sekuba:


packages/backend/discovery/metis/ethereum/diffHistory.md
```diff
@@ -1,3 +1,50 @@
+Generated with discovered.json: 0xf7442661475f0d41ad948d2a8d42265e1aaa8ee6
+
+# Diff at Mon, 01 Jul 2024 11:22:03 GMT:
+
+- author: sekuba (<29250140+sekuba@users.noreply.github.com>)
+- comparing to: main@b5540be6f0b9cb1f69a05dba32413b0eae0acbf4 block: 19926649
+- current block number: 20211335
+
+## Description
+
```
lucadonnoh, line 10: sounds big? which sequencers and which collateral? just theirs or others too? should the permission section be updated? need elaboration

sekuba: No it is just their own collateral that they had to stake for operating as a permissioned sequencer, and they can only withdraw after they are switched to inactive.
So it is inconsequential, and should have been there from the start.

```
function withdraw(
        uint256 _seqId,
        uint256 _amount
    ) external whenNotPaused whitelistRequired {
        Sequencer storage seq = sequencers[_seqId];
        if (seq.status != Status.Active) {
            revert SeqNotActive();
        }

        if (seq.owner != msg.sender) {
            revert NotSeqOwner();
        }

        // owner can only withdraw once in the current reward period
        require(curBatchState.id > seq.updatedBatch, "withdraw throttle");

        uint256 locked = seq.amount - _amount;
        uint256 nonce = seq.nonce + 1;

        seq.nonce = nonce;
        seq.amount = locked;
        seq.updatedBatch = curBatchState.id;

        escrow.withdrawLocking(_seqId, seq.owner, nonce, _amount, locked);
    }
  ```

lucadonnoh (approved):

PR #4368, opened 2024-07-03
Collapsible project descriptions

imxeno:


packages/frontend/src/components/ReadMore.tsx
```diff
@@ -0,0 +1,61 @@
...
+        {collapsible}
+      </span>
+      <span className="md:hidden"> </span>
+      <button
+        data-role="read-more-toggle"
+        className="cursor-pointer underline md:hidden"
+      >
+        <span className="group-data-[collapsed=false]:hidden">Read more</span>
+        <span className="hidden group-data-[collapsed=false]:inline">
+          Hide description
```
torztomasz, line 56: shouldn't this be `Read less`?

torztomasz (changes requested):
It doesn't work on storybook

torztomasz (approved):

PR #4388, opened 2024-07-05
Improve docker build stages

kondziorex7:
Use `turbo prune` functionality to exclude unnecessary packages and dependencies from final package.

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -10,3 +10,7 @@ node_modules
 .vscode
 .idea
 .*.swp
+
+# Docker
+.dockerignore
+Dockerfile.*
--- a/Dockerfile.backend
+++ b/Dockerfile.backend
@@ -1,10 +1,19 @@
-FROM node:18 as builder
 ARG ARCH=x86_64
 
-COPY . /src/
-RUN cd /src/ && \
-  yarn install && \
-  yarn build:backend
+FROM node:18 as pruner
+WORKDIR /src/
+COPY . .
+RUN yarn global add turbo
+RUN turbo prune --docker @l2beat/backend
+
+FROM node:18 as builder
+ARG ARCH
+WORKDIR /src/
+COPY --from=pruner /src/out/json/ .
+RUN yarn install
+COPY --from=pruner /src/out/full/ /src/tsconfig.json ./
+RUN yarn build:backend
+
 RUN mkdir /dist/ && \
   cd /dist/ && \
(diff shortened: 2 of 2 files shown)


.dockerignore
```diff
@@ -10,3 +10,7 @@ node_modules
 .vscode
 .idea
 .*.swp
+
+# Docker
+.dockerignore
```
imxeno, line 15: Aren't dotfiles ignored by default, or at least not added if you don't explicitly state you want them?

Dockerfile.backend
```diff
@@ -1,10 +1,20 @@
...
-COPY . /src/
-RUN cd /src/ && \
-  yarn install && \
-  yarn build:backend
+FROM node:18 as pruner
+WORKDIR /src/
+COPY . .
+RUN yarn global add turbo
+RUN yarn prune:backend
+# RUN turbo prune --docker @l2beat/backend
```
imxeno, line 8: I guess we should either remove this comment or uncomment it and remove the prune:backend script. It's so specific to docker it won't ever be used anywhere else (for purposes other than debugging docker issues).

Dockerfile.backend
```diff
@@ -1,10 +1,20 @@
...
+
+FROM node:18 as builder
+ARG ARCH
+WORKDIR /src/
+COPY --from=pruner /src/out/json/ /src/tsconfig.json ./
```
imxeno, line 13: Is tsconfig at this stage?

Dockerfile.backend
```diff
@@ -1,10 +1,20 @@
-FROM node:18 as builder
 ARG ARCH=x86_64
 
-COPY . /src/
-RUN cd /src/ && \
-  yarn install && \
-  yarn build:backend
+FROM node:18 as pruner
```
imxeno, line 3: Is there a reason we don't use an alpine linux images?

imxeno (approved):

Dockerfile.backend
kondziorex7: No, as discussed offline difftastic may be actually broken after such migration. I'll test it separately to this PR

Dockerfile.backend
kondziorex7: Moved to post-install/pre-build copy

Dockerfile.backend
kondziorex7: makes sense, also comment was left here unintentionally. Fixed

PR #4404, opened 2024-07-06
Get amounts by ids for breakdown

antooni:
Previously the SQL was getting all the amounts for timestamp, but that can lead to a bug when configuration gets deleted. It is the most safe approach to just get amounts by ids - this way we always get only amounts for configurations from runtime.

It is worth knowing that our ManagedMultiIndexer is able to handle this case - when config gets deleted amounts get deleted. But I believe it is more stable path forward to code this functionality via getting by ids, we do not know whether in the future all the amounts will be cleared after config removal


maciekop-l2b (approved):
LGTM

PR #4405, opened 2024-07-06
Remove unused logic from MultiIndexer

antooni:
`.getInitialConfigurations()` functionality was added to MultiIndexer in the past when we had an idea to have configurations with dynamic ids set by the DB autoincrement. We have ditched the idea later on, but the code has stayed and is not used anymore. This PR is a cleanup.


maciekop-l2b (approved):
LGTM

PR #4406, opened 2024-07-07
OpStack to read descriptions, relations and permissions from discovered.json

adamiak:
Resolves L2B-5088, L2B-6309, L2B-6409, L2B-6335, L2B-6310, L2B-6244

This PR provides implementation that allows to render Permissions and Contract sections on the project page directly from `discovered.json` file. This is done for zora chain only as an example.

A lot of templates have been updated with better descriptions which triggered a lot of changes in OP stack project discoveries.

Other features that were added:
* Gnosis Safe descriptions contain descriptions of all discovered modules
* Variables inside template descriptions are interpolated (e.g. `{{ minThreshold }}`)
* Required shape match was raised to 100%
* Hashes of used templates are saved to `discovered.json` to find if rediscovery is needed
* Hash of all shape files is saved to `discovered.json` to know when rediscovery is needed
* Tests were added to check if all hashes are up to date
* A new script `yarn refresh-discovery` was added that finds projects that need "refresh" (`discovery --refresh` which is like --dev but also updates diffHistory) and runs `discovery --refresh` on them


packages/discovery/src/cli/singleDiscoveryCommand.ts
```diff
@@ -35,7 +35,7 @@ export async function singleDiscoveryCommand(
   logger = logger.for('SingleDiscovery')
   logger.info('Starting')
 
-  const { result, blockNumber } = await discover(
+  const { result, blockNumber, templateService } = await discover(
```
mateuszradomski, line 38: Wouldn't it be better to return the hash here instead of the TemplateService?

mateuszradomski (approved):
Expect for one comment looks solid

packages/discovery/src/cli/singleDiscoveryCommand.ts
adamiak: :+1:

PR #4015, opened 2024-06-04
Add termstructure full

sekuba:
Closes L2B-5760

PR #4422, opened 2024-07-09
Add Term Structure (full)

lucadonnoh:
Resolves L2B-5760. Closes #4015. Some notes can be found here https://www.notion.so/l2beat/Term-Structure-858f8e4cb9a14678b55dacea07d9bbb4


sekuba (changes requested):
need to add eoa red banner
```
redWarning:
      'Critical contracts can be upgraded by an EOA which could result in the loss of all funds.',
```

sekuba (approved):

PR #4440, opened 2024-07-11
Add DA risks

mateuszradomski:


maciekzygmunt (approved):

packages/frontend/src/components/navbar/NavWrapper.tsx
```diff
@@ -353,6 +353,10 @@ export function Sidenav({
                 Jobs
                 {config.features.hiringBadge && <HiringBadge />}
               </NavSmallLink>
+              <NavSmallLink
```
sz-piotr, line 356: For some reason doesn't show up on frontend2

sz-piotr (changes requested):

sz-piotr (approved):

PR #4441, opened 2024-07-11
Merge upgradeability into values

sz-piotr:
This PR removes the upgradeability section in discovery output and replaces it with well defined:
- `$immutable`
- `$implementation`
- `$values`
fields.

Things left to do:
- [x] Handle solidity variables using `$`
- [x] Resolve issues with diffHistory hashes for all projects


mateuszradomski (approved):
LGTM

PR #4451, opened 2024-07-12
Make MXC underreview due to newly verified contracts

sekuba:
Their contracts are newly verified, will probably review them over the weekend.


radinat (approved):

sheenhx:
@sekuba hey what is wrong after this? contracts verified and then why remove us?

PR #4458, opened 2024-07-14
added RWA token to re.al tokens

chasebrownn:
Added RWA token to tokens for real chain.

### DIFF
--- a/packages/config/src/tokens/tokens.jsonc
+++ b/packages/config/src/tokens/tokens.jsonc
@@ -3263,6 +3263,13 @@
         "bridge": "Layer Zero",
         "slug": "omnichain"
       }
+    },
+    {
+      "symbol": "RWA",
+      "address": "0x4644066f535Ead0cde82D209dF78d94572fCbf14",
+      "source": "native",
+      "formula": "totalSupply",
+      "coingeckoId": "re-al"
     }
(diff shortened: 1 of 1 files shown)


sekuba:
Sorry, we are currently unable to support this token:
- it has full 200G supply on all supported chains (would be doublecounted if added)
- the circulating supply on coingecko is spread over different L2s and ethereum

For the next time, please try to follow [this lil guide](https://github.com/l2beat/l2beat/blob/main/docs/tvl.md#token-guide)

thanks ;)

chasebrownn:
RWA is not a cross chain token. It only exists on re.al so I’m confused as to your comment @sekuba

sekuba:
Sorry for that, i was thinking about the Xend RWA with the same ticker.
Do you have a working rpc for me that is not real.drpc.org?

chasebrownn:
@sekuba No public rpcs, but we have private rpcs that we can provide you guys as long as it's obfuscated. Can we reopen this PR?

sekuba:
so the drpc worked just now in contrast to this morning and i was able to add it
will see how it goes and msg you again in case we need a better one 
(if they work i prefer the public ones so we do not have to manage the secrets)
i continued at #4505 ;)

PR #4508, opened 2024-07-17
Use some new repositories in backend

sz-piotr:


packages/database/src/sequence-processor/repository.ts
```diff
@@ -24,7 +24,7 @@ export class SequenceProcessorRepository {
       .where('id', '=', id)
       .executeTakeFirst()
 
-    return row ? toRecord(row) : null
```
imxeno, line 27: Is this an intended change? `null` in this scenario is used in all repositories iirc

packages/database/src/Database.ts
imxeno, line 1: I guess this file should be named `database.ts`.

packages/database/src/test/setup.ts
```diff
@@ -1 +1,9 @@
+import { getLegacyTestDatabase } from '@l2beat/database-legacy'
+
 process.env.NODE_ENV = 'test'
+
```
imxeno, line 4: We didn't include
```js
waitForExpect.defaults.timeout = 1500
waitForExpect.defaults.interval = 10
```
there, was it on purpose?

packages/backend/src/modules/activity/counters/StarkexCounter.test.ts
```diff
@@ -2,37 +2,33 @@ import { Logger } from '@l2beat/backend-tools'
...
-      addOrUpdateMany: async () => -1,
-    })
     const txCount = 1
     const starkexClient = mockObject<StarkexClient>({
       getDailyCount: async () => txCount,
     })
-    const transaction = mockObject<Knex.Transaction>({})
-    const sequenceProcessorRepository = mockObject<SequenceProcessorRepository>(
-      {
+    const starkExTransactionCount = mockObject<
```
imxeno, line 19: Why is this mock defined outside the `db` object below?

sz-piotr: Otherwise earl expects seem to be not working correctly. Decided it wasn't important enough to worry about

packages/database/src/test/setup.ts
sz-piotr: We don't really use it outside of two specific tests in the backend package

imxeno (approved):

PR #4539, opened 2024-07-19
Update DeGate website and app links

BergDev22:


mateuszradomski:
Hello,

Sorry but we don't support tracker tags inside served links. We want to keep the URLs as clean as possible.

PR #4540, opened 2024-07-19
Added badges to Fuel

AK-Cza:


lucadonnoh:
please please please dont add badges to upcoming projects

PR #4542, opened 2024-07-19
Check if permissioned addresses are verified

adamiak:
When running `yarn check-verified-contracts`, also check addresses from "Permissions" section.

Resolves L2B-6489


torztomasz (approved):

torztomasz:
By the way, this part of the code is very crucial. It's because on the frontend the logic goes as follows: 
- if verified or **not on the list**: do not do anything
- if not verified (on the list but with false value): do smth

Can we add some serious test for that?

mateuszradomski (approved):

PR #4564, opened 2024-07-22
Add batchExecute and use it across new repositories

sz-piotr:
This PR also:
- replaces `eb.and` calls with multiple `.where` calls as they are equivalent.
- standardizes `.onConflict((eb) => ({ foo: eb.ref('excluded.foo') }))` instead of `.onConflict({ foo: (eb) => eb.ref('excluded.foo') })`


packages/database/src/aggregated-l2-cost/repository.ts
```diff
@@ -20,32 +19,31 @@ export class AggregatedL2CostRepository {
...
-              compute_gas_eth: (eb) => eb.ref('excluded.compute_gas_eth'),
-              compute_gas_usd: (eb) => eb.ref('excluded.compute_gas_usd'),
-              overhead_gas: (eb) => eb.ref('excluded.overhead_gas'),
-              overhead_gas_eth: (eb) => eb.ref('excluded.overhead_gas_eth'),
-              overhead_gas_usd: (eb) => eb.ref('excluded.overhead_gas_usd'),
-            }),
-          )
-          .execute()
-      }
+    // NOTE: This used to be 10_000, but was changed. (sz-piotr): No idea why
```
imxeno, line 22: I guess it has something to do with the prepared query param count limit? cc: @sdlyy

packages/database/src/value/repository.ts
```diff
@@ -49,34 +50,30 @@ export class ValueRepository {
   }
 
   async addOrUpdateMany(records: ValueRecord[]): Promise<number> {
-    await this.db.transaction().execute(async (trx) => {
-      for (let i = 0; i < records.length; i += BATCH_SIZE) {
-        await this._addOrUpdateMany(records.slice(i, i + BATCH_SIZE), trx)
-      }
+    const rows = records.map(toRow)
+
+    await batchExecute(this.db, rows, BATCH_SIZE, async (trx, batch) => {
```
imxeno, line 55: BATCH_SIZE could be inlined as in `aggregated-l2-cost`

packages/database/src/aggregated-l2-cost/repository.ts
sdlyy: yessirr

imxeno (approved):

sdlyy (approved):
🙇🏻

PR #4565, opened 2024-07-22
Add ActivityIndexer

maciekzygmunt:


packages/backend/src/modules/activity2/indexers/ActivityIndexer.ts
```diff
@@ -0,0 +1,69 @@
...
+}
+
+export class ActivityIndexer extends ManagedChildIndexer {
+  constructor(private readonly $: ActivityIndexerDeps) {
+    const logger = $.logger.tag($.projectId)
+    super({ ...$, logger, name: `activity_indexer_${$.projectId}` })
+  }
+
+  override async update(from: number, to: number): Promise<number> {
+    const batchSize = this.$.batchSize ?? 100
```
antooni, line 24: batch size should always be defined, this ?? will be used during object instantiation

packages/backend/src/modules/activity2/indexers/ActivityIndexer.ts
```diff
@@ -0,0 +1,69 @@
...
+
+    const adjustedTo = from + batchSize < to ? from + batchSize : to
+    const counts = await this.$.txsCountProvider.getTxsCount(from, adjustedTo)
```
antooni, line 27: return array of ActivityRecord from this method

packages/backend/src/modules/activity2/indexers/ActivityIndexer.ts
```diff
@@ -0,0 +1,69 @@
...
+
+    const timestamps = Array.from(counts.keys())
+
+    const currentValues =
```
antooni, line 31: move to method inside this class, e.g. getDatabaseEntries

antooni: and also return the array of ActivityRecord

packages/backend/src/modules/activity2/indexers/ActivityIndexer.ts
```diff
@@ -0,0 +1,69 @@
...
+        [
+          new UnixTime(Math.min(...timestamps)),
+          new UnixTime(Math.max(...timestamps)),
+        ],
+      )
+    const currentMap = new Map(
+      currentValues.map((v) => [v.timestamp.toNumber(), v.count]),
+    )
+
+    const dataToSave = Array.from(counts).map(([timestamp, count]) => {
```
antooni, line 43: zip arrays, the only thing to decide is what to do when there is no record in the DB, probably in this case we should create a zero record

maciekop-l2b (approved):
LGTM

PR #4577, opened 2024-07-22
Add DACs config

vincfurc:
Resolves L2B-6184


radinat (approved):

torztomasz (changes requested):
I wouldn't say I like how this looks in terms of config structure for DACs. I will talk with @sdlyy about possible solutions to make it look better

torztomasz (approved):

PR #4578, opened 2024-07-22
Port Finality page

sdlyy:
Resolves L2B-5918
Resolves L2B-6471


packages/frontend2/src/server/features/scaling/finality/get-liveness-by-type-since.ts
```diff
@@ -0,0 +1,41 @@
...
+    .filter((c) => c.subtype === subtype)
+    .map((c) => c.id)
+
+  const records = await db.liveness.getByConfigurationIdSince(
+    configurationIds,
+    since,
+  )
+
+  function toRecordWithConfiguration(record: LivenessRecord) {
+    const config = configurations.find((c) => c.id === record.configurationId)
```
imxeno, line 28: Would it be worth to turn `configurations` into an object first to speed up this?

packages/frontend2/src/app/(new)/(other)/scaling/finality/_components/table/grayed-out.tsx
```diff
@@ -0,0 +1,15 @@
+import React from 'react'
+
+export function GrayedOut(props: {
+  children: React.ReactNode
+  grayOut: boolean
```
imxeno, line 5: Why `GrayedOut` component has `grayOut` field. I suppose it's copied from old FE, but anyway I suggest to change the naming or just remove this prop.

imxeno (commented):
Should the changes in boba/metal be there? If yes, how does that affect the site?
Also I guess it would be best to start writing tests for the functions in `frontend2/src/server`, but this can be done in a separate PR.

imxeno (approved):

PR #4598, opened 2024-07-23
Complete database migration

sz-piotr:
For some reason tests for SequenceProcessor that involve the database hang. Happened also before this PR. I don't want to add `exit: true` to mocha config, would rather resolve the issue somehow.


packages/database/src/aggregated-liveness/repository.test.ts
```diff
@@ -22,7 +19,7 @@ describeDatabase(AggregatedLivenessRepository.name, (database) => {
       min: 10,
       avg: 10,
       max: 10,
-      updatedAt: START.add(-1, 'hours'),
+      timestamp: START.add(-1, 'hours'),
```
maciekop-l2b, line 22: The correct name of this field should be `updatedAt`. Let's change it `AggregatedLivenessRecord`. I think it was overlooked before. Underlying db filed is called `updated_at` so it will be consistent.

maciekop-l2b (approved):
Great job!

imxeno (approved):

PR #4618, opened 2024-07-24
Add Starknet Activity Indexer

maciekzygmunt:
Resolves L2B-6510
PROD: 

LOCAL:


packages/backend/src/modules/activity2/services/TxsCountProvider.ts
```diff
@@ -128,6 +132,34 @@ export class TxsCountProvider {
...
+    assert(
+      this.$.projectConfig.type === 'starknet',
+      'Method not supported for projects other than Starknet',
+    )
+    const projectConfig = this.$.projectConfig
+
+    const starknetClient = this.$.peripherals.getClient(StarknetClient, {
+      url: projectConfig.url,
+      callsPerMinute: projectConfig.callsPerMinute,
+    })
```
sz-piotr, line 148: Not a fan of constructing clients in runtime rather than in setup time

maciekop-l2b: In this case we've done it deliberately. Providers (this and others) will encapsulate multi-client logic:
1. Client is not injected because provider decides which client(s) to use 
2. Clients are not constructed in runtime because usage will depend on specific case (lazy vs eager)

maciekop-l2b (approved):

PR #4625, opened 2024-07-25
Port scaling summary

imxeno:
This PR:
- changes the logic of fetching data in scaling summary, now it uses tRPC with a cookie prefetch
- replaces separate methods for tvl chart with a monolithic getScalingSummary to optimize DB queries
- makes /scaling/summary-next actually work
- ~~moves the filters to a react context for ease of use in both charts and tables~~ (extracted to #4694)
- ~~fixes the chart re-rendering issue by splitting the big chart context into more granular ones~~ (extracted to #4691)
- ~~introduces cookie prefetching utils~~ (extracted to #4687)

Resolves L2B-6115.


packages/frontend2/src/server/features/scaling/tvl/utils/get-scaling-summary-chart-data.ts
```diff
@@ -0,0 +1,86 @@
...
+export const getCachedScalingChartData = cache(
+  async ({
+    range,
+    excludeAssociatedTokens,
+    ...rest
+  }: CommonOptions & DataType) => {
+    const projectsFilter = (() => {
+      if (rest.type === 'layer2') {
+        return (project: TvlProject) =>
+          ['layer2', 'layer3'].includes(project.type)
```
torztomasz, line 42: I dont like this approach. Let's say project.type is `"layer2" | "layer3"` and we refactor to `"Layer2" | "Layer3"`. In this case `if (project.type === 'layer2' || project.type === 'layer3')` would cause typescript to error but `.includes` would not. Another thing is that I don't know how JS handles temporary arrays and if it gets cleaned well. Overall I prefer more ifs.

packages/frontend2/src/server/features/scaling/tvl/utils/get-scaling-summary-chart-data.ts
```diff
@@ -0,0 +1,86 @@
...
+      }
+    }
+
+    const chart = Object.entries(timestampValues).map(([timestamp, values]) => {
+      const summed = sumValuesPerSource(values, {
+        forTotal: true,
+        excludeAssociatedTokens: !!excludeAssociatedTokens,
+      })
+      return [
+        +timestamp,
```
torztomasz, line 70: we usually do `timestamp.toNumber()` and we rarely use `+` notation

packages/frontend2/src/hooks/use-cookie-state.ts
```diff
@@ -11,14 +16,40 @@ import { getCookie, setCookie } from '~/utils/cookies/client'
...
+      document.dispatchEvent(
+        new CustomEvent('cookie', {
+          detail: {
+            name,
+            value: state,
+          },
+        }),
+      )
+    },
+  ]
```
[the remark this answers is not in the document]
imxeno: It's done to sync state of this hook to all instances with the same params.

packages/frontend2/src/server/features/scaling/tvl/utils/get-scaling-summary-chart-data.ts
imxeno: it's a string, and we use + to cast strings to numbers in a lot of places, especially in the backend

torztomasz (approved):

PR #4626, opened 2024-07-25
Add split mode to powerdiff

mateuszradomski:
Resolves L2B-6420

Added `--mode=split` to powerdiff. Instead of doing diffs `file <-> file` it does diffs `TopLevelDecl <-> TopLevelDecl`.


sz-piotr (approved):
Dissapointed in the lack of any tests. How am I supposed to know this works correctly?

Maybe some e2e tests at least - 2 folders and test different options?

PR #4632, opened 2024-07-25
Add trackedTxs for kinto blobs

sekuba:
Closes L2B-6535
no milestone since kinto did not post about it anywhere yet :'(

### DIFF
--- a/packages/config/src/projects/layer2s/kinto.ts
+++ b/packages/config/src/projects/layer2s/kinto.ts
@@ -97,6 +97,49 @@ export const kinto: Layer2 = orbitStackL2({
     },
     {
       uses: [
+        { type: 'liveness', subtype: 'batchSubmissions' },
+        { type: 'l2costs', subtype: 'batchSubmissions' },
+      ],
+      query: {
+        formula: 'functionCall',
+        address: EthereumAddress('0xF4Ef823D57819AC7202a081A5B49376BD28E7b3a'),
+        selector: '0xe0bc9729',
+        functionSignature:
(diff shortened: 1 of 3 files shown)


packages/config/src/projects/layer2s/kinto.ts
```diff
@@ -105,7 +109,26 @@ export const kinto: Layer2 = orbitStackL2({
...
       },
     },
+    {
+      uses: [
+        { type: 'liveness', subtype: 'batchSubmissions' },
+        { type: 'l2costs', subtype: 'batchSubmissions' },
+      ],
+      query: {
+        // tracks both blobs and calldata (works for op stack, does it work here too?)
+        formula: 'transfer',
```
lucadonnoh, line 119: why not tracking addSequencerL2BatchFromBlobs? i dont understand

lucadonnoh (changes requested):

packages/config/src/projects/layer2s/kinto.ts
sekuba: i am trying out this simpler logic because we are successfully using it in the op stack template and i thought it might work here too? If it does i will add it to parallel too

lucadonnoh: i mean, we don't want to count random txs unrelated to that function if sent from that address. I know that it's unlikely, but maybe not? i don't see why we should add random assumptions. In OP stack is that way because they are literally transfers to EOA

lucadonnoh (approved):

PR #4637, opened 2024-07-25
[L2B-6513] Add new fields to activity aggregate

maciekop-l2b:
Part of L2B-6513

Screenshot 2024-07-25 at 15 55 02


packages/database-legacy/src/migrations/125_alter_activity_table.ts
```diff
@@ -0,0 +1,28 @@
...
+should create a new migration file that fixes the issue.
+
+*/
+
+import { Knex } from 'knex'
+
+export async function up(knex: Knex) {
+  await knex.schema.alterTable('activity', function (table) {
+    table.integer('start').notNullable().defaultTo(0)
+    table.integer('end').notNullable().defaultTo(0)
```
maciekzygmunt, line 19: I think we should get rid of this default

maciekzygmunt (changes requested):
One comment, otherwise LGTM

maciekzygmunt (approved):

packages/database-legacy/src/migrations/125_alter_activity_table.ts
maciekop-l2b: If there is already data in the table migration will throw so we can either clear the table or provide default. IMHO default is safer.

PR #4639, opened 2024-07-25
Ignore terms linking within backticks

sdlyy:
Resolves L2B-6428


imxeno (approved):
~~Will this work? Won't this break in this scenario?
```
Text `some code` [a link](https://example.com) `some code`.
```~~

Nvm, checked it. Good job!

PR #4651, opened 2024-07-26
Migrate discovery to cmd-ts

mateuszradomski:


packages/discovery/src/config/chains.ts
```diff
@@ -213,5 +216,5 @@ export const chains: ChainConfig[] = [
       type: 'etherscan',
       url: 'https://api.blastscan.io/api',
     },
-  },
+  } as const satisfies ChainConfig,
 ]
```
sz-piotr, line 220: satisfies ChainConfig[]

sz-piotr (changes requested):

packages/discovery/src/config/chains.ts
mateuszradomski: Done, with the caveat that I had to export a different variable that was casted to `ChainConfig[]` type because otherwise the `as const` at the bottom pruned the type too much resulting in optional fields being seen as not provided/non existent for this type.

sz-piotr (approved):

PR #4655, opened 2024-07-29
Move smoke test from general CI to merge queue run

sdlyy:
Resolves L2B-6586
In case of forks it will be skipped, in case of merge-queues (including merging PRs from forks) it will run as long as we have runner .env set

Required checks can be skipped (checked in separate repo):
image

## Rational behind two files for two different scopes:
- We want this check to be required
- We want this check to put PR back if merge group CI fails
- We do not want this check to run in the context of the PR
- We cannot specify this check to be required only in merge group (GitHub limitation) - also GH matches steps as required based on step name
- Since the check is marked as required, we have also to run it in the context of the PR to be able to merge the PR
- We are skipping this check in the context of the PR - not to waste CI time - it serves dummy purposes, to preserve required constraint without actually running as a part of the PR

## Mock repo dump - no ability to specify context
image


.github/workflows/smoke-test.yml
```diff
@@ -0,0 +1,87 @@
+name: Smoke Test
+
+on:
+  pull_request:
+  push:
+    branches:
+      - master
+      - main
+  merge_group:
```
imxeno, line 9: Should we still run this on pull request & push? Isn't merge group enough?

sdlyy: Yes, it will be sufficient. I did mark this as ready for review too early, I suppose, since I'm still thinking of other solutions. ;p I wanted to cut the time somebody has to wait before getting feedback on whether the environments are good or not by running it as part of the PR.

sdlyy: Yes, it will be sufficient. I did mark this as ready for review too early, I suppose, since I'm still thinking of other solutions. ;p I wanted to cut the time somebody has to wait before getting feedback on whether the environments are good or not by running it as part of the PR.

imxeno (approved):

PR #4657, opened 2024-07-29
Sharp verifier upgrade

sekuba:
Closes L2B-6571

The diff seems minimal (see diffHistory), althought they had to redeploy the whole verifier system.
The old contracts are all prefixed 'Level2' to make the manual diff easier. (see protocolbeat for an overview)
I did not descend into the tree deeper than the CPUFrilessVerifier contracts as diffing is very manual here.

forum post?
https://community.starknet.io/t/starknet-v0-13-2-pre-release-notes/114223
https://community.starknet.io/t/upcoming-feature-starknet-applicative-recursion/113868
https://community.starknet.io/t/starknet-v0-13-2-and-roadmap-summer-update/114221


mateuszradomski (approved):
From the configuration perspective this is correct. Instead of using `target` to select a template it could use the `shape` match to select a template but overall it's fine.

PR #4674, opened 2024-07-31
Add Eclipse (under review)

lucadonnoh:
Resolves L2B-6595

RPC cannot be trivially added because it's not EVM (nor compatible)


packages/config/src/projects/layer2s/eclipse.ts
```diff
@@ -1,26 +1,36 @@
...
   display: {
     name: 'Eclipse',
     slug: 'eclipse',
     description:
       'Eclipse is a Layer 2 powered by the Solana Virtual Machine (SVM).',
     purposes: ['Universal'],
-    category: 'Optimistic Rollup',
+    category: 'Optimium',
     links: {
       websites: ['https://eclipse.builders/'],
```
radinat, line 15: The website and app links are outdated

packages/config/src/projects/layer2s/eclipse.ts
```diff
@@ -1,26 +1,36 @@
...
 import { Layer2 } from './types'
 
-export const eclipse: Layer2 = upcomingL2({
+export const eclipse: Layer2 = underReviewL2({
   id: 'eclipse',
   display: {
     name: 'Eclipse',
     slug: 'eclipse',
     description:
       'Eclipse is a Layer 2 powered by the Solana Virtual Machine (SVM).',
```
radinat, line 11: I'm still not sure if we stick to our L2 definition according to the FAQ, so leaving a comment just in case we do

radinat (approved):

PR #4691, opened 2024-08-01
Fix chart re-render loop in FE2

imxeno:
Related to L2B-6115.


packages/frontend2/src/app/_components/chart/core/chart-loading-context.tsx
```diff
@@ -0,0 +1,31 @@
...
+
+export function useChartLoading() {
+  const context = useContext(ChartLoadingContext)
+  return context
+}
+
+export function useChartSetLoading() {
+  const context = useContext(ChartSetLoadingContext)
+  return context
+}
```
torztomasz, line 31: why can't we keep it in one context?

torztomasz (approved):

PR #4697, opened 2024-08-01
Port bridges risks

sdlyy:
Resolves L2B-5922


packages/frontend2/src/app/(new)/(other)/bridges/risk/page.tsx
```diff
@@ -0,0 +1,40 @@
...
+  },
+})
+
+export default async function Page() {
+  const [tvl, projectsVerificationStatuses, implementationChangeReport] =
+    await Promise.all([
+      getLatestTvlUsd({ type: 'bridge' }),
+      getProjectsVerificationStatuses(),
+      getImplementationChangeReport(),
+    ])
```
torztomasz, line 23: We need to decide which pattern to use:
* fetch everything from db in page and pass to get entries fn
* fetch everything from db in get entries and on page just call get entries fn

Personally I would opt for 2nd approach

torztomasz (changes requested):

packages/frontend2/src/app/(new)/(other)/bridges/layout.tsx
```diff
@@ -0,0 +1,15 @@
...
+  return (
+    <ContentWrapper className="mt-4 md:mt-12">
+      <div className="mb-20">
+        {children}
+        <OtherSites />
+        <About />
+      </div>
+    </ContentWrapper>
+  )
+}
```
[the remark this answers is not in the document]
sdlyy: https://github.com/l2beat/l2beat/pull/4697/commits/b5e5cbefc1391b06d82d6a18e5e191ec2ee95381

packages/frontend2/src/server/features/bridges/types.ts
```diff
@@ -0,0 +1,36 @@
+import { type ScalingProjectRiskViewEntry } from '@l2beat/config'
+import { type TokenBreakdownProps } from '~/app/_components/breakdown/token-breakdown'
+
+export type CommonBridgeEntry = {
```
imxeno, line 4: Don't we want to follow the same approach as I used in new pages, and just define entries as `ReturnType` of the functions that create them?

packages/frontend2/src/server/features/bridges/get-bridge-risk-entries.ts
```diff
@@ -0,0 +1,57 @@
...
+import { getDestination } from './get-destination'
+import { type BridgesRiskEntry } from './types'
+
+export async function getBridgeRiskEntries(): Promise<BridgesRiskEntry[]> {
+  const [tvl, projectsVerificationStatuses, implementationChangeReport] =
+    await Promise.all([
+      getLatestTvlUsd(),
+      getProjectsVerificationStatuses(),
+      getImplementationChangeReport(),
+    ])
```
imxeno, line 17: I guess this should be moved to page for this function to be easily testable.

packages/frontend2/src/app/(new)/(other)/bridges/risk/page.tsx
```diff
@@ -0,0 +1,26 @@
+import { BridgesMvpWarning } from '~/app/_components/bridges-mvp-warning'
+import { SimplePageHeader } from '~/app/_components/simple-page-header'
+import { getBridgeRiskEntries } from '~/server/features/bridges/get-bridge-risk-entries'
+import { getDefaultMetadata } from '~/utils/get-default-metadata'
+import { BridgesFilterContextProvider } from '../_components/bridges-filter-context'
+import { BridgesRiskTables } from './_components/table/bridges-risks-tables'
+
+export const metadata = getDefaultMetadata({
+  openGraph: {
+    url: '/bridges/summary',
```
imxeno, line 10: This should be `/bridges/risk`

imxeno (changes requested):

packages/frontend2/src/server/features/bridges/get-bridge-risk-entries.ts
torztomasz: I guess we need to talk about it then as I was the person that opted for moving it inside files 😅

packages/frontend2/src/server/features/bridges/types.ts
sdlyy: I'll take care of that as a part of the bridges summary page - I'm not entirely sure about removing these entry-types since those serve as a kind of contract between actual frontend components and data we have to provide.

packages/frontend2/src/app/(new)/(other)/bridges/risk/page.tsx
sdlyy: https://github.com/l2beat/l2beat/pull/4697/commits/9f8d7d9f05ca0efe749006483a3985590bdb745d

imxeno (approved):
LGTM

PR #4706, opened 2024-08-02
Port Costs page

torztomasz:


packages/frontend2/tsconfig.json
```diff
@@ -41,7 +41,8 @@
     "**/*.cjs",
     "**/*.mjs",
     "**/*.js",
-    ".next/types/**/*.ts"
+    ".next/types/**/*.ts",
+    "src/server/features/costs/utils/add-if-definedts"
```
sdlyy, line 45: lacks `.` before extension - any particular reason for explicit inclusion here?

sdlyy (approved):
small build-related thing, also underlines are missing as we spoke. Other than that - lgtm & gj

packages/frontend2/tsconfig.json
torztomasz: tbh, no clue what happened there but this must have added itself automatically :O

PR #4725, opened 2024-08-06
Fix DA Layer filter

torztomasz:
Resolves L2B-6682, L2B-6681


packages/frontend2/src/app/(new)/(other)/_components/base-scaling-filters.tsx
```diff
@@ -11,6 +11,7 @@ interface Props {
   items: CommonScalingEntry[]
   showRollupsOnly?: boolean
   additionalFiltersLeft?: React.ReactNode
+  // NOTE (torztomasz): I don't want to remove this as it will break other PR that @imxeno made. But I think it is not needed here and the behavior is wrong.
```
sdlyy, line 14: Just create a spike task for @imxeno to decide whether we should remove it or not

sdlyy (approved):

PR #4742, opened 2024-08-08
Speed up getting latest tvl

torztomasz:


packages/database/src/tvl/value/select.ts
```diff
@@ -19,3 +19,6 @@ export const selectValue = [
   'ether',
   'stablecoin',
 ] as const satisfies (keyof Value)[]
+
+export const selectValueWithPrefix = <T extends string>(prefix: T) =>
```
sdlyy, line 23: Is it used anywhere?

packages/database/src/tvl/value/repository.ts
```diff
@@ -46,6 +48,40 @@ export class ValueRepository extends BaseRepository {
...
+
+    const rows = await this.db
+      .with('latest_values', latestValuesQuery)
+      .selectFrom('latest_values')
+      .select(selectValue)
+      .where('combination_number', '=', 1)
+      .execute()
+    return rows.map(toRecord)
+  }
+
```
sdlyy, line 84: We have almost 1:1 duplicate down below for da-beat - Let's either unify these or split this apart removing the optional param (`projectIds?: ProjectId[]`)

packages/frontend2/src/app/(new)/(other)/scaling/data-availability/page.tsx
```diff
@@ -12,8 +11,7 @@ export const metadata = getDefaultMetadata({
 })
 
 export default async function Page() {
-  const tvl = await getLatestTvlUsd({ type: 'all' })
-  const items = await getScalingDaEntries(tvl)
+  const items = await getScalingDaEntries()
```
sdlyy, line 14: Nit: Let's unify this and go with `entries` as well

sdlyy (changes requested):

packages/database/src/tvl/value/select.ts
torztomasz: I was using it, forgot to remove it

packages/database/src/tvl/value/repository.ts
torztomasz: yup, i wanted to unify that

sdlyy (approved):

PR #4755, opened 2024-08-09
Refactor TVL status logic

antooni:
Resolves L2B-6619
Resolves L2B-6725


packages/backend/src/modules/tvl/api/services/data/DataStatusService.ts
```diff
@@ -0,0 +1,229 @@
...
+      const config = findConfig(indexer, entries)
+      assert(config, `Config should be defined for ${indexer.indexerId}`)
+      processed.add(config.configId)
+
+      const syncStatus = new UnixTime(indexer.safeHeight)
+      if (syncStatus.gte(targetTimestamp)) {
+        continue
+      }
+
+      // TODO: what about max height?
```
maciekop-l2b, line 120: Should it be resolved now? If not do we have a ticket fot this?

maciekop-l2b (approved):
LGTM

packages/backend/src/modules/tvl/api/services/data/DataStatusService.ts
antooni: I will create a ticket

PR #4758, opened 2024-08-12
Parallel updates

lucadonnoh:
as requested via telegram

### DIFF
--- a/packages/config/src/projects/layer2s/parallel.ts
+++ b/packages/config/src/projects/layer2s/parallel.ts
@@ -2,7 +2,6 @@ import { EthereumAddress, UnixTime } from '@l2beat/shared-pure'
 
 import { subtractOne } from '../../common/assessCount'
 import { ProjectDiscovery } from '../../discovery/ProjectDiscovery'
-import { Badge } from '../badges'
 import { orbitStackL2 } from './templates/orbitStack'
 import { Layer2 } from './types'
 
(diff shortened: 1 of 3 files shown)


packages/config/src/projects/layer2s/parallel.ts
```diff
@@ -2,15 +2,14 @@ import { EthereumAddress, UnixTime } from '@l2beat/shared-pure'
...
-import { Badge } from '../badges'
 import { orbitStackL2 } from './templates/orbitStack'
 import { Layer2 } from './types'
 
 const discovery = new ProjectDiscovery('parallel')
 
 export const parallel: Layer2 = orbitStackL2({
   discovery,
-  badges: [Badge.RaaS.Conduit],
+  badges: [],
```
radinat, line 12: Can we add the relevant badges instead of leaving it empty?

lucadonnoh: they're added via template. i noticed that the explorer link doesnt work so im looking for one rn

radinat: ah indeed, nvm then

radinat (approved):

packages/config/src/projects/layer2s/parallel.ts
lucadonnoh: but the template field should definitely be "additionalBadges" because like this it is very confusing indeed

PR #4780, opened 2024-08-13
Introduce `scheduleTick` pattern to Indexer

antooni:


maciekop-l2b:
We decided to take a different approach

PR #4787, opened 2024-08-14
[L2B-6761] Change init logic

maciekop-l2b:
Part of L2B-6761

### DIFF
--- a/packages/uif/src/Indexer.ts
+++ b/packages/uif/src/Indexer.ts
@@ -64,7 +64,7 @@ export abstract class Indexer {
    *    hash can be used e.g. to detect changes in the indexer configuration between
    *    restarts.
    *
-   *  - undefined if the initialization is not needed. Be mindful that in such case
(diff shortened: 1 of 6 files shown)


packages/uif/src/Indexer.ts
```diff
@@ -42,25 +42,34 @@ export abstract class Indexer {
...
+   *
+   *    For root indexers it should return the initial target height for the entire
+   *    system. If `setSafeHeight` is implemented it should return the height that
+   *    was saved previously. If not it can `return this.tick()`.
+   *
+   *  - Optionally it can return a config hash that will be saved in the state. Config
+   *    hash can be used e.g. to detect changes in the indexer configuration between
+   *    restarts.
+   *
+   *  - undefined if the initialization is not needed. Be mindful that in such case
```
maciekzygmunt, line 67:
```suggestion
   *  - `undefined` if the initialization is not needed. Be mindful that in such case
```

maciekzygmunt (approved):

PR #4803, opened 2024-08-16
Use layer's value secured across layer and bridge combinations

sdlyy:
Resolves L2B-6780


radinat (changes requested):
We should apply this logic to the detailed page as well to show the TVS by the layer regardless of the selected bridge 
image

sdlyy:
> We should apply this logic to the detailed page as well to show the TVS by the layer regardless of the selected bridge image

Be my guest

radinat (approved):

PR #4816, opened 2024-08-19
Preserve selected tab in search params

imxeno:


packages/frontend2/src/app/_components/tabs.tsx
```diff
@@ -1,10 +1,61 @@
...
+
+          const stringified = search.toString()
+
+          if (stringified) {
+            router.replace(`${pathname}?${stringified}`)
+          } else {
+            router.replace(pathname)
+          }
+        }
+      : undefined
```
torztomasz, line 46: I would say that we will also preserve filters etc in query params so lets create `useSearchParamsState` or smth that will look like that:
`const [param, setParam] = useSearchParamsState('tab', passedDefaultValue)`

torztomasz (changes requested):

packages/frontend2/src/app/_components/tabs.tsx
imxeno: done

torztomasz (approved):

PR #4839, opened 2024-08-21
Unify `assert` usage in the codebase

antooni:
Resolves L2B-6730

We have multiple assert import sources in the codebase:
```
@l2beat/shared-pure
@l2beat/backend-tool
assert
node:assert
node:console
```

This PR will unify the usage, so we will only use the one from `shared-pure`

Usage of the ones built into node is banned via linter rules


imxeno:
Since `frontend2` uses ESLInt please also copy the rules to frontend2's `.eslintrc.json`.

maciekop-l2b (approved):
LGTM

mateuszradomski (approved):

adamiak (approved):

PR #4845, opened 2024-08-21
Port Bridges project page

torztomasz:
TODO (bridges-project-page)


imxeno:
I'll think about how the project can be structured to avoid copy-pasting layouts between two directories, but anyway this can be merged.

PR #4903, opened 2024-08-27
Blobstream SP1 update

vincfurc:
Resolves L2B-6957


packages/config/src/projects/other/da-beat/blockchain/celestia/bridges/blobstream/SP1Blobstream/ethereum.ts
```diff
@@ -0,0 +1,99 @@
...
+      documentation: ['https://docs.celestia.org/developers/blobstream'],
+      repositories: ['https://github.com/succinctlabs/sp1-blobstream'],
+      apps: [],
+      explorers: ['https://etherscan.io/'],
+      socialMedia: [],
+    },
+  },
+  technology: `
+     The Blobstream bridge is a data availability bridge that facilitates data availability commitments to be bridged between Celestia and Ethereum.
+     The Blobstream bridge is composed of three main components: the **Blobstream** contract, the **Succinct Gateway** contracts and the **Verifier** contracts.  <br />
```
sekuba, line 38: 'between Celestia and Ethereum' should be 'between Celestia and the targetchain'
If the text is 100% the same in all three deployments you could also save it somewhere else and only change this variable targetchain

sekuba (changes requested):

packages/config/src/projects/other/da-beat/blockchain/celestia/bridges/blobstream/SP1Blobstream/ethereum.ts
vincfurc: Thanks! added variables, we can't fetch chain name and update interval by chain anywhere, but they still help

sekuba (approved):

PR #4917, opened 2024-08-28
Port DA Risk Framework slides page

imxeno:


packages/frontend2/src/app/da-risk/opengraph-image.jpg
sdlyy, line 1: most of those are PNGs, I just noticed and don't know whether it makes a difference for next

sdlyy (approved):

packages/frontend2/src/app/da-risk/opengraph-image.jpg
imxeno: For Next it doesn't make much difference (images are optimised anyway), but it affects our repo size. This file is just copy-pasted, but when exporting new stuff from Figma I make sure they are jpegs if possible (& pls to do the same).

PR #4920, opened 2024-08-28
Port multisig report page

imxeno:


packages/frontend2/src/app/(legacy)/multisig-report/_components/report-download-button.tsx
```diff
@@ -0,0 +1,26 @@
...
+  fileUrl: string
+  className?: string
+}
+
+export function ReportDownloadButton({
+  fileUrl,
+  className,
+}: ReportDownloadButtonProps) {
+  return (
+    <a
```
sdlyy, line 14: Shouldn't it be `<Outlink>`?

sdlyy (approved):

PR #4928, opened 2024-08-29
Add stories to new frontend

torztomasz:


packages/frontend2/src/app/_components/projects/sections/milestones-and-incidents-section.tsx
imxeno, line 1: The problem with this implementation is that the component lacks context about the next (older) milestone/incident.

packages/frontend2/src/app/_components/tabs.stories.tsx
```diff
@@ -0,0 +1,56 @@
...
+
+const meta = {
+  title: 'Atoms/Tabs',
+  component: Tabs,
+  args: {
+    storeInSearchParams: false,
+  },
+  parameters: {
+    nextjs: {
+      appDirectory: true, // 👈 Set this
```
imxeno, line 14:
```suggestion
      appDirectory: true,
```

imxeno (approved):

PR #4963, opened 2024-09-02
Add SP1 Blobstream to ZK catalog

vincfurc:
Resolves L2B-6974


lucadonnoh (changes requested):
add icon (SP1Blobstream.png, same as slug) and SP1Blobstream.md

also i would rename the ts file to SP1Blobstream for consistency

lucadonnoh (changes requested):

lucadonnoh:
note: it's not clear whether they use aggregation. I'm asking them

lucadonnoh (approved):

PR #4969, opened 2024-09-02
DA-BEAT DAC improvements 

sdlyy:
Resolves L2B-6727
Resolves L2B-7055
Maybe it will Resolve L2B-6932
And definitely will Resolve L2B-7084


sdlyy:
nova causes verification check to fail - I'll re-run it/change it in the meantime

PR #4970, opened 2024-09-03
Add MGP token

radinat:
Resolves L2B-5139


packages/config/src/tokens/generated.json
```diff
@@ -7257,14 +7257,14 @@
...
       "bridgedUsing": {
         "bridges": [
           {
             "name": "XLink bridge"
           }
         ]
-      },
-      "excludeFromTotal": true
+      }
     },
```
sekuba, line 7268: what happened here in generated.json with the other tokens?

sekuba (changes requested):

packages/config/src/tokens/generated.json
radinat: wow weird

sekuba (approved):

PR #4975, opened 2024-09-03
Fix NaNs on the chart

imxeno:


sdlyy (approved):
Cool, the only thing we might want to improve is date range fallback:
image
 up to you whether it should be done here or separately

imxeno:
> Cool, the only thing we might want to improve is date range fallback: image up to you whether it should be done here or separately

I guess it would be best to fix that, when we'll be redesigning the charts.

PR #4986, opened 2024-09-03
Add state validation section to Orbit chains

lucadonnoh:
Resolves L2B-6232


luke7211 (approved):

luke7211:
We should make state validation diagram part of the Orbit template

PR #4991, opened 2024-09-03
Unify indexer tagging and naming by separating it into config

sdlyy:
Resolves L2B-6994

Tried my best to keep these in sync, still not ideal


antooni (approved):
looks fine, please fix the tests

PR #4996, opened 2024-09-04
Add incidents

sekuba:


packages/config/src/projects/bridges/ronin.ts
```diff
@@ -245,4 +245,18 @@ export const ronin: Bridge = {
       description: 'Addresses that can unlock withdrawals.',
     },
   ],
+  milestones: [
+    {
+      name: 'Whitehat hack for $12M',
+      date: '2023-08-06T00:00:00Z',
```
lucadonnoh, line 251: same date?

lucadonnoh (changes requested):
review above comment

packages/config/src/projects/bridges/ronin.ts
sekuba: thanks, fixd

lucadonnoh (approved):

PR #4999, opened 2024-09-04
Add GPTProtocol (full)

sekuba:
Closes L2B-6387

not sure about EOA warn here since main contract is only upgradeable by polygon, but there exist plenty of routes to exploit as LocalAdmin


lucadonnoh (approved):
nice diagram hehe

sekuba:
that vincenzos :D

PR #5004, opened 2024-09-04
Richer descriptions on the frontend (ultimate permission owners)

adamiak:
Resolves L2B-6940, L2B-6941

- `target.description` in config.jsonc has been moved to `target.permission.*.description`
- Descriptions on the frontend now include direct permissions, ultimate permissions, "via"s, delays and permission descriptions
- If description is long, it's displayed as bullet points
- `Can be upgraded by` and `Upgrade delay` on the frontend (in contracts section) correctly show ultimate permission holders and real delay values

Also, *important*, `act` permission was added to `opstack/ProxyAdmin` template, causing a lot of changes in all OP stacks (because descriptions etc. are moved to the ultimate permission holder).


packages/config/src/discovery/ProjectDiscovery.ts
```diff
@@ -744,24 +744,29 @@ export class ProjectDiscovery {
           return (
             value.permission +
             ':' +
-            (value.via !== undefined ? formatVia(value.via) : '')
+            (value.via !== undefined ? formatVia(value.via) : '') +
+            ':' +
+            (value.description ?? '')
```
mateuszradomski, line 749:
```suggestion
[
    value.permission,
    (value.via !== undefined ? formatVia(value.via) : ''),
    (value.description ?? '')
].join(":")
```

packages/config/src/discovery/ProjectDiscovery.ts
```diff
@@ -744,24 +744,29 @@ export class ProjectDiscovery {
...
           )
         },
       ),
     ).map(([key, entries]) => {
       const permission = key.split(':')[0] as PermissionType
       const via = key.split(':')[1] ?? ''
+      const description = key.split(':', 3)[2] ?? ''
```
mateuszradomski, line 756: Wouldn't it be better to return an object above and just do `.description`?

packages/config/src/discovery/ProjectDiscovery.ts
```diff
@@ -772,19 +777,27 @@ export class ProjectDiscovery {
...
+      const description = key.split(':', 2)[1] ?? ''
       const addressesString = entries
         .map((entry) => this.getContract(entry.target.toString()).name)
         .join(', ')
-      return `${
-        directPermissionToPrefix[permission as PermissionType]
-      } ${addressesString}.`
+      return (
+        `${directPermissionToPrefix[permission]} ${addressesString}` +
+        (description !== '' ? ` - ${trimTrailingDots(description)}` : '') +
```
mateuszradomski, line 798: This pattern is repeated again, maybe move it to a function like:

```typescript
function intoEntry(input: string): string {
    return input !== '' ? ` - ${trimTrailingDots(input)}` : ''
}
```

packages/config/src/discovery/ProjectDiscovery.ts
```diff
@@ -965,10 +992,11 @@ const roleDescriptions: { [key in StackRole]: string } = {
...
-  // return description.length > 1
-  //   ? description.map((s) => `* ${s}\n`).join('')
-  //   : description.join(' ')
+  return description.length > 1
+    ? description.map((s) => `* ${s}\n`).join('')
+    : description.join(' ')
+}
+
+export function trimTrailingDots(s: string): string {
+  return s.replace(/\.*$/, '')
```
mateuszradomski, line 1001: I would feel more comfortable if this was a simple loop going backwards and creating a substring. LLM can easily rewrite it. Regex = Bad

packages/config/src/discovery/ProjectDiscovery.ts
```diff
@@ -894,19 +907,33 @@ export class ProjectDiscovery {
...
+                this.getContractByAddress(p.target)?.name ?? p.target.toString()
+              const delay =
+                (p.delay ?? 0) + sum(p.via?.map((v) => v.delay ?? 0) ?? [])
+              return [address, delay]
+            }) ?? [],
+        )
+
+        const ultimateUpgraders = Object.keys(upgradersWithDelay)
+        const minDelay = Math.min(...Object.values(upgradersWithDelay))
+        const upgradableByStruct =
```
mateuszradomski, line 924: Naming is weird, why not just `upgardeableBy`

packages/config/src/discovery/ProjectDiscovery.ts
adamiak: Can't do it, because the concatenated string is used in `groupBy` - it's the key and it can't be an object. This makes it really simple to group projects that share the same description, path and delay. It's not the prettiest code, but gets thing done in a few simple lines of code.

packages/config/src/discovery/ProjectDiscovery.ts
```diff
@@ -772,19 +777,27 @@ export class ProjectDiscovery {
...
-      groupBy(contractOrEoa.directlyReceivedPermissions ?? [], 'permission'),
-    ).map(([permission, entries]) => {
+      groupBy(
+        contractOrEoa.directlyReceivedPermissions ?? [],
+        (value: ResolvedPermission) =>
+          value.permission + ':' + (value.description ?? ''),
+      ),
+    ).map(([key, entries]) => {
+      const permission = key.split(':')[0] as PermissionType
+      const description = key.split(':', 2)[1] ?? ''
```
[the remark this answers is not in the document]
adamiak: As before, it needs to be a string for groupBy to work. I think these 2 lines of code are simple enough to not make it any more complex.

packages/config/src/discovery/ProjectDiscovery.ts
adamiak: I think that code clarity beats more complex implementation here (looping, slicing, ...). This is one of the simplest regexes possible, you just know it works. But looping, slicing, requires much more analysis (are indexes used for slicing correct, are the zero based, what if it's an empty string, etc. etc.) I prefer to keep it like this.

packages/config/src/test/projects.test.ts
```diff
@@ -76,8 +76,13 @@ describe('projects', () => {
           []) {
           const description = contract.description
           if (description) {
-            it(`contracts[${i}].description ends with a dot`, () => {
-              expect(description.endsWith('.')).toEqual(true)
+            it(`contracts[${i}].description - each line ends with a dot`, () => {
+              for (const descLine of description.trimEnd().split('\n')) {
+                if (!descLine.trimEnd().endsWith('.')) {
+                  expect(description).toEqual('asdf')
```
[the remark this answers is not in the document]
adamiak: :see_no_evil: good spot, I used it for debugging and forgot to remove it.

mateuszradomski (approved):

PR #5005, opened 2024-09-04
Add fraxtalDA config

vincfurc:
Resolves L2B-6182


lucadonnoh (approved):

lucadonnoh (changes requested):
probably you need to move the image from frontend2

lucadonnoh (approved):

PR #5013, opened 2024-09-05
Blobstream SP1 verifier update

vincfurc:
Resolves L2B-7102

New verifier version, created blobstream/succinctGateway template


sekuba (approved):

sekuba (changes requested):
wait maybe add the new one to the zk catalog?

sekuba (approved):

PR #5030, opened 2024-09-09
Add SP1Vector to ZK Catalog

lucadonnoh:
Resolves L2B-6745


vincfurc:
Changed config name to Vector from VectorX and ZKcatalog entry to SP1Vector. I think this is more accurate since:
- Vector is the name of the bridge, can have several implementations that should share the same discovery config
- VectorX is the name of the original Succinct implementation, but it is now deprecated, and
- the current implementation and verifier are based on SP1 and called SP1Vector.
also now consistent with Blobstream config and SP1Blobstream bridge/zk catalog entry

vincfurc (approved):

PR #5051, opened 2024-09-10
Add state validation to Optimism

lucadonnoh:
Resolves L2B-6231


packages/config/src/projects/layer2s/optimism.ts
```diff
@@ -436,11 +482,50 @@ export const optimism: Layer2 = {
...
+        Actors might be involved in multiple (sub-)challenges at the same time, meaning that the protocol operates with [full concurrency](https://medium.com/l2beat/fraud-proof-wars-b0cb4d0f452a). Each claim has a clock that starts with ${formatSeconds(
+          maxClockDuration,
+        )}, and that gets passed down to grandchildren claims. If a clock expires, the claim is considered defeated if it was countered, or it gets confirmed if uncountered. Since honest parties can inherit clocks from malicious claims (therefore from malicious parties, see [freeloader claims](https://specs.optimism.io/fault-proof/stage-one/fault-dispute-game.html#freeloader-claims)), if a clock has less than ${formatSeconds(
+          permissionlessGameClockExtension,
+        )}, it generally gets extended by ${formatSeconds(
+          permissionlessGameClockExtension,
+        )} with the exception of ${formatSeconds(
+          permissionlessGameClockExtension * 2,
+        )} at depth ${permissionlessGameSplitDepth}, and ${formatSeconds(
```
sekuba, line 514: Should this be 'at depth **>** 30'? I am assuming the clock extension would not exclusively work at depth 30?

packages/config/src/projects/layer2s/optimism.ts
```diff
@@ -436,11 +482,50 @@ export const optimism: Layer2 = {
...
+          oracleChallengePeriod,
+        )} at the last depth. The maximum clock extension that a top level claim can get is therefore ${formatSeconds(
+          permissionlessGameMaxClockExtension,
+        )}. Since unconfirmed state roots are independent with one another, users can decide to exit with a subsequent state root if the previous one is delayed. Winners get the entire losers' stake, meaning that sybils can potentially play against each other at no cost. The final instruction found via the bisection game is then executed onchain in the MIPS one step prover contract who determines the winner. The protocol does not enforces valid bisections, meaning that actors can propose correct initial claims and then provide incorrect midpoints. The protocol can be subject to resource exhaustion attacks ([Spearbit 5.1.3](https://github.com/ethereum-optimism/optimism/blob/develop/docs/security-reviews/2024_08_report-cb-fault-proofs-non-mips.pdf)).`,
```
sekuba, line 518: 'are independent **of** one another' (typo)

sekuba: 'protocol does not **enforce** valid bisections' (typo)

packages/config/src/projects/layer2s/optimism.ts
lucadonnoh: i thought that too but it's not the case. their spec is a bit confusing and they tweaked the math in the latest upgrade. pls have a look at the FaultDisputeGame contract if you want to confirm

packages/config/src/projects/layer2s/optimism.ts
lucadonnoh:
```suggestion
        )}. Since unconfirmed state roots are independent of one another, users can decide to exit with a subsequent state root if the previous one is delayed. Winners get the entire losers' stake, meaning that sybils can potentially play against each other at no cost. The final instruction found via the bisection game is then executed onchain in the MIPS one step prover contract who determines the winner. The protocol does not enforce valid bisections, meaning that actors can propose correct initial claims and then provide incorrect midpoints. The protocol can be subject to resource exhaustion attacks ([Spearbit 5.1.3](https://github.com/ethereum-optimism/optimism/blob/develop/docs/security-reviews/2024_08_report-cb-fault-proofs-non-mips.pdf)).`,
```

packages/config/src/projects/layer2s/optimism.ts
```diff
@@ -436,11 +482,50 @@ export const optimism: Layer2 = {
...
+          permissionlessGameClockExtension,
+        )} with the exception of ${formatSeconds(
+          permissionlessGameClockExtension * 2,
+        )} at depth ${permissionlessGameSplitDepth}, and ${formatSeconds(
+          oracleChallengePeriod,
+        )} at the last depth. The maximum clock extension that a top level claim can get is therefore ${formatSeconds(
+          permissionlessGameMaxClockExtension,
+        )}. Since unconfirmed state roots are independent of one another, users can decide to exit with a subsequent state root if the previous one is delayed. Winners get the entire losers' stake, meaning that sybils can potentially play against each other at no cost. The final instruction found via the bisection game is then executed onchain in the MIPS one step prover contract who determines the winner. The protocol does not enforce valid bisections, meaning that actors can propose correct initial claims and then provide incorrect midpoints. The protocol can be subject to resource exhaustion attacks ([Spearbit 5.1.3](https://github.com/ethereum-optimism/optimism/blob/develop/docs/security-reviews/2024_08_report-cb-fault-proofs-non-mips.pdf)).`,
```
sekuba, line 518: > Since unconfirmed state roots are independent of one another, users can decide to exit with a subsequent state root if the previous one is delayed.

'users can decide to exit with a subsequent **confirmed** state root'?

lucadonnoh:
```suggestion
        )}. Since unconfirmed state roots are independent of one another, users can decide to exit with a subsequent confirmed state root if the previous one is delayed. Winners get the entire losers' stake, meaning that sybils can potentially play against each other at no cost. The final instruction found via the bisection game is then executed onchain in the MIPS one step prover contract who determines the winner. The protocol does not enforce valid bisections, meaning that actors can propose correct initial claims and then provide incorrect midpoints. The protocol can be subject to resource exhaustion attacks ([Spearbit 5.1.3](https://github.com/ethereum-optimism/optimism/blob/develop/docs/security-reviews/2024_08_report-cb-fault-proofs-non-mips.pdf)).`,
```

packages/config/src/projects/layer2s/optimism.ts
```diff
@@ -436,11 +482,50 @@ export const optimism: Layer2 = {
...
+          formatEther(permissionlessGameFullCost),
+        ).toFixed(2)} ETH from depth 0. Actors can participate in any challenge by calling the \`defend\` or \`attack\` functions, depending whether they agree or disagree with the claim and want to move the bisection game forward.
+        Actors might be involved in multiple (sub-)challenges at the same time, meaning that the protocol operates with [full concurrency](https://medium.com/l2beat/fraud-proof-wars-b0cb4d0f452a). Each claim has a clock that starts with ${formatSeconds(
+          maxClockDuration,
+        )}, and that gets passed down to grandchildren claims. If a clock expires, the claim is considered defeated if it was countered, or it gets confirmed if uncountered. Since honest parties can inherit clocks from malicious claims (therefore from malicious parties, see [freeloader claims](https://specs.optimism.io/fault-proof/stage-one/fault-dispute-game.html#freeloader-claims)), if a clock has less than ${formatSeconds(
```
sekuba, line 508: >if a clock has less than

-->'if a clock gets inherited with less than'

lucadonnoh:
```suggestion
        )}, and that gets passed down to grandchildren claims. If a clock expires, the claim is considered defeated if it was countered, or it gets confirmed if uncountered. Since honest parties can inherit clocks from malicious claims (therefore from malicious parties, see [freeloader claims](https://specs.optimism.io/fault-proof/stage-one/fault-dispute-game.html#freeloader-claims)), if a clock gets inherited with less than ${formatSeconds(
```

packages/config/src/projects/layer2s/optimism.ts
sekuba: wow ok this is complicated. I read the contract and because of this part:
```
if (nextPositionDepth == MAX_GAME_DEPTH - 1) {
            // If the next position is `MAX_GAME_DEPTH - 1` then we're about to execute a step. Our
            // clock extension must therefore account for the LPP challenge period in addition to
            // the standard clock extension.
            actualExtension = CLOCK_EXTENSION.raw() + uint64(VM.oracle().challengePeriod());
        } else if (nextPositionDepth == SPLIT_DEPTH - 1) {
            // If the next position is `SPLIT_DEPTH - 1` then we're about to begin an execution
            // trace bisection and we need to give extra time for the off-chain challenge agent to
            // be able to generate the initial instruction trace on the native FPVM.
            actualExtension = CLOCK_EXTENSION.raw() * 2;
            ```

i think the two depth vars in the text should be both reduced by 1: `at depth ${permissionlessGameSplitDepth-1}` and `at the last-1 depth`

luke7211 (approved):

PR #5062, opened 2024-09-10
Add RaaS filter

sdlyy:
Resolves L2B-6833


torztomasz (approved):
I think we can merge it but after merging we need to handle empty chart state. Look at chart when you select Gateway

sdlyy:
❗ DNM - Merge along with L2B-6909

PR #5082, opened 2024-09-12
NProgress integration

imxeno:


sdlyy (approved):
nice, just a color - it seems odd to me, maybe we should ask around and seek alternatives

PR #5101, opened 2024-09-13
Fix prefetches and stale time

torztomasz:
Resolves L2B-7258


packages/frontend2/src/trpc/query-client.ts
```diff
@@ -7,7 +7,7 @@ export const createQueryClient = () =>
       queries: {
         // With SSR, we usually want to set some default staleTime
         // above 0 to avoid refetching immediately on the client
-        staleTime: 30 * 1000,
+        staleTime: 10 * 60 * 1000,
```
imxeno, line 10: Consider using UnixTime units

imxeno (approved):

sdlyy (approved):

PR #5109, opened 2024-09-15
Trivial discovery updates 09/15

sekuba:
Closes L2B-7251, L2B-7250, L2B-7248, L2B-7247


lucadonnoh (changes requested):
see my draft pr. the guardian module template is not recognized anymore and the new implementation needs to be added as a shape. also module contracts should be added in the contract section.

lucadonnoh (approved):

PR #5120, opened 2024-09-16
Add Polynomial (full)

lucadonnoh:
Resolves L2B-7268


packages/config/src/projects/layer2s/polynomial.ts
```diff
@@ -0,0 +1,55 @@
...
+        'https://x.com/PolynomialFi',
+        'https://discord.gg/Mr9XKU5W',
+      ],
+    },
+    activityDataSource: 'Blockchain RPC',
+  },
+  rpcUrl: 'https://rpc.polynomial.fi',
+  finality: {
+    type: 'OPStack-blob',
+    genesisTimestamp: new UnixTime(1718038307),
```
sekuba, line 36: with their rpc i got 1718038177 for block no. 1 but probably does not matter?

sekuba (changes requested):
Proposer, Challenger, Guardian do not show up in perms (and the proposer is not ConduitMS like the other two)

sekuba (approved):

PR #5140, opened 2024-09-17
Add MST token

radinat:
Resolves L2B-7234

It's not ERC20, to be checked if we can fetch the data and researched

closes L2B-7234


sekuba:
uses the [ERC-404 standard](https://github.com/Pandora-Labs-Org/erc404/blob/main/contracts/ERC404.sol) which combines erc 20 and 721 standard and seems to be in beta. Should be fine especially if we are using CG data here

sekuba (approved):

PR #5162, opened 2024-09-19
Fix missing provider consideration in type sort sorting logic

hyperchris123:
Currently, the type sort function only sorts by category, leading to inconsistent ordering when provider is not taken into account. 
This fix introduces sorting by both category and provider to ensure the correct order and eliminate sorting issues.

### Before
image

### After
image

### DIFF
--- a/packages/frontend/src/app/(side-nav)/scaling/summary/_components/table/columns.tsx
+++ b/packages/frontend/src/app/(side-nav)/scaling/summary/_components/table/columns.tsx
@@ -41,6 +41,17 @@ export const scalingLayer2sColumns = [
     meta: {
       tooltip: <TypeExplanationTooltip />,
     },
+    sortingFn: (rowA, rowB) => {
+      // Sort by category first, then by provider
+      const categoryComparison = (rowA.original.category ?? '').localeCompare(
+        rowB.original.category ?? '',
+      )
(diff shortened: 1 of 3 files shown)


packages/frontend/src/app/(side-nav)/scaling/summary/_components/table/columns.tsx
```diff
@@ -41,6 +41,17 @@ export const scalingLayer2sColumns = [
...
+      // Sort by category first, then by provider
+      const categoryComparison = (rowA.original.category ?? '').localeCompare(
+        rowB.original.category ?? '',
+      )
+      return categoryComparison !== 0
+        ? categoryComparison
+        : (rowA.original.provider ?? '').localeCompare(
+            rowB.original.provider ?? '',
+          )
+    },
```
torztomasz, line 54: let's extract the fn to separate file called `order-by-category-and-provider` or smth

torztomasz (changes requested):
image

Projects without providers should also be grouped

lucadonnoh:
my opinion is that projects should be sorted by column order first and then tvl

lucadonnoh:
imo not needed, @torztomasz feel free to reopen if you think otherwise

PR #5163, opened 2024-09-19
Improve the galxegravity information

Peter6501:


sekuba:
heyo, pls group low effort PRs in the future so they become medium effort!
Also in this specific one i am closing because the SAFE frontend is kinda random/unrelated and unclear from what source code.

PR #5185, opened 2024-09-21
Fix SVG fill issue for NavCollapseIcon

hyperchris123:
Resolve the issue where the triangle in the NavCollapseIcon appears solid due to an unintended fill color. Update the SVG component to explicitly set fill="none" to prevent the shape from being filled and ensure the icon renders as expected.

### Before vs After

<div style="display: flex; ">
image
image
</div>


torztomasz:
Great catch, although we have removed collapsing of the side bar.

PR #5188, opened 2024-09-22
fix: links in scroll L2 page

colinlyguo:
This PR:
* fixes outdated links to contract repo.
* adds some opinionated changes (may influence the UX, feel free to change them).

btw, because I could not change this graph, it's also nice to update the "A diagram of the smart contract architecture" graph with the following changes (based on the rules, maybe raising an issue for this change request is better?):
* add `finalizeBundle` in the link points to the `ScrollChain` contract.
* add `verifyBundleProof` in the link points to verifier contracts.

### DIFF
--- a/packages/config/src/projects/layer2s/scroll.ts
+++ b/packages/config/src/projects/layer2s/scroll.ts
@@ -385,7 +385,7 @@ export const scroll: Layer2 = {
       ...STATE_CORRECTNESS.VALIDITY_PROOFS,
       references: [
         {
-          text: 'ScrollChain.sol - Etherscan source code, verifyAggregateProof4844() and verifyBundleProof() calls',
+          text: 'ScrollChain.sol - Etherscan source code, verifyAggregateProof() and verifyBundleProof() calls',
           href: 'https://etherscan.io/address/0x9bB163401E8C72573854c4Cd968aFA7A7b02D25f#code',
         },
       ],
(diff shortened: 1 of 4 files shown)


packages/config/src/projects/layer2s/scroll.ts
```diff
@@ -407,9 +412,13 @@ export const scroll: Layer2 = {
       ...FORCE_TRANSACTIONS.SEQUENCER_NO_MECHANISM,
       references: [
         {
-          text: 'EnforcedTxGateway.sol - Etherscan source code, EnforcedTxGateway is paused',
+          text: 'EnforcedTxGateway.sol - Etherscan source code',
           href: 'https://etherscan.io/address/0x642af405bF64660665B37977449C9C536B806318#code',
         },
+        {
+          text: 'EnforcedTxGateway is paused - Etherscan proxy contract',
```
colinlyguo, line 419: this is for clarity, because when opening `https://etherscan.io/address/0x642af405bF64660665B37977449C9C536B806318#code` to check the "paused" value, it shows "false" because the states are all stored in the proxy contract. there are dummy values in the implementation contract.

packages/config/src/projects/layer2s/scroll.ts
```diff
@@ -173,6 +173,11 @@ export const scroll: Layer2 = {
         tokens: ['USDC'],
         ...upgradesScrollMultisig,
       }),
+      discovery.getEscrowDetails({
```
colinlyguo, line 176: sorry just found during testing, that adding more gateways might do the same steps as this PR: https://github.com/l2beat/l2beat/pull/3754

colinlyguo:
@sekuba pls take a look.

sekuba (approved):

PR #5194, opened 2024-09-23
Remove project RedSonic

benjamincyberb:
Remove project RedSonic


sekuba:
We do not remove projects but put them in the archive (see aztec) in case they are discontinued

PR #5195, opened 2024-09-23
Blobstream verifier update

vincfurc:


lucadonnoh (changes requested):
zk catalog entries need to be updated

lucadonnoh (approved):
lgtm

PR #5205, opened 2024-09-24
Move user operations dashboard to main repo

maciekop-l2b:


torztomasz (approved):

torztomasz:
There is a lot of files coming from Next.js template. I would consider deleting them but not needed

PR #5230, opened 2024-09-26
Asset risks token fetching

imxeno:
Resolves L2B-7445.


packages/frontend/src/app/asset-risks/[address]/_components/clientside-logic.tsx
```diff
@@ -0,0 +1,26 @@
...
+  const [refetched, setRefetched] = useState(false)
+
+  useEffect(() => {
+    if (refetched) return
+    setRefetched(true)
+    void (async () => {
+      await refreshTokens.mutateAsync({ address })
+      await report.refetch()
+      await refreshBalances.mutateAsync({ address })
+      await report.refetch()
```
Comment (sdlyy) [inline packages/frontend/src/app/asset-risks/[address]/_components/clientside-logic.tsx:21]: Why do we refetch the report twice here?

packages/frontend/src/server/features/asset-risks/refresh-tokens-of-address.ts
```diff
@@ -0,0 +1,160 @@
...
+}
+
+async function getAllLogsInner(
+  client: PublicClient,
+  address: Address,
+  fromBlock: bigint,
+  toBlock: bigint,
+  mode?: 'from' | 'to',
+) {
+  console.log('Getting logs from', fromBlock, 'to', toBlock, 'mode', mode)
```
sdlyy, line 116: clog

sdlyy (approved):
nits only, lg

packages/frontend/src/app/asset-risks/[address]/_components/clientside-logic.tsx
Comment (imxeno) [reply packages/frontend/src/app/asset-risks/[address]/_components/clientside-logic.tsx:21]: Because these are two separate steps, ideally in the UI we'd display a list of tokens before we display the balances, it's not coded yet though (and it's good since we're changing the design)

PR #5242, opened 2024-09-26
Add Hemi to upcoming chains

jcvernaleo:


sekuba (approved):

sekuba:
heyo, please in the future allow contributors to contribute to your PR otherwise we cannot make quick fixes before approving ;)

jcvernaleo:
@sekuba really sorry about missing the `allow contributors edit` option.  I totally missed it somehow.  I'll make sure to catch it in the future.

jcvernaleo:
Okay, just want to fill in, but it appears that github does not offer that option when the fork the PR comes from is under an organization account.  It only appears for forks of a repo in an individual account.  Seems like a strange and somewhat annoying limitation but I'll be sure to fork under my own account to avoid the issue next time.

PR #5252, opened 2024-09-29
Add World Chain badges

QuantumFatima:


lucadonnoh:
we dont add badges to upcoming projects

PR #5256, opened 2024-09-30
[L2B-7355] New execute method + minor fixes

maciekop-l2b:


packages/uops-dashboard/src/components/transactionDetails.tsx
```diff
@@ -28,7 +28,10 @@ export function TransactionDetails({
       <>
         <p
           key={op.id}
-          className={`${padding}mt-0 text-gray-500 dark:text-gray-400`}
+          className={`${
+            padding
+            // biome-ignore lint/nursery/useSortedClasses: space in front unnecessarily triggers the rule
+          } mt-0 text-gray-500 dark:text-gray-400`}
```
maciekzygmunt, line 34: We can consider installing `clsx` and use it there

PR #5258, opened 2024-09-30
Make naming consistent in token-db

imxeno:


imxeno:
> why all of them have `js` in import?

it's an esmodule and we're not using a bundler

sdlyy (approved):

PR #5259, opened 2024-09-30
Basic frontend pool config

sdlyy:
10 is the default for max connection pool size - we were keeping up just fine with it so we can explicitly stick to it.
The only thing that changed is the local pool size - also identification will now be way easier through app ids so we can go back to it later on to see how connection distribution is doing.

image

### DIFF
--- a/packages/frontend/src/server/database.ts
+++ b/packages/frontend/src/server/database.ts
@@ -21,7 +21,8 @@ function createThrowingProxy() {
 }
 
 function createConnectionTag() {
-  const base = `l2beat-frontend-${env.NODE_ENV}`
+  // Tag is limited to 63 characters, so it will cut off the excess
+  const base = `FE-${toShort()}`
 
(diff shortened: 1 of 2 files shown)


packages/frontend/src/server/database.ts
```diff
@@ -21,3 +19,30 @@ function createThrowingProxy() {
     },
   })
 }
+
+function createConnectionTag() {
+  const base = 'l2beat-frontend'
+
+  return `${base}-${env.NODE_ENV}`
```
imxeno, line 26: isn't it better to use vercel's preview id somehow?

sdlyy: https://github.com/l2beat/l2beat/pull/5259/commits/2c090f79bbf1c70814c841abeadee0d2e8e99759

imxeno (approved):

PR #5280, opened 2024-10-01
Add ability to hide some projects' TVL

imxeno:


sdlyy (approved):
Briefly went through and tested it by excluding Arbitrum - chart goes down, no errors are being thrown - looks good
image

image

antooni (approved):

PR #5282, opened 2024-10-01
[L2B-7075] Update finality analyzers for starknet and paradex

maciekop-l2b:


packages/backend/src/modules/finality/analyzers/StarknetFinalityAnalyzer.ts
```diff
@@ -52,14 +52,14 @@ export class StarknetFinalityAnalyzer extends BaseAnalyzer {
 }
 
 function decodeTransaction(data: string) {
-  const signature = `updateStateKzgDA(uint256[] programOutput, bytes kzgProof)`
+  const signature = `updateStateKzgDA(uint256[] programOutput, bytes[] kzgProofs)`
```
antooni, line 55: this kzgProofs is not used at all, lets remove this field from parsing and logic

packages/config/src/projects/layer2s/paradex.ts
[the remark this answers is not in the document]
maciekop-l2b: it is updated already :)

packages/backend/src/modules/finality/analyzers/StarknetFinalityAnalyzer.ts
maciekop-l2b: Removed

antooni (approved):

PR #5287, opened 2024-10-02
Migrate Backend's CD pipeline to GH Actions

kondziorex7:


maciekop-l2b (changes requested):
@kondziorex7  this is not how it supposed to be implemented. We have deliberately decoupled database migrations from app deployments so it SHOULD NOT be executed on app deployment. Please leave db workflow as it is and create separate flow for BE apps

maciekop-l2b (approved):
LGTM

PR #5312, opened 2024-10-03
Update EigenDA permissions 

vincfurc:
some risks are better suited to the technology section but the sc section is the only place for them rn


lucadonnoh (approved):

lucadonnoh (approved):
can you also add upgradableBy in contracts?

PR #5335, opened 2024-10-07
Test templatization: PGN (`discoveryDrivenData: true`)

sekuba:


lucadonnoh (changes requested):
image

I suggest to rename this to PgnMultisig, but also, why is the description so shallow 😅 should be something like "It controls Guardian and Challenger roles" or similar. Outside the scope of this PR but maybe we can signal it to @adamiak

sekuba:
looking at zora it seems as if we haven't got longer descriptions defined for roles yet 
image

lucadonnoh (approved):

PR #5343, opened 2024-10-08
Create "interpretation layer" in token-db

imxeno:


packages/database/src/token-db/token-meta/repository.test.ts
```diff
@@ -0,0 +1,266 @@
...
+    tokenId,
+    source,
+    externalId,
+    name,
+    symbol,
+    decimals,
+    logoUrl,
+    contractName,
+  }
+}
```
sdlyy, line 251: Wrap this down to `record: UpsertableTokenMetaRecord -> return record`

packages/token-db/src/queues/token-meta-aggregator.ts
```diff
@@ -0,0 +1,53 @@
...
+
+  const tokenMetaAggregator = queueWithProcessor<TokenPayload>({
+    name: 'TokenMetaAggregator.Processor',
+    processor: (job) =>
+      buildTokenMetaAggregatorSource({
+        logger,
+        db,
+        queue: tokenMetaAggregatorQueue,
+      })(job.data.tokenId),
+  })
```
sdlyy, line 38:
```suggestion
  const tokenMetaAggregator = queueWithProcessor<TokenPayload>({
    name: 'TokenMetaAggregator.Processor',
    processor: (job) =>
      buildTokenMetaAggregatorSource({
        logger,
        db,
      })(job.data.tokenId),
  })
```

I'll test it tomorrow but:
- A separate inbox is not needed in this case
- separate inbox is not attached nor routed to any source inbox
- queueWithProcessor result (`.queue`) should be returned as a source inbox (below)
- metadata aggregator does not need to propagate events any further (at least for now) - it simply ingests events once independent sources provided basic information (and some meta) about the token

packages/token-db/src/sources/token-meta-aggregator.ts
```diff
@@ -0,0 +1,65 @@
+import { Logger } from '@l2beat/backend-tools'
+import { Database } from '@l2beat/database'
+import { TokenUpdateQueue } from '../utils/queue/wrap.js'
+
+type Dependencies = {
+  logger: Logger
+  db: Database
+  queue: TokenUpdateQueue
```
sdlyy, line 8:
```suggestion
```

not needed, evens are not propagated further, remove it down the road

packages/token-db/src/queues/token-meta-aggregator.ts
```diff
@@ -0,0 +1,53 @@
...
+    statusLogger.info('Starting')
+
+    tokenMetaAggregator.worker.run()
+
+    statusLogger.info('Started')
+  }
+
+  return {
+    start,
+    inbox: tokenMetaAggregatorInbox,
```
sdlyy, line 51:
```suggestion
    inbox: tokenMetaAggregator.queue,
```

sdlyy (changes requested):
Just a clean-up thing - additional inbox is not needed
Re-run some basic sources to make sure everything works as intended since in the current setup:
```
coingecko-source -> tokenUpdateQueue --- broadcast --> [onChain, deployment, meta aggregator.inbox]
```
Messages in the inbox had no worker assigned and thus were not processed

sdlyy (approved):

PR #5349, opened 2024-10-08
[L2B-7322] Implement custom tracked txs logic for zkstack chains

maciekop-l2b:


packages/backend/src/test/sharedBridge.ts
```diff
@@ -0,0 +1,25 @@
+export const sharedBridgeChainId = 324
+
+export const sharedBridgeCommitBatchesSelector = '0x6edd4f12'
+export const sharedBridgeCommitBatchesSignature =
+  'function commitBatchesSharedBridge(uint256 _chainId, (uint64 batchNumber, bytes32 batchHash, uint64 indexRepeatedStorageChanges, uint256 numberOfLayer1Txs, bytes32 priorityOperationsHash, bytes32 l2LogsTreeRoot, uint256 timestamp, bytes32 commitment) _lastCommittedBatchData, (uint64 batchNumber, uint64 timestamp, uint64 indexRepeatedStorageChanges, bytes32 newStateRoot, uint256 numberOfLayer1Txs, bytes32 priorityOperationsHash, bytes32 bootloaderHeapInitialContentsHash, bytes32 eventsQueueStateHash, bytes systemLogs, bytes pubdataCommitments)[] _newBatchesData)'
+// input has been taken from a real tx on ethereum
+// 0xa4f216973754831a5760bb1cd7c78ad4c7c3cf7403c59dcd18aaa0ebf9e217b2
+export const sharedBridgeCommitBatchesInput =
```
maciekzygmunt, line 8: Try to not paste raw data here. Rather try to encode it on your own. (See: LineaFinalityAnalyzer.test.ts Line: 95)

maciekzygmunt (changes requested):
Left 2 comments. Rest looks good

packages/config/src/projects/layer2s/types/Layer2TrackedTxsConfig.ts
```diff
@@ -47,3 +47,15 @@ interface SharpSubmission {
   /** Inclusive */
   untilTimestamp?: UnixTime
 }
+
+interface SharedBridge {
+  formula: 'sharedBridge'
```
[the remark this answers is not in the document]
maciekop-l2b: Hm... I actually wanted it to be generic. As lon as we can extract `chainId` from calldata we can use the same formula (It does not use static function signature as `sharpSubmission` is doing but it takes it from actual tracked tx definition )

packages/backend/src/test/sharedBridge.ts
maciekop-l2b: I saw that across our codebase we use two approaches:
1. Static literal data
2. Generated by encoding
I'm leaning towards the first one because then we are testing against real on-chain data. Otherwise we are just proving that we can decode data that we encoded ourselves. 

In this case we would also have to generate a lot of data which is out of scope of the test itself (we use just a single param)

maciekzygmunt (approved):

PR #5350, opened 2024-10-08
Implement new RPC client

antooni:
- [x] logger
- [x] throttling
- [x] error handling for erorrs in body

Resolves L2B-7550


packages/shared/src/clients/rpc/RpcClient2.test.ts
```diff
@@ -0,0 +1,123 @@
+import { Logger } from '@l2beat/backend-tools'
+import { expect, mockObject } from 'earl'
+import { HttpClient2 } from '../http/HttpClient2'
+import { RpcClient2 } from './RpcClient2'
+
+describe(RpcClient2.name, () => {
+  describe(RpcClient2.prototype.getBlock.name, () => {
+    it('fetches block from rpc are parsers response', async () => {
```
maciekop-l2b, line 8:
```suggestion
    it('fetches block from rpc and parses response', async () => {
```

maciekop-l2b (approved):
LGTM but let's add more logging as we discussed

packages/shared/src/clients/rpc/RpcClient2.ts
```diff
@@ -0,0 +1,76 @@
...
+    const method = 'eth_getBlockByNumber'
+    const encodedNumber =
+      blockNumber === 'latest' ? 'latest' : `0x${blockNumber.toString(16)}`
+    const blockResponse = await this.query(method, [encodedNumber, false])
+
+    const block = Block.safeParse(blockResponse)
+    if (!block.success) {
+      this.$.logger.error(JSON.stringify(blockResponse))
+      throw new Error('Error during parsing of eth_getBlockByNumber response')
+    }
```
sz-piotr, line 35: This architecture of retries being implemented on HttpClient2 makes us unable to retry in this case. Also IIRC Etherscan is happy to return 200 OK { success: false } or something like that in their responses.

As an additional note it might be the case that some endpoints require more or less timeouts/retries and having a single http client underneath makes handling that impossible.

CC: @maciekop-l2b

packages/shared/src/clients/rpc/types.ts
```diff
@@ -0,0 +1,15 @@
...
+  result: z.union([z.string(), z.number(), z.record(z.string(), z.any())]),
+})
+
+export type Block = z.infer<typeof Block>
+export const Block = z.object({
+  // TODO: add support for including txs body
+  transactions: z.array(z.string()),
+  timestamp: stringAsInt(),
+  hash: z.string(),
+  number: stringAsInt(),
```
sz-piotr, line 14: This is not the way to do it. JSON-RPC describes a type called Quantity https://ethereum.org/en/developers/docs/apis/json-rpc/#quantities-encoding. We should be able to say `number: quantity` and also when encoding something like `quantity.encode(number)`

PR #5351, opened 2024-10-08
Optimize getTvlProjects

mateuszradomski:
Part of L2B-7570

- [x] Not recreating the all projects array (-40ms) 
- [x] Removing the n^2 filter from the array (-150ms)
- [x] Filtering sooner so we don't do useless computation (1.5x)
- [x] getTvlAmountsConfig doesn't require all the projects

### DIFF
--- a/packages/config/src/tvl/getTvlAmountsConfig.ts
+++ b/packages/config/src/tvl/getTvlAmountsConfig.ts
@@ -52,11 +52,6 @@ export function getTvlAmountsConfig(
     ),
   )
 
-  const chainMap: Record<number, ChainConfig> = {}
-  for (const chain of chains) {
-    chainMap[chain.chainId] = chain
-  }
-
   for (const project of projects) {
     for (const escrow of project.escrows) {
       switch (escrow.sharedEscrow?.type) {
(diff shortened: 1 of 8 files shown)


lucadonnoh:
filter 👏 before 👏 mapping 👏 

maybe my university scala lessons were not so useless after all

packages/config/src/tvl/getTvlAmountsConfig.ts
```diff
@@ -127,6 +132,11 @@ export function getTvlAmountsConfigForProject(
     entries.push(configEntry)
   }
 
+  const chainMap: Record<number, ChainConfig> = {}
+  for (const chain of chains) {
+    chainMap[chain.chainId] = chain
+  }
```
torztomasz, line 138: let's use groupBy for this one

packages/frontend/src/server/features/scaling/tvl/utils/get-tvl-values-for-projects.ts
```diff
@@ -20,13 +20,15 @@ export async function getTvlValuesForProjects(
...
+  const valueRecords = await db.value.getForProjects(
+    projects.map((p) => p.projectId),
+  )
 
   const valuesByProject = groupBy(valueRecords, 'projectId')
 
   const result: Dictionary<Dictionary<ValueRecord[]>> = {}
   for (const [projectId, projectValues] of Object.entries(valuesByProject)) {
-    const project = projects.find((p) => p.id === projectId)
+    const project = projects.find((p) => p.projectId === projectId)
```
torztomasz, line 31: shouldn't we use the map here too to eliminate `.find` lookup

torztomasz (approved):

PR #5352, opened 2024-10-08
Move OP stack to discovery driven data (batch 2)

sekuba:
Part of L2B-7544

all op stacks discoverydriven starting from blast tvl downwards


lucadonnoh (changes requested):
image

i know this was present already, but why does it happen? (blast)

lucadonnoh:
rest lgtm

sekuba:
Dai vault had empty token brackets since Blast was added :p

lucadonnoh:
> Dai vault had empty token brackets since Blast was added :p

isn’t this because we’re tracking USDB already and we shouldn’t double count?

sekuba:
yeeees of course! what should we do?
1) yeet whole escrow (it is still described in the contracts section as usdyieldprovider)
2) idk 1 is very good choice prly

lucadonnoh:
i don’t know. please revert the change and discuss this tomorrow during standup (i won’t be there cuz traveling) and i’ll approva this PR in the meantime

lucadonnoh (approved):

PR #5354, opened 2024-10-08
Adjusts costs chart time range

sdlyy:
Resolves L2B-7386

### DIFF
--- a/packages/frontend/src/server/features/scaling/costs/get-costs-chart.ts
+++ b/packages/frontend/src/server/features/scaling/costs/get-costs-chart.ts
@@ -1,4 +1,4 @@
-import { type AggregatedL2CostRecord } from '@l2beat/database'
+import type { AggregatedL2CostRecord } from '@l2beat/database'
 import { UnixTime } from '@l2beat/shared-pure'
 import {
   unstable_cache as cache,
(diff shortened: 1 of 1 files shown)


packages/frontend/src/server/features/scaling/costs/get-costs-chart.ts
```diff
@@ -46,20 +46,26 @@ const getCachedCostsChart = cache(
...
 
     if (data.length === 0) {
       return []
     }
 
     const summedByTimestamp = sumByTimestamp(data, resolution)
-    const [from, to] = range
     const timestamps = generateTimestamps(
-      [from, to.add(-1, resolution === 'daily' ? 'days' : 'hours')],
+      [fromToQuery, to.add(-1, resolution === 'daily' ? 'days' : 'hours')],
```
torztomasz, line 68:
```suggestion
      rangeForQuery,
```

torztomasz (dismissed):
Overall I think it is wrong solution, let's talk when you are available

packages/frontend/src/server/features/scaling/costs/get-costs-chart.ts
sdlyy: How can this be wrong? Database query convention stays the same, and without extending the range we'll be lacking data

imxeno (approved):

PR #5369, opened 2024-10-09
Opstack discoverydriven (batch 3)

sekuba:
Small fixes for e.g. Base Challenger inheritance are incoming, but i think we can merge like this

only OP stack not included is optimism because of the native permissions

Closes L2B-7544


lucadonnoh (approved):

lucadonnoh:
ggwp

PR #5387, opened 2024-10-11
Replace type with proof system

torztomasz:


lucadonnoh:
IMG_2457


not vertically aligned for me

PR #5397, opened 2024-10-11
Plugin RpcAnalyzer logic to BlockActivityIndexer

maciekzygmunt:
Resolves L2B-7619


packages/backend/src/modules/activity/services/providers/RpcTxsCountProvider.ts
```diff
@@ -17,15 +19,24 @@ export class RpcTxsCountProvider {
...
+        this.uopsAnalyzer.analyzeBlock(block)
       const timestamp = new UnixTime(block.timestamp)
 
+      const txsCount =
+        this.assessCount?.(transactionsLength, blockNumber) ??
+        transactionsLength
+      const uopsCount =
+        this.assessCount?.(uopsLength, blockNumber) ?? uopsLength
+
+      const ratio = uopsCount / txsCount
```
maciekop-l2b, line 33: Let's remove ration calculation from here as it's no being used and return types is `ActivityRecordWithoutRatio`

maciekop-l2b (changes requested):

maciekop-l2b (approved):

PR #5405, opened 2024-10-12
[External] Add 2 tokens to re.al

sekuba:
from external PR without push permissions, I reviewed the tokens tho

CVR is a liquid wrapper for DEFI positions in pearl, thus cannot be added

UKRE is an RWA without circ supply data (but seems to be fully circulating from explorer data) - added

### DIFF
--- a/packages/config/src/tokens/generated.json
+++ b/packages/config/src/tokens/generated.json
@@ -8796,6 +8796,28 @@
       }
     },
     {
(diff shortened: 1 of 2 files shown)


radinat (approved):

chasebrownn:
Thank you for your help @sekuba . Will reference the guidelines on next PR.

PR #5406, opened 2024-10-12
Update Nil Foundation

rosaria292:
Add explorer and repo links


packages/config/src/projects/layer2s/nil.ts
```diff
@@ -16,8 +16,8 @@ export const nil: Layer2 = upcomingL2({
       websites: ['https://nil.foundation/'],
       apps: [],
       documentation: ['https://docs.nil.foundation/nil/'],
```
lucadonnoh, line 18:
```suggestion
      documentation: ['https://docs.nil.foundation/'],
```

lucadonnoh (changes requested):

lucadonnoh (approved):

PR #5407, opened 2024-10-12
Add Ithaca chain

carl587:


packages/config/src/projects/layer2s/ithaca.ts
```diff
@@ -0,0 +1,26 @@
+import { underReviewL2 } from './templates/underReview'
+import { Layer2 } from './types'
+
+export const ithaca: Layer2 = underReviewL2({
```
lucadonnoh, line 4: Ithaca is the name of the company, not the L2. Rename this to Odyssey and other instances of the name here too

lucadonnoh (changes requested):

packages/config/src/projects/layer2s/ithaca.ts
carl587: Thank you for the correction! I’ve made the changes and renamed it to Odyssey in all relevant instances.

lucadonnoh (approved):

PR #5411, opened 2024-10-14
Update Activity repo for new UOPS fields 

maciekzygmunt:
Resolves L2B-7621


packages/backend/src/modules/activity/indexers/DayActivityIndexer.ts
```diff
@@ -34,7 +34,12 @@ export class DayActivityIndexer extends ManagedChildIndexer {
       adjustedTo,
     )
 
-    await this.$.db.activity.upsertMany(counts)
+    await this.$.db.activity.upsertMany(
+      counts.map((count) => ({
+        ...count,
+        ratio: count.uopsCount ? count.uopsCount / count.count : null,
```
maciekop-l2b, line 40: Let's have it fixed to two decimal points

maciekop-l2b (changes requested):

PR #5415, opened 2024-10-14
Add search bar

torztomasz:


packages/frontend/src/app/layout.tsx
```diff
@@ -31,6 +48,11 @@ export default async function RootLayout({
       <head>
         {/* The rest of the icons are handled by the App Router */}
         <link rel="mask-icon" href="/mask-icon.svg" />
+        {/* Prevent zooming on input click on iOS by adding maximum-scale=1 */}
+        <meta
+          name="viewport"
+          content="width=device-width, initial-scale=1, maximum-scale=1"
```
imxeno, line 54: it doesn't work on iOS 18 in Safari (btw. I don't think we should fix it like that)

imxeno (approved):

PR #5418, opened 2024-10-14
Add templates and shapes for discoverydriven orbit stack

sekuba:
Part of L2B-7474
Closes L2B-7769


sekuba:
check out [galxegravity as a discodriven example](https://l2beat-production-git-orbit-disco-driven-l2beat.vercel.app/scaling/projects/galxegravity#permissions)

sekuba:
The upgrade permission of the UpgradeExecutor is currently not correctly inherited since we cannot target from inside accesscontrol fields yet.

lucadonnoh (approved):

PR #5436, opened 2024-10-15
Adding new token to re.al tokens

chasebrownn:
Adding USDR to real tokens list

`yarn tokens` was not working for me so if I can get an admin to perform that action that would be very helpful.

Thanks

### DIFF
--- a/packages/config/src/tokens/tokens.jsonc
+++ b/packages/config/src/tokens/tokens.jsonc
@@ -4039,6 +4039,21 @@
       "coingeckoId": "uk-real-estate"
     },
     {
+      "symbol": "USDR",
+      "address": "0xb2d75f8Aa33608cF15940Ed47bF139F7CD15d073",
+      "source": "external",
+      "supply": "totalSupply",
+      "bridgedUsing": {
+        "bridges": [
+          {
+            "name": "Layer Zero",
+            "slug": "omnichain"
+          }
+        ]
+      },
+      "coingeckoId": "real-usd"
+    },
+    {
       "symbol": "USTB",
       "address": "0x83fedbc0b85c6e29b589aa6bdefb1cc581935ecd",
       "source": "external",


sekuba:
this stablecoin is depegged and they are planning a migration to a different CA

sekuba:
yes, like for every coin we need valid coingecko price data though.. the only CG page i found was for the old coin

chasebrownn:
@sekuba Its the same token.. just bridged. It’s not a different contract address.

sekuba:
so the [price of 40 cents ](https://www.coingecko.com/en/coins/real-usd)is correct?
we would be very hesitant to add a token with a price feed from a different chain while the price is unpegged.

chasebrownn:
Yes the price is correct. And what “migration” are you referring to?

chasebrownn:
@sekuba Is there anything additionally I can provide for you?

chasebrownn:
@sekuba Just nudging here

sekuba:
sir, we are not adding the token in this state. 
* formally, we do not add any tokens with nonexisting price data for the chain they are added to. (this is why you could not succeed with the tokens script)
* informally, the info around this asset is chaotic at best and actually misleading from my short research (why is it - one year later - still not fully redeemed if that was what the team communicated as their goal... why are there farming campaigns for an asset that is illiquid, unpegged, and supposed to be sunset)

If you really are confident it is an essential asset in you eco, please put in some effort and help us understand what use-case there is for this token and what is the actual goal of the 'redemptions'/'migration' and we can make an exception for the formal problem mentioned above

chasebrownn:
@sekuba We understand. Once redemption is completed the remaining USDR is going to be burned and deprecated anyway so it's not worth the hassle. Thanks for your help nonetheless.

Danni1326:
I am completely new to cryptocurrency and and wasn't aware of the many
layers too it. On top of I am completely technologically declined. I
started a coinbase account, than a wallet, than I started to see NFT's and
Dapps.....needless to say I am trying to break generational poverty and to
try something completely knew, with a lack of computer skills and noone to
really learn from

Danni1326:
Apologies for the misspelled words. I obtain a new phone am struggling a
bit to get everything half operable again. I do appreciate any guidance or
knowledge so this doesn't happen again I found a couple more address on the
block chains that look to be getting steady income on my behalf. I need to
relinquish the wallet apparently?!

PR #5447, opened 2024-10-15
Add Shibarium bridge

MaximousDarwin:
Creating pull request to add Shibarium bridge to the l2beat bridge list, since Shibarium can't be added to the chains list for being a sidechain L2


MaximousDarwin:
Cancelling temporarily.

PR #5455, opened 2024-10-16
Replace roles with permissions

mateuszradomski:
Resolves L2B-6946


packages/backend/discovery/_templates/opstack/L2OutputOracle/template.jsonc
```diff
@@ -17,22 +17,22 @@
   "fields": {
     "proposer": {
       "target": {
-        "role": "Proposer"
+        "permissions": [{ "type": "propose", "delay": 0 }]
```
adamiak, line 20: Let's remove `"delay": 0` in all places. This just adds unnecessary noise. We also don't add it in other places.

sekuba (approved):
looks nice to me
stuff until permissions heaven:
- delete legacy roles section, ascend addresses with roles to top of permissions section or keep legacy roles section similar to now but minimal and just short names linking to the full entries in permissions section
- replace orbitstack 'validate' permission (is rly just sequence + challenge)

packages/config/src/discovery/ProjectDiscovery.ts
```diff
@@ -981,16 +1015,37 @@ export function stringFormat(str: string, ...val: string[]) {
...
-    'Sequencer is an actor allowed to commit transactions from current layer to the host chain.',
-  Proposer:
-    'Proposer is an actor allowed to post new state roots of current layer to the host chain.',
-  Challenger:
-    'Challenger is an actor allowed to delete state roots proposed by a Proposer.',
-  Guardian: 'Guardian is an actor allowed to pause deposits and withdrawals.',
-  Validator:
-    'Validator is an actor that validates the correctness of state transitions.',
+const roleDescriptions: {
+  [key in (typeof RolePermissionEntries)[number]]: {
```
adamiak, line 1019: Can we use description of the permission as defined in the template?

packages/discovery/schemas/contract.v2.schema.json
```diff
@@ -98,7 +98,7 @@
       "properties": {
         "type": {
           "type": "string",
-          "enum": ["configure", "upgrade", "act"]
+          "enum": ["guard", "configure", "upgrade", "act"]
```
adamiak, line 101: There are a few values missing here.

packages/discovery/src/discovery/permission-resolving/resolvePermissions.ts
```diff
@@ -23,7 +23,16 @@ import { assert, EthereumAddress } from '@l2beat/shared-pure'
 // what to do in a scenario when one members has a smaller delay than others
 // and such a case should be discussed.
 
-export type Permission = 'member' | 'act' | 'configure' | 'upgrade'
+export const RolePermissionEntries = [
```
adamiak, line 26: Please leverage TypeScript type system, so that I can't simply put an invalid value in `RolePermissionEntries`. Either decalre that values are from `RawPermissionConfiguration`, or - even better - define `RawPermissionConfiguration` as a union between RolePermissionEntries and e.g. NonRolePermissionEntries (act, configure, update).

packages/config/src/discovery/ProjectDiscovery.ts
sekuba: just a note: Validator role is only temporary and will be removed later exactly because of this ambiguity between the validators in Arbitrum and validators of say ethereum or any other pos network

mateuszradomski: It's going to duplicate the code even more. Since we're repeating the permission validator in many different places since most of the code is not yet templatized that well.

adamiak (approved):

PR #5473, opened 2024-10-17
Update blobstream to single file

vincfurc:
resolves L2B-7712


imxeno (approved):
Approved, but shouldn't we move the SP1Blobstream.ts file to the parent directory, and remove the SP1Blobstream dir?

PR #5499, opened 2024-10-18
Worldchain discovery multisig update

vincfurc:


kurbanski (approved):
This might need a further review later as well.

PR #5505, opened 2024-10-21
Add significant tokens for fuel TVL (multichain-->canonical bridged)

lucadonnoh:
please check that this wouldn't double count the SolvBTC we already have on L2s. This is mainly to solve missing TVL to Fuel

Resolves L2B-7803,L2B-7834

### DIFF
--- a/packages/config/src/projects/layer2s/arbitrum.ts
+++ b/packages/config/src/projects/layer2s/arbitrum.ts
@@ -446,6 +446,7 @@ export const arbitrum: Layer2 = orbitStackL2({
       // ERC20 Gateway
       address: EthereumAddress('0xa3A7B6F88361F48403514059F1F16C8E78d60EeC'),
       tokens: '*',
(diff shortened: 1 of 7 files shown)


sekuba:
solvBTC is kind of a problem because you can wrap / unwrap it on multiple chains, for example on Ethereum, Arbitrum and Mantle
underlyings on these chains are tokens that we count, e.g. wBTC, cbBTC

sekuba:
we can hope that nobody bridges canonically to the projects that allow wrapping (at least on arb there is currently nothing in the escrow)

but apart from that, you are adding the wrong solvBTC i think. Main token on fuel is solvBTC.BBN
image

i am looking into it

sekuba:
so for solvBTC.BBN there is currently no way to redeem to the underlying solvBTC (nor mint new solvBTC.BBN either)
image

before it was available to wrap/unwrap on the same chains as solvBTC above.

I guess we could add the two tokens on ethereum and explicitly exclude it from Ethereum, Arbitrum, Base, BOB and Mantle

PR #5559, opened 2024-10-23
Add FUL, MOON, ybUSD and FRTN to Cronos zkEVM TVL

lezzokafka:
Hello L2beat team, this PR aims to:
- Add FUL, MOON, ybUSD, and FRTN.
- Include the above tokens in the TVL of Cronos zkEVM.


sekuba:
MOON does not have price data on CG (only geckoterminal)
ybUSD is listed on CG but is currently in preview mode. (in this case we cannot easily add it with a price of another token because our tests will fail)

sekuba (approved):

PR #5607, opened 2024-10-28
Fix DA Bridge resolution for Redstone and Xterio

sdlyy:
Resolves L2B-7870


sdlyy:
> I don't think this is needed after #5605

exactly, closing

PR #5612, opened 2024-10-28
Better state updates

mateuszradomski:
Resolves L2B-4742


packages/backend/src/modules/finality/analyzers/types/BaseAnalyzer.ts
```diff
@@ -88,5 +123,69 @@ export abstract class BaseAnalyzer {
    * @param transaction
    * @returns TTI/SUD delays in seconds for each transaction
```
maciekzygmunt, line 124:
```suggestion
   * @returns array of L2Block number and timestamp
```

packages/backend/src/modules/finality/analyzers/types/BaseAnalyzer.ts
```diff
@@ -88,5 +123,69 @@ export abstract class BaseAnalyzer {
...
    */
-  abstract analyze(transaction: Transaction): Promise<number[]>
+  abstract analyze(
+    previousTransaction: Transaction,
+    transaction: Transaction,
+  ): Promise<L2Block[]>
+}
```
maciekzygmunt, line 130: Maybe we can mae previousTransaction optional as we do not even use it in T2I analyzers

packages/backend/src/modules/finality/analyzers/LoopringFinalityAnalyzer.ts
```diff
@@ -1,16 +1,12 @@
...
 import { BigNumber, utils } from 'ethers'
 
 import { Database } from '@l2beat/database'
 import { DegateClient } from '../../../peripherals/degate'
 import { LoopringClient } from '../../../peripherals/loopring/LoopringClient'
 import { RpcClient } from '../../../peripherals/rpcclient/RpcClient'
 import { BaseAnalyzer } from './types/BaseAnalyzer'
+import type { L2Block, Transaction } from './types/BaseAnalyzer'
 
 export class LoopringFinalityAnalyzer extends BaseAnalyzer {
```
maciekzygmunt, line 11: I think we should rename it to T2I analyzer as finality analyzer too general

packages/backend/src/modules/finality/analyzers/types/BaseAnalyzer.ts
mateuszradomski: I don't want to make it optional because the `BaseAnalyzer.ts` is always passing it down

maciekzygmunt (approved):

PR #5625, opened 2024-10-28
Enforce strict rules for canonically bridged assets

sekuba:
canonically bridged asset: 
- an asset bridged via canonical AMB
- escrowed in a contract with the same governance as the L2
- with a standard contract on L2 without minters (e.g. NOT USDC bridged standard)

other assets are tagged as `source: 'external'` with this PR

Closes L2B-7965, L2B-7977


packages/config/src/projects/layer2s/arbitrum.ts
```diff
@@ -439,6 +439,7 @@ export const arbitrum: Layer2 = orbitStackL2({
       // Custom ERC20 Gateway
       address: EthereumAddress('0xcEe284F754E854890e311e3280b767F80797180d'),
       tokens: '*',
+      source: 'external',
```
lucadonnoh, line 442: why

sekuba: the custom gateway is used for custom l2 contracts (and usually custom governance, especially minting/upgrades)
USDT on arb is owned by tether, USDC by circle etc.

sekuba: so USDT and USDC at the source as well as at the destination are fully externally controlled (same for other tokens in custom gateways)
the fact that the custom gateway itself is owned by the rollup seems more like a technicality in this context

PR #5626, opened 2024-10-28
Add public activity API endpoints

sdlyy:
Resolves L2B-7732


packages/frontend/src/app/api/(public)/scaling/activity/[slug]/route.ts
```diff
@@ -0,0 +1,67 @@
...
+      success: false,
+      error: 'Project not found.',
+    })
+  }
+
+  const isEthereum = project.slug === 'ethereum'
+
+  const filter: ActivityProjectFilter = isEthereum
+    ? { type: 'all' }
+    : { type: 'projects', projectIds: [project.id] }
```
Comment (torztomasz) [inline packages/frontend/src/app/api/(public)/scaling/activity/[slug]/route.ts:38]: we want to show all project if ethereum is passed as slug?

torztomasz (changes requested):

packages/frontend/src/app/api/(public)/scaling/activity/[slug]/route.ts
Comment (sdlyy) [reply packages/frontend/src/app/api/(public)/scaling/activity/[slug]/route.ts:38]: If we could use ethereum as ProjectId then I would have used it - but that's no the case - we pull 'something' and extract ethereum data points - look below

Comment (torztomasz) [reply packages/frontend/src/app/api/(public)/scaling/activity/[slug]/route.ts:38]: got it

torztomasz (approved):

PR #5634, opened 2024-10-29
Add banner under permissions and smart contracts for DACs

imxeno:


sdlyy:
Looks odd on mobile - maybe we can reuse/alter host-chains-risks component - it is similar if not the same 
image

sdlyy (approved):

PR #5642, opened 2024-10-29
Update GH Actions to newer versions

kondziorex7:
This is because they rely on older NodeJS version that will soon become deprecated by Github


sz-piotr:
@kondziorex7 the node actions are already updated on main branch but you can update the PR for the checkout actions only

maciekop-l2b (approved):

sz-piotr (approved):

PR #5652, opened 2024-10-30
Add DA references

vincfurc:
resolves L2B-7605


lucadonnoh:
this is not what i meant, i imagined references: [] as we have in scaling projects, at the end if each section

radinat:
> this is not what i meant, i imagined references: [] as we have in scaling projects, at the end if each section

But is it an issue if it's like this? I think this way it's even easier to know what the references are about?

lucadonnoh:
@radinat are you suggesting we should update all scaling pages?

radinat:
> @radinat are you suggesting we should update all scaling pages?

@lucadonnoh not necessarily, in the markdown sections on the scaling pages we're already adding references in the same way (State derivation, State validation)

lucadonnoh:
i think we should be consistent

PR #5679, opened 2024-10-31
Add metadata to DA pages

imxeno:


sdlyy:
I can't get em to work properly - summery and risks work properly but the detailed page is missing of image tags in this configuration

sdlyy (approved):

PR #5710, opened 2024-11-04
Change wording

sdlyy:
~I dislike it without at least a tooltip explaining that given projects are not listed on the main page thus not mentioned here under `used by`~


radinat:
The tooltip on the detailed page summary header should say "There are no scaling projects listed on L2BEAT that use this DA layer", without "and bridge"

radinat:
@sdlyy the wrong tooltip was updated 😃 
image
I meant the one in the top summary that's for the layer only, not the table

sdlyy:
Went for unified one aka 'solution'

radinat (approved):

PR #5715, opened 2024-11-05
Implement auth in Token UI

imxeno:


packages/cms/src/app/(auth)/layout.tsx
```diff
@@ -35,21 +36,26 @@ const menu = [
...
 
   // biome-ignore lint/suspicious/useAwait: server action must be async
   async function logout() {
     'use server'
-    cookies().set('auth_session', 'unauthenticated', { expires: new Date(0) })
+    deleteSession()
     redirect('/auth')
   }
 
+  console.log(session)
```
sdlyy, line 54: 👀

packages/cms/src/app/api/auth/callback/route.ts
```diff
@@ -0,0 +1,101 @@
...
+  sub: z.string(),
+  hd: z.string().optional(),
+  email: z.string().email(),
+  email_verified: z.boolean(),
+  name: z.string(),
+  picture: z.string().optional(),
+})
+
+// 12 hours
+const expiresAt = new Date(Date.now() + 1000 * 60 * 60 * 12)
```
sdlyy, line 21: Are route files evaluated each time somebody calls an endpoint? If not, then `expirestAt` will be evaluated once and will become obsolete within 12 hours after the deployment, no?

imxeno: uhm, yeah, good catch!

PR #5726, opened 2024-11-05
Declutter repository

torztomasz:


packages/backend/discovery/kroma/ethereum/script-ts/typechain-gen/Colosseum.ts
```diff
@@ -432,96 +432,96 @@ export interface ColosseumInterface extends utils.Interface {
...
-export type ChallengeCreatedEvent = TypedEvent<
+type ChallengeCreatedEvent = TypedEvent<
   [BigNumber, string, string, BigNumber],
   ChallengeCreatedEventObject
 >
 
 export type ChallengeCreatedEventFilter =
   TypedEventFilter<ChallengeCreatedEvent>
 
-export interface ChallengeDismissedEventObject {
```
sz-piotr, line 475: Those typechain files are autogenerated

packages/frontend/public/images/finality/state-diff-zk-mobile.png
sz-piotr, line 1: should those files be removed?

packages/l2b/package.json
```diff
@@ -26,7 +26,6 @@
     "@l2beat/config": "workspace:*",
     "@l2beat/discovery": "workspace:*",
     "@l2beat/discovery-types": "workspace:*",
-    "@l2beat/protocolbeat": "workspace:*",
```
sz-piotr, line 29: This is actually used but not imported. See discovery-ui/main.ts

packages/l2b/src/implementations/discovery-ui/types.ts
```diff
@@ -2,7 +2,7 @@
 
 export type ApiProjectsResponse = ApiProjectEntry[]
 
-export interface ApiProjectEntry {
+interface ApiProjectEntry {
```
sz-piotr, line 5: And again here IMO going too far. Files that define types should export everything whether it's used or not.

sz-piotr (changes requested):

packages/config/src/projects/bridges/common/riskView.ts
```diff
@@ -11,28 +11,28 @@ export function NATIVE_AND_CANONICAL(
...
-export const WRAPPED: ScalingProjectRiskViewEntry = {
+const WRAPPED: ScalingProjectRiskViewEntry = {
   value: 'Wrapped',
   description:
     'Tokens transferred by the bridge are not canonical. Users who wish to obtain the canonical counterparts need to do so by trading.',
   sentiment: 'bad',
 }
 
-export const UPGRADABLE_NO: ScalingProjectRiskViewEntry = {
+const UPGRADABLE_NO: ScalingProjectRiskViewEntry = {
```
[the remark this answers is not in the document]
torztomasz: Yeah, I agree, I will go through files and rethink this really quick. Btw. in this case the export is in the bottom

packages/frontend/public/images/finality/state-diff-zk-mobile.png
torztomasz: I mean, I was hesitant to delete it for the past 6 months because we may use it. But it doesnt look like we will push finality diagrams in the near future so using this PR I just got rid of it

packages/l2b/package.json
torztomasz: good catch

sz-piotr (approved):

PR #5740, opened 2024-11-06
Add lz USDC to re.al

sekuba:
Closes L2B-8124


radinat (approved):

chasebrownn:
@sekuba Is the TVL of lz USDC excluded?

sekuba:
just from the total tvl for all scaling projects, not from re.al
This is because this token is locked on other L2s, where it is already counted towards the total

PR #5749, opened 2024-11-06
Add Educhain (full)

sekuba:
Closes L2B-8036


sekuba:
project launch postponed to q1 next year (source: telegram)

PR #5765, opened 2024-11-07
Add CROAK

Thisuu:
Add croak by (Efrogs)[https://x.com/efrogs_on_linea] project


mateuszradomski:
Hi,

Thank you for the PR. The token does not have any circulating supply on Coingecko so we can't add it since we use the Market Cap data to determine the supply.

Thisuu:
> Hi,
> 
> Thank you for the PR. The token does not have any circulating supply on Coingecko so we can't add it since we use the Market Cap data to determine the supply.

Hi, The circulating supply issue on Coingecko is now fixed. Should I open a new PR?

PR #5767, opened 2024-11-07
Test to keep scaling and DA-BEAT projects in sync

sdlyy:
Resolves L2B-7838


sdlyy:
We have to wait until fluence, fraxtal, oevnetwork and apechain have corresponding DA-BEAT entries assigned ;p

PR #5804, opened 2024-11-12
Allow editing external ids in CMS

imxeno:


packages/cms/src/app/(auth)/tokens/[id]/_actions/index.ts
```diff
@@ -74,16 +76,20 @@ export const updateToken = actionClient
...
-          tokenId: id,
-          externalId: '',
-          source: 'Overrides',
-          symbol: customMeta?.symbol ?? null,
-          name: customMeta?.name ?? null,
-          logoUrl: customMeta?.logoUrl ?? null,
-          decimals: customMeta?.decimals ?? null,
-          contractName: customMeta?.contractName ?? null,
-        })
+        await db.tokenMeta.deleteByTokenId(id)
```
Comment (torztomasz) [inline packages/cms/src/app/(auth)/tokens/[id]/_actions/index.ts:79]: why delete when upserting?

Comment (imxeno) [reply packages/cms/src/app/(auth)/tokens/[id]/_actions/index.ts:79]: Without delete here we're left with e.g. CoinGecko source meta without externalId.

PR #5810, opened 2024-11-12
Taiko: ZK proofs upgrade (Ontake)

sekuba:


packages/backend/discovery/taiko/ethereum/config.jsonc
```diff
@@ -13,18 +13,27 @@
     "0x996282cA11E5DEb6B5D122CC3B9A1FcAAD4415Ab": "SharedERC20Vault",
     "0x06a9Ab27c7e2255df1815E6CC0168d7755Feb19a": "TaikoL1Contract",
     "0x68d30f47F19c07bCCEf4Ac7FAE2Dc12FCa3e0dC9": "ProverSetProxy",
```
lucadonnoh, line 15: it's worth it to rename this to make it more obvious that it is specific to the DAO_FALLBACK_PROPOSER as rn it seems a core contract and it isn't, also since this is a special address we can get the provers (in `isProver` mapping)

lucadonnoh (changes requested):

packages/frontend/public/images/architecture/taiko.png
lucadonnoh, line 1: add function name to propose blocks

lucadonnoh (changes requested):

lucadonnoh:
rest lgtm

packages/backend/discovery/taiko/ethereum/config.jsonc
sekuba: do you have a better name than DAOProverSetProxy?

lucadonnoh: what about DAOFallbackProposer

sekuba: okay, but i will sync this name in the text and diagram then

lucadonnoh (approved):

PR #5829, opened 2024-11-13
Fix callout text & icon positioning

imxeno:


torztomasz (commented):
Check chromatic, it looks really off after the changes

torztomasz (commented):
Check chromatic, it looks really off after the changes

PR #5848, opened 2024-11-15
Run CLI in Discovery UI

adamiak:


packages/protocolbeat/src/panel-terminal/TerminalPanel.tsx
```diff
@@ -0,0 +1,102 @@
...
+  useEffect(() => {
+    return () => {
+      // Cleanup on unmount
+      abortControllerRef.current?.abort()
+    }
+  }, [])
+
+  useEffect(() => {
+    if (outputRef.current) {
+      outputRef.current.scrollTop = outputRef.current.scrollHeight
```
sz-piotr, line 40: This is always frustrating if implemented in this way. You only want to scroll if it was already fully scrolled. Otherwise you want to keep where the user scrolled to.

packages/protocolbeat/src/panel-terminal/TerminalPanel.tsx
```diff
@@ -0,0 +1,102 @@
...
+    setIsRunning(true)
+
+    try {
+      const stream = await executeCommand('discover', project, discoveryChain)
+      abortControllerRef.current = new AbortController()
+
+      const reader = stream.getReader()
+      while (!abortControllerRef.current.signal.aborted) {
+        const { value, done } = await reader.read()
+        if (done) break
```
sz-piotr, line 62: Shouldn't we somehow close the stream? Or pass the abortcontroller to fetch in executeCommand?

packages/l2b/src/implementations/discovery-ui/main.ts
```diff
@@ -46,6 +49,25 @@ export function runDiscoveryUi() {
...
+    if (!command || !project || !chain) {
+      res.status(400).send('Missing required parameters')
+      return
+    }
+    if (command !== 'discover') {
+      res.status(400).send('Invalid command')
+    }
+
+    res.setHeader('Content-Type', 'text/plain')
+    executeTerminalCommand(
```
sz-piotr, line 65: I was thinking we could utilize https://developer.mozilla.org/en-US/docs/Web/API/Server-sent_events/Using_server-sent_events for this. However I don't want it getting too complex, maybe this is actually enough for what we need, we just need a better cancellation policy.

packages/protocolbeat/src/store/store.ts
```diff
@@ -12,3 +12,19 @@ export const usePanelStore = create<State & Actions>((set) => ({
...
+  clear: () => void
+}
+
+export const useTerminalStore = create<TerminalState>((set) => ({
+  output: '',
+  isRunning: false,
+  addOutput: (text) => set((state) => ({ output: state.output + text })),
+  setIsRunning: (running) => set({ isRunning: running }),
+  clear: () => set({ output: '' }),
+}))
```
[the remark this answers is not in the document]
adamiak: done

packages/l2b/src/implementations/discovery-ui/main.ts
adamiak: Reimplemented

packages/protocolbeat/src/panel-terminal/TerminalPanel.tsx
adamiak: This is no longer relevant since I switched to server-sent events.

packages/protocolbeat/src/panel-terminal/TerminalPanel.tsx
adamiak: This is not fixed as I find it a bit too challenging for now (my solutions never work reliably). I'll do it as a separate PR when I have more time or pass it to someone better with frontend code.

packages/protocolbeat/src/panel-terminal/TerminalPanel.tsx
```diff
@@ -0,0 +1,110 @@
...
+      eventSource.onmessage = (event) => {
+        const text = event.data.toString()
+        addOutput(text.endsWith('\n') ? text : text + '\n')
+      }
+
+      eventSource.onerror = () => {
+        eventSource.close()
+        setIsRunning(false)
+        queryClient.invalidateQueries({ queryKey: ['projects', project] })
+      }
```
sz-piotr, line 69: onerror fires also on onclose?

sz-piotr (approved):

packages/protocolbeat/src/panel-terminal/TerminalPanel.tsx
adamiak: This is a known quirk of the SSE protocol - normal completion and errors both go through the error handler.

Added a comment.,

PR #5856, opened 2024-11-18
Add tests to frontend server

torztomasz:
Resolves L2B-5892


packages/frontend/src/server/features/scaling/activity/utils/get-fully-synced-activity-range.test.ts
sz-piotr, line 1: empty file

packages/frontend/src/server/features/scaling/costs/utils/sum-cost-values.test.ts
```diff
@@ -0,0 +1,225 @@
...
+
+  it('should handle empty records array', () => {
+    const result = sumValues([])
+
+    expect(result).toEqual({
+      gas: {
+        overhead: 0,
+        calldata: 0,
+        compute: 0,
+        blobs: undefined,
```
sz-piotr, line 111: why undefined? why not 0?

packages/frontend/src/server/features/scaling/costs/utils/sum-cost-values.test.ts
```diff
@@ -0,0 +1,225 @@
...
+    })
+  })
+
+  it('should handle null blob values correctly', () => {
+    const records = [
+      {
+        overheadGas: 100,
+        calldataGas: 200,
+        computeGas: 300,
+        blobsGas: null,
```
sz-piotr, line 134: why null?

packages/frontend/src/server/features/scaling/costs/utils/add-if-defined.test.ts
```diff
@@ -0,0 +1,76 @@
...
+    a: null,
+    b: 3,
+    expected: 3,
+    description: 'should return the second number if the first is null',
+  },
+  {
+    a: undefined,
+    b: undefined,
+    expected: undefined,
+    description: 'should return undefined if both numbers are undefined',
```
sz-piotr, line 46: Wouldn't zero be cleaner? To me this abstraction feels weird.

packages/frontend/src/server/api/trpc.ts
```diff
@@ -1,10 +1,12 @@
 import { initTRPC } from '@trpc/server'
 import superjson from 'superjson'
 import { ZodError } from 'zod'
+import { db } from '../database'
```
sz-piotr, line 4: we can remove those tests and this abstraction

sz-piotr (changes requested):

packages/frontend/src/server/features/scaling/costs/utils/sum-cost-values.test.ts
torztomasz: it's before the duncan upgrade

packages/frontend/src/server/features/scaling/costs/utils/sum-cost-values.test.ts
torztomasz: same, pre duncan upgrade. however it's written in a weird way where we get null from the db and then cast it to undefined

packages/frontend/src/server/features/scaling/costs/utils/add-if-defined.test.ts
torztomasz: It adds up only if something is defined. Take a look at the files that are using it - it would be nightmare to inline this, that's why I did this abstraction

sz-piotr (approved):

PR #5868, opened 2024-11-19
update governance calendar times

L2BeatSinkas:


torztomasz (changes requested):
I would not delete if not needed. Let's keep the history, maybe it will be useful one day

torztomasz (approved):

PR #5869, opened 2024-11-19
Use RpcClient2 in TvlModule

maciekzygmunt:
Resolves L2B-8228


packages/shared/src/clients/rpc/types.ts
```diff
@@ -49,6 +49,10 @@ export const EVMBalanceResponse = z.object({
   result: Quantity.decode.transform((n) => Number(n)),
```
antooni, line 49: casting to number can result in loss of information

packages/backend/src/modules/tvl/services/AmountService.ts
```diff
@@ -79,7 +79,7 @@ export class AmountService {
         return {
           configId: configuration.id,
           type: configuration.type,
-          amount: amount.toBigInt(),
+          amount: BigInt(amount),
```
antooni, line 82: please remove other .toBigInt() invocations

antooni: remove them from all TvlModule files

antooni (changes requested):
Looks good. Please add requested changes & run local sync for the latest hour to see whether everything works as intended

PR #5878, opened 2024-11-19
Morph review (full)

luke7211:
Morph is a Scroll fork that:
- uses decentralised sequencer (stakerSet)
- does not require posting a zkPoof for each L2 state commit
- has a 3 day challenge window during which whitelisted challengers can challenge (rn one EOA address)
- after challenge anyone can submit a zkProof


lucadonnoh (changes requested):
TODO: have a look at the EnforcedTxGateway which is not paused unlike Scroll. I don't think we ever fully reviewed it in Scroll as it was never unpaused.

sekuba:
closing to avoid confusion, here is the new one #5891

PR #5883, opened 2024-11-20
Show inline function signature / event topic

mateuszradomski:
Resolves L2B-8152

### DIFF
--- a/packages/protocolbeat/src/colors.json
+++ b/packages/protocolbeat/src/colors.json
@@ -23,6 +23,7 @@
     "yellow": "#FABD30",
     "cyan": "#1c92a8",
     "pink": "#e27991",
-    "purple": "#a73db5"
+    "purple": "#a73db5",
+    "grey": "#928474"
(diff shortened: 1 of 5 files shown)


packages/protocolbeat/src/colors.json
```diff
@@ -23,6 +23,7 @@
     "yellow": "#FABD30",
     "cyan": "#1c92a8",
     "pink": "#e27991",
-    "purple": "#a73db5"
+    "purple": "#a73db5",
+    "grey": "#928474"
```
sz-piotr, line 27: Can't we use the same color we use for comments in the editor?

sz-piotr (changes requested):

packages/protocolbeat/src/colors.json
mateuszradomski: Good point, fixed in the last commit

sz-piotr (approved):

PR #5929, opened 2024-11-25
TVL for Taiko, add SolvBTC

lorenz234:
We noticed that the TVL for Taiko is off by a factor of approximately 2x because the token SolvBTC is not being tracked.

I’ve added it as natively minted on Taiko. However, there also appears to be a bridged version with around $9M in TVL, which I’m unsure how to add. (Source: [Avalon Finance](https://app.avalonfinance.xyz/?marketName=proto_taiko_v3))

I hope we can get this in, so the TVL is more accurate. Thanks!

### DIFF
--- a/packages/config/src/tokens/tokens.jsonc
+++ b/packages/config/src/tokens/tokens.jsonc
@@ -4586,6 +4586,12 @@
       "supply": "totalSupply",
       "excludeFromTotal": true,
       "coingeckoId": "tether"
+    },
+    {
+      "symbol": "solvBTC",
+      "address": "0x541fd749419ca806a8bc7da8ac23d346f2df8b77",
+      "source": "native",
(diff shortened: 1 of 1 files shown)


lorenz234:
We looked through and found some more tokens that are missing and would add significantly to the TVL
https://docs.google.com/spreadsheets/d/1xNykt2DyFfcN2KrmKz8O7BouVIjGcjMnVG2X1jo6Itg/edit?usp=sharing

sekuba:
heyoo, tokens can be added as [described here](https://github.com/l2beat/l2beat/blob/main/CONTRIBUTING.md#add-tokens-to-a-project) or via our [token request form](https://forms.gle/fQFsC5g1LgG5z12T7) :)

for now i am adding solv  tokens to taiko at #5933

PR #5935, opened 2024-11-26
glossary: added Predicates

Rubyt0x:


sekuba:
this is a pretty narrow definition of 'predicate' and it is unclear how it is relevant as the word is not used on our site

PR #5984, opened 2024-11-29
Adds Token Gateway (Hyperbridge)

olahfemi:
You can find our contracts here: [https://docs.hyperbridge.network/developers/evm/contracts/mainnet](https://docs.hyperbridge.network/developers/evm/contracts/mainnet).

**Other Contracts:**
Multi-chain DOT: 0x8d010bf9C26881788b4e6bf5Fd1bdC358c8F90b8

I'm trying to add our [project](https://hyperbridge.network/token-gateway), but I keep getting an error when I try to run it locally. I have attached a file with the log of the error.


mateuszradomski:
Hi,

provide the error as a pastebin.com url

olahfemi:
> pastebin.com

Here is it: [https://pastebin.com/e7aeL5S9](https://pastebin.com/e7aeL5S9)

olahfemi:
> https://docs.hyperbridge.network/developers/evm/contracts/mainnet

What if we try to implement the logic ourselves?

olahfemi:
> sorry i don't understand, do you mean making a complete PR here by yourselves?

Yes, exactly. And if we need to go beyond that, we are willing.

sekuba:
Yes of course it is possible, but be aware of the following: External PRs with full reviews have barely ever been successful for projects of your scale as our discovery tooling is not easy to get to work (we are working on that). Also we are currently reforming our bridge page in general and would recommend to do the review afterwards as we might even have stricter thresholds regarding our ethereum-centric-ness or e.g. have the resources to do it ourselves. Finally, your bridge is incompatible with our current TVL counting, which we are also planning to change.

So all in all it would make a lot of sense to wait and we would definitely recommend you to!

PR #6002, opened 2024-12-02
feat: add GAEA network

Peter6501:


radinat:
@Peter6501 is there a testnet available? We'll be changing our Upcoming projects requirements to only list chains that have a testnet available and it might get removed

Peter6501:
> @Peter6501 is there a testnet available? We'll be changing our Upcoming projects requirements to only list chains that have a testnet available and it might get removed

Thank you for your response. I'll close this PR for now and will add it back once the testnet is available.

PR #6017, opened 2024-12-03
Show members' keys

sdlyy:
Resolves L2B-8353

image

CopyIcon was used only for the glossary. Replaced with actual copy icon and renamed old one to copybook.


packages/frontend/src/components/copy-button.tsx
```diff
@@ -1,20 +1,52 @@
...
+    active: <SatisfiedIcon className="size-4" />,
+    inactive: <CopyIcon className="size-4 fill-gray-750 dark:fill-white/80" />,
+  },
+  glossary: {
+    active: <SatisfiedIcon className="fill-green-700  dark:fill-green-450" />,
+    inactive: (
+      <CopyBookIcon className="fill-blue-700 hover:fill-blue-600 dark:fill-blue-500 dark:hover:fill-blue-550" />
+    ),
+  },
+}
```
torztomasz, line 41: i don't like the idea of having two different styles for copy button, let's use one

torztomasz (changes requested):

packages/frontend/src/components/copy-button.tsx
sdlyy: https://github.com/l2beat/l2beat/pull/6017/commits/4e81a041ca5ea8c1224b4cf05ad37cfdc1174ed2

torztomasz (approved):

PR #6018, opened 2024-12-04
Cronos zkEVM updates

lezzokafka:
Hello L2beat team, as similar to #5559 - 
This PR aims to update canonical token bridged to Cronos zkEVM.
Thanks again for the on-going support.


sekuba:
hey thanks for the pr!
unfortunately this token is what we call 'preminted' by an EOA. That EOA bridged the entire supply to cronoszkevm, although this token is also on cronos. (liquidity on cronos seems to be higher than on -zkevm)

Since coingecko reports the token as 100% circulating, we cannot add it as it would significantly overcount the liquid supply on cronos zkEVM

closes L2B-8462

PR #6029, opened 2024-12-05
Switch arguments around

mateuszradomski:
Resolves L2B-8407


lucadonnoh:
the only drawback i see is that now I wouldn't be able to use my `pde` (`pnpm discover ethereum`) alias anymore

mateuszradomski:
The majority voted against the switch, closing.

PR #6046, opened 2024-12-06
Add project logo to Discovery UI

maciekzygmunt:


mateuszradomski:
We should use the slug not the project name as those differ in certain projects. There is no easy way to get that in l2b without adding a dependency onto the layer2s/layer3s array which I don't want to do because it will kill the start up performance of that command and it will introduce the dependency of path you run it from. We will probably add it in the future. Closing for now.

PR #6052, opened 2024-12-09
Make Avail discovery-driven

sdlyy:
Resolves L2B-8314


packages/backend/discovery/vector/ethereum/config.jsonc
```diff
@@ -7,26 +7,12 @@
...
-        },
-        "isRelayerApproved": {
-          "handler": {
-            "type": "call",
-            "method": "function approvedRelayers(address) view returns (bool)",
-            "args": ["0x44eb418a966ff47f5af6f48aea6afde0bf193a8d"]
-          }
-        }
-      }
+      "extends": "succinct/SP1Vector"
```
adamiak, line 15: When working with config.jsonc files, please remove all "extends" keywords (and also don't add or approve new "extends"). It's an old approach that we're getting rid of. Instead, add flattened .sol file to template shape, so that it's discovered by implementation match.

packages/backend/discovery/_templates/succinct/SP1SuccinctGateway/template.jsonc
```diff
@@ -54,6 +55,24 @@
...
+          }
+        ]
+      }
+    },
+    "owner": {
+      "target": {
+        "permissions": [
+          {
+            "type": "configure",
+            "description": "manager of router for proof verification, it holds the power to affect the liveness and safety of the bridge"
```
adamiak, line 73: Please make this description more specific - what can owner do?

adamiak:
Can we fix this error? 
image

adamiak:
These SP1Verifier entries should have some description or better name, because now it's not obvious why they're here or what they're doing. If they're referenced as "old", use the "configure" permission with proper description and let's discuss this case later.

packages/backend/discovery/vector/ethereum/config.jsonc
sdlyy: https://github.com/l2beat/l2beat/pull/6052/commits/e606330a4a3b6025053a4d960236b9fd26dc5047
https://github.com/l2beat/l2beat/pull/6052/commits/16ff8bc3b220b425db62556d9a5e71cb2d62b464

adamiak (approved):

PR #6065, opened 2024-12-09
Ethereum DA page

sdlyy:
Resolves L2B-8291

### DIFF
--- a/packages/frontend/src/app/(side-nav)/data-availability/summary/_components/ethereum-da-entry.tsx
+++ b/packages/frontend/src/app/(side-nav)/data-availability/summary/_components/ethereum-da-entry.tsx
@@ -1,4 +1,5 @@
 import Image from 'next/image'
+import Link from 'next/link'
 import {
   Tooltip,
   TooltipContent,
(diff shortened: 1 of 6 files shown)


packages/frontend/src/app/(side-nav)/data-availability/summary/_components/ethereum-da-entry.tsx
```diff
@@ -15,56 +16,59 @@ export function EthereumDaEntry({ entry }: { entry: DaSummaryEthereumEntry }) {
...
-              {formatCurrency(tvsToFormat, 'usd')}
+    <Link href={entry.href ?? ''} className="group">
+      <div className="mb-3 overflow-auto max-md:-mr-4">
+        <div className="relative flex min-w-[940px] items-center justify-between gap-3 rounded-lg bg-blue-300 px-4 py-1 transition-colors duration-200 group-hover:bg-blue-400 dark:bg-blue-950 group-hover:dark:bg-blue-900 max-md:mr-4 md:py-3">
+          <Tooltip>
+            <TooltipTrigger>
+              <InfoIcon variant="blue" />
+            </TooltipTrigger>
+            <TooltipContent className="z-[1000]">
+              From the rollup perspective, Ethereum&apos;s canonical chain
```
sz-piotr, line 27: Quite weird that this is here and not in the config

sz-piotr (changes requested):
I don't think I like this approach to implementing the ethereum page. Looking at the UI it seems to me that while it shares some similarities to other DA pages it is actually a different page that reuses components. I would prefer that implementation to the current one which has a ton of ifs for ethereum because it treats it as if it was the same as the others

packages/frontend/src/app/(side-nav)/data-availability/summary/_components/ethereum-da-entry.tsx
sdlyy: changed, now part of ethereum da config

PR #6077, opened 2024-12-10
Fix `pnpm lint` in frontend

adamiak:
Resolves L2B-8489

fix is to use biome for linting, exactly like we do it in the backend


torztomasz (changes requested):
We need to have eslint on frontend as biome does not support plugins like react-hooks etc.

adamiak:
Looks like we want to use `next lint` here. I'll check with the team later.

PR #6084, opened 2024-12-11
Refactor filters

sz-piotr:


packages/frontend/src/app/(side-nav)/scaling/_components/scaling-filter-context.tsx
```diff
@@ -153,10 +89,10 @@ export function useScalingUpcomingFilter() {
     (entry: ScalingUpcomingEntry) => {
       const checks = [
         scalingFilters.rollupsOnly !== false
-          ? entry.category?.includes('Rollup')
+          ? !!entry.category?.includes('Rollup')
           : undefined,
-        scalingFilters.category !== undefined
-          ? entry.category === scalingFilters.category
+        scalingFilters.type !== undefined
+          ? entry.category === scalingFilters.type
```
torztomasz, line 95: let's refactor this part too

packages/frontend/src/server/features/scaling/activity/get-scaling-activity-entries.ts
```diff
@@ -81,16 +88,21 @@ export async function getScalingActivityEntries() {
...
->
+export interface ScalingActivityEntry extends CommonScalingEntry {
+  entryType: 'activity'
+  dataSource: ScalingProjectDisplay['activityDataSource']
+  dataAvailability: {
+    layer: DataAvailabilityWithSentiment['layer'] | undefined
+  }
+  data: ActivityProjectTableData
+}
+
```
torztomasz, line 99: the pattern that we decided to use is to determine type based on function return value rather than typing the function, it is done like that everywhere and worked pretty smoothly so I would not change it

packages/frontend/src/server/features/scaling/get-common-scaling-entry.ts
```diff
@@ -80,7 +122,36 @@ export function getCommonScalingEntry(
...
+      isRollup: project.display.category.includes('Rollup'),
+      type: project.display.category,
+      stack: project.display.provider ?? 'No stack',
+      stage: getStage(project.stage),
+      purposes: project.display.purposes,
+      hostChain: project.type === 'layer2' ? 'Ethereum' : getHostChain(project),
+      daLayer:
+        getCurrentEntry(project.dataAvailability)?.layer.value ?? 'Unknown',
+      raas: getRaas(project.badges ?? []),
+    },
```
torztomasz, line 137: mb let's move this to separate function that takes `project: Layer2 | Layer3 | Bridge` as an argument and returns filterable object

torztomasz (changes requested):
Shouldn't we also refactor bridges filters the same way?

Also the filters do not work currently
image

packages/frontend/src/app/(side-nav)/scaling/_components/scaling-filter-context.tsx
sz-piotr: Once upcoming uses the same common base `useScalingUpcomingFilter` can be removed and replaced with `useScalingFilter`

packages/frontend/src/server/features/scaling/get-common-scaling-entry.ts
sz-piotr: A separate filterable for bridges is the way to go but needs a common entry first

torztomasz (approved):

PR #6120, opened 2024-12-13
Add geist network as coming

Peter6501:


radinat:
There is an older PR that adds the project

PR #6122, opened 2024-12-13
chore: fix boolean typo

majorteach:


sz-piotr:
This doesn't affect anything - the test doesn't rely on it not being a typo, this doesn't show up anywhere, no need to fix

PR #6136, opened 2024-12-16
Use ProjectService in getScalingRiskEntry

sz-piotr:


packages/config/src/projects/project/Project.ts
```diff
@@ -11,18 +13,60 @@ export interface Project {
...
+  daLayer: string
+  stage: string
+  purposes: string[]
+}
+
+export interface ProjectScalingRisks {
+  self: ScalingProjectRiskView
+  host: ScalingProjectRiskView | undefined
+  stacked: ScalingProjectRiskView | undefined
+}
```
torztomasz, line 64: maybe better to use union?
```suggestion
export type ProjectScalingRisks = {
  type: 'single'
  self: ScalingProjectRiskView
} | {
  type: 'multiple'
  host: ScalingProjectRiskView
  stacked: ScalingProjectRiskView
}
```

packages/frontend/src/server/features/scaling/get-common-scaling-entry.ts
```diff
@@ -102,11 +166,11 @@ export function getCommonScalingEntry({
   }
 }
 
-function getStageOrder(stage: StageConfig | undefined): number {
-  if (stage?.stage === 'Stage 2' || stage?.stage === 'Stage 1') {
+function getStageOrder(stage: string | undefined): number {
```
torztomasz, line 169: can't we have it as StageConfig? I would prefer to have the checks below typesafe

torztomasz (approved):

packages/config/src/projects/project/Project.ts
sz-piotr: I errored on the ease of use. Also when I was writing there was this edge case of a project having stacked risks but no host risks. After removing hostChain = multiple this is no longer the case so maybe union? But then it's harder to use.

packages/frontend/src/server/features/scaling/get-common-scaling-entry.ts
sz-piotr: I'm going to make another PR where I make some of those small things cleaner

PR #6160, opened 2024-12-17
chore: fix 404 status URL

linchizhen:


mateuszradomski:
These URLs are not shown anywhere. These files are shapes and cannot be modified.

PR #6180, opened 2024-12-17
chore:spellchecker for *.md

argentpapa:
- Due to the large number of spelling errors requiring correction, manual filtering of false positives is needed in addition to tool checks.  Currently, errors are being corrected by file type. 
- It is then suggested to add checks to GitHub Actions to address incremental issues. ref:https://github.com/codespell-project/actions-codespell


mateuszradomski:
Changes in `docs.md` are bad. We are not going to be changing `diffHistory.md` files as it's a capsule to track changes when they happened and it's internal only.

PR #6182, opened 2024-12-18
Add zytron to L3s.

fryingcakemaster:


sekuba:
heyo thanks for the PR! the TVL of zytron is very low and all [rollup contracts are unverified](https://lineascan.build/address/0x30f7623dcbdc41351fcbd3068d04b9917d5957fa#code) on lineascan. we can revisit this pr if they are verified in the future and the TVL has increased.

PR #6099, opened 2024-12-12
Update Summary page charts

maciekzygmunt:
Resolves L2B-8503 L2B-8588

PR #6187, opened 2024-12-18
Round UOPS labels 

torztomasz:
This PR makes sure that UOPS are rounded if all UOPS labels do not have fraction part


maciekzygmunt (changes requested):
LGTM, but I would wait with it for https://github.com/l2beat/l2beat/pull/6099 to be merged as it introduces new activity recategorized chart.

maciekzygmunt (approved):

PR #6226, opened 2024-12-23
Add Hyperliquid (full)

lucadonnoh:


luke7211 (approved):

packages/config/src/projects/layer3s/hyperliquid.ts
```diff
@@ -0,0 +1,181 @@
...
+  display: {
+    name: 'Hyperliquid',
+    slug: 'hyperliquid',
+    description:
+      'Hyperliquid is a performant exchange on Arbitrum, utilizing a custom consensus algorithm called HyperBFT.',
+    purposes: ['Exchange'],
+    category: 'Other',
+    reasonsForBeingOther: [
+      REASON_FOR_BEING_OTHER.NO_PROOFS,
+      REASON_FOR_BEING_OTHER.SMALL_DAC,
```
sekuba, line 71: wouldn't we use `REASON_FOR_BEING_OTHER.NO_DA_ORACLE` here because there are no commitments?

lucadonnoh: yeah there is nothing so it’s basically the same, but signatures get actually verified and they can be seen as an oracle for something. idk doesn’t matter imo

PR #6228, opened 2024-12-23
SKALE IMA: new tokens added

manuelbarbas:
### Developer: Manuel Barbas from SKALE Network

This PR aims to update the tokens being tracked by L2Beat on the SKALE IMA Bridge page.

The tokens were both added to `tokens.jsonc` and `skaleIMA.ts`. 

However when running `pnpm tokens` command I got the error `WARN [ RetryHandler ] Scheduling retry  { attempt: 1, delay: 1 }` until it throws the error `Error: HTTP error: 401 Unauthorized`. 

I'm probably missing the API key


radinat:
The changes were made as part of #6288 - now all tokens should be automatically fetched. Thanks for reporting it!

PR #6237, opened 2024-12-26
"the"

Vikt0rPavlik:
The documentation contained a duplicated word "the" which impacts readability and professionalism of the codebase. This small but important fix:


mateuszradomski:
These files are 1-1 copies from source code submitted to Etherscan. We can't change anything in there since it's the basis of our shape matching.

PR #6239, opened 2024-12-26
add COW,DODO,Ondo,Puffer,Usual and some wormhole wrapped tokens

clar:


sekuba:
some tokens from this pr will be added in #6275

for the wormhole tokens, they do not seem to be bridged to any L2 using canonical bridges? so adding them would not have any effect..

sekuba:
oh i see they are for degate bridge sry, will look into the others!

PR #6248, opened 2025-01-01
Add DataDance as coming soon project

ian9403:


radinat:
@ian9403 does the project have a testnet and a bridge interface where we can access it from? We're starting to enforce new requirements for the upcoming projects and one of them is to have a testnet available. If there's not one yet, I'd suggest to close the PR and reopen it once the project is live on testnet.

PR #6254, opened 2025-01-02
Fix activity with stage sorting on

torztomasz:
Resolves L2B-8631


radinat:
@torztomasz On the production build Ethereum is ranked 5th but is placed on the top, I think we should either keep the logic we currently have live until we have the dividers, or remove the rank number for Ethereum (replace with -)

One thing to discuss (if not discussed already) - wouldn't it be better to sort within the groups? Looking at the staging preview, it's a bit confusing that a project with larger Past day UOPS has higher number

sz-piotr (approved):

PR #6261, opened 2025-01-03
Add Status as upcoming project

hejpiotrze:
Resolves L2B-8666


radinat:
@Huberacht @hejpiotrze does the project have a testnet?

PR #6264, opened 2025-01-03
[L2B-7416] Add block activity backfill indexer

maciekop-l2b:


maciekzygmunt (approved):

radinat:
Maciek confirmed that a different approach is needed, closing this.

PR #6272, opened 2025-01-06
Update OPStack finality analyzer

maciekzygmunt:
Resolves L2B-8348

This PR updates OPStack finality T2I analyzer to also decode and calculate type 2 txs.


maciekop-l2b (approved):
LGTM

PR #6286, opened 2025-01-07
Add ability to specify available space for DA Layers

sdlyy:
Resolves L2B-8535

See Celestia on the preview


packages/frontend/src/app/(top-nav)/data-availability/projects/[layer]/_components/da-project-stats.tsx
Comment (vincfurc) [inline packages/frontend/src/app/(top-nav)/data-availability/projects/[layer]/_components/da-project-stats.tsx:1]: we need to specify the timeframe, Celestia right now does 0.33 MB/s

Comment (radinat) [reply packages/frontend/src/app/(top-nav)/data-availability/projects/[layer]/_components/da-project-stats.tsx:1]: Also would it make sense to rename the property to be "Block space", "Space available" might not be clear enough? Or if we're about to configure it in MB/s, then "Max throughput"?

Comment (sdlyy) [reply packages/frontend/src/app/(top-nav)/data-availability/projects/[layer]/_components/da-project-stats.tsx:1]: I'm fine with both - naming is a flexible thing here - I just need a confirmation for the unit. 
So the question remains -  which one will bring more value?

radinat (approved):

vincfurc (approved):

PR #6313, opened 2025-01-09
Add Phala Network (full)

vincfurc:


lucadonnoh (changes requested):
just noticed we're missing risks everywhere

lucadonnoh (changes requested):
please add succintgateway operator name to the gnosis safe and add some funds can be frozen if risks

lucadonnoh (approved):

lucadonnoh (approved):

PR #6315, opened 2025-01-09
Corn Review

januszgrze:
I'm working through the write up on Corn. I've confirmed a few of the trust assumptions, but am working through confirming the delay to self-sequence + validating if the self-proposing functionality is live.

I can confirm it's an Optimium using AnyTrust DA built with Conduit.

Will have this ready for review in a few hours.


sekuba:
gm sir, i did not see this and went ahead reviewing corn at #6403

thanks for the PR though!

sekuba:
btw we did not have to re-review corn risks since its orbit contracts match the source of our previously-reviewed contracts :)

PR #6320, opened 2025-01-10
Refactor icon test to use ProjectService

sz-piotr:


packages/frontend/scripts/logos/tinifiyLogos.test.ts
```diff
@@ -1,46 +1,37 @@
...
-  }
+  const icons = readdirSync(path.join(__dirname, `../../public/icons`), {
+    withFileTypes: true,
+  })
+    .filter((x) => x.isFile() && x.name.endsWith('.png'))
+    .map((x) => x.name.slice(0, -4))
+    .sort()
+
+  it('every project has an icon', async () => {
+    const projects = await ProjectService.STATIC.getProjects({})
```
torztomasz, line 16: mb we should make the `getProjects` options optional?

torztomasz (approved):

packages/frontend/scripts/logos/tinifiyLogos.test.ts
sz-piotr: You almost never want to call it like this.

PR #6328, opened 2025-01-10
Refactor scaling da entries to use ProjectService

sz-piotr:


packages/config/src/projects/project/Project.ts
```diff
@@ -23,6 +24,7 @@ export interface Project {
   statuses?: ProjectStatuses
   scalingInfo?: ProjectScalingInfo
   scalingRisks?: ProjectScalingRisks
+  scalingDa?: DataAvailabilityWithSentiment
```
torztomasz, line 27: mb `ProjectDataAvailability` to keep the naming convention

torztomasz (approved):

PR #6350, opened 2025-01-14
Move ZK Catalog to side navigation

torztomasz:


packages/frontend/src/app/(side-nav)/zk-catalog/_utils/get-zk-catalog-entries.ts
```diff
@@ -0,0 +1,24 @@
+import { type ProjectWith } from '@l2beat/config'
+import { type VerifiersStatuses } from '~/server/features/zk-catalog/get-verifiers'
+import { getProofVerification } from './get-proof-verification'
+import { getTrustedSetup } from './get-trusted-setup'
+
+export type ZkCatalogEntry = ReturnType<typeof getZkCatalogEntries>[number]
```
sz-piotr, line 6: I'd prefer an explicit type here

sz-piotr (approved):

PR #6357, opened 2025-01-15
Add Saakuru as upcomingL2

ian9403:


radinat:
Seems like Saakuru is going to be a layer 2 on Oasys, not Ethereum. For now, L2BEAT only focuses on chains built on top of Ethereum.

PR #6369, opened 2025-01-15
Remove sentry

torztomasz:


Yutaro-Mori-eng:
hi @torztomasz 
I noticed that Sentry has been removed. Could you share the reasons for this decision? Is it because monitoring is no longer needed, or has it been replaced with a better alternative?

torztomasz:
> hi @torztomasz I noticed that Sentry has been removed. Could you share the reasons for this decision? Is it because monitoring is no longer needed, or has it been replaced with a better alternative?

It has not provided any useful info, 99% of errors are basically a noise from user extensions. Observability is of course needed but unfortunately I have not found better alternative yet.

Yutaro-Mori-eng:
> > hi @torztomasz I noticed that Sentry has been removed. Could you share the reasons for this decision? Is it because monitoring is no longer needed, or has it been replaced with a better alternative?
> 
> It has not provided any useful info, 99% of errors are basically a noise from user extensions. Observability is of course needed but unfortunately I have not found better alternative yet.

Thanks for your response! My experience has been mostly in Web2, where Sentry is very useful for detecting and reproducing errors. However, in the Web3 space, Sentry often gets overwhelmed with errors, and as you mentioned, 99% of them are just noise. I’m now considering how to use Sentry more effectively—or maybe even dropping it altogether. I’ll keep an eye on what you’re doing to see if there’s a better approach. Thanks again!

torztomasz:
> > > hi @torztomasz I noticed that Sentry has been removed. Could you share the reasons for this decision? Is it because monitoring is no longer needed, or has it been replaced with a better alternative?
> > 
> > 
> > It has not provided any useful info, 99% of errors are basically a noise from user extensions. Observability is of course needed but unfortunately I have not found better alternative yet.
> 
> Thanks for your response! My experience has been mostly in Web2, where Sentry is very useful for detecting and reproducing errors. However, in the Web3 space, Sentry often gets overwhelmed with errors, and as you mentioned, 99% of them are just noise. I’m now considering how to use Sentry more effectively—or maybe even dropping it altogether. I’ll keep an eye on what you’re doing to see if there’s a better approach. Thanks again!

If you will find something to recommend, feel free to text me 😅

maciekop-l2b (approved):

PR #6379, opened 2025-01-15
Morph forcetxs (queue)

sekuba:
There is no mechanism that enforces regular dequeuing, but skipping txs is disabled (yellow?)

closes L2B-8787


lucadonnoh (changes requested):
seq fail should probably stay red as the enforced tx gateway is paused

lucadonnoh (approved):

PR #6396, opened 2025-01-16
Use checkbox instead link for bridge selector

torztomasz:


sz-piotr (approved):

sz-piotr (changes requested):
I click the button and there is no async action indicator. Nothing happens and then the ui just changes

sz-piotr (approved):

PR #6398, opened 2025-01-16
refactor: format number

Holybasil:
Also, I’m a bit curious why the rest decimals are being directly discarded [here](https://github.com/l2beat/l2beat/blob/main/packages/frontend/src/utils/number-format/format-number.ts#L32) instead of using rounding.


mateuszradomski:
Hi,

Thank you for your contribution but I don't see what this improves. There is less code, but the code that is there is now harder to understand and modify if we ever decide to do anything else.

PR #6404, opened 2025-01-17
[TVL4] Implement LocalStorage that can be persisted to local json file

maciekzygmunt:
Resolves L2B-8809

### DIFF
--- a/packages/backend/src/modules/tvs/types.ts
+++ b/packages/backend/src/modules/tvs/types.ts
@@ -94,3 +94,8 @@ export interface TokenValue {
   projectId: ProjectId
   valueUsd: number
 }
+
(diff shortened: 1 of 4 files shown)


packages/backend/src/modules/tvs/types.ts
```diff
@@ -94,3 +94,8 @@ export interface TokenValue {
   projectId: ProjectId
   valueUsd: number
 }
+
+export interface LocalStorageJSON {
```
maciekop-l2b, line 98: Let's move this type to `LocalStorage.ts` as it is implementation detail that is not used anywhere else

maciekop-l2b (changes requested):

PR #6419, opened 2025-01-17
Add public blockchains DA throughput 

vincfurc:
resolves L2B-8745
resolves L2B-8650


vincfurc:
ethereum shows KB/s? @sdlyy

sdlyy:
> ethereum shows KB/s? @sdlyy
@vincfurc 

ah yeah, we try to 'expand' the unit as much as possible but preserve a pretty format, without too many digits in place - I guess we would like to always expand it towards MB/s?

lucadonnoh (approved):

PR #6428, opened 2025-01-20
[L2B-8805] Add support for totalSupply and circulatingSupply tokens

maciekop-l2b:


packages/backend/src/modules/tvs/mapConfig.ts
```diff
@@ -29,77 +40,128 @@ export function mapConfig(project: BackendProject, chain: string): TvsConfig {
...
+        chain.minTimestampForTvl,
+        'Chain with token should have minTimestamp',
+      )
+
       amountFormula = {
         type: 'circulatingSupply',
         ticker: legacyToken.symbol,
       } as CirculatingSupplyAmountFormula
+
+      sinceTimestamp = UnixTime.max(
```
antooni, line 161: IMO this should be fixed "higher" in the abstraction chain, token should not be allowed to be earlier than the chain. 
I think somewhere before this mapConfig function we will have function e.g. mergeConfigs(chains,tokens) which will handle this one.
I propose we emulate this here already, so we do not forget in the future. Because setting those min and max timestamps for configs on chain is one of the biggest pinpoints of current infra

antooni (approved):

PR #6435, opened 2025-01-20
Add tracking to UI components

torztomasz:


packages/frontend/src/hooks/use-custom-event.tsx
```diff
@@ -0,0 +1,15 @@
...
+type MyEvents = {
+  filterChanged: { name: string; value: string }
+  checkboxChanged: { name: string; value: string }
+  radioGroupChanged: { name: string; value: string }
+  directoryTabsChanged: { value: string }
+  searchBarProjectSelected: { name: string }
+  uopsExplorerSelected: never
+}
+
+export function useCustomEvent() {
```
sz-piotr, line 12:
```suggestion
export function useTrackingEvent() {
```

packages/frontend/src/components/core/radio-group.tsx
```diff
@@ -2,24 +2,43 @@
...
+      name={name}
       className={cn(
         'group/radio-group inline-flex h-8 w-max items-center gap-1 rounded-lg p-1 font-medium',
         'bg-surface-primary primary-card:bg-surface-secondary',
         className,
       )}
       {...props}
+      onValueChange={(val) => {
+        onValueChange?.(val)
+        if (enableTracking) {
```
sz-piotr, line 33: Why here? Maybe the hook should check? Or maybe a tracking event id should be passed and only if passed this would be called.

sz-piotr (approved):

PR #6449, opened 2025-01-21
Convert throughput to MB/s

sdlyy:


packages/frontend/src/utils/number-format/format-throughput.test.ts
```diff
@@ -1,45 +1,36 @@
 import { expect } from 'earl'
-import { type SizeUnit, formatThroughput } from './format-throughput'
+import { formatThroughput } from './format-throughput'
 
 describe('formatThroughput', () => {
   it('formats basic throughput correctly', () => {
-    expect(formatThroughput(100, 1)).toEqual('100 B/s')
-    expect(formatThroughput(1000, 1)).toEqual('1 KB/s')
-    expect(formatThroughput(1_000_000, 1)).toEqual('1 MB/s')
+    expect(formatThroughput(1_000, 1)).toEqual('1 MB/s')
```
mateuszradomski, line 6: What about GB/s? I'd test for value that is in the GB/s range to show the reader that GB/s are formatted as MB/s with thousands. If we always return MB/s from this function rename it to formatThroughputMBPS or something like that

mateuszradomski (changes requested):

packages/frontend/src/utils/number-format/format-throughput.test.ts
sdlyy: we always kick off with KB and want MB so here is the rename:
https://github.com/l2beat/l2beat/pull/6449/commits/3f15626bb2bafd3a729e6764eda8abde03933bd4

mateuszradomski (approved):

PR #6466, opened 2025-01-22
Update fallback error message in PriceProvider

antooni:
We are missing some context in Elastic, this PR aims to improve it
image

### DIFF
--- a/packages/backend/src/modules/tvl/services/PriceService.ts
+++ b/packages/backend/src/modules/tvl/services/PriceService.ts
@@ -20,7 +20,11 @@ export interface PriceServiceDependencies {
 }
 
 export class PriceService {
-  constructor(private readonly $: PriceServiceDependencies) {}
+  logger: Logger
+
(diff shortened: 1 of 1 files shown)


maciekzygmunt (approved):

packages/backend/src/modules/tvl/services/PriceService.ts
```diff
@@ -96,8 +100,8 @@ export class PriceService {
       `Latest price not found for ${coingeckoId} @ ${latestHour.toNumber()}`,
     )
 
-    this.$.logger.error(
-      'DB fallback triggered: failed to fetch price from Coingecko',
+    this.logger.error(
```
maciekop-l2b, line 103: let's make it a warning as we can automatically handle it without breaking the flow

maciekop-l2b (changes requested):

maciekop-l2b (approved):

PR #6468, opened 2025-01-22
Remove `activityDataSource`

maciekzygmunt:
Resolves L2B-8917


torztomasz:
Now there is too much projects in activity because of wrong condition for `getProjects` select property

torztomasz (changes requested):
As above

torztomasz (approved):

PR #6470, opened 2025-01-22
Discovery updates 01/22

sekuba:
closes L2B-8900

re Scroll: makes sense to discodrive now, will start on that soon, also add a upgrades&gov section


lucadonnoh (changes requested):
can you fix the L1USDCBridge filename

sekuba:
fixdededed

lucadonnoh (approved):

PR #6505, opened 2025-01-24
Refactor da-beat config

sz-piotr:


packages/config/src/projects/da-beat/index.ts
```diff
@@ -1,13 +1,15 @@
...
+import { espressoDA } from './espressoDA/espressoDA'
+import { ethereum } from './ethereum/ethereum'
+import { memo } from './memo/memo'
+import { near } from './near/near'
 
 export * from './types'
-export * from './utils/to-used-in-project'
+// TODO: do not export from common and usedInProject
+export { DaCommitteeSecurityRisk } from './common'
+export { toUsedInProject } from './utils/to-used-in-project'
```
sdlyy, line 12: Something wrong here?

sdlyy (approved):
🙇🏻

packages/config/src/projects/da-beat/index.ts
sz-piotr: It is about the future direction. I'd like the config package to ideally end up only exporting `ProjectService` and types. We'll see how close I can get.

PR #6513, opened 2025-01-24
Change Per L2 tx to Per L2 User Op

maciekzygmunt:
Resolvse L2B-8886

### DIFF
--- a/packages/database/src/activity/repository.ts
+++ b/packages/database/src/activity/repository.ts
@@ -146,17 +146,19 @@ export class ActivityRepository extends BaseRepository {
     )
   }
 
-  async getSummedCountForProjectsAndTimeRange(
+  async getSummedUopsCountForProjectsAndTimeRange(
     projectIds: ProjectId[],
     timeRange: [UnixTime, UnixTime],
-  ): Promise<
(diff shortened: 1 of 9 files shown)


packages/database/src/activity/repository.ts
```diff
@@ -146,17 +146,19 @@ export class ActivityRepository extends BaseRepository {
...
+  ): Promise<{ projectId: ProjectId; uopsCount: number }[]> {
     const [from, to] = timeRange
     const rows = await this.db
       .selectFrom('Activity')
       .select(['projectId'])
-      .select((eb) => eb.fn.sum('count').as('count'))
+      .select((eb) =>
+        eb.fn
+          .sum(eb.fn.coalesce('Activity.uopsCount', 'Activity.count'))
+          .as('count'),
```
torztomasz, line 160: why are you selectin count if we want uopsCount

torztomasz (approved):

packages/database/src/activity/repository.ts
maciekzygmunt: Because some projects do not have `uopsCount` or have it as null until some timestamp. So we want to take `count` if that is the case which is always present

PR #6516, opened 2025-01-24
add Arena-Z (full)

sekuba:
unsure about this one
image
can it be set to true (due to only SC being the upgrader) while the operator could censor (bc of permissioned games)?


lucadonnoh (approved):
it’s good like this

PR #6538, opened 2025-01-27
Upgrade to tailwindcss v4

torztomasz:


torztomasz:
apparently biome doesnt support new tailwind syntax in css files :(

PR #6558, opened 2025-01-28
Add basic Ethereum DA indexing infra

sdlyy:
Resolves L2B-8865


packages/backend/src/modules/data-availability/indexers/BlockTargetIndexer.ts
```diff
@@ -0,0 +1,48 @@
+import type { Logger } from '@l2beat/backend-tools'
+import { assert, type ProjectId } from '@l2beat/shared-pure'
+import { Indexer, RootIndexer } from '@l2beat/uif'
+import type { Clock } from '../../../tools/Clock'
+import type { BlockTimestampProvider } from '../../tvl/services/BlockTimestampProvider'
+
+export class BlockTargetIndexer extends RootIndexer {
```
maciekop-l2b, line 7: If this is 1:1 copy of `BlockTargetIndexer` please create 'shared' directory under 'modules', move it there and reference in activity and data-availability. We will quickly run into problems with current setup as this copy is already outdated comparing to source

packages/shared/src/clients/blobscan/BlobscanClient.ts
```diff
@@ -0,0 +1,150 @@
...
+      },
+      ...options,
+    })
+  }
+
+  override validateResponse(_response: json): {
+    success: boolean
+    message?: string
+  } {
+    // TODO: Implement
```
maciekop-l2b, line 147: Let's implement some simple validation here

packages/backend/src/modules/data-availability/providers/BlobscanDaProvider.ts
```diff
@@ -0,0 +1,42 @@
+import type { BlobScanClient, BlobSizeData, DaProvider } from '@l2beat/shared'
+import { UnixTime } from '@l2beat/shared-pure'
+
+// each blob is 128 KiB so 131,072 B
+const BLOB_SIZE_BYTES = 131072n
+
+export class BlobScanDaProvider implements DaProvider {
```
maciekop-l2b, line 7: This implementation should be in `DaProvider` in shared. We should also consider if this could be a part of `BlobProvider` as it deals with blobs but not sure if we can have a common result type ...

maciekop-l2b (changes requested):

packages/shared/src/clients/blobscan/BlobscanClient.ts
sdlyy: https://github.com/l2beat/l2beat/pull/6558/commits/1f5ee829d8668584f7cccad54b75fc525223cf5d

maciekop-l2b (approved):

PR #6565, opened 2025-01-28
Add capability and notice properties

torztomasz:


packages/config/src/projects/layer2s/templates/opStack.ts
```diff
@@ -235,6 +235,7 @@ function opStackCommon(
   return {
     isArchived: templateVars.isArchived,
     id: ProjectId(templateVars.discovery.projectName),
+    capability: 'universal',
```
sz-piotr, line 238: I think we need to allow overrides in every template

torztomasz: ofc, my bad

sz-piotr (approved):

PR #6589, opened 2025-01-29
Update withdrawals wording to messaging

lucadonnoh:
Resolves L2B-8989. In the withdrawals section, chains that have an AMB now use the messaging wording, while chains that don't use the old withdrawals centric wording. This is a temporary solution (so don't expect it to be perfect) as we're likely to do some bigger refactoring of these sections


sekuba (approved):
assuming that the yeetage of merkle proof string is intended, lgtmmm

PR #6591, opened 2025-01-29
Use TableReadyValue everywhere and unify frontend handling

sz-piotr:


packages/config/src/projects/da-beat/common/DaCommitteeSecurityRisk.ts
```diff
@@ -107,6 +118,33 @@ function getDacSentiment(params: {
   return 'warning'
 }
 
+/**
+ * This function approximates the probability of a compromise in a DAC.
+ * With the assumption that the probability of a single member being compromised is 1%.
+ */
+export function probabilityOfCompromise(required: number, total: number) {
```
mateuszradomski, line 125: I'd test that the numbers grow/shrink as we assume. Plus I'd test if the number doesn't breakdown if we pass a number bigger than 100/1000 since we have factorial

mateuszradomski (approved):

PR #6594, opened 2025-01-29
FE: tvl -> tvs

maciekzygmunt:
Resolves L2B-8715


torztomasz (approved):
The only thing I have seen not renamed is env variable `EXCLUDED_TVL_PROJECTS`

PR #6596, opened 2025-01-29
add Sonic Gateway (full review)

sekuba:
closes L2B-8969


packages/config/src/projects/bridges/sonicgateway.ts
prototypo, line 1: Shouldn't the S token be included in the token list? https://docs.soniclabs.com/sonic/s-token

prototypo, line 1: 3/8 validators can violate liveness.

prototypo (commented):
One question and one additional risk comment added.

packages/config/src/projects/bridges/sonicgateway.ts
sekuba: no, you can read in the principle of operation section how the conversion works. on the ethereum side, there is only FTM

sekuba: yes, we have this risk here
image
do you think it suffices? ultimately this bridge cannot steal with liveness failures because of the 'dead' system, so i did not add a different risk compared to other bridges.

in general i think liveness should have a place in our bridges assessment btw (currently it hasn't). and it can be nicely combined with censorship resistance and metrics like average, best, worst case bridging speed

prototypo: Ah, yes. Thanks!

luke7211 (approved):

PR #6611, opened 2025-01-30
Remove Slashable column from Custom DAs

maciekzygmunt:
Resolves L2B-9039


packages/frontend/src/app/(side-nav)/data-availability/summary/_components/table/columns.tsx
```diff
@@ -144,7 +144,6 @@ export const customColumns = [
   membersColumn,
   fallbackColumn,
   challengeMechanismColumn,
-  slashableStakeColumn,
```
torztomasz, line 147: shouldn't we also remove the data from entries?

torztomasz (changes requested):

torztomasz (approved):

PR #6616, opened 2025-01-30
Split roles and actors into different fields

mateuszradomski:
Part of L2B-8704


packages/frontend/src/components/projects/sections/permissions/multichain-permissions-section.tsx
```diff
@@ -31,24 +34,42 @@ export function MultichainPermissionsSection({
 
       <div className="my-4">
         {Object.entries(permissions).map(([chainName, permissions]) => {
-          if (permissions.length === 0) {
-            return null
+          if (permissions.roles.length > 0) {
+            return (
+              <div key={chainName}>
+                <h3 className="font-bold">
+                  The system consists of the following permissions on{' '}
```
sz-piotr, line 41: copy seems wrong

sz-piotr (approved):

PR #6640, opened 2025-01-31
Discodrive polygon cdk stack

sekuba:
closes L2B-8422


lucadonnoh (changes requested):
lgtm but let's ignore permit2 as it is not part of polygon stuff

lucadonnoh (approved):

PR #6670, opened 2025-02-04
Verification status for permissions comes from config

mateuszradomski:
Part of L2B-8704


packages/frontend/src/components/projects/sections/contract-entry.tsx
```diff
@@ -166,12 +166,7 @@ function getCalloutProps(
   const isAnyAddressUnverified = contract.addresses.some(
     (c) => c.verificationStatus === 'unverified' && !c.isAdmin,
```
sz-piotr, line 167: but why !c.isAdmin?

packages/frontend/src/components/projects/sections/permissions/participants.tsx
```diff
@@ -21,20 +16,9 @@ export function ParticipantsEntry({
       {participants.map((address, i) => (
         <HighlightableLink
           key={i}
-          variant={
-            address.verificationStatus === 'unverified' ? 'danger' : undefined
-          }
           href={address.href}
           className="flex items-center gap-0.5"
         >
-          {address.verificationStatus === 'unverified' ? (
```
sz-piotr, line 30: Not needed anymore?

mateuszradomski: Talked with basti, he said that showing if a participant is unverified doesn't make sense

sz-piotr (approved):

packages/frontend/src/components/projects/sections/contract-entry.tsx
mateuszradomski: Removed that flag, replaced with admins array that's simpler and does the same thing

PR #6673, opened 2025-02-04
Remove stories

mateuszradomski:
Since we don't use chromatic they are useless. Also removed dependencies connected with them.


maciekzygmunt (approved):

sz-piotr (changes requested):
wait

sz-piotr (approved):
good to go, sorry

torztomasz (approved):

PR #6676, opened 2025-02-04
Move permission name/url resolution to config

mateuszradomski:
Part of L2B-8704

### DIFF
--- a/packages/config/src/chains/explorerUrls.ts
+++ b/packages/config/src/chains/explorerUrls.ts
@@ -14,7 +14,7 @@
 // In the third pass we'll understand what projects are available and we can
 // inline the explorer url at that time.
 
-export const ExplorerUrlMap: Record<string, string> = {
+export const EXPLORER_URLS: Record<string, string> = {
   ethereum: 'https://etherscan.io',
   zksync2: 'https://era.zksync.network',
   arbitrum: 'https://arbiscan.io',
--- a/packages/config/src/chains/explorerUrls.ts
+++ b/packages/config/src/chains/explorerUrls.ts
@@ -1,0 +1,29 @@
+// NOTE(radomski): This should be removed. It's only here because config is in
+// a transitional period where the way it works is not yet where we would want
+// it to be. In ProjectDiscovery we need to have access to the url of the
+// explorer for different chains but we can't access layer2s/layer3s/bridges
+// since entries of these arrays are being built by ProjectDiscovery.
+//
+// The ideal solution that config will employ is a three pass approach. 
+//
+// - The first pass reads in the data required for operation.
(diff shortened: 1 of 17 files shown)


packages/config/src/chains/explorerUrls.ts
```diff
@@ -0,0 +1,29 @@
...
+//
+// - The first pass reads in the data required for operation.
+// - The second pass builds each project
+// - The third pass links up projects between each other, e.g. DA Bridges used
+// by projects.
+//
+// In the third pass we'll understand what projects are available and we can
+// inline the explorer url at that time.
+
+export const ExplorerUrlMap: Record<string, string> = {
```
sz-piotr, line 17:
```suggestion
export const EXPLORER_URLS: Record<string, string> = {
```

packages/config/src/discovery/ProjectDiscovery.ts
```diff
@@ -1356,3 +1362,32 @@ function isEntryVerified(entry: ContractParameters | EoaParameters): boolean {
...
+    actorNameLUT[actor.accounts[0].address] = actor.name
+  }
+
+  for (const account of accountsToLink) {
+    const entry = structuredClone(account)
+
+    const actorName = actorNameLUT[account.address]
+    if (actorName !== undefined) {
+      entry.name = actorName
+      entry.url = `#${actorName}`
```
sz-piotr, line 1387: are we sure that
a) actor names are properly formatted for urls (e.g. no spaces)
b) are unique across actors
?

packages/config/src/test/stubs/layer2aWithDups.ts
```diff
@@ -343,17 +361,19 @@ export const layer2aWithDups: Layer2 = {
...
           ),
+          chain: 'ethereum',
           isVerified: true,
         },
       ],
     },
     risks: [],
   },
-}
+} as unknown as Layer2
```
sz-piotr, line 379: why?

sz-piotr (approved):

adamiak (approved):

packages/config/src/test/stubs/layer2aWithDups.ts
mateuszradomski: This object is used only for testing if we are able to extract different chains from permissions/chains. I don't want to change this big object every time I do a small change with data that is never touched

packages/config/src/discovery/ProjectDiscovery.ts
mateuszradomski: As for (b), I've added an assertion to make sure that it's true. As for (a), they are not, we have spaces but does it matter?

PR #6686, opened 2025-02-05
Orbit tradeable exit section

vincfurc:
resolves L2B-9034

I wouldn't remove the whole section


sekuba:
yoooooo but we agreed to delete no?

sekuba (approved):

PR #6690, opened 2025-02-05
Simplify contracts & permissions

sz-piotr:


packages/config/src/projects/layer2s/layer2finance.ts
```diff
@@ -66,11 +65,5 @@ export const layer2finance: Layer2 = {
         isCritical: true,
       },
     ],
-    references: [
```
mateuszradomski, line 69: Move this reference to the `discovery.getContractDetails('RollupChain')`

packages/config/src/projects/layer2s/zkfair.ts
```diff
@@ -343,13 +343,6 @@ export const zkfair: Layer2 = {
         }),
       ],
     },
-    references: [
```
mateuszradomski, line 346: Move this reference to the `discovery.getContractDetails('ZKFairValidium', {})`

mateuszradomski (approved):

PR #6703, opened 2025-02-05
Discodrive Arbitrum

vincfurc:
Part of L2B-8452


vincfurc:
reopened at https://github.com/l2beat/l2beat/pull/6729

PR #6715, opened 2025-02-06
[L2B-9142] Implement DaService

maciekop-l2b:


packages/shared/src/providers/da/DaProvider.ts
```diff
@@ -1,10 +1,28 @@
 import type { UnixTime } from '@l2beat/shared-pure'
 
-export type DaBlob = {
+export type DaBlobBase = {
   blockTimestamp: UnixTime
   size: bigint
 }
 
+export type EthereumBlob = DaBlobBase & {
```
antooni, line 8: what about interface & extend? we are using this pattern in other parts of code e.g. AmountConfigEntry.ts

packages/backend/src/modules/data-availability/services/DaService.ts
```diff
@@ -0,0 +1,122 @@
...
+  constructor(
+    private readonly projectConfigs: {
+      id: ProjectId
+      config: ProjectDaTrackingConfig
+    }[],
+  ) {}
+
+  processBlobs(
+    blobs: DaBlob[],
+    records: DataAvailabilityRecord[],
```
antooni, line 21: pass a Map<project<Map<timestamp, record>> here instead of array

maciekop-l2b: tried with Map od Maps and Map with composite key but the code becomes messy ...

packages/backend/src/modules/data-availability/services/DaService.ts
```diff
@@ -0,0 +1,122 @@
...
+
+export class DaService {
+  constructor(
+    private readonly projectConfigs: {
+      id: ProjectId
+      config: ProjectDaTrackingConfig
+    }[],
+  ) {}
+
+  processBlobs(
```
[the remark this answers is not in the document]
maciekop-l2b: Renamed to: `generateRecords`

antooni (approved):

PR #6720, opened 2025-02-07
Add Spheron as upcomingL3

akitothemoon:


akitothemoon:
> @akitothemoon is there an announcement or another source that makes it clear that they're working on a Layer3 solution? Couldn't find any

Hi, yes, there are relevant explanations in the documentation.
https://docs.spheron.network/user-guide/protocol-cli#escrow-account-overview

image
image

radinat (approved):

PR #6722, opened 2025-02-07
feat(chain) move form chain from upcoming

tmarwen:
- Update Form chain as opStackL2 instead of upcomingL2
- Add Form chain project with available contracts


mateuszradomski:
Hi,

Sorry, but this job needs to be done by researchers.

tmarwen:
@mateuszradomski what should be the process to go through the `Form Network` chain promotion?

PR #6727, opened 2025-02-10
Make da bridges into projects

sz-piotr:


packages/config/src/projects/da-beat/avail/avail.ts
```diff
@@ -21,21 +17,15 @@ export const avail: BaseProject = {
...
       'Avail is a public blockchain and data availability network combining erasure coding, KZG polynomial commitments, and data availability sampling.',
     links: {
       websites: ['https://www.availproject.org/'],
-      documentation: [
-        'https://docs.availproject.org/',
-        'https://docs.succinct.xyz/',
-      ],
-      repositories: [
-        'https://github.com/availproject/',
-        'https://github.com/succinctlabs/sp1-vector',
```
torztomasz, line 37: why removing it?

torztomasz (approved):
I don't know if its intended but "Bridge" column for Ethereum entry in DA summary table has changed to "Ethereum" from "Enshrined bridge". Otherwise LGTM

packages/config/src/projects/da-beat/avail/avail.ts
sz-piotr: Not removed, moved to vector

mateuszradomski (approved):

PR #6772, opened 2025-02-13
Discodrive Nova

vincfurc:
Resolves L2B-9211


sekuba (changes requested):
amazing, but i shall humbly request to change the l2 `config.jsonc` or template for l2baseFee and l2surplusfee contracts, specifically their `recipients` array. Either define a permission like 'receive fees' or "ignorerelatives": ["recipients"] to make the two spammy multisigs disappear

this will also affect arbitrum i guess but that is good imo

sekuba (approved):

PR #6797, opened 2025-02-14
update events in governance page

L2BeatSinkas:


torztomasz (changes requested):
I understand removing non existing recurring meetings but why would you remove past events?

L2BeatSinkas:
> I understand removing non existing recurring meetings but why would you remove past events?

They're not visible on the website anyway. Why keep them?

torztomasz:
I think @kurbanski wanted to keep the history, in case we want to present what the gov team has done in the past.

PR #6815, opened 2025-02-17
Eigen batch verification

sdlyy:
Resolves L2B-9237

### DIFF
--- a/packages/discovery/src/discovery/handlers/user/OpDAHandler.ts
+++ b/packages/discovery/src/discovery/handlers/user/OpDAHandler.ts
@@ -1,6 +1,7 @@
-import type { EthereumAddress } from '@l2beat/shared-pure'
-import * as z from 'zod'
+import { assert, EthereumAddress, UnixTime } from '@l2beat/shared-pure'
 
+import { AbiCoder } from 'ethers/lib/utils'
+import { z } from 'zod'
 import type { IProvider } from '../../provider/IProvider'
 import type { Handler, HandlerResult } from '../Handler'
 import {
@@ -16,6 +17,7 @@ export type OpStackDAHandlerDefinition = z.infer<
 export const OpStackDAHandlerDefinition = z.strictObject({
   type: z.literal('opStackDA'),
   sequencerAddress: z.string(),
+  inboxAddress: z.string(),
 })
 
 /**
@@ -58,9 +60,13 @@ export class OpStackDAHandler implements Handler {
     readonly field: string,
     readonly definition: OpStackDAHandlerDefinition,
   ) {
-    const dependency = getReferencedName(this.definition.sequencerAddress)
-    if (dependency) {
-      this.dependencies.push(dependency)
+    const dependencies = [
+      getReferencedName(this.definition.sequencerAddress),
+      getReferencedName(this.definition.inboxAddress),
+    ].filter((d) => d !== undefined)
+
+    if (dependencies.length > 0) {
+      this.dependencies.push(...dependencies)
     }
   }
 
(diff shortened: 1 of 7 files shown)


packages/discovery/src/discovery/handlers/user/OpDAHandler.ts
```diff
@@ -126,4 +137,278 @@ export class OpStackDAHandler implements Handler {
...
+  // 1150 is arbitrary, we treat anything longer than that as extended
+  if (hexData.length > 1150) {
+    return parseExtendedCommitment(hexData)
+  }
+
+  return parseStandardCommitment(hexData)
+}
+
+function decodeBatchConfirmation(inputData: string): string | undefined {
+  if (!inputData.startsWith('0x7794965a')) {
```
mateuszradomski, line 394: Do we know what the function is? If yes we can just inline it, look at ArbitrumSequencerVersionHandler and see how functions/selectors are defined there

packages/discovery/src/discovery/handlers/user/OpDAHandler.ts
```diff
@@ -126,4 +137,278 @@ export class OpStackDAHandler implements Handler {
...
+    return
+  }
+
+  // Remove '0x' and function selector
+  const data = inputData.slice(10)
+  const abiCoder = new AbiCoder()
+
+  const batchHeaderType =
+    'tuple(bytes32 blobHeadersRoot, bytes quorumNumbers, bytes signedStakeForQuorums, uint32 referenceBlockNumber)'
+  const nonSignerStakesType =
```
mateuszradomski, line 405: What is this, it seems like it's never used. because we extract blobHeadersRoot and it's not present in this type

packages/discovery/src/discovery/handlers/user/OpDAHandler.ts
```diff
@@ -44,8 +47,17 @@ const BLOB_TX_TYPE = 3
  * automata: 	0x01 00 a5 		(challenges)
  * syndicate: 0x01 00 9c		(keccak256)
  */
-const EIGEN_DA_COMMITMENT_PREFIX = '0x01'
-const EIGEN_DA_COMMITMENT_THIRD_BYTE = '00'
+const EIGEN_DA_CONSTANTS = {
+  COMMITMENT_PREFIX: '0x01',
+  COMMITMENT_THIRD_BYTE: '00',
+  VERIFICATION_THRESHOLD: 0.8, // 80% success rate required
```
mateuszradomski, line 53: 80% seems kinda low, no? Why can the check fail?

packages/discovery/src/discovery/handlers/user/OpDAHandler.ts
```diff
@@ -16,6 +18,7 @@ export type OpStackDAHandlerDefinition = z.infer<
 export const OpStackDAHandlerDefinition = z.strictObject({
   type: z.literal('opStackDA'),
   sequencerAddress: z.string(),
+  inboxAddress: z.string(),
```
mateuszradomski, line 21: I don't get it, why do we need the `inboxAddress`? Isn't sequencer posting data to the inbox? Isn't it true that all transactions that are outgoing from the sequencer are the same as the incoming transactions for the inbox?

packages/discovery/src/discovery/handlers/user/OpDAHandler.ts
```diff
@@ -126,4 +137,278 @@ export class OpStackDAHandler implements Handler {
...
+    timestamp: latestBlock.timestamp,
+  }
+
+  const confirmBatchTxs = await getBatchConfirmTransactions(
+    provider,
+    fromBlock,
+    toBlock,
+  )
+
+  const inboxTxs = await provider.raw(
```
mateuszradomski, line 215: inboxTxs are the same as the sequencerTxs, if you print them out you'll see that 

image

mateuszradomski (changes requested):
I found a better way to check that, I've talked with Vincenzo and agreed to use the newer method

packages/discovery/src/discovery/handlers/user/OpDAHandler.ts
sdlyy: That part has been removed

packages/discovery/src/discovery/handlers/user/OpDAHandler.ts
sdlyy: wiped as well

packages/discovery/src/discovery/handlers/user/OpDAHandler.ts
sdlyy: removed completely

packages/discovery/src/discovery/handlers/user/OpDAHandler.ts
sdlyy: fixed as discussed, no more inbox transactions

PR #6848, opened 2025-02-19
Add invalidation support to DaIndexer

antooni:


packages/database/scripts/tvl-restore/tvl-restore.sh
```diff
@@ -1,19 +1,95 @@
 #!/bin/bash
-source .env &&
```
maciekop-l2b, line 2: let's rename it to `db-restore.sh`

packages/backend/src/config/features/da.ts
```diff
@@ -49,7 +49,7 @@ export async function getDaTrackingConfig(
         20_000,
       ),
       batchSize: env.integer('CELESTIA_BLOBS_BATCH_SIZE', 100),
-      startingBlock: 1,
+      startingBlock: 983042,
```
maciekop-l2b, line 52: We should do it as a separate PR

maciekop-l2b (changes requested):

maciekop-l2b (approved):

PR #6854, opened 2025-02-19
Untracked gas tokens

sdlyy:
Resolves L2B-8856


mateuszradomski (approved):
Write a test that checks if tokens in `tracked` are actually in `tokens.jsonc`

PR #6878, opened 2025-02-21
[L2B-9343] Add support for new method in FinalityIndexer:arbitrum

maciekop-l2b:


.gitignore
```diff
@@ -29,3 +29,4 @@ packages/backend/src/modules/tvs/local-data-test.json
 packages/backend/src/modules/tvs/local-data.json
 packages/backend/src/token-config.json
 packages/backend/src/token-breakdown.json
+packages/database/tvl.pgdump
```
antooni, line 32: Isn't it located inside scripts folder? Also, lets rename it to `db.pgdump` or something more generic hence it is no longer a tvl script

maciekop-l2b: Fixed, the file is removed after restore is completed so no need for gitignore

antooni (approved):

PR #6884, opened 2025-02-21
Use ProjectService for trackedTxs

sz-piotr:


packages/backend/scripts/starkwareMultiplier.ts
```diff
@@ -32,7 +32,7 @@ const cmd = command({
   name: 'starkware-multiplier',
   args,
   handler: async (args) => {
-    const env = getEnv()
+    const env = getEnv({ FEATURES: 'tracked-txs' })
```
antooni, line 35: is this change intended?

antooni (approved):

PR #6891, opened 2025-02-22
fix: Deduplicate project links in DA project view

Daljeet819:
## Problem
The Data Availability (DA) project view currently shows duplicate links because the same links are being combined from both layer and bridges sources. This occurs in `getProjectLinks` where we merge links from both the DA layer and all bridge configurations.

image

### DIFF
--- a/packages/frontend/src/server/features/data-availability/project/get-da-project-entry.ts
+++ b/packages/frontend/src/server/features/data-availability/project/get-da-project-entry.ts
@@ -170,7 +170,7 @@ export async function getDaProjectEntry(
     header: {
       links: getProjectLinks(
         layer.display.links,
-        ...bridges.map((x) => x.display.links),
+        ...bridges.filter((x) => x.id !== layer.id).map((x) => x.display.links),
       ),
       daLayerGrissiniValues: layerGrissiniValues,
       daBridgeGrissiniValues: bridgeGrissiniValues,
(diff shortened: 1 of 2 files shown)


torztomasz (changes requested):
Hey. Thanks for contributing! This solution fixes the issue but it seems that you are solving [problem Y](https://xyproblem.info/). The problem X is actually not proper links being passed to the function you refactored. Proper solution would be to change the line `173` in `packages/frontend/src/server/features/data-availability/project/get-da-project-entry.ts` to filter out bridges which project has the same id as layer: `...bridges.filter(x => x.id !== layer.id).map((x) => x.display.links)`.

Daljeet819:
> Hey. Thanks for contributing! This solution fixes the issue but it seems that you are solving [problem Y](https://xyproblem.info/). The problem X is actually not proper links being passed to the function you refactored. Proper solution would be to change the line `173` in `packages/frontend/src/server/features/data-availability/project/get-da-project-entry.ts` to filter out bridges which project has the same id as layer: `...bridges.filter(x => x.id !== layer.id).map((x) => x.display.links)`.
> 
> PS. do not forget to run linter and formatter at the end to make the CI pass

GM @torztomasz Thank you for the review! I've updated the code according to your suggestion to filter out bridges with the same ID as the layer. 

I've also run the linter and formatter to ensure the code meets the project's style requirements and will pass CI checks.
Please let me know if there's anything else that needs to be addressed!

torztomasz (approved):

PR #6917, opened 2025-02-24
Include shared modules in Disco UI

adamiak:
Resolves L2B-9383

If there's a shared module, it's listed in smart contracts (separate section) and shown in Protocol Beat.


adamiak:
Fixed by https://github.com/l2beat/l2beat/pull/7000

PR #6935, opened 2025-02-25
Refactor chain apis

sz-piotr:


packages/backend/src/config/chain/getChainConfig.ts
```diff
@@ -1,245 +1,156 @@
...
+          })
+          break
+        case 'blockscout':
+          indexerApis.push({
+            type: api.type,
+            url: api.url,
+          })
+          break
+        case 'blockscoutV2':
+          // TODO: not sure why we don't support this
```
maciekop-l2b, line 39: `blockscoutV2` is used only in `VerifiersStatusRefresher` which is not using `getChainConfig`

maciekop-l2b (approved):

PR #6937, opened 2025-02-25
Use clingo-wasm to build project page facts

adamiak:
Resolves L2B-9395

When discovery is called with `--build-project-page-facts`, it uses clingo model files to generate facts that can be used to display permissions section on the frontend.

Unfortunately I can't add tests for this part of the code because `clingo-wasm` (or more precisely, `emscripten`'s binding to wasm) seems to modify the global nodejs's `Error` prototype chain to add custom serialization/deserialization logic for WebAssembly exceptions. This seems to conflicts with `makeEthersError` in our codebase. I'll tackle this issue in a separate PR.


packages/discovery/src/discovery/modelling/projectPageFacts.test.ts
```diff
@@ -0,0 +1,17 @@
+import { expect } from 'earl'
+import { runClingo } from './projectPageFacts'
+
+// TODO:
+// `clingo-wasm` (or more precisely, emscripten's binding to wasm)
```
mateuszradomski, line 5: remove todo

mateuszradomski (approved):

PR #6995, opened 2025-03-03
Add Gas token info to TVS Breakdown

maciekzygmunt:
Resolves L2B-9454


torztomasz (changes requested):
Code looks fine although shouldn't it be "FUEL (gas)"?
PixelSnap 2025-03-05 at 12 24 47

maciekzygmunt:
> Code looks fine although shouldn't it be "FUEL (gas)"? PixelSnap 2025-03-05 at 12 24 47

1. ETH is gas for fuel.
2. ETH is not configured in fuel.ts that why it is not next to Ether

PR #7001, opened 2025-03-03
Op-stack Celestia verification

sdlyy:
Resolves L2B-9285


packages/discovery/src/discovery/provider/AllProviders.ts
```diff
@@ -69,13 +69,23 @@ export class AllProviders {
         })
       }
 
+      const celestiaApiClient = new CelestiaApiClient({
```
mateuszradomski, line 72: It's kinda stupid that it's required. I'd want it to be optional like beaconApiUrl above.

mateuszradomski (approved):
LGTM, fix this one problem with optional provider

PR #7016, opened 2025-03-04
Allow for undefined gas token

maciekzygmunt:
Resolves L2B-9480


packages/config/src/projects/layer2s/templates/orbitStack.ts
```diff
@@ -157,12 +157,8 @@ interface OrbitStackConfigCommon {
   additionalPurposes?: ScalingProjectPurpose[]
   overridingPurposes?: ScalingProjectPurpose[]
   isArchived?: boolean
-  gasTokens?: {
-    /** Gas tokens that have been added to tokens.jsonc - will be listed under the escrow */
-    tracked?: string[]
-    /** Gas tokens that are applicable yet cannot be added to tokens.jsonc for some reason (e.g. lack of GC support) */
-    untracked?: string[]
-  }
+  /** Gas tokens that are applicable yet cannot be added to tokens.jsonc for some reason (e.g. lack of GC support) */
```
sz-piotr, line 160:
```suggestion
  /** Gas tokens that are applicable yet cannot be added to tokens.jsonc for some reason (e.g. lack of CG support) */
```

packages/config/src/projects/layer2s/templates/orbitStack.ts
```diff
@@ -451,6 +447,15 @@ function orbitStackCommon(
           ? 'orbit-optimium'
           : 'orbit-rollup'
 
+  const trackedGasTokens = templateVars.chainConfig?.gasTokens?.filter(
+    (t) => !templateVars.untrackedGasTokens?.includes(t),
+  )
+
+  const allGasTokens = uniq([
```
sz-piotr, line 454: I would remove this and use templateVars.chainConfig?.gasTokens instead

sz-piotr (approved):

packages/config/src/projects/layer2s/templates/orbitStack.ts
maciekzygmunt: I did it in case someone configures `untracked` without configuring tokens in `chainConfig`, but maybe we can add test so it is not possible.

sz-piotr: yeah, I'd prefer duplicating the declaration for this. So you'd say gasTokes: ['reETH'] and then unconfiguredGasTokens: ['reETH']

PR #7031, opened 2025-03-04
Allow to omit `multicall` instead of setting garbage value

sdlyy:
Resolves L2B-9468


packages/discovery/src/discovery/provider/multicall/MulticallClient.ts
```diff
@@ -17,7 +17,7 @@ export interface CallProvider {
 export class MulticallClient {
   constructor(
     private readonly provider: CallProvider,
-    private readonly config: MulticallConfig,
+    private readonly config?: MulticallConfig,
```
mateuszradomski, line 20: Since we're passing this config as a function argument later it's better to remove this variable as a field of this class so we can't access it through `this.config` as a mistake

mateuszradomski (changes requested):

mateuszradomski (approved):

PR #7035, opened 2025-03-05
Finish refactoring details pages to ProjectService

sz-piotr:


packages/config/src/adjustments.ts
```diff
@@ -21,6 +23,9 @@ import { isProjectVerified, isVerified } from './verification/isVerified'
  * process.
  */
 export function runConfigAdjustments() {
+  if (once) return
+  once = true
+
```
torztomasz, line 28: why would it be run twice?

packages/frontend/src/utils/project/contracts-and-permissions/get-contract-utils.ts
```diff
@@ -0,0 +1,140 @@
...
+    address: EthereumAddress | string,
+  ): UsedInProject[]
+}
+
+let contractUtils: ContractUtils | undefined
+
+export async function getContractUtils(): Promise<ContractUtils> {
+  if (contractUtils) {
+    return contractUtils
+  }
```
torztomasz, line 20: idk if we need a singleton here, probably will never happen bcs of serverless

torztomasz (approved):
LFG 🔥

packages/config/src/adjustments.ts
sz-piotr: Just a precaution. Since this is no longer run on import but on `getProjects`. Maybe tests? Maybe some future invocation.

packages/frontend/src/utils/project/contracts-and-permissions/get-contract-utils.ts
sz-piotr: Even if that's the case it probably doesn't hurt and adds 3 lines of code. If you really insist I'll remove

PR #7045, opened 2025-03-05
Refactor UnixTime into number

sz-piotr:


maciekop-l2b (approved):

packages/backend/scripts/tvs/execute.ts
```diff
@@ -36,7 +36,8 @@ const cmd = command({
     const ps = new ProjectService()
     const localExecutor = new LocalExecutor(ps, env, logger)
 
-    const timestamp = UnixTime.now().toStartOf('hour').add(-3, 'hours')
+    const timestamp =
+      UnixTime.toStartOf(UnixTime.now(), 'hour') - UnixTime(3, 'hours')
```
torztomasz, line 40: Shouldn't this be `UnixTime.toStartOf(UnixTime.now(), 'hour') - 3 * UnixTime.HOUR`

`(3, 'hours')` was a nice syntax for subtracting but now its redundant

sz-piotr: I am worried about introducing conversion bugs if I change that. I was mostly doing those changes with regexes and the case of `UnixTime(2+2,'days')` will be incorrectly translated to `2 + 2 * UnixTime.DAY` and I am really scared of breaking stuff

packages/backend/src/modules/activity/indexers/BlockActivityIndexer.ts
```diff
@@ -65,20 +65,20 @@ export class BlockActivityIndexer extends ManagedChildIndexer {
...
-      const timestamp = record.timestamp.toNumber()
+      const timestamp = record.timestamp
       if (timestamp < min) min = timestamp
       if (timestamp > max) max = timestamp
     }
 
     const currentValues = await this.$.db.activity.getByProjectAndTimeRange(
       this.$.projectId,
-      [new UnixTime(min), new UnixTime(max)],
+      [UnixTime(min), UnixTime(max)],
```
torztomasz, line 79: hmm, now `[min, max]` also will work

sz-piotr: true. In almost all cases dropping `UnixTime` would work. I'd error on the side of safety since I have no way of figuring out the places where it was necessary.

packages/backend/scripts/tvs/execute.ts
sz-piotr: Okay I figured out that if I replace the occurrences with a regex but wrap with () then it's safe. So I replaced it all

PR #7046, opened 2025-03-05
Add PermissionsFromModel

adamiak:
Part of L2B-9400

This PR enables displaying (rendering) permissions on the project page based on facts generated by Clingo.

`ProjectDiscovery` has:
 * `PermissionRegistry` field
  
 `PermissionRegistry` is an interface with 2 implementations:
* `PermissionsFromDiscovery` - used by default right now, only lightly refactored in this PR
* `PermissionsFromModel` - this is the new one, used if there's a `/displayUsingModel` file in project's discovery folder

This PR includes a bunch of refactoring of Clingo files to make them easier to render, plus refactoring of PermissionRegistry to make it simpler.

Also, sorting by priority has been moved back to ProjectDiscovery to make sure sorting works the same way regardless of the source of data.

I've decided to check the existence of `displayUsingModel` file rather than some entry in `config.jsonc` file, because config files are chain-specific while displaying using model is chain-agnostic (works only for the whole project).

`PermissionFromModel` accesses fact elements by their position in the params list. This can be done in a much nicer way but will be a bigger effort and is left for later.


packages/discovery/src/discovery/modelling/interpolate.ts
```diff
@@ -11,9 +11,10 @@ export function interpolateModelTemplate(
     tryCastingToName(String(values['$.address']), addressToNameMap, false),
   )
   const withValuesReplaced = withSelfReplaced.replace(
-    /#([a-zA-Z0-9_$.]+)(:raw)?/g,
-    (_match, key, raw) => {
+    /#([a-zA-Z0-9_$.]+)(:raw)?(\|lower)?/g,
```
mateuszradomski, line 14: This is getting very close to being better to hand roll a parser for instead of tacking on additional regex stuff to it

mateuszradomski (approved):

PR #7069, opened 2025-03-07
Add costs chart stats on project page

torztomasz:


packages/frontend/src/components/chart/costs/project-costs-chart.tsx
```diff
@@ -21,9 +27,9 @@ interface Props {
 export function ProjectCostsChart({ milestones, projectId }: Props) {
   const [range, setRange] = useState<CostsTimeRange>('1y')
   const [unit, setUnit] = useState<CostsUnit>('usd')
-  const [showDataPosted, setShowDataPosted] = useState(false)
+  const [showDataPosted, setShowDataPosted] = useState(true)
```
maciekzygmunt, line 30: Should it be shown by default?

maciekzygmunt (approved):
LGTM but check the comment

packages/frontend/src/components/chart/costs/project-costs-chart.tsx
torztomasz: yeah, we wanted to change that

PR #7071, opened 2025-03-07
Add scope of assessment to Stage 1 and 2 projects

lucadonnoh:
Resolves L2B-9523


packages/config/src/projects/layer2s/aztecV1.ts
```diff
@@ -155,6 +156,15 @@ export const aztecV1: Layer2 = {
     dataFormat:
       'The data format used can be found [here](https://github.com/AztecProtocol/aztec-2.0/blob/master/blockchain/contracts/Decoder.sol).',
   },
+  scopeOfAssessment: {
+    checked: [SOA.l1Contracts, SOA.gasToken],
+    notChecked: [
+      SOA.l2Node,
```
sekuba, line 162: aztec v1 had a trusted setup that we can list under 'not checked' i think

lucadonnoh: done

sekuba (approved):

PR #7094, opened 2025-03-10
Make all table rows have the same height

maciekzygmunt:
Resolves L2B-9162


packages/frontend/src/app/(side-nav)/scaling/summary/_components/table/columns.tsx
```diff
@@ -78,6 +78,7 @@ export const scalingSummaryColumns = [
       sortUndefined: 'last',
       meta: {
         hash: 'stage',
+        cellClassName: 'h-10',
```
torztomasz, line 81: why do you need to explicitly declare that here? shouldn't it be set in `table.tsx`?

torztomasz (changes requested):

torztomasz (approved):

PR #7119, opened 2025-03-11
StarkEx diagram updates

sekuba:
Also clarifies the operator role description so that starknet (onchain DA) can be differentiated from starkex (offchain DA)


packages/frontend/public/images/architecture/starkex.png
lucadonnoh, line 1: where are the escrowed assets?

lucadonnoh (changes requested):

packages/frontend/public/images/architecture/starkex.png
sekuba: fixo

lucadonnoh (approved):

PR #7125, opened 2025-03-12
Remove Stealthchain

ahmed9202543:
Website & Explorer do not work, no social media posts


lucadonnoh:
i don't think we can archive an upcoming project, it should just be removed

ahmed9202543:
> i don't think we can archive an upcoming project, it should just be removed

@lucadonnoh  Thanks for the feedback! I've updated the PR to remove the `Stealthchain` instead of archiving it, as suggested. Let me know if there's anything else to adjust.

lucadonnoh (approved):

PR #7132, opened 2025-03-12
Align colors on costs table

maciekzygmunt:


packages/frontend/src/app/(side-nav)/scaling/costs/_components/table/columns.tsx
```diff
@@ -89,8 +89,7 @@ export function getScalingCostsColumns(metric: CostsMetric) {
         align: 'right',
         headClassName: getColumnHeaderUnderline(
           'w-[132px]',
-          'before:bg-sky-550',
-          'dark:before:bg-sky-500',
+          'before:bg-[hsl(var(--chart-stacked-blue))]',
```
torztomasz, line 92: it should be added to tailwind config so that you can use `bg-chart-stacked-blue`

torztomasz (approved):

PR #7137, opened 2025-03-12
Filters rework

torztomasz:
🔥🚀


packages/frontend/src/components/table/filters/utils/labels.ts
```diff
@@ -0,0 +1,35 @@
...
+  stack: 'stacks',
+  stage: 'stages',
+  purpose: 'purposes',
+  hostChain: 'chains',
+  daLayer: 'layers',
+  raas: 'providers',
+  validatedBy: 'validators',
+}
+
+export const emtpyStateLabel = (filterId: FilterableValueId | undefined) => {
```
maciekzygmunt, line 25:
```suggestion
export const emptyStateLabel = (filterId: FilterableValueId | undefined) => {
```

packages/frontend/src/app/(side-nav)/scaling/data-availability/_components/scaling-da-tables.tsx
```diff
@@ -58,14 +58,12 @@ export function ScalingDaTables(props: Props) {
...
-          ...entries.others,
+      <TableFilters
+        entries={[
+          ...props.rollups,
+          ...props.validiumsAndOptimiums,
+          ...props.others,
         ]}
-        className="max-md:mt-4"
-        showDALayerFilter
       />
```
maciekzygmunt, line 67: ?

maciekzygmunt:
Works great, but I have 2 issues:

When you open the filters and search for a filter starting with "F," it immediately closes. I understand this happens because "F" is a filter shortcut, but perhaps this behavior could be handled differently.

When clicking on a filter type (e.g., "Stack") without selecting anything, there's no way to go back—I have to close and reopen the popup.

packages/frontend/src/components/table/filters/use-include-filters.tsx
```diff
@@ -0,0 +1,31 @@
...
+        }
+        // You can have multiple filterable values with the same id, e.g. purpose
+        const filterableValues = filterable.filter((f) => f.id === filter.id)
+        if (filterableValues.length === 0) {
+          return false
+        }
+
+        const valueMatches = filterableValues.some((f) =>
+          filter.values.includes(f.value),
+        )
```
sz-piotr, line 23:
```suggestion
        if (!filterable.some((f) => f.id === filter.id)) {
          return false
        }

        const matches = filterable.some((f) =>
          f.id === filter.id && filter.values.includes(f.value),
        )
```

packages/frontend/src/components/table/filters/use-include-filters.tsx
```diff
@@ -0,0 +1,31 @@
+import { useCallback } from 'react'
+import { useTableFilterContext } from './table-filter-context'
+import type { FilterableEntry } from './types'
+
+export function useIncludeFilters() {
```
sz-piotr, line 5: Why not `useTableFilters()`?

maciekzygmunt (approved):

packages/frontend/src/components/table/filters/types.ts
```diff
@@ -0,0 +1,12 @@
...
+export interface FilterableEntry {
+  filterable: FilterableValue[] | undefined
+}
+
+export type FilterableValue = {
+  id: FilterableValueId
+  value: string
+}
+
+export type FilterableValueId = keyof typeof filterIdToLabel
```
sz-piotr, line 12: I would combine utils/labels, utils/filter-values and types into a single FilterableValue file

PR #7155, opened 2025-03-13
do not subtract count for Zircuit activity

sekuba:


lucadonnoh (approved):

sekuba:
waiting for tomorrow to see if we implement
- sinceBlock 
- system tx of <1 per block
for this single bad boi

PR #7161, opened 2025-03-14
[L2B-9443] Implement TvsPriceIndexer

maciekop-l2b:


packages/backend/src/modules/tvs/indexers/TvsPriceIndexer.ts
```diff
@@ -0,0 +1,100 @@
...
+
+  override async removeData(configurations: RemovalConfiguration[]) {
+    for (const configuration of configurations) {
+      const deletedRecords = await this.$.db.tvsPrice.deleteByConfigInTimeRange(
+        configuration.id,
+        UnixTime(configuration.from),
+        UnixTime(configuration.to),
+      )
+
+      if (deletedRecords > 0) {
```
antooni, line 90: I know we have this pattern in other indexers, but IMO we should change it to always log, even if deleted === 0

maciekop-l2b: done

PR #7172, opened 2025-03-17
Adjust project filenames and ids

sz-piotr:


mateuszradomski (approved):
If we are changing the ids of projects then it'd be good to also consolidate other projects like
- `polygon-pos` has a `-` but all other polygon projects do not
- possibly make all project ChainIDs lowercase, this would change `pNetwork`, `pulseChain` and `opticsV1`, `opticsV2`

PR #7239, opened 2025-03-21
[L2B-9678] Handle since and until within ValueService

maciekop-l2b:


packages/backend/src/modules/tvs/services/ValueService.ts
```diff
@@ -74,7 +74,20 @@ export class ValueService {
...
     const config = createAmountConfig(formula)
     const amount = await this.storage.getAmount(config.id, timestamp)
-    assert(amount !== undefined, `${formula.type} ${config.id}`)
+
+    if (amount === undefined) {
+      if (
+        timestamp < config.sinceTimestamp ||
+        (config.untilTimestamp && timestamp > config.untilTimestamp)
+      ) {
+        return BigIntWithDecimals(BigInt(0), config.decimals)
```
antooni, line 83: this can be problematic for min() and max() formulas

antooni (changes requested):

packages/backend/src/modules/tvs/services/ValueService.ts
maciekop-l2b: fixed

PR #7262, opened 2025-03-24
Fix stage notice color

maciekzygmunt:
Resolves L2B-9695

### DIFF
--- a/packages/frontend/src/components/table/cells/stage/stage-cell.tsx
+++ b/packages/frontend/src/components/table/cells/stage/stage-cell.tsx
@@ -38,11 +38,10 @@ export function StageCell({ stageConfig, isAppchain, href }: StageCellProps) {
               <CircleQuestionMarkIcon
                 className={cn(
                   '-mt-px size-5 fill-current md:mt-px',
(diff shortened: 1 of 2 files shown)


packages/frontend/src/components/table/cells/stage/stage-cell.tsx
```diff
@@ -59,3 +60,19 @@ export function StageCell({ stageConfig, isAppchain, href }: StageCellProps) {
...
+    case 'Stage 2':
+      return questionMark ? 'fill-white' : 'fill-green-900'
+    case 'Stage 1':
+      return questionMark ? 'fill-yellow-900' : 'fill-[#FFC61B]'
+    case 'Stage 0':
+      return questionMark ? 'text-white' : 'fill-orange-600 dark:fill-[#E03109]'
+    default:
+      return ''
+  }
+}
```
torztomasz, line 78: create two different fns, this is looking ugly

torztomasz (changes requested):

torztomasz (approved):

PR #7265, opened 2025-03-24
Sidebar rewrite

torztomasz:


maciekzygmunt (changes requested):
Code looks good, but I do not like spacing, it is much different then current one. Also styling of bottom section is different (glossary, gov etc.).

maciekzygmunt:
Also we need to remove styles from `ul` on mobile

PR #7276, opened 2025-03-26
Review celo l2

sekuba:
Resolves L2B-9693


luke7211 (dismissed):
Permissions section is missing challenger

PR #7313, opened 2025-03-30
Add documentation for registry function in StarkgateManager

GarmashAlex:
Added NatSpec documentation for the internal registry() function in StarkgateManager.sol, explaining its purpose, implementation details and return value. This addresses the TODO comment that was previously in place.


lucadonnoh:
external contract, do not modify

PR #7328, opened 2025-03-31
Add corn token

vincfurc:
resolves L2B-9738

### DIFF
--- a/packages/config/src/tokens/tokens.jsonc
+++ b/packages/config/src/tokens/tokens.jsonc
@@ -5815,6 +5815,21 @@
   ],
   "corn": [
     {
+      "symbol": "CORN",
+      "address": "0x44f49ff0da2498bCb1D3Dc7C0f999578F67FD8C6",
+      "coingeckoId": "corn-3",
+      "source": "external",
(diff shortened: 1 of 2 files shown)


packages/config/src/tokens/tokens.jsonc
```diff
@@ -5814,6 +5814,21 @@
     }
   ],
   "corn": [
+    {
+      "symbol": "CORN",
+      "address": "0x44f49ff0da2498bCb1D3Dc7C0f999578F67FD8C6",
+      "coingeckoId": "corn-3",
+      "source": "external",
+      "supply": "circulatingSupply",
```
sekuba, line 5822: hmm this combo is weird, is it preminted? (should probably be totalSupply)

sekuba (changes requested):

packages/config/src/tokens/tokens.jsonc
vincfurc: Not fully preminted, I used circulating supply to avoid prematurely inflate the TVS since (most of) the rest is not minted

sekuba (approved):

PR #7364, opened 2025-04-02
Add config diff step

sz-piotr:


packages/config/scripts/diff/html.ts
```diff
@@ -0,0 +1,172 @@
...
+  <meta charset="UTF-8" />
+  <title>Diff @${hashAfter}</title>
+  <style>${getStyles()}</style>
+</head>
+
+<body>
+  <script>${alpineSource}</script>
+  <h1>
+    Diff between
+    <code class="removed">${hashBefore}</code> (main)
```
mateuszradomski, line 30: Suggestion: Make this a link to the commit on GH?

packages/config/scripts/diff/diff.ts
```diff
@@ -0,0 +1,159 @@
...
+  let beforeIndex = 0
+  let afterIndex = 0
+  while (
+    beforeIndex < projectsBefore.length ||
+    afterIndex < projectsAfter.length
+  ) {
+    const before = projectsBefore[beforeIndex]
+    const after = projectsAfter[afterIndex]
+
+    if (!before) {
```
mateuszradomski, line 39: What will happen if in the following scenario

```
| before projects | after projects |
|-----------------|----------------|
|        arbitrum |       arbitrum |
|        starknet |                |
|            zora |           zora |
```

We want to see `starknet` as removed but since we iterate over indexes we're going to show a diff for starknet between it and zora plus later we're going to say that zora was removed. Or am i mistaken?

packages/config/scripts/diff/diff.ts
mateuszradomski, line 1: We need a clear stance on where each script goes. I think that this makes sense to be here, but other stuff inside this folder does not belong

mateuszradomski (approved):

PR #7369, opened 2025-04-02
Basic ecosystem page and config

torztomasz:
Resolves L2B-9697


packages/config/src/types.ts
```diff
@@ -873,6 +875,32 @@ export interface AvailDaTrackingConfig {
   sinceBlock: number
   untilBlock?: number
 }
+
+export interface ProjectEcosystemInfo {
+  id: ProjectId
+  sinceTimestamp: UnixTime
+  untilTimestamp?: UnixTime
```
sz-piotr, line 882: Why do we need this?

sz-piotr (approved):

packages/config/src/types.ts
torztomasz: For this chart
image

PR #7380, opened 2025-04-03
Improve new project under review state

maciekzygmunt:
Resolves L2B-9272

### DIFF
--- a/packages/frontend/src/utils/project/technology/get-technology-section-props.ts
+++ b/packages/frontend/src/utils/project/technology/get-technology-section-props.ts
@@ -8,9 +8,11 @@ export function getTechnologySectionProps(
   if (items.length === 0) {
     return undefined
   }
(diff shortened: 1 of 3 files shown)


packages/frontend/src/utils/project/technology/get-technology-section-props.ts
```diff
@@ -8,9 +8,11 @@ export function getTechnologySectionProps(
...
     return undefined
   }
+  const nonEmptyItems = items.filter((i) => i.name !== 'Empty')
   return {
     isUnderReview:
-      items.every((x) => x.isUnderReview) || project.statuses.isUnderReview,
-    items,
+      nonEmptyItems.every((x) => x.isUnderReview) ||
+      project.statuses.isUnderReview,
+    items: nonEmptyItems,
```
torztomasz, line 16: we can keep it for now but this is so wrong, we should allow for undefined or smth, please take care of that

torztomasz (approved):

PR #7386, opened 2025-04-03
Replace drag-and-drop shape definition

mateuszradomski:
Resolves L2B-9588


packages/config/src/projects/_templates/facet/FacetEtherBridge/shape/FacetEtherBridgeV6.sol
```diff
@@ -1,3728 +0,0 @@
-// SPDX-License-Identifier: Unknown
```
sz-piotr, line 1: no shapes.json

sz-piotr (commented):
Looks good but I am worried about the removed code without shapes. Maybe historical discoveries have the contracts? We can probably find those automatically by figuring out when this file was created and checking the discovery then.

lucadonnoh:
what about having a .flat folder here as well. i usually run diffs against contracts in shape since that is where i know contracts we have already seen are

adamiak (approved):

mateuszradomski:
I've found the implementations that produce hashes present in the shapes that didn't match to current state. For example, Base had a contract `Challenger1of2` and we templatized it but it has since been deleted but the shape was untouched by us. I found all ~60 implementations that had this status and hardcoded them into the migration script. I double checked if importing those addresses produces the same hash and everything is fine. There where three examples of hashes that were unused for misconfigured:

KeyValueStore - never used/matched to anything (0x56713b5534c12ed120f7815d30e2b2dc41289446324aec923a867cfd300f1be5)
MIPS - never used/matched to anything (0x062a3f9d395c0c0119e9ee90d3f5c2f8d62124b52ceaa8cdb4df0e107c8f8c4f)

(0x46eae4f2e7b6b0f5096ba403f113ff3c86205368a36af081ab280441d64942fc) - misconfigured, it's a flat file of the proxy, not the implementation

PR #7413, opened 2025-04-06
Discovery updates 04/06

sekuba:


lucadonnoh (changes requested):
IMG_6005


this is wrong, we need to add proper untilTimestamp and the new entry

sekuba:
i think the diff is broken

lucadonnoh:
yes for now the diff ci job needs to be rerun manually until peter fixes it on monday

sekuba:
yush i failed to rerun it even

lucadonnoh (approved):

PR #7414, opened 2025-04-06
Review lens chain

sekuba:
- diamond proxy unverified
- across integration for most user bridging, external usdc escrow
- currently tracking 3 tokens: ETH, USDC, LGHO (gastoken)
- added gastoken, but not as associated bc it is a collateralized stablecoin
Resolves L2B-9832


sekuba:
wow, good catch, so they are even sharing the **custom** escrows now
lets try to count this on L2 then

lucadonnoh:
don't you also need to move it for sophon?

lucadonnoh (approved):

PR #7415, opened 2025-04-07
Publish diff to surge instead

sz-piotr:


.github/workflows/config-diff.yml
kondziorex7, line 1: Please remove unnecessary permissions

kondziorex7 (approved):

PR #7419, opened 2025-04-07
Allow to specify bridges tab

maciekzygmunt:
Resolves L2B-9776


packages/config/src/internalTypes.ts
```diff
@@ -263,6 +264,7 @@ export interface BridgeDisplay {
   description: string
   detailedDescription?: string
   category: BridgeCategory
+  bridgeTab: BridgeTab
```
torztomasz, line 267: it should not be named bridgeTab, config is not aware of frontend, it should be named after whatever is "Single-chain" and "Multichain"

torztomasz (approved):

sekuba (approved):

PR #7420, opened 2025-04-07
More ecosystem widgets

torztomasz:


lucadonnoh:
we should add archived timestamp to archived projects so that the project count is not up only

packages/frontend/src/hooks/use-tailwind-breakpoint.ts
```diff
@@ -0,0 +1,27 @@
+import { useEffect, useState } from 'react'
+import { useEventListener } from './use-event-listener'
+
+type Breakpoint = 'xs' | 'sm' | 'md' | 'lg' | 'xl'
+
+export function useTailwindBreakpoint() {
```
maciekzygmunt, line 6: We should not create new one, but reuse useBreakpoint hook

maciekzygmunt (approved):

PR #7435, opened 2025-04-08
Add iconUrl to TVS config

antooni:
Resolves L2B-9836


torztomasz (approved):

maciekop-l2b (changes requested):
This should not be a separate script. Please integrate it with `generate-config`

maciekop-l2b (approved):

PR #7439, opened 2025-04-09
Update type in ProjectValue table

antooni:


packages/shared-pure/src/types/ProjectValueType.ts
```diff
@@ -0,0 +1,5 @@
+export type ProjectValueType =
+  | 'PROJECT'
```
maciekop-l2b, line 2: Please add comments explaining what each type is doing

packages/database/prisma/schema.prisma
```diff
@@ -298,17 +298,24 @@ model TokenValue {
...
+  native     Float            @db.Real
+  ether      Float            @db.Real
+  stablecoin Float            @db.Real
+  other      Float            @db.Real
+  associated Float            @db.Real
 
   @@id([timestamp, project, type])
 }
+
+enum ProjectValueType {
```
maciekop-l2b, line 316: I would opt for using regulars string as DB level enum will be problematic ( for example when adding/removing elements)

maciekop-l2b (changes requested):

maciekop-l2b (approved):

PR #7445, opened 2025-04-09
Refactor isArchived to archivedAt

torztomasz:


packages/frontend/src/server/features/scaling/summary/get-scaling-api-entries.ts
```diff
@@ -59,7 +60,7 @@ export async function getScalingApiEntries(): Promise<ScalingApiEntry[]> {
         category: project.scalingInfo.type,
         provider: project.scalingInfo.stack,
         purposes: project.scalingInfo.purposes,
-        isArchived: false,
+        archivedAt: undefined,
```
sz-piotr, line 63: don't change it here. No need to break api

sz-piotr (approved):

packages/frontend/src/server/features/scaling/summary/get-scaling-api-entries.ts
torztomasz: yup, overlooked it :P

PR #7457, opened 2025-04-10
Add l2b download-shapes command

adamiak:
Calling `l2b download-shapes <template>` removes existing <template>/shapes folder and recreates if by downloading all source file referenced in shapes.json. It also verifies that hashes of downloaded files match those in shapes.json.

### DIFF
--- a/packages/l2b/src/commands/DownloadShapes.ts
+++ b/packages/l2b/src/commands/DownloadShapes.ts
@@ -1,0 +1,78 @@
+import { writeFileSync } from 'fs'
+import { mkdirSync } from 'fs'
+import { join } from 'path'
+import {
+  TemplateService,
+  flattenStartingFrom,
+  getChainConfig,
+  getDiscoveryPaths,
+} from '@l2beat/discovery'
+import { flatteningHash } from '@l2beat/discovery/dist/flatten/utils'
+import { getExplorerClient } from '@l2beat/discovery/dist/utils/IEtherscanClient'
+import { CliLogger, HttpClient } from '@l2beat/shared'
+import { command, positional, string } from 'cmd-ts'
+import { rimraf } from 'rimraf'
+
+export const DownloadShapes = command({
+  name: 'download-shapes',
+  description: 'Download all Solidity files for shapes defined in a template',
+  args: {
+    template: positional({
+      type: string,
+      displayName: 'template',
+      description: 'name of the template to download shapes for',
+    }),
+  },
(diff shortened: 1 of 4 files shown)


packages/l2b/src/commands/DownloadShapes.ts
```diff
@@ -0,0 +1,78 @@
...
+    }
+
+    const templatePath = templateService.getTemplatePath(args.template)
+    const shapeSchema = templateService.readShapeSchema(
+      join(templatePath, 'shapes.json'),
+    )
+
+    // 1. Remove and recreate the shapes folder
+    // (helps if there are renames or removed shapes)
+    const shapesFolder = join(templatePath, 'shapes')
```
mateuszradomski, line 42: Do this at the end when we successfully downloaded all the new shapes. Otherwise if we failed to get the shapes we don't want to leave the user without their old shapes.

packages/l2b/src/commands/DownloadShapes.ts
```diff
@@ -0,0 +1,78 @@
...
+    rimraf.sync(shapesFolder)
+    mkdirSync(shapesFolder, { recursive: true })
+
+    // 2. Download the source code and flatten it
+    for (const shape of shapeSchema) {
+      const chainConfig = getChainConfig(shape.chain)
+      const httpClient = new HttpClient()
```
mateuszradomski, line 49: I'd group by chain and not recreate all of these things all the time, but not super important

packages/l2b/src/commands/DownloadShapes.ts
```diff
@@ -0,0 +1,78 @@
...
+import { getExplorerClient } from '@l2beat/discovery/dist/utils/IEtherscanClient'
+import { CliLogger, HttpClient } from '@l2beat/shared'
+import { command, positional, string } from 'cmd-ts'
+import { rimraf } from 'rimraf'
+
+export const DownloadShapes = command({
+  name: 'download-shapes',
+  description: 'Download all Solidity files for shapes defined in a template',
+  args: {
+    template: positional({
```
mateuszradomski, line 20: I'd allow the user to pass `--all` and it would iterate all templates with shapes and download all things

packages/l2b/src/commands/DownloadShapes.ts
adamiak: Let's optimize as a separate PR.

PR #7472, opened 2025-04-11
Near upgrades to 8 shards

vincfurc:
resolves L2B-9896

### DIFF
--- a/packages/config/src/projects/near/near-da.ts
+++ b/packages/config/src/projects/near/near-da.ts
@@ -147,9 +147,14 @@ Regarding data retrieval, full nodes prune Receipts after 3 epochs (approximatel
     pruningWindow: 43200 * 3, // minimum 3 epochs (12 hours each), claimed in practice around 5 epochs (due to nodes garbage collection)
     throughput: [
       {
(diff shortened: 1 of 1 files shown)


packages/config/src/projects/near/near-da.ts
```diff
@@ -147,10 +147,15 @@ Regarding data retrieval, full nodes prune Receipts after 3 epochs (approximatel
     pruningWindow: 43200 * 3, // minimum 3 epochs (12 hours each), claimed in practice around 5 epochs (due to nodes garbage collection)
     throughput: [
       {
-        size: 16777216, // 16 MiB , 4MiB per 4 shard
+        size: 16777216, // 16 MiB , 4MiB per 4 shards
```
lucadonnoh, line 150: until timestamp?

lucadonnoh (changes requested):

packages/config/src/projects/near/near-da.ts
vincfurc: not needed
image

lucadonnoh (approved):

PR #7481, opened 2025-04-14
Bridges config changes

maciekzygmunt:
Resolves L2B-9905


packages/config/src/internalTypes.ts
```diff
@@ -256,6 +256,7 @@ export interface Bridge {
   permissions?: Record<string, ProjectPermissions>
   milestones?: Milestone[]
   discoveryInfo?: ProjectDiscoveryInfo
+  upgradesAndGovernance?: string
```
torztomasz, line 259: It's weird that this is in here and image is in display, probably it should be merged to a object.
```tsx
...
upgradesAndGovernance?: {
	content: string | undefined
	image?: string
}
...
```

torztomasz (approved):

PR #7483, opened 2025-04-14
Add "Viewed" checkbox to config diff

torztomasz:


packages/config/scripts/diff/html.ts
```diff
@@ -169,8 +173,11 @@ ul {
   padding-left: 0;
   margin: 0;
 }
+input[type="checkbox"] {
+  margin: 0;
+}
 .field {
   margin-left: 16px;
-}
+}
```
sz-piotr, line 181: unnecessary whitespace

sz-piotr (approved):
just revert the changes to arbitrum

PR #7498, opened 2025-04-15
Fix permissions overflowin

maciekzygmunt:
Resolves L2B-9932

### DIFF
--- a/packages/frontend/src/components/projects/sections/contract-entry.tsx
+++ b/packages/frontend/src/components/projects/sections/contract-entry.tsx
@@ -74,7 +74,13 @@ export function ContractEntry({
       body={
         <>
           <div className="flex flex-wrap items-center gap-x-2 !leading-[1.15]">
-            <strong id={contract.name} className="scroll-mt-14 md:scroll-mt-10">
+            <strong
+              id={contract.name}
+              className="scroll-mt-14 md:scroll-mt-10"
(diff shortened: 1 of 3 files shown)


packages/frontend/src/components/projects/sections/contract-entry.tsx
```diff
@@ -74,7 +74,13 @@ export function ContractEntry({
       body={
         <>
           <div className="flex flex-wrap items-center gap-x-2 !leading-[1.15]">
-            <strong id={contract.name} className="scroll-mt-14 md:scroll-mt-10">
+            <strong
+              id={contract.name}
+              className="scroll-mt-14 md:scroll-mt-10"
+              style={{
+                wordBreak: 'break-word',
+              }}
```
torztomasz, line 82: why via style? :O

torztomasz (changes requested):

packages/frontend/src/components/projects/sections/contract-entry.tsx
maciekzygmunt: there is no tailwind class for this

torztomasz: then add it under `@utilities` in `global.css`

torztomasz (approved):

PR #7505, opened 2025-04-16
use opstack.ts for base

sekuba:


sekuba:
diff is a bit noisy here

lucadonnoh (approved):

PR #7507, opened 2025-04-16
Add ability to configure project colors

torztomasz:


maciekzygmunt (approved):

packages/config/src/ProjectDatabase.ts
```diff
@@ -25,6 +25,7 @@ const schema = {
   name: 'TEXT NOT NULL',
   shortName: 'TEXT',
   addedAt: 'INTEGER NOT NULL',
+  colors: 'TEXT',
```
sz-piotr, line 28: here also :)

PR #7511, opened 2025-04-16
New permission modelling flow using Clingo

adamiak:
Resolves L2B-9709, L2B-9827

This PR achieves four things:
1. Removes permissions solver and all code around it (e.g. self/target meta and permission merging in Discovery). Structure discovery no longer sees permission related config and doesn't compute anything permission-related.
2. Implements permission modelling using Clingo, achieving exactly the same result as the old solver on all projects.  
3. Introduces a new `l2b model-permissions` flow (similar to `l2b colorize`) that is optional on top of structure discovery. It's also always run as part of `l2b discover`, just like colorization.
4. Keeps format of permissions "combined" into `discovered.json` mostly as it was before (keeps `.receivedPermissions` and `.directlyReceivedPermissions`, but drops `.issuedPermissions` which is trivially derived from previous two). This approach shows that no errors were introduced and ProjectDiscovery etc. works without any additional modifications.

The modelling process with Clingo follows the following steps:
```
modelPermissions()

* gather all relevant config, templates and .lp files, combine and interpolate
  |
  --> produces `clingo.input.lp` (in memory, unless `--debug` flag is passed)
    |
    --> runs Clingo on that file
      |
      --> produces `clingo.output.lp` (in memory) with "facts" called `ultimatePermission`
        |
        --> transforms facts into `.receivedPermissions` and `.directlyReceivedPermissions`
          |
          --> combines those into discovered.json
```

The `permissionsConfigHash` is calculated as a hash of `clingo.input.lp`. In the future a more elaborate and faster process can be introduced, but current one is surprisingly efficient - it catches changes in templates, .lp files, and even .ts code - anything that would generate different input for Clingo.

### DIFF
--- a/packages/config/src/projects/linea/ethereum/diffHistory.md
+++ b/packages/config/src/projects/linea/ethereum/diffHistory.md
@@ -1,6 +1,6 @@
-Generated with discovered.json: 0x719cc041ce1a45d6a7efcb1dd319d810455f945a
+Generated with discovered.json: 0x4ce0363175c7691159b2deeb25bdfc9377c9046e
 
-# Diff at Tue, 29 Apr 2025 08:19:06 GMT:
+# Diff at Tue, 29 Apr 2025 09:46:34 GMT:
 
 - author: Adrian Adamiak (<adrian@adamiak.net>)
 - comparing to: main@ef7477af00fe0b57a2f7cacf7e958c12494af662 block: 22273396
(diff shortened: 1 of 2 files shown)


packages/shared-pure/src/types/EthereumAddress.test.ts
```diff
@@ -68,4 +68,13 @@ describe(EthereumAddress.name, () => {
       '0x0000000000000000000000000001234567890AbC',
     )
   })
+
+  it('properly checks ignoring case', () => {
```
mateuszradomski, line 72: This seems like it's not necessary since the default implmentation ignores casing if it's all lower/upper

packages/l2b/src/implementations/discovery/updateDiffHistory.ts
```diff
@@ -27,15 +27,34 @@ import { updateDiffHistoryHash } from './hashing'
...
+  configReader: ConfigReader,
   projectName: string,
   chain: string,
   description?: string,
   overwriteCache: boolean = false,
 ) {
   // Get discovered.json from main branch and compare to current
-  console.log(`Project: ${projectName}`)
+  console.log(`Updating diffHistory for: ${projectName} on ${chain}`)
   const paths = getDiscoveryPaths()
```
mateuszradomski, line 57: We can send those paths as an argument

packages/config/src/projects/_templates/orbitstack/Outbox/model.lp
```diff
@@ -1,6 +1,7 @@
-permission(L2Entrypoint, "act", @self) :-
-  l2Entrypoint(L2Entrypoint).
-
-permissionDelay(L2Entrypoint, "act", @self, Seconds) :-
-  challengePeriodSeconds(Seconds),
-  permission(L2Entrypoint, "act", @self).
+% permission(L2Entrypoint, "act", @self) :-
```
mateuszradomski, line 1: If we're commenting this out maybe we can just delete it?

packages/discovery/src/cli/modelPermissionsCommand.ts
```diff
@@ -0,0 +1,63 @@
...
+  type DiscoveryPaths,
+  getDiscoveryPaths,
+} from '../discovery/config/getDiscoveryPaths'
+import { combinePermissionsIntoDiscovery } from '../discovery/modelling/combinePermissionsIntoDiscovery'
+import { modelPermissions } from '../discovery/modelling/modelPermissions'
+import { saveDiscoveredJson } from '../discovery/output/saveDiscoveryResult'
+import { sortEntry } from '../discovery/output/toDiscoveryOutput'
+import type { PermissionsOutput } from '../discovery/output/types'
+
+export async function modelPermissionsCommand(
```
mateuszradomski, line 14: I'd rename this to just `modelPermissions`

packages/config/src/test/discovery/config.test.ts
```diff
@@ -247,4 +249,28 @@ describe('discovery config.jsonc', () => {
       }
     }
   })
+
+  it('model-permissions is up to date', () => {
```
mateuszradomski, line 253: On my laptop this test takes around 1-1.5s. It might become a problem on the CI, there probably are ways to speed it up. But that's for later, just something to be aware of.

For safety we might want to add `.timeout(5000)` to this test, just to give ourselves some time

mateuszradomski (approved):

packages/shared-pure/src/types/EthereumAddress.test.ts
adamiak: That's no my experience. Using `EthereumAddress.check` with incorrect casing returns false. That's why I added `checkIgnoringCase()`

packages/discovery/src/cli/modelPermissionsCommand.ts
adamiak: There's already `modelPermissions` so I want to avoid confusion. It also follows naming of the other files and functions in this folder: `discoverCommand()` and `singleDiscoveryCommand()`

PR #7534, opened 2025-04-17
Allow to use all params via public api

torztomasz:


packages/frontend/src/app/api/(public)/scaling/activity/[slug]/route.ts
```diff
@@ -2,44 +2,54 @@ import { UnixTime } from '@l2beat/shared-pure'
...
   request: NextRequest,
   props: { params: Promise<{ slug: string }> },
 ) {
-  const params = await props.params
+  const { slug } = await props.params
   const searchParams = request.nextUrl.searchParams
-  const range = ActivityTimeRange.catch('30d').parse(searchParams.get('range'))
-  const response = await getCachedResponse(params.slug, range)
 
+  const params = {
```
Comment (maciekzygmunt) [inline packages/frontend/src/app/api/(public)/scaling/activity/[slug]/route.ts:18]: Maybe this object should be typed as `ActivityChartParams` so we get an type error here when we add new property to chart params. Same for other routes

maciekzygmunt (approved):

packages/frontend/src/app/api/(public)/scaling/activity/[slug]/route.ts
Comment (torztomasz) [reply packages/frontend/src/app/api/(public)/scaling/activity/[slug]/route.ts]: good catch, it would cause a lot of issues left like that

PR #7546, opened 2025-04-18
Improve Ethereum DA syncing performance

antooni:
Resolves L2B-9963


maciekop-l2b:
Unfortunately we are not able to squeeze any more from the blobscout API so the only solution to improve sync time is to switch to a RPC + Blobscan

PR #7556, opened 2025-04-22
Edit (JSONata)

mateuszradomski:


mateuszradomski:
Decided against using JSONata, going to use BLIP instead.

PR #7559, opened 2025-04-22
Scroll Euclid upgrade phase 2 (Stage 1 review)

vincfurc:
Resolves L2B-9975
Resolves L2B-8871


sekuba:
canActIndependently seems to still be broken (scrollOwner appears in permissions and has the wrong strings)

PR #7584, opened 2025-04-24
Align tooltips styling

maciekzygmunt:
Resolves L2B-9858


torztomasz (approved):
lgtm, although I am starting to dislike the pattern of `label-...` or `heading-...`, I feel that something is wrong about that

PR #7591, opened 2025-04-24
[L2B-10018] Remove TVL code - part 2

maciekop-l2b:
- Remove `tvlConfig` from `ProjectService` package and make it a part of TVS generation script
- Rename `tvlInfo` to `tvsInfo`
- Remove `backend-shared` package as it is no longer needed (no BE code is shared with FE)


packages/backend/src/tools/uif/indexerIdentity.ts
maciekzygmunt, line 1: `getPremintedIndexerId` and everything below can be removed as this is part of old TVL. And some of values in `INDEXER_NAMES`

maciekzygmunt (changes requested):

maciekzygmunt (approved):

PR #7631, opened 2025-04-28
Add limited Sourcify support

sdlyy:


packages/l2b/src/commands/args.ts
```diff
@@ -23,10 +23,10 @@ export const explorerUrl = option({
 })
 
 export const explorerType = option({
-  type: oneOf(['etherscan', 'blockscout']),
+  type: oneOf(['etherscan', 'blockscout'] as const),
```
mateuszradomski, line 26: I guess we also need to add `sourcify` as an option

packages/l2b/src/implementations/tokenMinters.ts
```diff
@@ -17,10 +16,10 @@ export async function getTokenMinterEvents(
   rpcUrl: string,
   explorerUrl?: string,
   explorerApiKey?: string,
-  explorerType?: string,
+  explorerType?: 'etherscan' | 'blockscout',
```
mateuszradomski, line 19: also `sourcify` here, but maybe it's better to export the type from discovery

packages/discovery/src/utils/SourcifyClient.ts
```diff
@@ -0,0 +1,230 @@
...
+      )
+    }
+
+    const json = await response.json()
+
+    return json
+  }
+
+  getFirstTxTimestamp(_address: EthereumAddress): Promise<UnixTime> {
+    throw new Error('getFirstTxTimestamp: Not implemented')
```
mateuszradomski, line 147: Maybe instead of `Not implemented` we can give a little more context, like
> "Sourcify only supports source code fetching. getFirstTxTimestamp is not supported by this explorer"

mateuszradomski (approved):

packages/l2b/src/commands/args.ts
sdlyy: Done by etherscan V2 integration changes

packages/discovery/src/utils/SourcifyClient.ts
sdlyy: https://github.com/l2beat/l2beat/pull/7631/commits/5858d40db1aebcf31bd41713d4d259bccf65da9a

PR #7640, opened 2025-04-29
Add option to enable emergency warning

maciekzygmunt:
Resolves L2B-10000
Screenshot 2025-04-29 at 15 49 13


torztomasz (commented):
looks good, although it seems a little bit odd to keep the `emergencyWarning` in statuses instead of keeping it closer to the `stage` field as it is tightly related to stage

torztomasz (approved):

PR #7700, opened 2025-05-06
Moonchain full review (copied from external)

sekuba:
Resolves L2B-10197


sheenhx:
@sekuba hi, any update?

sekuba:
gm, am i reading it correctly that the operator is able to finalize invalid state roots in your setup? (since the optimistic tier is 1: permissioned and 2: the highest tier)

sheenhx:
@sekuba gm, any update?

sheenhx:
@sekuba gm , any update?

sekuba:
hey i outlined our prioritisation based on TVS/activity above, this is the same rule for all projects.

sheenhx:
> hey i outlined our prioritisation based on TVS/activity above, this is the same rule for all projects.

ok what is the current estimation of the review time, also if we don't get exposed by l2beats our activities can't be very high as no one will notice.

sekuba:
We allocate our resources based on user activity and ecosystem relevance of deployed scaling systems. We will keep monitoring the metrics and reopen in case there is a change.

PR #7705, opened 2025-05-06
Ethereum blob throughput increase [EIP-7691]

vincfurc:
resolves L2B-10114


maciekzygmunt:
Maybe we should also add it to milestones on charts

maciekzygmunt (changes requested):
Maybe we should also add this milestone to `config/src/global/milestones.ts` as those milestones are showing on main page charts (activity, tvs etc.) not sure

maciekzygmunt (approved):

PR #7706, opened 2025-05-07
Update ink.ts

intensodefi:
Update documentation for ink


mateuszradomski:
Can you just create a single PR instead of flooding the repo with tens of PRs where the only change is a single line?

PR #7725, opened 2025-05-08
Add Unichain wBTC

vincfurc:
resolves L2B-10159


sekuba (changes requested):
hmmm why is there no unichain.json did you run tvs:generate?

sekuba (approved):

PR #7732, opened 2025-05-09
Reduce heavy deps from FE

torztomasz:


packages/backend/src/modules/update-monitor/DiscoveryRunner.ts
```diff
@@ -11,12 +11,12 @@ import {
   modelPermissionsForIsolatedDiscovery,
   toRawDiscoveryOutput,
 } from '@l2beat/discovery'
-import type { Analysis } from '@l2beat/discovery/dist/discovery/analysis/AddressAnalyzer'
+import type { Analysis } from '@l2beat/discovery'
```
mateuszradomski, line 14: Move this import lower

mateuszradomski (approved):

PR #7734, opened 2025-05-09
Move RedisCache implementation from discovery to backend

mateuszradomski:
This moves redis package dependency closer to the consumer

Part of L2B-10169


packages/backend/src/modules/update-monitor/DatabaseCache.ts
```diff
@@ -2,7 +2,7 @@ import type { DiscoveryCache as DiscoveryCacheInterface } from '@l2beat/discover
 
 import type { Database } from '@l2beat/database'
 
-export class DiscoveryCache implements DiscoveryCacheInterface {
```
torztomasz, line 5: why rename?

torztomasz (approved):

packages/backend/src/modules/update-monitor/DatabaseCache.ts
mateuszradomski: Because it tells the user what's the backend of the cache.

redis backend - `RedisCache`
database backend - `DatabaseCache`

PR #7750, opened 2025-05-12
Enable adding shapes via UI

sdlyy:


packages/l2b/src/implementations/discovery-ui/main.ts
```diff
@@ -120,6 +121,8 @@ export function runDiscoveryUi({ readonly }: { readonly: boolean }) {
     res.json(response)
   })
 
+  attachTemplateRouter(app, templateService)
```
mateuszradomski, line 124: I'd move this to read only mode

packages/protocolbeat/src/panel-values/ValuesPanel.tsx
```diff
@@ -95,15 +93,26 @@ function Display({
...
   const address = getAddressToCopy(selected)
 
-  const copy = address && canCopy(selected) && !isReadOnly && (
-    <CopyAddShapeCommand
-      chain={chain}
-      address={address}
-      name={selected.name}
-      blockNumber={blockNumber}
-    />
+  const dialog = address && canAddShape(selected) && !isReadOnly && (
```
mateuszradomski, line 102: Hide in read only

GIGA nit pick: the animation has a hard step, maybe it's not that hard to remove it and make it smooooooooth from start to finish. Although I don't know how to achieve that

mateuszradomski (approved):

PR #7777, opened 2025-05-14
Update Polygon stack & badges

radinat:


packages/frontend/src/components/table/cells/type-info.tsx
```diff
@@ -78,9 +78,9 @@ export const providerMap: Record<
     Icon: ArbitrumIcon,
     text: "This project is based on Arbitrum's code base.",
   },
-  Polygon: {
+  'Agglayer CDK': {
     Icon: PolygonIcon,
-    text: "This project is based on Polygon's code base.",
+    text: "This project is based on Polygon's Agglayer CDK code base.",
```
lucadonnoh, line 83:
```suggestion
    text: "This project is based on the Agglayer CDK code base.",
```

my understanding is that Agglayer doesn't want to be associated with Polygon

lucadonnoh (changes requested):

lucadonnoh (approved):

PR #7796, opened 2025-05-16
Improve costs chart performance

torztomasz:


packages/frontend/src/server/features/scaling/activity/get-activity-for-project-and-range.ts
```diff
@@ -0,0 +1,34 @@
...
+
+import { UnixTime } from '@l2beat/shared-pure'
+import { unstable_cache as cache } from 'next/cache'
+import { env } from '~/env'
+import { getRangeWithMax } from '~/utils/range/range'
+import { type CostsTimeRange, rangeToResolution } from '../costs/utils/range'
+
+export function getActivityForProjectAndRange(
+  projectId: string,
+  range: CostsTimeRange,
```
maciekzygmunt, line 12: Use correct type

maciekzygmunt (changes requested):

packages/frontend/src/server/features/scaling/activity/get-activity-for-project-and-range.ts
torztomasz: maybe we should call it CostsChartTimeRange

maciekzygmunt (approved):

PR #7813, opened 2025-05-19
Add Orbit finality to docs

vincfurc:
resolves L2B-10193


lucadonnoh (changes requested):
you never mention the grace period hence i was a bit confused whether you are correctly accounting for that as well as the actual delay is challenge period + grace period. please check if the math is correct and at least mention it

lucadonnoh (approved):

PR #7816, opened 2025-05-19
New Contracts Section design

adamiak:
Resolves L2B-9657, L2B-10323

Uses the new ".role" field on permissions to render Contracts section according to the spec from https://docs.l2beat.com/l2b_specs/contracts.html

Example:

image


packages/config/src/processing/getProjects.test.ts
```diff
@@ -178,15 +178,6 @@ describe('getProjects', () => {
         const contracts = project.contracts?.addresses ?? {}
         for (const [chain, perChain] of Object.entries(contracts)) {
           for (const [i, contract] of perChain.entries()) {
-            const description = contract.description
```
mateuszradomski, line 181: Why are we deleting this test?

packages/config/src/discovery/ProjectDiscovery.ts
```diff
@@ -646,8 +654,8 @@ export class ProjectDiscovery {
       )
       .map((contract) => contract.name)
     return safesWithThisMember.length === 0
-      ? []
-      : ['Member of ' + safesWithThisMember.join(', ') + '.']
+      ? undefined
+      : '* Member of ' + safesWithThisMember.join(', ') + '.'
```
mateuszradomski, line 658: It's better to use the format string instead of the addition operator

packages/config/src/discovery/PermissionsFromDiscovery.ts
```diff
@@ -76,7 +68,7 @@ export class PermissionsFromDiscovery implements PermissionRegistry {
           .map((d) => formatPermissionDelay(Number(d)))
           .join(' or ')
         return [
-          `Can upgrade **${delaysString}**`,
+          `* Can upgrade **${delaysString}**`,
```
mateuszradomski, line 71: Some descriptions have dots at the end and some don't. I don't know if it's intentional

mateuszradomski (approved):

packages/config/src/discovery/ProjectDiscovery.ts
adamiak: fixed

packages/config/src/discovery/PermissionsFromDiscovery.ts
adamiak: Yes, it's intentional. A checked with Claude 4 and when there are bullet-points listing things you're not supposed to end them with a period (dot).

PR #7832, opened 2025-05-20
Implement prefetching in REWRITE

torztomasz:


packages/frontend/rewrite/src/pages/scaling/project/getScalingProjectData.ts
```diff
@@ -58,6 +75,7 @@ export async function getScalingProjectData(
       props: {
         ...appLayoutProps,
         projectEntry,
+        dehydratedState: helpers.dehydrate(),
```
sz-piotr, line 78: reactQueryState

PR #7840, opened 2025-05-21
Update post-Pectra calldata costs logic

maciekzygmunt:
Resolves L2B-9267


packages/backend/src/modules/tracked-txs/utils/calculateCalldataGasUsed.ts
```diff
@@ -0,0 +1,22 @@
+export function calculateCalldataGasUsed(
+  blockNumber: number,
+  dataLength: number,
+  nonZeroBytes: number,
+  gasUsed: number,
+) {
+  const zeroBytes = dataLength - nonZeroBytes
+  const standardCalldata = 16 * nonZeroBytes + 4 * zeroBytes
+
+  if (blockNumber < 22431084) {
```
maciekop-l2b, line 10: instead of comment let's introduce const `PECTRA_START_BLOCK` or similar

maciekop-l2b (changes requested):

maciekop-l2b (approved):

PR #7841, opened 2025-05-21
Use cookies for recategorisation preview

torztomasz:


packages/frontend/src/consts/cookies.ts
```diff
@@ -6,6 +6,11 @@ import { z } from 'zod'
  */
 export const knownCookies = {
   example: knownCookie('example', z.string(), 'example'),
```
sz-piotr, line 8: Remove example or turn it into a comment

sz-piotr (approved):

PR #7850, opened 2025-05-21
Add tools api with a new decoder

sz-piotr:
If you want to check it out you need to run it locally.
1. Go to `/packages/tools-api` and run `pnpm dev`
2. Go to `/packages/tools` and run `pnpm start`
3. Open the website go to `/decoder-new`
4. Use this tx hash [0x0fdae8349cb42d2eb465a627468ae0bf5ea13302030ea426d353a271e9ae78a0](https://etherscan.io/tx/0x0fdae8349cb42d2eb465a627468ae0bf5ea13302030ea426d353a271e9ae78a0) as a demo


packages/tools-api/package.json
```diff
@@ -0,0 +1,35 @@
...
+  "main": "./build/data.json",
+  "types": "./index.d.ts",
+  "scripts": {
+    "build": "tsc",
+    "clean": "rm -rf build",
+    "format": "biome format .",
+    "format:fix": "biome format --write .",
+    "lint": "biome check --formatter-enabled=false .",
+    "lint:fix": "biome check --formatter-enabled=false --write .",
+    "start": "node -r source-map-support/register build/index.js",
```
mateuszradomski, line 14: Running `start` locally does not work because `source-map-support` is not installed. I don't know if it's intendent

mateuszradomski (approved):

packages/tools-api/package.json
sz-piotr: Must be a weird leftover from something. I'll remove this.

PR #7858, opened 2025-05-22
Migrate to Zod v4

mateuszradomski:
Resolves L2B-9940

Small speed up for Zod heavy things

What | Before ms | After ms | Speed up
-- | -- | -- | --
l2b colorize | 430 | 297 | 144.78%
config build | 1770 | 1620 | 109.26%


packages/config/src/tvs/json/schema/tvs-config-schema.json
```diff
@@ -13,12 +13,9 @@
         "type": "object",
         "properties": {
           "mode": {
-            "type": "string",
             "enum": ["auto", "custom"]
           },
-          "id": {
-            "type": "string"
-          },
+          "id": {},
```
sdlyy, line 18: Double check if it's intended - it won't yield editor hint saying value is incorrect.

sdlyy (approved):
Please do a final sweep and convert any remaining imports from 'zod' to 'zod/v4', where applicable. Other than that, everything looks good.

PR #7881, opened 2025-05-26
[L2B-10322] Add support for batch() in UOPS

maciekop-l2b:


packages/shared/src/uops/protocols/eip-7821/methods.ts
```diff
@@ -69,3 +80,43 @@ function executionModeId(mode: string): number {
...
   }
 }
+
+function decodeBatchInput(data: `0x${string}`): TransferOperation[] {
+  if (!data) {
+    return []
+  }
+
+  const reader = new BinaryReader(data)
+  reader.read(4)
```
torztomasz, line 90: maybe offset instead reading and voiding results?

```suggestion
  const reader = new BinaryReader(data, {offset: 4})
```

torztomasz (approved):

packages/shared/src/uops/protocols/eip-7821/methods.ts
maciekop-l2b: Done

PR #7899, opened 2025-05-27
Further UI improvements 

sdlyy:


mateuszradomski (approved):
We'll run out of dots soon :D

But for now it's a good solution

PR #7902, opened 2025-05-27
Fixes, zkSync plugin, condensed view

sz-piotr:


packages/tools-api/src/domain/encoding.ts
```diff
@@ -30,7 +30,27 @@ export function decodeType(type: string, encoded: `0x${string}`): AbiValue {
   return decodeParsed(parseType(type), encoded)
 }
 
+const DEBUG = false
```
mateuszradomski, line 33: Either leftover or extract it to some higher place that allows every file to see it/use it

mateuszradomski (approved):

PR #7936, opened 2025-05-29
feat: add SOPH native token

lgingerich:


sekuba:
thx! continued at #7937

PR #7941, opened 2025-05-30
[L2B-10276] Add support for events

maciekop-l2b:


packages/shared/src/providers/da/EthereumDaProvider.ts
```diff
@@ -18,10 +19,27 @@ export class EthereumDaProvider implements DaBlobProvider {
...
+    // to be able to track internal call we need to get logs
+    let logs: EVMLog[] = []
+    if (logFilters) {
+      const getLogs = []
+      for (const filter of logFilters) {
+        getLogs.push(
+          this.rpcClient.getLogs(filter.address, filter.topics, from, to),
+        )
+      }
+      logs = (await Promise.all(getLogs)).flat()
```
maciekzygmunt, line 36: `eth_getLogs` can take array of addresses, so maybe we can make it as just one query, with list of addresses and topics

maciekzygmunt (changes requested):

packages/shared/src/providers/da/EthereumDaProvider.ts
maciekop-l2b: Done

maciekzygmunt (approved):

PR #7945, opened 2025-05-30
Move Under review projects to a new tab

maciekzygmunt:
Resolves L2B-10277


packages/frontend/src/server/routers/InternalApiRouter/getHotPages.ts
```diff
@@ -12,7 +12,7 @@ export async function getHotPages() {
   const [scaling, bridges, daLayers, daBridges, latestValues] =
     await Promise.all([
       ps.getProjects({
-        select: ['scalingInfo'],
+        select: ['scalingInfo', 'statuses'],
         where: ['isScaling'],
       }),
```
torztomasz, line 17: why? its not used later

maciekzygmunt: Projects are later mapped using `getScalingTab` function which requires `statuses`

torztomasz (approved):

PR #7948, opened 2025-05-30
Add Template View - Readonly 

sdlyy:


packages/l2b/src/implementations/discovery-ui/templates/router.ts
```diff
@@ -5,12 +5,25 @@ import { z } from 'zod'
 import { createShape } from './create-shape'
 import { listDirectories } from './list-directories'
 
+const templateIdRegex = new RegExp(
+  '^(?!\\/)(?!.*\\/\\/)(?!.*\\s)(?!.*\\\\)(?:[a-zA-Z0-9]+([-_][a-zA-Z0-9]+)*\\/)*[a-zA-Z0-9]+([-_][a-zA-Z0-9]+)*\\/?$',
```
mateuszradomski, line 9: 😭

mateuszradomski (approved):
Question is do we want to show it in read only mode? Other than that LGTM

PR #7961, opened 2025-06-02
Add Logger with ES on FE

torztomasz:


packages/frontend/src/server/utils/logger.ts
```diff
@@ -0,0 +1,58 @@
...
+      elasticSearchModule.ElasticSearchTransport as ElasticSearchTransport
+    const options: ElasticSearchTransportOptions = {
+      node: env.ES_NODE,
+      apiKey: env.ES_API_KEY,
+      indexPrefix: env.ES_INDEX_PREFIX,
+      flushInterval: env.ES_FLUSH_INTERVAL,
+    }
+
+    loggerTransports.push({
+      // @ts-expect-error - dynamic import messes it up although it's correct
```
sz-piotr: maybe we can fix by using `await import`?

sz-piotr (approved):

PR #7980, opened 2025-06-03
Fix timeout handling

torztomasz:


packages/frontend/src/server/middlewares/ErrorHandler.ts
```diff
@@ -1,19 +1,15 @@
...
-    method: req.method,
-    url: req.originalUrl,
-    status: res.statusCode,
-  })
-  res.send('Internal Server Error')
+export function ErrorHandler(appLogger: Logger) {
+  return (err: Error, req: Request, res: Response, _: NextFunction) => {
+    res.status(500)
+    appLogger.error('Error processing request', {
+      error: err instanceof Error ? err.message : String(err),
```
sz-piotr, line 8: generate a uuid and add it to the log and send the UUID in the response so that when you get a 500 you know which log to search for

sz-piotr (approved):

PR #8005, opened 2025-06-04
Add deployment env

torztomasz:


Dockerfile.frontend
```diff
@@ -36,4 +36,4 @@ COPY --from=builder /src/ .
 COPY docker/.profile.d .profile.d
 RUN rm /bin/sh && ln -s /bin/bash /bin/sh
 WORKDIR /app/packages/frontend/
-CMD NODE_ENV=production node ./dist/server/index.js
```
sz-piotr, line 39: you still want to pass this one cause libraries behave differently

sz-piotr (approved):

Dockerfile.frontend
torztomasz: It will be set directly in Heroku envs

PR #8006, opened 2025-06-04
Pass some headers to logs

torztomasz:


packages/frontend/src/server/devReload.ts
```diff
@@ -2,12 +2,14 @@ import { createServer } from 'http'
...
+
 export function setupDevReload(logger: Logger) {
   const appLogger = logger.for('DevReload')
   const server = createServer()
   new WebSocketServer({ server })
 
-  server.listen(9999, () => {
-    appLogger.info(`Started at http://localhost:9999`)
+  server.listen(PORT, () => {
+    appLogger.info(`Started`, { port: PORT })
```
sz-piotr, line 13: Can you mention that this one is for the dev reload in the message? Got me confused once

torztomasz: well it is mentioned  
image

PR #8021, opened 2025-06-05
Improve project queries

torztomasz:


packages/database/src/other/aggregated-liveness/repository.ts
```diff
@@ -125,11 +126,15 @@ export class AggregatedLivenessRepository extends BaseRepository {
...
-      .groupBy(['projectId', 'subtype'])
 
     if (from) {
       query = query.where('timestamp', '>=', UnixTime.toDate(from))
     }
+    if (projectId) {
+      query = query.where('projectId', '=', projectId)
+    }
+
+    query = query.groupBy(['projectId', 'subtype'])
```
maciekop-l2b, line 137: let's not group by project if we're querying only for single project

PR #8023, opened 2025-06-05
Prefetch correctly

torztomasz:


packages/frontend/src/pages/scaling/ScalingRouter.ts
```diff
@@ -93,14 +94,21 @@ export function createScalingRouter(
...
         {
           key: ['scaling', 'projects', req.params.slug],
           ttl: 5 * 60,
           staleWhileRevalidate: 25 * 60,
         },
-        () => getScalingProjectData(manifest, req.params.slug, req.originalUrl),
+        () =>
+          getScalingProjectData(
+            manifest,
+            req.context.helpers,
```
sz-piotr: IMO the req.context is unnecessary. Just pass `getExpressHelpers()` here explicitly. Otherwise the "second idea" is cleaner, because you don't merge anything

sz-piotr (changes requested):

packages/frontend/src/pages/scaling/ScalingRouter.ts
sz-piotr: Just so that my thoughts about the entire thing are clear. For a request that needs helpers the helpers should of course be created once. Then anything that uses trpc receives those helpers as parameter and only the top-level call calls `.dehydrate()`. But using `req.context` doesn't bring any additional benefits

packages/frontend/src/pages/scaling/project/getScalingProjectData.ts
```diff
@@ -77,7 +60,7 @@ export async function getScalingProjectData(
       page: 'ScalingProjectPage',
       props: {
         ...appLayoutProps,
-        projectEntry,
+        projectEntry: projectEntry,
```
maciekzygmunt:
```suggestion
        projectEntry,
```

maciekzygmunt (approved):

PR #8066, opened 2025-06-11
Center node view on selected node

sdlyy:


packages/protocolbeat/src/panel-nodes/store/actions/selectAndFocus.ts
```diff
@@ -2,34 +2,43 @@ import type { State } from '../State'
 
 export function selectAndFocus(
   state: State,
-  selected: readonly string[],
+  selected: string,
+  container?: HTMLElement,
```
mateuszradomski: Maybe we don't need this argument, since I don't see any state where we would want to center based on a different container

PR #8073, opened 2025-06-11
Add ability to archive discovery configs

sdlyy:


packages/discovery/src/discovery/config/ConfigReader.ts
```diff
@@ -57,9 +57,18 @@ export class ConfigReader {
       )
     }
 
+    // Extract project-level properties (excluding 'chains', 'import', etc.)
+    const {
+      chains,
+      import: imports,
+      name: configName,
+      ...projectLevelProps
+    } = rawConfig
```
mateuszradomski: I wouldn't do it like this because then the config has every field that is configured. I'd just explicitly pass the `archived` flag, just like `chain` and `name` is done on lines 69-70

mateuszradomski (changes requested):

packages/discovery/src/discovery/config/ConfigReader.ts
sdlyy: https://github.com/l2beat/l2beat/pull/8073/commits/a4f6d0afd745d95d6e4f9e50f85c5f681d35ebe2

mateuszradomski (approved):

PR #8082, opened 2025-06-12
feat(cartesi-prt-honeypot): add cartesi-prt-honeypot

GCdePaula:
# Add Cartesi PRT Honeypot _(Stage 2 Rollup)_

This PR adds Cartesi PRT Honeypot, a newly deployed Cartesi appchain rollup with an operational fraud proof system.

It is submitted as a separate project from the original Honeypot, which was explicitly marked as immutable and non-upgradeable. By launching a v2, we preserve that intent and clearly separate the original experiment from this new stage-2-compliant deployment.

---

We got this error running `l2b discover ethereum cartesi-prt-honeypot`:

```
Starting discovery of cartesi-prt-honeypot on ethereum
1/2 0x4c1E74EF88a75C24e49eddD9f70D82A94D19251c Application
      ├─ P immutable
      ├─ R 0x6CE590b9F0697327f18c601DF6f0baE4a0801B68
      ├─ R 0x0000000000000000000000000000000000000000
      └─ E wasOutputExecuted - Too many values. Update configuration to explore fully
```

This error seems to be causing an error in `pnpm test`:

```
Error: Assertion Error: cartesi-prt-honeypot discovery.json includes errors. Run "l2b discover cartesi-prt-honeypot".
```

---

We’re available for any questions or clarifications :)

Thanks to the L2Beat team for your work! 💗


packages/config/src/projects/cartesi-prt-honeypot/cartesi-prt-honeypot.ts
```diff
@@ -0,0 +1,305 @@
...
+      description: 'Cartesi PRT Honeypot deployed to Ethereum mainnet',
+      type: 'general',
+    },
+  ],
+
+  badges: [
+    BADGES.VM.CartesiVM,
+    BADGES.VM.AppChain,
+    BADGES.DA.EthereumCalldata,
+    BADGES.Stack.Cartesi,
```
bbjubjub2494, line 303: Shouldn't this rollup have the based sequencing badge? There's no privileged actor as far as the sequencing contract is concerned...

felipeargento: Hey @bbjubjub2494 , I don't think so. You're right that there are no privileged actors, but the transactions are included individually in the smart contract. There are no L1 nodes building L2 blocks, batching transactions or anything of the sort...so my understanding is that it wouldnt count as based sequencing.

bbjubjub2494: Hi @felipeargento, that doesn't seem right to me. Observe that since the L1 builder decides the order of L1 transactions, they also decide the order of the subset of L1 transactions that are L2 inputs. Thus, they could extract MEV and issue preconfs, even though they are currently not able to do so afaik. They could also batch user ops, or anything like that. I don't think the fact that there's currently no public special software that L1 builders can run in order to be decode those transactions and be intentional about their power changes anything. Note that [the page about Facet](https://l2beat.com/scaling/projects/facet) has the based sequencing badge and Facet does basically the same thing in terms of sequencing (the only difference is there's no explicit contract on L1 just like OP chains, which is not significant imo)

felipeargento: Thats fair, to be honest I'm a bit confused about the definition of based rollups regarding the need of an actually implemented software that does what you mentioned.
But I do agree that the L1 builder does define the order of L2 inputs and that, in theory, they could extract MEV and issue preconfs. 

I thought that to qualify as a based rollup one would have to implement the software that L1 validators use to do those things...but if its just being compatible with it, then I think you're right.

Thanks for the explanation!

sekuba:
continued at #8093 (no write permissions here)

PR #8094, opened 2025-06-16
metal and mode: permissioned opfp

sekuba:


vincfurc:
why in config and not a template?

sekuba:
> why in config and not a template?

because template would mess with the permissionless ones

PR #8129, opened 2025-06-20
Recategorization cleanup

maciekzygmunt:


packages/config/src/processing/getProjects.ts
```diff
@@ -87,10 +77,7 @@ function layer2Or3ToProject(
       layer: p.type,
       type: p.display.category,
       capability: p.capability,
-      isOther:
-        p.display.category === 'Other' ||
-        (PROJECT_COUNTDOWNS.otherMigration < UnixTime.now() &&
-          !!p.reasonsForBeingOther),
+      isOther: p.display.category === 'Other' || !!p.reasonsForBeingOther,
```
sz-piotr, line 80: Should be just `p.display.category === 'Other'`. We can add a test that checks that if `Other` then we have reasons, otherwise we have no reasons

packages/frontend/src/server/features/scaling/utils/isProjectOther.ts
```diff
@@ -1,14 +1,8 @@
...
-export function isProjectOther(
-  scalingInfo: ProjectScalingInfo,
-  previewRecategorisation?: boolean,
-) {
-  const migrated = featureFlags.othersMigrated() || !!previewRecategorisation
+export function isProjectOther(scalingInfo: ProjectScalingInfo) {
   return (
-    (migrated &&
-      !!scalingInfo.reasonsForBeingOther &&
+    (!!scalingInfo.reasonsForBeingOther &&
```
sz-piotr, line 5: Just `scalingInfo.type === 'Other'`

sz-piotr (changes requested):

sz-piotr (approved):

PR #8231, opened 2025-07-01
[L2B-10703] Add daily report for real-time anomalies

maciekop-l2b:


packages/backend/src/modules/shared/notifiers/AnomalyNotifier.ts
```diff
@@ -0,0 +1,125 @@
...
+    interval: number,
+    z: number,
+    projectId: ProjectId,
+    subtype: TrackedTxsConfigSubtype,
+    block: Block,
+    latestRecord: RealTimeLivenessRecord,
+    latestStat: AnomalyStatsRecord,
+  ) {
+    const message =
+      `**${projectId}** stopped posting **${formatSubtype(subtype)}** - typically posts every **${formatDuration(latestStat.mean)}**, hasn't posted for **${formatDuration(interval)}**\n\n` +
```
maciekzygmunt: Maybe we can update it to the one propsed by Luca:
> nitpick but it should probably be "stopped batch submissions" instead of "stopped posting batch submissions"

maciekzygmunt (approved):
LGTM, but pls check the comment

packages/backend/src/modules/shared/notifiers/AnomalyNotifier.ts
maciekop-l2b: Applied

PR #8236, opened 2025-07-01
Stacked diff controls and display 

sdlyy:
GH diff looks ugly - but all I did is order and clean up


mateuszradomski (changes requested):
Let's have it on a single line. It takes too much vertical space

PR #8248, opened 2025-07-02
Implement ongoing anomalies

torztomasz:


packages/database/src/other/real-time-anomalies/repository.ts
```diff
@@ -45,6 +45,19 @@ export class RealTimeAnomaliesRepository extends BaseRepository {
...
   }
 
+  async getApprovedAnomaliesByProjectIds(
+    projectIds: string[],
+  ): Promise<RealTimeAnomalyRecord[]> {
+    const rows = await this.db
+      .selectFrom('RealTimeAnomaly')
+      .select(selectRealtimeAnomaly)
+      .where('status', '=', 'approved')
+      .where('projectId', 'in', projectIds)
```
maciekzygmunt: When `projectIds` will be empty array it will throw. Add if statement before query and return empty array if it is a case. Also pls add tests for this method

torztomasz: good catch

PR #8253, opened 2025-07-02
DiscoUI locally won't return the code unless it's up to date

mateuszradomski:
Resolves L2B-10397

### DIFF
--- a/packages/l2b/src/implementations/discovery-ui/getCode.ts
+++ b/packages/l2b/src/implementations/discovery-ui/getCode.ts
@@ -1,8 +1,10 @@
 import { existsSync, readFileSync, readdirSync } from 'fs'
 import { join } from 'path'
+import { isDeepStrictEqual } from 'util'
 import {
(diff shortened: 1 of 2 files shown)


packages/l2b/src/implementations/discovery-ui/getCode.ts
```diff
@@ -45,6 +74,17 @@ export function getCode(
...
+    configReader,
+    project,
+    address,
+    codePaths,
+  )
+
+  if (checkFlatCode && !flatCodeCurrent) {
+    throw new Error('Flat code is outdated')
+  }
+
```
sdlyy: Nit: we can spare some reads here - won't make a huge difference tho

```suggestion
  if (checkFlatCode) {
    if (!isFlatCodeCurrent(configReader, project, address, codePaths)) {
      throw new Error('Flat code is outdated')
    }
  }
```

sdlyy (approved):

PR #8255, opened 2025-07-02
[L2B-10746] Improve anomaly notifications

maciekop-l2b:
We want to only notify about anomalies which last more than configured duration (default 1h)

### DIFF
--- a/packages/database/prisma/migrations/20250702093752_add_notifications/migration.sql
+++ b/packages/database/prisma/migrations/20250702093752_add_notifications/migration.sql
@@ -1,0 +1,10 @@
+-- CreateTable
+CREATE TABLE "Notification" (
+    "id" VARCHAR(255) NOT NULL,
+    "channel" VARCHAR(255) NOT NULL,
+    "type" VARCHAR(255) NOT NULL,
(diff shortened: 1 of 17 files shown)


packages/database/prisma/migrations/20250702093752_add_notifications/migration.sql
antooni, line 1: I would opt for something more precise for table name e.g. RealitimeNotification

packages/database/src/other/notifications/select.ts
antooni, line 1: I know we have it in many parts of our repo, but I do not like this seperate select file pattern. Maybe we could just inline this code in repository.ts?

packages/database/prisma/migrations/20250702093752_add_notifications/migration.sql
maciekop-l2b: This was meant to be very generic to allow usage in all other features and communication channels

packages/database/src/other/notifications/select.ts
maciekop-l2b: hmm.. I was not the one that introduce it so can't properly justify why this pattern was used. Let's discuss in wider group

PR #8279, opened 2025-07-04
Turn Degate shutdown milestone to incident

lucadonnoh:


sekuba:
i made it a standard milestone because it was announced and intentional and users can exit, incident sounds more like something harmful or unintended no?

lucadonnoh:
good point, the way i see it is green gud and red bad, and shutting down imo bad. i have no strong opinion tho

radinat (approved):

PR #8289, opened 2025-07-07
Add ability to configure new ZK Catalog page and configs

maciekzygmunt:
Resolves L2B-10350

### DIFF
--- a/packages/config/src/common/attesters.ts
+++ b/packages/config/src/common/attesters.ts
@@ -1,0 +1,13 @@
+export interface ZkCatalogAttester {
+  id: string
+  name: string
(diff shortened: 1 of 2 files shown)


packages/config/src/common/attesters.ts
```diff
@@ -0,0 +1,13 @@
+export interface ZkCatalogAttester {
+  id: string
+  name: string
+  link: string
+}
+
+export const ATTESTERS = {
```
torztomasz: ZK_CATALOG prefix pls

torztomasz (approved):

PR #8295, opened 2025-07-07
Emit chain specific addresses in discovered.json

mateuszradomski:
Resolves L2B-10787


packages/backend/src/modules/update-monitor/DiscoveryRunner.ts
```diff
@@ -17,7 +17,7 @@ import {
   modelPermissions,
   toRawDiscoveryOutput,
 } from '@l2beat/discovery'
-import { assert, withoutUndefinedKeys } from '@l2beat/shared-pure'
+import { assert, rawAddress, withoutUndefinedKeys } from '@l2beat/shared-pure'
```
adamiak: I think we should follow the good example of EthereumAddress branded type, and scope functions that are obviously related into the type. So:
- ChainSpecificAddress.chain()
- ChainSpecificAddress.address()

I wouldn't mind that it's long. The truth is, we should almost never use them, especially the raw address.

packages/shared-pure/src/types/ChainSpecificAddress.ts
```diff
@@ -58,3 +59,26 @@ export function ChainSpecificAddress(value: string): ChainSpecificAddress {
...
+}
+
+export function fromParts(
+  shortChainName: string,
+  pureAddress: string | EthereumAddress,
+): ChainSpecificAddress {
+  return ChainSpecificAddress(`${shortChainName}:${pureAddress}`)
+}
+
+export function rawAddress(value: ChainSpecificAddress): EthereumAddress {
```
adamiak: Please add tests for all new functions.

packages/backend/src/modules/update-monitor/DiscoveryRunner.ts
```diff
@@ -218,7 +218,7 @@ function remapNames(
     }
 
     const matchingEntry = discoveryOutput.entries.find(
-      (e) => e.address === entry.address,
+      (e) => rawAddress(e.address) === entry.address,
```
adamiak: I think that we need to validate the chain part too. Previously we could just match addresses, but now that .entries will have a mix of different chains, just comparing raw address is probably a mistake.

packages/backend/src/modules/update-monitor/UpdateDiffer.ts
```diff
@@ -152,7 +152,7 @@ export class UpdateDiffer {
         const index = parseInt(indexString)
 
         const entry = latestContracts.find(
-          (e) => e.address === discoveryDiff.address,
+          (e) => rawAddress(e.address) === discoveryDiff.address,
```
adamiak: I find the use of rawAddress to be a "bad smell". Why would we ever need to disregard the chain part? Apparently we're mixing EthereumAddress with ChainSpecificAddresses here, and instead of ignoring the chain part, we should make sure that we get ChainSpecificAddresses in the first place.

packages/backend/src/modules/update-monitor/UpdateMonitor.test.ts
```diff
@@ -1049,7 +1050,7 @@ function mockContract(name: string, address: EthereumAddress): EntryParameters {
   return {
     type: 'Contract',
     name,
-    address,
+    address: fromParts('eth', address),
```
adamiak: Just like with with rawAddress comment above, using ChainSpecificAddress.from would be more consistent with our other branded types, and more readable here.

packages/config/src/discovery/ProjectDiscovery.ts
```diff
@@ -306,7 +312,7 @@ export class ProjectDiscovery {
 
     const contract = identifier.includes(':')
       ? this.getContractByChainSpecificAddress(ChainSpecificAddress(identifier))
-      : this.getContractByAddress(EthereumAddress(identifier))
+      : this.getContractByAddress(ChainSpecificAddress(identifier))
```
adamiak, line 313: This is a general comment for the whole `ProjectDiscovery.ts` file. 

Currently it has this hacky support for raw addresses and chain specific addresses. So it's full of hacky ternaries like the one above: `identifier.includes(':') ? it chain-specific : it's raw`

That was necessary because we used to handle a mix of such addresses.

But now we no longer need this at all since all addresses should be chain specific. So instead of just switching the type like here, all those ternaries and many functions should be removed and the rest refactored.  

It's not a simple change, but the otherwise the code above just makes no sense.

adamiak (changes requested):
Publishing my first part of comments. I've only gone through some of the files, so I'll review the rest a bit later.

packages/backend/src/modules/update-monitor/utils/findUnknownEntries.test.ts
```diff
@@ -1,12 +1,18 @@
...
+import { ChainSpecificAddress, EthereumAddress } from '@l2beat/shared-pure'
 import { expect, mockFn, mockObject } from 'earl'
 
 import { findUnknownEntries } from './findUnknownEntries'
 
-const A = { address: EthereumAddress.random() }
-const B = { address: EthereumAddress.random() }
-const C = { address: EthereumAddress.random() }
+const A = {
+  address: ChainSpecificAddress.from('eth', EthereumAddress.random()),
```
adamiak, line 8: Just a thought - we use `ChainSpecificAddress.from('eth, addr)` so much, that maybe it's worth adding `ChainSpecificAddress.eth(addr)` helper. But no strong opinion on this. It just would be shorter.

Btw, I think in the future we should rename `ChainSpecificAddress` to just `Address` :)

packages/config/src/discovery/starkware/getSHARPVerifier.ts
```diff
@@ -97,12 +98,15 @@ export function getSHARPVerifierContracts(
...
 export function getSHARPVerifierGovernors(
   projectDiscovery: ProjectDiscovery,
-  verifierAddress: EthereumAddress,
+  verifierAddress: ChainSpecificAddress,
 ): ProjectPermission[] {
   assert(
-    verifierAddress === SHARP_VERIFIER_PROXY.address &&
+    ChainSpecificAddress.address(verifierAddress) ===
+      SHARP_VERIFIER_PROXY.address &&
       getProxyGovernance(discovery, 'SHARPVerifierCallProxy')[0].address ===
```
adamiak, line 106: Instead of extracting the address on the left hand side, it would be better to return ChainSpecificAddress from getProxyGovernance. The same for every other change in this file.

packages/config/src/projects/abstract/ethereum/diffHistory.md
```diff
@@ -1,3 +1,177 @@
...
+Discovery rerun on the same block number with only config-related changes.
+
+## Config/verification related changes
+
+Following changes come from updates made to the config file,
+or/and contracts becoming verified, not from differences found during
+discovery. Values are for block 22865542 (main branch discovery), not current.
+
+```diff
+-   Status: DELETED
```
adamiak: I think this diffHistory updates are incorrect. They all show contracts being removed and then re-readded...

packages/config/src/projects/abstract/ethereum/diffHistory.md
```diff
@@ -1,3 +1,177 @@
...
+
+# Diff at Thu, 10 Jul 2025 08:15:56 GMT:
+
+- author: Mateusz Radomski (<radomski.main@protonmail.com>)
+- comparing to: main@b92b4c59c14c9cd2f4e072498dc1bcf695d33787 block: 22865542
+- current block number: 22865542
+
+## Description
+
+Discovery rerun on the same block number with only config-related changes.
```
adamiak, line 11: Please provide a meaningful description of the change.

packages/config/src/templates/opStack.ts
```diff
@@ -305,7 +305,7 @@ function opStackCommon(
       escrows: [
         templateVars.discovery.getEscrowDetails({
           includeInTotal: type === 'layer2',
-          address: portal.address,
+          address: ChainSpecificAddress.address(portal.address),
```
adamiak, line 308: ProjectDiscovery should operate on ChainSpecificAddresses, so there should be no need to extract address here, nor in any other place in this file. Same with orbitStack.ts and other stack .ts files. In particular, `ProjectDiscovery.getEscrowDetails` uses getContract internally, which should support ChainSpecificAddress, and so on.

packages/config/src/templates/utils.ts
```diff
@@ -60,7 +60,7 @@ export function safeGetImplementation(entry: EntryParameters): EthereumAddress {
   if (!implementation) {
     throw new Error(`No implementation found for ${entry.name}`)
   }
-  return implementation
+  return ChainSpecificAddress.address(implementation)
```
adamiak, line 63: Instead of extracting raw address, return ChainSpecificAddress and appropriately refactor the callers.

packages/discovery/src/discovery/analysis/TemplateService.ts
```diff
@@ -260,7 +265,7 @@ export class TemplateService {
       const sourcesHash = Hash256(hash)
       const matchingTemplates = this.findMatchingTemplatesByHash(
         sourcesHash,
-        contract.address,
+        ChainSpecificAddress.address(contract.address),
```
adamiak, line 268: Don't extract raw address. Pass ChainSpecificAddress and appropriately refactor `findMatchingTemplatesByHash`

packages/discovery/src/discovery/utils/extractorsPublic.ts
```diff
@@ -0,0 +1,48 @@
+// NOTE(radomski): There is a difference between the internals operate on raw
```
adamiak, line 1: Since the idea is for the whole discovery to operate only on chain-specific addresses (so that you never have to think what type of address you have), the discovery "internals" should not be using raw addresses. I strongly feel this file should not exists and whatever internals operate on EthereumAddress should be refactored to operate on ChainSpecificAddress.

packages/discovery/src/discovery/colorize/colorize.ts
```diff
@@ -25,7 +26,7 @@ export function colorize(
   for (const e of structure.entries) {
     const entryConfig = makeEntryColorConfig(
```
adamiak, line 27: I'm not sure why makeEntryColorConfig should receive on `EthereumAddress`. I'ts not used outside discovery. It should accept ChainSpecificAddress type.

adamiak (changes requested):
I have left some additional comments, but there is a general one that I didn't want to leave in many places. 

Discovery should operate on ChainSpecificAddresses internally exclusively, and fall back to EthereumAddress maybe in some edge cases when a function is exported and used outside of discovery. But I don't think we have many such cases - actually I didn't find a single one reviewing this PR. 

In many places this PR keeps internal functions accepting `EthereumAddress` and when they're called, the caller uses `ChainSpecificAddress.address()` to extract the raw address. This should be the other way around. Caller should pass ChainSpecificAddress, and the function should be refactored to accept it.

In general, I think there should be nearly zero uses of `ChainSpecificAddress.address(...)` and almost zero cases of internal discovery functions accepting or returning `EthereumAddress`.

PR #8297, opened 2025-07-07
Across v4 review

sekuba:
closes L2B-10812


vincfurc (approved):
to check if we can also change the folder and project id name to just Across

PR #8305, opened 2025-07-08
Add INTMAX (full)

lucadonnoh:
Resolves L2B-10707


packages/frontend/static/images/architecture/intmax.png
sergeyshemyakov, line 1: I would add an arrow from liquidity contract to user as well to show that the withdrawal flow goes from Liquidity on L1.
Also I propose to completely remove Claim contract as it is not really relevant for the interactions with the Intmax network.

packages/config/src/projects/intmax/config.jsonc
```diff
@@ -1,5 +1,11 @@
...
   "$schema": "../../../../discovery/schemas/config.v2.schema.json",
   "name": "intmax",
-  "import": ["../globalConfig.jsonc"]
+  "import": ["../globalConfig.jsonc"],
+  "maxAddresses": 200,
+  "initialAddresses": [
+    "eth:0xF65e73aAc9182e353600a916a6c7681F810f79C3",
+    "scr:0x22ac649b3229eC099C32D790e9e46FbA2CE6C9A5",
+    "scr:0x86B06D2604D9A6f9760E8f691F86d5B2a7C9c449"
+  ]
```
sergeyshemyakov, line 10: I suggest to add another initial address: `scr:0x6f09FFEb68e8914944eCe03BB487ac35e8a73911`. It is the address of [Block Building Registry](https://docs.network.intmax.io/smart-contracts#block-builder-registry) ([this contract in the docs](https://intmax-wallet.gitbook.io/intmax-developers-hub/nodes/smart-contracts#block-builder-registry))

packages/config/src/projects/intmax/intmax.ts
```diff
@@ -21,14 +36,160 @@ export const intmax: ScalingProject = underReviewL2({
       ],
       repositories: ['https://github.com/InternetMaximalism'],
       socialMedia: ['https://twitter.com/intmaxIO'],
+      bridges: ['https://app.intmax.io/bridge'],
```
sergeyshemyakov, line 41: They don't seem to have a block explorer for the mainnet yet, but maybe it would still make sense to link testnet explorer https://beta.testnet.explorer.intmax.io? But I don't have a strong opinion here, it is just for a testnet and could confuse someone

packages/config/src/projects/intmax/intmax.ts
```diff
@@ -21,14 +36,160 @@ export const intmax: ScalingProject = underReviewL2({
       ],
```
sergeyshemyakov, line 38: They have nice docs now: https://intmax-wallet.gitbook.io/intmax-developers-hub. Also this: https://docs.network.intmax.io

packages/config/src/projects/intmax/intmax.ts
```diff
@@ -21,14 +36,160 @@ export const intmax: ScalingProject = underReviewL2({
...
+          title: 'INTMAX Block Builder - INTMAX docs',
+          url: 'https://intmax-wallet.gitbook.io/intmax-developers-hub/intmax-block-builder',
+        },
+      ],
+    },
+    exitMechanisms: [
+      {
+        name: 'Users can autonomously exit their funds',
+        description:
+          'Users can autonomously exit by providing a ZK proof of sufficient balance. This requires keeping track of all funds received and sent.',
```
sergeyshemyakov: I would stress that the locked tokens are on Ethereum L1 (in Liquidity SC) since this is something Intmax highlight several times in their docs. I think they don't want to be seen as a full L3 because of this and it makes sense.

I think this section could be one fitting place, but maybe we can mention it somewhere else as well

sergeyshemyakov (changes requested):
Generally looks good!

packages/config/src/projects/intmax/config.jsonc
lucadonnoh: not adding this rn because showing as compromised and unverified in disco. its ok because not core

packages/config/src/projects/intmax/intmax.ts
lucadonnoh: yep i think it's confusing

PR #8335, opened 2025-07-09
Update market share label

maciekzygmunt:
Resolves L2B-10385


packages/frontend/src/pages/ecosystems/project/components/charts/EcosystemsMarketShare.tsx
```diff
@@ -10,7 +10,7 @@ export function EcosystemsMarketShare({
 
   return (
     <div className="font-medium text-[--ecosystem-primary] text-xs">
-      {formatPercent(marketShare)} market share
+      {formatPercent(marketShare)} scaling market share
```
torztomasz, line 13: does that fit?

PR #8336, opened 2025-07-09
SP1 config for zk catalog and shared modules

sergeyshemyakov:
This PR:

- Completes `ZkCatalogTags.ts` file by adding all zk catalog tags and their descriptions.
- Adds ZK Catalog config for SP1 (`sp1.ts`) with all information from the research
- Pulls SP1 Plonk discovery into a shared module, updates in all places that use the standard Plonk `SP1VerifierGateway`. Projects that refer to the verifier without the standard `SP1VerifierGateway` are not touched.


maciekzygmunt (approved):
LGTM

sekuba (approved):

PR #8358, opened 2025-07-11
Replace prettier with internal json formatting

mateuszradomski:
Resolves L2B-10901

-500KB package size
~15-20x faster


sdlyy (approved):
n1

PR #8367, opened 2025-07-14
[L2B-10905] Remove archived project's anomalies

maciekop-l2b:


packages/backend/src/modules/shared/processors/RealTimeLivenessProcessor.test.ts
```diff
@@ -518,18 +518,51 @@ describe(RealTimeLivenessProcessor.prototype.constructor.name, () => {
       expect(mockNotifier.anomalyRecovered).toHaveBeenCalled()
     })
   })
+
+  describe(RealTimeLivenessProcessor.prototype.deleteForArchivedProjects
+    .name, () => {
+    it('should match liveness txs and detect anomalies', async () => {
```
antooni: The name seems wrong

packages/backend/src/modules/shared/processors/RealTimeLivenessProcessor.test.ts
```diff
@@ -518,18 +518,51 @@ describe(RealTimeLivenessProcessor.prototype.constructor.name, () => {
       expect(mockNotifier.anomalyRecovered).toHaveBeenCalled()
     })
   })
+
+  describe(RealTimeLivenessProcessor.prototype.deleteForArchivedProjects
```
antooni: I would also test .init() function, possibly in the same test case so we cover them both

packages/backend/src/modules/shared/processors/RealTimeLivenessProcessor.ts
```diff
@@ -347,4 +353,28 @@ export class RealTimeLivenessProcessor implements BlockProcessor {
       } => c.params.formula === 'sharedBridge',
     )
   }
+
+  async deleteForArchivedProjects() {
+    const archivedProjectIds = this.trackedTxsConfig.projects
+      .filter((project) => project.isArchived)
+      .map((project) => project.id.toString())
+
+    const projectIdsInDb = await this.db.realTimeAnomalies.getProjectIds()
```
antooni, line 362: do we have an index for `projectId` column?

packages/backend/src/modules/shared/processors/RealTimeLivenessProcessor.ts
```diff
@@ -347,4 +353,28 @@ export class RealTimeLivenessProcessor implements BlockProcessor {
...
+    )
+
+    if (projectIdsToDelete.length === 0) {
+      return
+    }
+
+    const deletedCount =
+      await this.db.realTimeAnomalies.deleteByProjectId(projectIdsToDelete)
+
+    this.logger.info(`Deleted ${deletedCount} records for archived projects`, {
```
antooni, line 375: I would log only if deletedCount > 0

packages/backend/src/modules/shared/processors/RealTimeLivenessProcessor.ts
maciekop-l2b: Yes, we have

packages/backend/src/modules/shared/processors/RealTimeLivenessProcessor.test.ts
maciekop-l2b: fixed

packages/backend/src/modules/shared/processors/RealTimeLivenessProcessor.test.ts
maciekop-l2b: test added

packages/backend/src/modules/shared/processors/RealTimeLivenessProcessor.ts
maciekop-l2b: It will always be greater than zero if `projectIdsToDelete` is greater than zero

antooni (approved):

PR #8396, opened 2025-07-15
[L2B-10718] Show charts with MAX range for archived projects

maciekop-l2b:


packages/frontend/src/components/chart/liveness/ProjectLivenessChart.tsx
```diff
@@ -92,6 +96,7 @@ export function ProjectLivenessChart({
         anomalies={anomalies}
         configuredSubtypes={configuredSubtypes}
         hasTrackedContractsChanged={hasTrackedContractsChanged}
+        isArchived
```
torztomasz: that resolves to always true, what you want is:

```suggestion
        isArchived={isArchived}
```

packages/frontend/src/server/features/scaling/project/getScalingProjectEntry.ts
```diff
@@ -307,6 +307,7 @@ export async function getScalingProjectEntry(
         tokens,
         tvsProjectStats,
         tvsInfo: project.tvsInfo,
+        defaultRange: project.archivedAt ? 'max' : '1y',
```
torztomasz: well you need to use this range to fetch the data too, that's why I would go with separate files per chart section

packages/frontend/src/server/features/scaling/project/getScalingProjectEntry.ts
```diff
@@ -321,6 +322,7 @@ export async function getScalingProjectEntry(
         milestones: sortedMilestones,
         category: project.scalingInfo.type,
         projectName: project.name,
+        defaultRange: project.archivedAt ? 'max' : '1y',
```
torztomasz: well you need to use this range to fetch the data too, that's why I would go with separate files per chart section

packages/frontend/src/server/features/scaling/project/getScalingProjectEntry.ts
```diff
@@ -340,6 +342,7 @@ export async function getScalingProjectEntry(
         projectId: project.id,
         milestones: sortedMilestones,
         trackedTransactions,
+        defaultRange: project.archivedAt ? 'max' : '1y',
```
torztomasz: well you need to use this range to fetch the data too, that's why I would go with separate files per chart section

packages/frontend/src/utils/project/liveness/getLivenessSection.ts
```diff
@@ -60,5 +62,7 @@ export async function getLivenessSection(
     anomalies: liveness?.anomalies ?? [],
     hasTrackedContractsChanged,
     trackedTransactions,
+    defaultRange: range,
+    isArchived: project.archivedAt !== undefined,
```
torztomasz, line 66: what if project.archivedAt is equal to false?

maciekop-l2b: `project.archivedAt` is a number so can't be set to false ;)

packages/frontend/src/server/features/scaling/project/getScalingProjectEntry.ts
maciekop-l2b: fixed

packages/frontend/src/server/features/scaling/project/getScalingProjectEntry.ts
maciekop-l2b: fixed

packages/frontend/src/components/chart/liveness/ProjectLivenessChart.tsx
maciekop-l2b: Fixed

packages/frontend/src/server/features/scaling/project/getScalingProjectEntry.ts
maciekop-l2b: fixed

PR #8398, opened 2025-07-15
add pepeunchained2 full

sekuba:


packages/config/src/projects/pepeunchained2/pepeunchained2.ts
```diff
@@ -38,20 +46,24 @@ export const pepeunchained2: ScalingProject = underReviewL2({
...
     {
-      address: EthereumAddress('0xd3643255ea784c75a5325CC5a4A549C7CD62E499'), // bridge
-      sinceTimestamp: UnixTime(1748837581),
-      tokens: ['PEPU'],
-      chain: 'ethereum',
+      title: 'Mainnet migration',
+      description:
+        'Pepe Unchained migrates their optimium from op stack to orbit stack.',
+      url: 'https://x.com/pepe_unchained/status/1928485153132916993',
+      date: '2025-07-24T00:00:00Z',
```
sergeyshemyakov: The date is wrong for sure (it is in the future)
I suggest we take the timestamp of block #1 https://pepuscan.com/block/1, which would be `'2025-06-02T21:03:23Z'`.

sergeyshemyakov (commented):
Just one comment, PR could be merged after it is resolved

PR #8410, opened 2025-07-16
Pentagon chain full review

sekuba:
closes L2B-10943


packages/config/src/projects/penchain/penchain.ts
```diff
@@ -1,32 +1,139 @@
...
+        },
+      ],
+    },
   },
+  rollupModuleContract,
+  rollupVerifierContract: discovery.getContract('Verifier'),
+  isForcedBatchDisallowed,
+  chainConfig: {
+    name: 'penchain',
+    chainId: 623,
```
sergeyshemyakov: I think it should be 3344: https://chainlist.org/chain/3344

packages/config/src/projects/penchain/config.jsonc
```diff
@@ -2,5 +2,14 @@
   "$schema": "../../../../discovery/schemas/config.v2.schema.json",
   "name": "penchain",
   "import": ["../globalConfig.jsonc"],
-  "chains": {}
+  "chains": {
+    "ethereum": {
+      "sharedModules": ["shared-polygon-cdk"],
+      "initialAddresses": [
+        "0xB4cEb70E8778a9928feD6ECBa1b03706a57b0ce8", // validium
```
sergeyshemyakov: I think this is a wrong contract. IMO it should be [0xb1714954bbc0162a36fb44934f3216ace81c40d7](https://etherscan.io/address/0xb1714954bbc0162a36fb44934f3216ace81c40d7)
1. See here: https://bridge.pentagon.games/login
2. On Polygon rollup manager entry 14 in `rollupsDataV2` refers to the chain with id 623 and address `0xB4cEb70E8778a9928feD6ECBa1b03706a57b0ce8` but entry 15 is with id 3344 and address 0xb1714954bbc0162a36fb44934f3216ace81c40d7.

sergeyshemyakov:
I looked into the proofs of Pentagon, and it doesn't seem to use pessimistic proofs and SP1Verifier in particular. Only the polygon Fflonk verifier (so this is old pre-agglayer setup).
The state validation section of the Pentagon page now mentions the pessimistic proofs and SP1. IMO this is not too big of a problem, but technically this is incorrect. I think this is coming from the Polygon CDK shared module that includes SP1Verifier.

sekuba:
> I looked into the proofs of Pentagon, and it doesn't seem to use pessimistic proofs and SP1Verifier in particular. Only the polygon Fflonk verifier (so this is old pre-agglayer setup). The state validation section of the Pentagon page now mentions the pessimistic proofs and SP1. IMO this is not too big of a problem, but technically this is incorrect. I think this is coming from the Polygon CDK shared module that includes SP1Verifier.

u can steal via pessimistic proofs, that is why all agglayer chains have this explanation due to the additional trust assumption

packages/config/src/projects/penchain/config.jsonc
sekuba: sheeesh thank you, this is their testnet. i changed everything around because i thought the OTHER chain was their testnet, whyyyy

fixed now ( i hope)

sergeyshemyakov (approved):
LGTM!

PR #8418, opened 2025-07-17
Base rebrand

Barkin-Tuncer:


lucadonnoh (changes requested):
blocking this as im checking with them to confirm. they might also want to change the name to base chain

lucadonnoh (approved):

PR #8430, opened 2025-07-17
Archive OEV Network

Barkin-Tuncer:
Archiving OEVNetwork — the project is no longer active and its links (website/docs/X/etc.) are either down or no longer relevant.


lucadonnoh:
the project is still producing blocks and the explorer is up. you can also see it from the liveness section. maybe we just need to find better links, but the project seems to be live https://oev.explorer.api3.org/

PR #8453, opened 2025-07-21
Archive Hyphen

Barkin-Tuncer:
Hyphen has been decommissioned as of December 31, 2024.

image

### DIFF
--- a/packages/config/src/projects/hyphen/hyphen.ts
+++ b/packages/config/src/projects/hyphen/hyphen.ts
@@ -14,6 +14,8 @@ export const hyphen: Bridge = {
   display: {
     name: 'Hyphen',
     slug: 'hyphen',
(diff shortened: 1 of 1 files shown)


Barkin-Tuncer:
> please find an announcement and add a `headerWarning` that links to it. for ref see here https://l2beat.com/bridges/projects/near

Thanks for the feedback! I’ve checked both Twitter and Discord but couldn’t find a dedicated announcement—only the notice on the official website. Based on that, I’ve added a headerWarning with the link. Let me know if this works for you or if you'd prefer a different approach.

lucadonnoh (approved):

Barkin-Tuncer:
It seems like it timed out? I didn’t encounter any issues when running the test locally.

PR #8455, opened 2025-07-21
Remove Axonum

Barkin-Tuncer:
The website and explorer are no longer accessible, and both GitHub and Twitter haven't been updated about one year.


Barkin-Tuncer:
@lucadonnoh  Apologies! I missed deleting an unused image earlier, which caused the GitHub Action test to fail. I've fixed it now. Please let me know if there's anything else I should adjust.

PR #8469, opened 2025-07-22
[L2B-10926] Add ability to configure new token categories

maciekop-l2b:


packages/backend/scripts/tvs/execute.ts
antooni, line 1: There is a missing summary for rwa in return

antooni, line 1: There is a missing case for btc in switch

packages/config/src/tvs/json/schema/tvs-config-schema.json
```diff
@@ -708,7 +708,7 @@
             ]
           },
           "category": {
-            "enum": ["ether", "stablecoin", "other"]
+            "enum": ["ether", "stablecoin", "btc", "rwa", "other"]
```
antooni: shouldn't be rwaPublic and rwaPrivate?

packages/database/prisma/schema.prisma
```diff
@@ -282,6 +282,9 @@ model ProjectValue {
   stablecoin Float    @db.Real
   other      Float    @db.Real
   associated Float    @db.Real
+  btc        Float    @default(0) @db.Real
```
antooni: why is this default needed?

maciekop-l2b: because we're adding mandatory field to a table that already has records

packages/config/src/tvs/json/schema/tvs-config-schema.json
maciekop-l2b: ah, forgot to update schema after rename, will fix

packages/backend/scripts/tvs/execute.ts
maciekop-l2b: fixed

PR #8501, opened 2025-07-23
fix: update broken OP Stack docs link

reallesee:
replaced outdated OP Stack URL with the correct working link from the official BOB documentation


radinat:
Thanks, but we updated to a different link already. Closing the PR

PR #8511, opened 2025-07-24
Take into account sovereign projects in DA data size

maciekzygmunt:
Resolves L2B-11087

### DIFF
--- a/packages/frontend/src/server/features/data-availability/throughput/getDaThroughputTable.ts
+++ b/packages/frontend/src/server/features/data-availability/throughput/getDaThroughputTable.ts
@@ -327,11 +327,14 @@ async function getLargestPosters(
         ]
       }
 
-      if (sovereignProjectsNamesMap.has(largestPoster.projectId)) {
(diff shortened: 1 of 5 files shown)


packages/frontend/src/server/features/data-availability/throughput/getDaThroughputTable.ts
```diff
@@ -293,15 +316,30 @@ async function getLargestPosters(
...
+      if (sovereignProjectsNamesMap.has(largestPoster.projectId)) {
+        return [
+          daLayer,
+          {
+            name: sovereignProjectsNamesMap.get(largestPoster.projectId),
+            slug: undefined,
+            ...largestPoster,
+          },
+        ]
+      }
```
torztomasz:
```suggestion
      const sovereignProject = 	sovereignProjectsNamesMap.get(largestPoster.projectId)
      if (sovereignProject) {
        return [
          daLayer,
          {
            name: sovereignProject,
            slug: undefined,
            ...largestPoster,
          },
        ]
      }
```

one lookup this way

PR #8520, opened 2025-07-25
Show project type in search bar

maciekzygmunt:
Resolves L2B-11112


lucadonnoh:
i would show "upcoming" instead of type for upcoming projects. we plan to remove that type from upcoming anyway

PR #8532, opened 2025-07-25
Op upgrade 16 review

sekuba:


packages/config/src/projects/_templates/opstack/OptimismPortal2/shapes.json
```diff
@@ -34,5 +34,11 @@
     "address": "0x215A5fF85308A72A772F09B520dA71D3520e9aC7",
     "chain": "ethereum",
     "blockNumber": 22882513
+  },
+  "OptimismPortal2_v4_6_0": {
```
lucadonnoh: it's not true anymore that the game type is specified here

lucadonnoh (changes requested):

packages/config/src/projects/_templates/opstack/OptimismPortal2/shapes.json
sekuba: you mean because of this?
image
should be fine no?

sekuba: ah you mean for changing it? i can add this string to anchorstateReg

lucadonnoh: the method is never called anywhere and it's not used. the source of truth now is the anchor registry

lucadonnoh: this is what it was done before but now the check has been moved to the anchor registry https://disco.l2beat.com/diff/eth:0xB443Da3e07052204A02d630a8933dAc05a0d6fB4/eth:0xEFEd7F38BB9BE74bBa583a1A5B7D0fe7C9D5787a?lines=L3453

sekuba: migrated the description to anchorstate and linked the correct diff in diffHistory

lucadonnoh (approved):

PR #8544, opened 2025-07-28
Add % to past day data posted

lucadonnoh:


maciekzygmunt (changes requested):
You changed styling of this cell, prevously it was 15px, font medium, now it is 16px font bold.

Also what about adding text to tooltip explaining what this percentage is? Same as on TVS.

Image

lucadonnoh:
ok gimme one second
image

torztomasz (approved):

torztomasz (changes requested):
the cell formatting is still not fixed

lucadonnoh:
fixing, but why do we have different styles in the first place in columns? like summary of scaling and summary of dabeat

torztomasz (approved):

PR #8552, opened 2025-07-28
Rework charts to show current timestamp

torztomasz:


maciekzygmunt (commented):
Fix this no data issue
Image

And i guess here it also should have this gray background
Image

packages/frontend/src/components/chart/utils/getLastValidTimestamp.ts
```diff
@@ -0,0 +1,12 @@
...
+>(data: T[] | undefined): number | undefined {
+  if (!data) {
+    return undefined
+  }
+  const lastValidTimestamp = data.findLast(([_, ...rest]) =>
+    rest.every((v) => v !== null),
+  )?.[0]
+
+  return lastValidTimestamp
+}
```
maciekzygmunt: Remove it as it is only used in one place

maciekzygmunt (approved):

PR #8553, opened 2025-07-28
feat: improve error logging in ZkStackDataFetcher.safeContractCall me…

strmfos:
Enhanced error logging in the ZkStackDataFetcher's safeContractCall method to include the actual error message. Previously, the method was swallowing error details, making debugging difficult. The improved implementation now appends the original error message to the custom error message, providing more context for troubleshooting without changing the function's behavior


lucadonnoh:
not needed atm

PR #8567, opened 2025-07-29
Add sis chain

akitothemoon:
Related news: https://x.com/symbiosis_fi/status/1943261860855242930


packages/config/src/projects/sischain/sischain.ts
```diff
@@ -0,0 +1,46 @@
...
+export const sischain: ScalingProject = underReviewL2({
+  id: ProjectId('sischain'),
+  addedAt: UnixTime(1753778036),
+  capability: 'universal',
+  display: {
+    name: 'SIS Chain',
+    slug: 'sischain',
+    description:
+      'The SIS Chain is purpose-built for cross-chain activity – with fast execution, simple tooling, and native token utility at its core.',
+    category: 'ZK Rollup',
```
lucadonnoh: why zk rollup? i would assume than an orbit (arbitrum) stack would be optimistic

lucadonnoh (changes requested):

packages/config/src/projects/sischain/sischain.ts
akitothemoon: Ah you're right — my bad!I've already updated it. Thanks for pointing it out!

lucadonnoh (approved):

PR #8573, opened 2025-07-29
Add option to Show all in DABEAT throughput breakdown per project

maciekzygmunt:
Resolves L2B-11159


torztomasz:
PixelSnap 2025-07-29 at 13 56 13

what about this one?

PR #8597, opened 2025-07-30
Gateway migration review

sekuba:
Resolves L2B-11218, L2B-11231


sergeyshemyakov (approved):
Looks good as far as I can tell

PR #8608, opened 2025-07-31
Add CVEX chain

alejandrogogogo:
https://x.com/Jetstreamtrade/status/1908893340281774568


radinat:
The project has a different name now https://hub.conduit.xyz/crypto-valley-chain-0 and should be added as underReviewL3 instead as it's on mainnet.

ChaSerrano:
> The project has a different name now https://hub.conduit.xyz/crypto-valley-chain-0 and should be added as underReviewL3 instead as it's on mainnet.

Thanks for the heads-up. After checking, CVEX has rebranded to Jetstream Chain. I've made the corresponding update accordingly.

alejandrogogogo:
> The project has a different name now https://hub.conduit.xyz/crypto-valley-chain-0 and should be added as underReviewL3 instead as it's on mainnet.

Thanks for the heads-up. After checking, CVEX has rebranded to Jetstream Chain. I've made the corresponding update accordingly.

packages/config/src/projects/jetstreamchain/jetstreamchain.ts
```diff
@@ -0,0 +1,32 @@
...
+import type { ScalingProject } from '../../internalTypes'
+import { underReviewL3 } from '../../templates/underReview'
+
+export const jetstreamchain: ScalingProject = underReviewL3({
+  id: 'jetstreamchain',
+  capability: 'universal',
+  addedAt: UnixTime(1753947052),
+  hostChain: ProjectId('arbitrum'),
+  display: {
+    name: 'Jetstream Chain',
```
lucadonnoh: add a `shortName` with just "Jetstream" given that this name is already too long

lucadonnoh (changes requested):

packages/config/src/projects/jetstreamchain/jetstreamchain.ts
alejandrogogogo: Done. Added shortName: `Jetstream`.

lucadonnoh (approved):

PR #8658, opened 2025-08-04
Add Axelar to Bridges CLI

antooni:
Resolves L2B-10978


sekuba (approved):
merging because it includes changes to other decoders

PR #8693, opened 2025-08-05
Fix TVS Coingecko API issue

antooni:
Resolves L2B-11281


packages/shared/src/clients/coingecko/CoingeckoClient.ts
```diff
@@ -36,12 +36,14 @@ export class CoingeckoClient extends ClientCore {
     from: UnixTime,
     to: UnixTime,
   ): Promise<CoinMarketChartRangeData> {
+    const adjustedTo = to > UnixTime.now() ? UnixTime.now() : to
```
torztomasz, line 39: isnt 
```suggestion
    const adjustedTo = Math.min(to, UnixTime.now())
```
easier to read?

antooni: yup, thx

PR #8699, opened 2025-08-06
Interactive legends

torztomasz:


maciekzygmunt:
Adjust time range switch on ecosystems

torztomasz:
> On summary i think it should sum UOPS based on selected lines Image

Imo it should not. @radinat wdyt?

PR #8707, opened 2025-08-06
fix: update outdated SHARP aggregation link in Starknet description

Forostovec:
Replaced outdated link:
https://book.starknet.io/ch03-03-provers.html
with updated link:
https://docs.starknet.io/architecture/sharp/


lucadonnoh:
this page will be deprecated, see here https://l2beat.com/zk-catalog

PR #8739, opened 2025-08-07
Lido v2 discovery

lucadonnoh:


lucadonnoh:
closing but adding a ref to notion if someone wants to pick it up in the future

PR #8758, opened 2025-08-08
fix(multichain, omnichain): address sorting, Etherscan URL typos, and correct JSON file extension

fifalodm:
1. multichain/generateIntermediateConfig.ts:
   - Fixed potential runtime error in `.sort()` by converting `EthereumAddress` objects to strings before calling `localeCompare`.

2. omnichain/getLZ.ts: 
   - Removed stray `)}` from Etherscan API URLs in `fetchInternalTxs`, `fetchDeployer`, and `fetchNameAndABI` functions, ensuring valid HTTP requests.

3. omnichain/index.ts:
   - Changed output file extension from `.csv` to `.json` for `libraries` data, since the content is actually JSON and there are no `.csv` consumers in the repository.


lucadonnoh:
this is not causing problems right now

PR #8763, opened 2025-08-08
Update Stages links

maciekzygmunt:
Resolves L2B-11312


packages/frontend/src/pages/stages/getStagesData.ts
```diff
@@ -20,7 +20,6 @@ export async function getStagesData(
       manifest,
       metadata: getMetadata(manifest, {
         title: 'Stages - L2BEAT',
-        description: 'Page explaining the L2BEAT Stages framework.',
```
torztomasz, line 23: why did you get rid of the description?

torztomasz (changes requested):

PR #8768, opened 2025-08-08
tests(elastic-search): fix typo in mock param name

sashaodessa:
Renamed the mock parameter and its usages from `indextExist` to `indexExist` in `createClientMock`.


lucadonnoh:
already fixed

PR #8769, opened 2025-08-09
fix: correct typo in avalanche bridge description

hyperraza:
Change `loose` to `lose` in the risk description for avalanche bridge. This improves the grammar accuracy of the warning message.


lucadonnoh:
already fixed

PR #8777, opened 2025-08-09
Some typo fixes

lucadonnoh:


packages/uif/package.json
```diff
@@ -19,7 +19,7 @@
     "!dist/test"
   ],
   "scripts": {
-    "test": "mocha",
+    "test": "NODE_OPTIONS='--no-experimental-strip-types' mocha",
```
sekuba: is this intended?

lucadonnoh: tests are failing and this was one solution, but i’ll wait for website team to better undeestand what’s going on

PR #8807, opened 2025-08-12
fix: update broken Paradex Chain documentation link

aso20455:
- Old: https://docs.paradex.trade/documentation/paradex-chain/node-setup
+ New: https://docs.paradex.trade/paradex-chain/paradex-chain-rpc-node


lucadonnoh:
not the correct link. we need a link to node setup docs page.

aso20455:
@lucadonnoh I searched and it’s missing, what I suggested is the best alternative, but we can leave it as is. Thanks

PR #8825, opened 2025-08-12
Add Reference entry type in `discovered.json`

adamiak:
Resolves L2B-10986


adamiak:
Cancelled in favor of a different branch with the same changes

PR #8843, opened 2025-08-13
fix(logger): sync docs with code and add METRIC log color

dneptolus:
These changes were made to align documentation with the actual logger implementation and ensure consistency between code and docs.

- The `logger.tag()` method now requires an object with a `tag` key (code changed earlier, docs outdated).
- The default formatter in code is JSON, not `pretty` (docs were misleading).
- The default `colors` value is `true` in code, while docs stated `false`.
- Added `METRIC` log level color (`cyan`) in `LogFormatterPretty.ts` to maintain visual consistency with other log levels.

Without these updates, the documentation does not match the actual behavior of the logger, which can cause confusion for developers.

### DIFF
--- a/packages/backend-tools/src/logger/LogFormatterPretty.ts
+++ b/packages/backend-tools/src/logger/LogFormatterPretty.ts
@@ -72,6 +72,8 @@ export class LogFormatterPretty implements LogFormatter {
           return chalk.yellow(chalk.bold(level.toUpperCase()))
         case 'INFO':
           return chalk.green(chalk.bold(level.toUpperCase()))
+        case 'METRIC':
+          return chalk.cyan(chalk.bold(level.toUpperCase()))
         case 'DEBUG':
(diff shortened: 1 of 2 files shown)


torztomasz:
want to elaborate why the changes were made?

torztomasz:
if you provide some meaningful reasons I will re-open

dneptolus:
@torztomasz 

These changes were made to align documentation with the actual logger implementation and ensure consistency between code and docs.

- The `logger.tag()` method now requires an object with a `tag` key (code changed earlier, docs outdated).
- The default formatter in code is JSON, not `pretty` (docs were misleading).
- The default `colors` value is `true` in code, while docs stated `false`.
- Added `METRIC` log level color (`cyan`) in `LogFormatterPretty.ts` to maintain visual consistency with other log levels.

Without these updates, the documentation does not match the actual behavior of the logger, which can cause confusion for developers.

torztomasz (approved):

PR #8846, opened 2025-08-13
fix: update broken Celo Plumo link in zkCatalogTrustedSetups

honeypop-fe:
Replace broken celo.org/plumo link with working celo.org/papers/plumo link for ceremony details reference.


lucadonnoh:
the link you provided doesnt show ceremony participation

PR #8864, opened 2025-08-14
Disco very updates 08/14

sekuba:


packages/config/src/projects/lisk/ethereum/diffHistory.md
```diff
@@ -1,3 +1,26 @@
...
+
+# Diff at Thu, 14 Aug 2025 07:44:37 GMT:
+
+- author: sekuba (<29250140+sekuba@users.noreply.github.com>)
+- comparing to: main@200c2747a4a049cdea3746f37927303721bc165b block: 1753681871
+- current timestamp: 1755157463
+
+## Description
+
+Provide description of changes. This section will be preserved.
```
lucadonnoh:
```suggestion
Fee pricing update.
```

sekuba:
added the SC reference that metis sent via telegram
the 'sc correctly setup' is not true yet but we can already link to the members

PR #8867, opened 2025-08-14
docs: typo fixes

brawncode:
"Each file represent" → "represents".
"in a form" → "in the form".
"along with it's value" → "along with its value"
"when token was used" → "when a token was used".
"when token was bridged" → "when a token was bridged".


lucadonnoh:
already fixed

PR #8874, opened 2025-08-14
fix: add missing shebang to frontend build script

dizer-ti:
Adds #!/bin/bash shebang to ensure the script runs in bash interpreter.
The script uses bash-specific features like background processes (&) and wait command, so the shebang is necessary for reliable execution across different environments and CI systems.


lucadonnoh:
not needed

PR #8932, opened 2025-08-20
Op canonical tokens

sdlyy:
- [x] remove mog
- [x] handle case when source is explicitly set to 'canonical' - since it defeats the purpose of calling different things


packages/discovery/src/discovery/handlers/user/ERC20DataHandler/SubHandlers.ts
```diff
@@ -0,0 +1,57 @@
...
+  execute: (deps: ERC20SubHandlerDeps) => Promise<T> | T
+}
+
+type OpCanonicalResult = {
+  canonical: boolean
+  underlying?: ChainSpecificAddress
+}
+
+export const OpCanonical: ERC20SubHandler<OpCanonicalResult> = {
+  classify: ({ abi }) => {
```
adamiak: Check if the chain of the address is optimism. Otherwise it's an our-right false.

adamiak (approved):

PR #8959, opened 2025-08-21
Update TVS categories

maciekop-l2b:


radinat:
I reverted the Facet changes as they were not expected, we can investigate next week why they happened

PR #8964, opened 2025-08-21
Link table and token chart on TVS Breakdown page

torztomasz:


maciekzygmunt (changes requested):
Looks good but doesnt work on mobile, pls fix

maciekzygmunt (approved):

PR #8992, opened 2025-08-25
Discovery is now cross-chain internally

mateuszradomski:
Part of L2B-10790
Resolves L2B-11279

### DIFF
--- a/packages/backend/src/modules/update-monitor/UpdateMonitor.ts
+++ b/packages/backend/src/modules/update-monitor/UpdateMonitor.ts
@@ -174,8 +174,6 @@ export class UpdateMonitor {
       )
 
       for (const chain of chains) {
-        const { discovery } = discoveryPerChain[chain]
-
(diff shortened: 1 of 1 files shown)


packages/discovery/src/discovery/runDiscovery.ts
```diff
@@ -23,21 +28,25 @@ import type { DiscoveryOutput } from './output/types'
...
   config: DiscoveryModuleConfig,
-): Promise<Date> {
-  if (config.blockNumber !== undefined) {
-    const provider = new providers.StaticJsonRpcProvider(config.chain.rpcUrl)
-    return UnixTime.toDate(
-      (await provider.getBlock(config.blockNumber)).timestamp,
-    )
-  }
+): Date {
+  // TODO(radomski): I don't know how to handle discovery on a block with different chains
```
adamiak, line 35: I suggest - for now - to throw an error if someone passes a block number, instead of silently ignoring it.

adamiak (approved):

PR #8996, opened 2025-08-25
Minters script V1

sdlyy:
Resolves L2B-11470

### DIFF
--- a/packages/l2b/src/implementations/minters/getMinters.ts
+++ b/packages/l2b/src/implementations/minters/getMinters.ts
@@ -10,17 +10,17 @@ import {
 } from '@l2beat/shared-pure'
 import { utils } from 'ethers'
 
-const abi = [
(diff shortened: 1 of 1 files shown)


packages/l2b/src/commands/Minters.ts
```diff
@@ -0,0 +1,88 @@
...
+async function analyzeAll(
+  provider: IProvider,
+  logger: Logger,
+  transactions: string[],
+) {
+  const minters = new Set<string>()
+  let counter = 0
+  let nextPercentToLog = 10
+
+  await Promise.all(
```
adamiak: It's not safe to Promise.all here. In one run I got over 80k transactions. Calling Promise.all on all of them will most likely break something. I think that a normal loop is a safer bet for now.

adamiak (approved):

PR #9000, opened 2025-08-25
Improve TVS Breakdown token formula visuals

torztomasz:


packages/frontend/src/server/features/scaling/tvs/breakdown/getProjectTokensEntries.ts
```diff
@@ -142,27 +132,63 @@ function getEntries(
...
     return 'multiple'
   }
   if (addresses.length === 1 && addresses[0]) {
     const address = addresses[0]
-    const contractName = projectContracts?.[address.chain]?.find(
-      (c) =>
-        ChainSpecificAddress.address(c.address).toLowerCase() ===
-        address.address.toLowerCase(),
-    )?.name
     const explorer = chains.find((c) => c.name === address.chain)?.explorerUrl
```
maciekzygmunt: Maybe we should keep it and show escrow name in formula? Also please update this logic here, as i think we are no longer showing multiple?

maciekzygmunt (changes requested):
CHeck the comment

packages/frontend/src/server/features/scaling/tvs/breakdown/getProjectTokensEntries.ts
torztomasz: 1. It was not used cuz the argument was not passed at all so thats why I removed it
2. We do show the `multiple`
image

maciekzygmunt (approved):

PR #9025, opened 2025-08-27
Add No cap option for max throughput and show max registered

maciekzygmunt:
Resolves L2B-11533


lucadonnoh:
🗣️💯🚫🧢

packages/config/src/types.ts
```diff
@@ -632,7 +632,7 @@ export interface DaLayerThroughput {
    * Batch size for data availability. Together with batchFrequency it determines max throughput.
    * @unit B - bytes
    */
-  size: number
+  size: number | 'NO_CAP'
```
torztomasz, line 635: let's go with Number.PositiveInfinity

torztomasz (changes requested):

lucadonnoh:
is it possible to show the date as in activity for the max registered? 
image

torztomasz (approved):

torztomasz:
Please do check if the EXPLAIN ANALYZE of new query returns quick execution time

PR #9029, opened 2025-08-28
docs: replace TODO with technical description for ErasureCodingProof

aso20455:
Replace the TODO placeholder in DaFraudDetectionRisk.ts with a precisetechnical description explaining how erasure coding proofs work and why they receive a 'bad' security rating


lucadonnoh:
not needed

PR #9068, opened 2025-09-01
[L2B-8153] Use project config for UOPS Explorer

maciekop-l2b:


kondziorex7 (approved):

packages/uops-dashboard/src/chains.ts
```diff
@@ -1,3 +1,5 @@
+import chains from './chains.json'
```
antooni, line 1: chains.json is not commited

antooni (approved):

lucadonnoh (approved):

packages/uops-dashboard/src/chains.ts
maciekop-l2b: It's a built time artifact thus it's not being committed

PR #9083, opened 2025-09-01
Add missing AA methods (Base)

lucadonnoh:


sekuba:
this vercel preview of the tools does not recognize them yet but that is probably because it uses the old db

PR #9086, opened 2025-09-01
feat: add Zircuit explorer to project configuration

pipkaandres:
Add explorer.zircuit.com to the explorers list in the Zircuit scaling project configuration to provide users with access to the official block explorer for transaction and contract verification.


lucadonnoh:
it's already there

PR #9090, opened 2025-09-02
feat: add ZK Stack stacks to projects

aditya008-beep:


radinat:
Thanks for your willingness to contribute! The stack is inherited from the zkStackL2 template we're using for these projects, so it doesn't have to be specified in the project .ts files. I'll close the PR.

PR #9106, opened 2025-09-02
Create DABEAT Liveness page

maciekzygmunt:
Resolves L2B-11589


packages/config/src/types.ts
```diff
@@ -30,6 +30,13 @@ export interface TableReadyValue {
   orderHint?: number
 }
 
+export type TableReadyValueWithPredefinedValue<T extends string> = Omit<
+  TableReadyValue,
+  'value'
+> & {
+  value: T
+}
+
```
torztomasz: no need for separate type, just modify `TableReadyValue`:

```
export interface TableReadyValue<T extends string = string> {
  value: T
  secondLine?: string
  description?: string
  sentiment?: Sentiment
  warning?: WarningWithSentiment
  /** Taken into account when sorting. Defaults to 0. */
  orderHint?: number
}
```

packages/frontend/src/pages/data-availability/liveness/components/table/DaLivenessIntervalCell.tsx
```diff
@@ -0,0 +1,89 @@
...
+
+  if (!isClient) {
+    return <Skeleton className="h-6 w-[100px]" />
+  }
+
+  const durationInSeconds = data?.averageInSeconds
+
+  if (!durationInSeconds) {
+    return <NotApplicableBadge />
+  }
```
torztomasz: what if durationInSeconds = 0?

torztomasz (changes requested):

PR #9117, opened 2025-09-03
feat: remove Tusima project due to service unavailability

joey0293:
Tusima project website, bridge, and services are no longer accessible, and the project has not been updated for over a year.


lucadonnoh:
websites are accessible and twitter has not be inactive for over one year

PR #9123, opened 2025-09-03
Dailies (arb, bob, scroll, xai)

lucadonnoh:


sekuba:
added a handler for the scheduled verifier (BOB) because we will not get an alert as soon as it is added otherwise

PR #9141, opened 2025-09-04
Update b3.ts

erkidonik:
Added repo `https://github.com/b3-fun` in **packages/config/src/projects/b3/b3.ts**


lucadonnoh:
tests failing

PR #9149, opened 2025-09-04
Clarify call rate bound message

AdamDari99:
## Description

Refines the validation error in PositiveRpcBoundNumber to state “Call rate bound per minute must be a positive integer in [1, 1,000,000]”, making the inclusive endpoints and positivity explicit and fully aligned with the enforcement logic.


lucadonnoh:
not needed

PR #9155, opened 2025-09-08
Archive Kinto (Kinto is shutting down)

aditya008-beep:
Kinto [is shutting down](https://x.com/KintoXYZ/status/1964721235675537573).


lucadonnoh:
the warning is not descriptive enough. also the project is still processing deposits so it should not be archived yet.

PR #9170, opened 2025-09-08
Remove debug console.log from terminal store error handling

eeemmmmmm:
The error is already properly handled through:
- UI output display to user  
- Promise rejection for caller handling
- State cleanup for stream management

Debug console statements should not be present in production code as they can expose internal details and clutter browser console.


lucadonnoh:
not needed

PR #9178, opened 2025-09-09
fix: remove debug console.log statements from L2CostsUpdater

zeroprooff:
Remove leftover debug console.log statements from production code. The class already uses proper logging via this.logger.info().


lucadonnoh:
not needed

PR #9190, opened 2025-09-10
Add Onyx (full)

lucadonnoh:


packages/config/src/projects/onyx/diffHistory.md
```diff
@@ -0,0 +1,126 @@
+Generated with discovered.json: 0x62fc30863a8a2cff36b7baa63cd32deb8d89396e
+
+# Diff at Wed, 10 Sep 2025 13:27:44 GMT:
+
+- author: Luca Donno (<donnoh99@gmail.com>)
+- current timestamp: 1757509668
+
+## Description
+
+Discovery rerun on the same block number with only config-related changes.
```
sekuba: Image

sekuba (changes requested):
also created [this issue for later](https://linear.app/l2beat/issue/L2B-11715/eoa-referenced-but-not-listed-because-it-has-a-role)

packages/config/src/projects/onyx/diffHistory.md
lucadonnoh: same block as what? it's the first one

sekuba (approved):

PR #9200, opened 2025-09-11
Add Wormhole Token Bridge to Interop

luke7211:


packages/backend/src/modules/bridges/examples/examples.json
```diff
@@ -59,6 +59,26 @@
       "tx": "0x0eb86d1c6690d314971c0f1b038c06f278a7217a414e0b6ea28bcecb3f005743"
     }
   ],
+  "test": [
```
sz-piotr, line 62: Use a more descriptive name

PR #9201, opened 2025-09-11
Sort DABEAT liveness by TVS

maciekzygmunt:


packages/frontend/src/server/features/data-availability/liveness/getDaLivenessEntries.ts
```diff
@@ -34,22 +36,28 @@ export async function getDaLivenessEntries(): Promise<
     getLiveness(),
   ])
 
+  const uniqueProjectsInUse = getDaUsers(layers, bridges, [])
+  const tvsPerProject = await getDaProjectsTvs(uniqueProjectsInUse)
+  const getTvs = pickTvsForProjects(tvsPerProject)
```
torztomasz: move move getDaProjectTvs into Promise.all

maciekzygmunt: We cant as it depends on layers and bridges which are also async

torztomasz: move layers and bridges to separate promise.all, these calls are super fast so that you can group the heavier calls together

PR #9229, opened 2025-09-12
Add OP stack plugin (weak version but good enough)

lucadonnoh:


sekuba (approved):
as soon as we can identify the state root containing the withdrawal we could maybe say the state root finalization is the MESSAGE and the withdrawal finalization is the TRANSFER

PR #9236, opened 2025-09-14
Fix Array.isArray misuse and whatType prototype detection

AdamDari99:
Fix the misuse of Array.isArray on the wrong variable inside object/record validators, which could misclassify arrays and break strict checks.
Fix whatType() so it properly detects arrays and reads the prototype name from the actual value.


lucadonnoh:
not needed

PR #9267, opened 2025-09-16
Allow for initial discovery through the UI

sdlyy:


packages/discovery/src/discovery/config/ConfigWriter.ts
```diff
@@ -1,13 +1,19 @@
...
-  constructor(private configReader: ConfigReader) {}
+  constructor(public configReader: ConfigReader) {}
 
   writeConfigFile(project: string, contents: string) {
     const projectPath = this.configReader.resolveProjectPath(project)
     const configPath = join(projectPath, 'config.jsonc')
     writeFileSync(configPath, contents)
   }
+
+  createConfigFile(projectPath: string, contents: string) {
```
mateuszradomski: This probably shouldn't be an API for the ConfigWriter. `writeConfigFile` should check if the file exists, if it does not exist it should call createConfigFile. Otherwise the user is responsible for knowing to call `createConfigFile`

packages/l2b/src/implementations/discovery-ui/configs/router.ts
```diff
@@ -26,4 +49,111 @@ export function attachConfigRouter(app: Express, configWriter: ConfigWriter) {
...
+      maxAddresses: body.data.maxAddresses,
+    })
+
+    configWriter.createConfigFile(projectPath, formatJson(contents))
+
+    res.json({ success: true })
+  })
+}
+
+function configExists(configReader: ConfigReader, project: string) {
```
mateuszradomski: This should be a function of `ConfigReader`, also it shouldn't be implemented this way

packages/l2b/src/implementations/discovery-ui/configs/router.ts
```diff
@@ -26,4 +49,111 @@ export function attachConfigRouter(app: Express, configWriter: ConfigWriter) {
...
+      ...(maxDepth ? { maxDepth } : {}),
+      ...(maxAddresses ? { maxAddresses } : {}),
+      initialAddresses,
+    }
+
+    return config
+  },
+}
+
+const TokenConfigTemplate: ConfigTemplate = {
```
mateuszradomski: These things don't feel like they belong here. This file has many things inside that could be separate, not that urgent though

packages/l2b/src/implementations/discovery-ui/configs/router.ts
```diff
@@ -1,16 +1,39 @@
...
+  maxDepth: z.number().optional(),
+  maxAddresses: z.number().optional(),
+  overwrite: z.boolean().optional(),
+})
+
+export function attachConfigRouter(
+  app: Express,
+  configReader: ConfigReader,
+  configWriter: ConfigWriter,
+  paths: DiscoveryPaths,
```
mateuszradomski: This function ideally wouldn't accept paths. Since it's only used for ConfigWriter, maybe it's better to pass it there in the constructor and use it internally

PR #9344, opened 2025-09-19
Allow for other content in Publications

maciekzygmunt:
Resolves L2B-11804


packages/frontend/scripts/opengraph-images/projects/generateExternalPublicationThumbnails.tsx
```diff
@@ -0,0 +1,109 @@
+import { existsSync, mkdirSync, writeFileSync } from 'fs'
+import path from 'path'
+import { getCollection } from '~/content/getCollection'
+
+export async function generateExternalPublicationThumbnails() {
```
torztomasz, line 5: wouldnt it be easier to just enter the link manually?

PR #9348, opened 2025-09-20
fix: add radix parameter to Number.parseInt in blockCountInput

strmfos:
Add explicit radix parameter (10) to Number.parseInt call to follow best practices and prevent potential parsing issues. This ensures consistent decimal parsing behavior across different JavaScript engines


lucadonnoh:
not needed

PR #9358, opened 2025-09-22
Review Metis update

vincfurc:
Part of L2B-11932


packages/config/src/projects/metis/metis.ts
```diff
@@ -111,6 +111,10 @@ export const metis: ScalingProject = {
     },
     {
       rollupNodeLink: 'https://github.com/MetisProtocol/mvm',
+      securityCouncilReference:
+        'https://docs.metis.io/andromeda/network/council',
+      stage1PrincipleDescription:
+        'While Metis is considered Stage 1, the Security Council minority is employed to enforce censorship resistance in case the permissioned operator fails to include transactions. The process through which a censored user can contact the Security Council is not defined and currently unclear.',
```
sekuba: metis is not considered stage 1
- no forced transactions (sequencer cannot be changed by minority)
- no valid proof system (state roots can be injected by the game creator)
- proposer wrongly setup (2/8 should have the proposer role, but not be able to change proposers)
- principle not satisfied
- l2 contracts upgradable by EOA
- game creator can block the proof system regardless of above injection bug (currently metis multisig)

the current system also needs the 'can finalize malicious state roots' in the state validation slice, because the proposer can be set by the SC minority, who can then collude with metis multisig to propose and finalize malicious state

packages/config/src/projects/metis/metis.ts
```diff
@@ -99,8 +99,8 @@ export const metis: ScalingProject = {
       },
       stage1: {
         principle: false,
-        usersHave7DaysToExit: false,
-        usersCanExitWithoutCooperation: false,
+        usersHave7DaysToExit: true,
+        usersCanExitWithoutCooperation: true,
```
sekuba: `usersCanExitWithoutCooperation: false` because sequencer cannot be rotated by SC minority
`usersHave7DaysToExit: false` because the sc minority can affect liveness indefinitely, shortening any exit window, and l2 contract upgrades

packages/config/src/projects/metis/metis.ts
vincfurc: yep leftover, thanks

sekuba (approved):
approved with that comment about the sequencer failure slice
> op proof system should not be able to bypass the sequencer in the state proposal. ofc we do not know what they are proving at metis bc we don't know the program :/
> so maybe u can leave it like that

PR #9363, opened 2025-09-22
Show past upgrades for contracts

maciekzygmunt:
Resolves L2B-11803

### DIFF
--- a/packages/config/src/discovery/ProjectDiscovery.ts
+++ b/packages/config/src/discovery/ProjectDiscovery.ts
@@ -35,6 +35,7 @@ import type {
 } from '../types'
 import { RoleDescriptions } from './descriptions'
 import { get$Admins, get$Implementations, toAddressArray } from './extractors'
+import { pastUpgradesSchema } from './models'
 import type { PermissionRegistry } from './PermissionRegistry'
 import { PermissionsFromDiscovery } from './PermissionsFromDiscovery'
 import {
@@ -154,6 +155,7 @@ export class ProjectDiscovery {
         title: x.text,
         url: x.href,
       })),
+      pastUpgrades: getPastUpgrades(contract),
       ...descriptionOrOptions,
     }
   }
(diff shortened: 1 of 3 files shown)


packages/frontend/src/components/projects/sections/PastUpgradesDialog.tsx
```diff
@@ -0,0 +1,157 @@
...
+        </DialogContent>
+      </Dialog>
+      <Drawer>
+        <DrawerTrigger className="md:hidden">{trigger}</DrawerTrigger>
+        <DrawerContent className="max-h-[90dvh]">
+          <div className="flex items-center justify-between">
+            <DrawerTitle className="font-semibold text-label-value-18">
+              Past upgrades
+            </DrawerTitle>
+          </div>
```
torztomasz: DrawerHeader?

torztomasz (changes requested):

torztomasz:
Where is the bar?
image

Also, can we get rid of the "Close" button? We do not use it anywhere else and it looks really odd and redundant inside this component

torztomasz:
This needs a fix


https://github.com/user-attachments/assets/1e300d43-6cc4-4a7d-a797-3600a58487be

packages/frontend/src/components/projects/sections/PastUpgradesDialog.tsx
```diff
@@ -49,49 +65,82 @@ export function PastUpgradesDialog({
...
-            <DialogClose>
-              <CloseIcon className="size-4" />
-            </DialogClose>
-          </div>
-          <DialogDescription className="sr-only">
-            List of past upgrades
-          </DialogDescription>
-          <div className="space-y-3">
-            {pastUpgrades?.map((upgrade) => (
+        <DialogContent className="flex max-h-[90dvh] w-full flex-col overflow-y-hidden bg-surface-primary md:top-1/2 md:max-w-[720px]">
```
torztomasz: do you need that `md:top-1/2`, I think we talked about it that it should be the defualt inside `DialogContent`

packages/frontend/src/utils/project/contracts-and-permissions/getContractsSection.ts
```diff
@@ -239,7 +227,61 @@ function makeTechnologyContract(
...
+  return {
+    upgrades: pastUpgrades,
+    stats: {
+      count: pastUpgrades.length - 1,
+      avgInterval: intervals.length > 0 ? mean(intervals) : null,
+      lastInterval:
+        pastUpgrades.length > 1 ? UnixTime.now() - lastUpgrade.timestamp : null,
+    },
   }
 }
```
torztomasz: functions like this can be so much cleaner if you split them

torztomasz (changes requested):

packages/frontend/src/utils/project/contracts-and-permissions/getPastUpgradesData.ts
```diff
@@ -0,0 +1,76 @@
...
+            href: `${explorerUrl}/tx/${upgrade.transactionHash}`,
+          },
+          implementations: upgrade.implementations.map(
+            (implementation, implementationIndex) =>
+              getImplementationData(
+                explorerUrl,
+                implementation,
+                previousUpgrade?.implementations[implementationIndex],
+              ),
+          ),
```
torztomasz: if there is index used etc, just move the whole map into the function

torztomasz (approved):

PR #9414, opened 2025-09-25
[L2B-11875] Add logic to update TokenMetadata on BE start

maciekop-l2b:


packages/backend/src/modules/tvs/TvsModule.ts
```diff
@@ -239,7 +240,9 @@ export function initTvsModule({
     valueIndexers.push(projectValueIndexer)
   }
 
+  const tvsProjects = config.tvs.projects
   const start = async () => {
+    await updateTokenMetadata(tvsProjects, db, logger)
```
torztomasz, line 245: why not inlined?

maciekop-l2b: because it complains about `config.tvs` being of type `TvsConfig | false`. We do check for `false` on top but `start` function closure does not propagate it

PR #9422, opened 2025-09-25
Fix BlockIndexer dropping logs

sz-piotr:


packages/backend/src/modules/block-sync/BlockSyncModule.ts
```diff
@@ -50,7 +50,6 @@ export function createBlockSyncModule({
       parents: [blockNumberIndexer],
       blockProcessors: blockProcessors.filter((x) => x.chain === chain),
       source: chain,
-      mode: 'CONTINUOUS',
```
antooni, line 53: after this PR we will have only CONTINUOUS mode?

packages/backend/src/modules/block-sync/BlockIndexer.ts
```diff
@@ -34,16 +32,9 @@ export class BlockIndexer extends ManagedChildIndexer {
 
   override async update(from: number, to: number): Promise<number> {
     let adjustedFrom = from
-    if (
-      adjustedFrom !== to &&
-      (from === this.$.minHeight || this.$.mode === 'LATEST_ONLY')
-    ) {
+    if (adjustedFrom !== to && from === this.$.minHeight) {
```
antooni, line 35: shouldn't we delete this whole block of code given that we remove latest mode?

packages/backend/src/modules/block-sync/BlockIndexer.ts
```diff
@@ -55,31 +46,42 @@ export class BlockIndexer extends ManagedChildIndexer {
...
+    const [blocks, logs] = await Promise.all([
+      Promise.all(
+        blockNumbers.map((n) =>
+          this.$.blockProvider.getBlockWithTransactions(n),
+        ),
+      ),
+      this.$.logsProvider.getLogs(adjustedFrom, adjustedTo),
+    ])
+    const consistentBlocks = onlyConsistent(blocks, logs)
+    if (consistentBlocks.length === 0) {
```
antooni, line 67: shouldn't we check that `consistentBlocks.length !== blocks.length` ?

packages/backend/src/modules/block-sync/BlockSyncModule.ts
sz-piotr: Nobody ever used the other mode, I see no reason to support it

packages/backend/src/modules/block-sync/BlockIndexer.ts
sz-piotr: No, because this code ensures that we do not sync the entire history

packages/backend/src/modules/block-sync/BlockIndexer.ts
sz-piotr: No, the idea of this part of the code is that it can be less than blocks.length and we can still proceed up to that point

PR #9426, opened 2025-09-25
Upgrade Bridges Dashboard main page

antooni:


packages/database/src/repositories/BridgeMessageRepository.ts
```diff
@@ -83,31 +89,72 @@ export class BridgeMessageRepository extends BaseRepository {
...
+        sql<number>`percentile_cont(0.5) within group (order by duration)`.as(
+          'medianDuration',
+        ),
+      ])
+      .where('srcChain', 'is not', null)
+      .where('dstChain', 'is not', null)
+      .groupBy(['type', 'srcChain', 'dstChain'])
+      .execute()
+
+    return overallStats.map((overall) => ({
```
sz-piotr: There shouldn't be any business logic in repositories

PR #9434, opened 2025-09-26
Fix cctp match, add srcChain and dstChain to Mayan

luke7211:


packages/backend/src/modules/bridges/plugins/mayan-swift.ts
```diff
@@ -38,10 +49,43 @@ export class MayanSwiftPlugin implements BridgePlugin {
       })
     }
 
+    const abi =
```
sz-piotr, line 52: The code that follows is absolutely horrible and I'm only approving it with the idea that it will be refactored

PR #9445, opened 2025-09-26
Token Database

adamiak:
Resolves L2B-11568

### DIFF
--- a/packages/database/prisma/schema.prisma
+++ b/packages/database/prisma/schema.prisma
@@ -439,3 +439,48 @@ model EcosystemToken {
   @@id([projectId, coingeckoId, configurationId])
   @@index([projectId])
 }
+
+model AbstractToken {
+  id                        String    @db.Char(6)
+  issuer                    String?   @db.VarChar(255)
+  symbol                    String    @db.VarChar(255) 
+  category                  String    @db.VarChar(255)
+  iconUrl                   String?   @db.VarChar(255)
+  coingeckoId               String?   @db.VarChar(255)
+  coingeckoListingTimestamp DateTime? @db.Timestamp(6)
+  comment                   String?   @db.Text
+
+  deployedTokens            DeployedToken[]
+
+  @@id([id])
+}
+
+model DeployedToken {
+  id                  String            @db.VarChar(255)
+  abstractTokenId     String?           @db.Char(6)
+  symbol              String            @db.VarChar(255) 
+  decimals            Int
+  deploymentTimestamp DateTime          @db.Timestamp(6)
+  
+  abstractToken       AbstractToken?    @relation(fields: [abstractTokenId], references: [id], onDelete: Restrict)
+  connectionsFrom     TokenConnection[] @relation("TokenConnection_tokenFrom")
+  connectionsTo       TokenConnection[] @relation("TokenConnection_tokenTo")
+  
+  @@id([id])
+  @@index([abstractTokenId])
+}
+
+model TokenConnection {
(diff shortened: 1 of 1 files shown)


packages/token-service/src/commands.ts
```diff
@@ -0,0 +1,44 @@
...
+
+export interface AddAbstractTokenCommand {
+  type: 'AddAbstractTokenCommand'
+  record: AbstractTokenInsertable
+}
+
+export interface UpdateAbstractTokenCommand {
+  type: 'UpdateAbstractTokenCommand'
+  before: AbstractTokenInsertable
+  update: AbstractTokenUpdate
```
torztomasz: maybe like this?
```suggestion
  before: AbstractTokenInsertable
  after: AbstractTokenUpdate
```

packages/token-service/src/intents.ts
```diff
@@ -0,0 +1,42 @@
...
+  | DeleteAllDeployedTokensIntent
+
+export interface AddAbstractTokenIntent {
+  type: 'AddAbstractTokenIntent'
+  record: AbstractTokenInsertable
+}
+
+export interface UpdateAbstractTokenIntent {
+  type: 'UpdateAbstractTokenIntent'
+  update: AbstractTokenUpdate
```
torztomasz: same here, if `Add` has record, then maybe here it should be record too?

packages/token-backend/src/planning.ts
```diff
@@ -0,0 +1,193 @@
...
+
+function planDeleteAllDeployedTokens(
+  intent: DeleteAllDeployedTokensIntent,
+): Command[] {
+  return [
+    {
+      type: 'DeleteAllDeployedTokensCommand',
+    },
+  ]
+}
```
torztomasz, line 192: maybe we should have `/plans` directory and every plan should be a single file? Would be readable as hell imo

torztomasz (changes requested):

packages/token-backend/src/planning.ts
adamiak: I guess it's a matter of opinion. For such tiny functions I prefer them in one file, but as they grow we might want to split into separate files.

packages/database/src/repositories/AbstractTokenRepository.ts
```diff
@@ -1,19 +1,13 @@
...
-  type AsInsertable,
-  type AsSelectable,
-  type AsUpdate,
-  toRecord,
-} from '../utils/typeUtils'
 
-export type AbstractTokenInsertable = AsInsertable<AbstractToken>
-export type AbstractTokenSelectable = AsSelectable<AbstractToken>
-export type AbstractTokenUpdate = AsUpdate<AbstractToken, 'id'>
+export type AbstractTokenRecord = Selectable<AbstractToken>
```
torztomasz: we need to return UnixTime

packages/database/src/repositories/DeployedTokenRepository.ts
```diff
@@ -1,70 +1,64 @@
...
 
     return Number(result.numUpdatedRows)
   }
 
   async findByChainAndAddress(
-    chain: string,
-    address: string,
-  ): Promise<DeployedTokenSelectable | undefined> {
-    const result = await this.db
+    pk: DeployedTokenPrimaryKey,
```
torztomasz, line 72: never seen this pattern in our db, im fine with it but usally we just do chain:string, address: string

PR #9446, opened 2025-09-26
Fix dead links and harden external references across docs

prestoalvarez:
Replaced dead external links


radinat:
The provided link doesn't work

PR #9448, opened 2025-09-27
Update links in zkswap2.ts

eeemmmmmm:
Update website/bridge/twitter


radinat:
ZKBase is a different product and the ZKSwap page is already archived, so it doesn't have to be updated.

PR #9461, opened 2025-09-29
Add chain info to Events Dashboard

antooni:


packages/backend/src/modules/bridges/BridgeBlockProcessor.ts
```diff
@@ -26,17 +26,25 @@ export class BridgeBlockProcessor implements BlockProcessor {
...
+      for (const logToDecode of toDecode) {
         try {
           const event = await plugin.capture?.(logToDecode)
           if (event) {
-            events.push(event)
+            pluginEvents.push(event)
           }
         } catch (e) {
-          this.logger.error(e)
+          this.logger.tag({ project: plugin.name }).error(e)
```
sz-piotr:
```suggestion
          this.logger.error(e, { project: plugin.name })
```

packages/backend/src/modules/bridges/BridgeMatcher.ts
```diff
@@ -122,7 +122,7 @@ export async function match(
         try {
           result = await plugin.match?.(event, db)
         } catch (e) {
-          logger.error(e)
+          logger.tag({ project: plugin.name }).error(e)
```
sz-piotr: same here

PR #9465, opened 2025-09-29
Public API setup

torztomasz:


packages/public-api/src/OpenApi.ts
```diff
@@ -0,0 +1,306 @@
...
+      },
+      // TODO: add proper server url
+      servers: [{ url: 'http://localhost:3000' }],
+      tags: [
+        {
+          name: 'projects',
+          description:
+            'Endpoints for listing projects and retrieving detailed information about individual projects.',
+        },
+      ] satisfies { name: Tags; description: string }[],
```
sz-piotr: Maybe this could go in the constructor?

sz-piotr (approved):

packages/public-api/src/OpenApi.ts
torztomasz: had the same feeling, will do that

PR #9471, opened 2025-09-30
Add comparison for Interop

antooni:
image

Resolves L2B-11840


packages/backend/src/modules/bridges/BridgeComparator.ts
```diff
@@ -0,0 +1,74 @@
...
+    runCompare()
+    this.logger.info('Started')
+  }
+
+  async compare() {
+    await Promise.all(
+      this.plugins.map(async (plugin) => {
+        const items = await plugin.getExternalItems()
+
+        await new Promise((resolve) => setTimeout(resolve, this.waitingTimeMs))
```
sz-piotr: 1. Error handling. The current setup will be a nightmare to debug
2. I would do two different loops. One for getting the external items and one for checking. Then you don't have this setTimeout.

packages/backend/src/modules/bridges/BridgeModule.ts
```diff
@@ -46,15 +47,24 @@ export function createBridgeModule({
 
   const bridgeRouter = createBridgeRouter(db, config.bridges, processors)
 
+  const comparisonPlugins = createBridgeComparisonPlugins()
+
+  const bridgeComparator = new BridgeComparator(db, comparisonPlugins, logger)
+
   const bridgeCleaner = new BridgeCleaner(bridgeStore, db, logger)
 
+  console.log('here')
```
sz-piotr: not here

packages/backend/src/modules/bridges/compare/across.ts
```diff
@@ -0,0 +1,52 @@
...
+  types = ['across.Swap']
+
+  async getExternalItems(): Promise<BridgeExternalItem[]> {
+    const response = await fetch(
+      'https://explorer.range.org/transactions?max=100000000&b=across&sn=eth,base,arb1,oeth&dn=eth,arb1,base,oeth&tt=CROSSCHAIN&l=100',
+    )
+    const htmlContent = await response.text()
+
+    const detailedPairRegex =
+      /\\"sender_network\\":\\"([^\\"]*)\\".+?\\"sender_tx_hash\\":\\"([^\\"]*)\\".+?\\"sender_amount\\":([^,}]+).+?\\"sender_symbol\\":\\"([^\\"]*)\\".+?\\"receiver_network\\":\\"([^\\"]*)\\".+?\\"receiver_tx_hash\\":\\"([^\\"]*)\\".+?\\"receiver_amount\\":([^,}]+).+?\\"receiver_symbol\\":\\"([^\\"]*)\\"/g
```
sz-piotr: 😱  There has to be a better way

packages/database/src/repositories/BridgeMessageRepository.ts
```diff
@@ -133,6 +133,36 @@ export class BridgeMessageRepository extends BaseRepository {
     }))
   }
 
+  async getExistingItems(
+    items: {
+      srcTxHash: string
+      dstTxHash: string
+    }[],
```
sz-piotr: we're ignoring the chains? Maybe if we want a simple find we can even have one list of hashes?

antooni: we can include chains, but at the same time pair of hashes is unique enough

packages/backend/src/modules/bridges/BridgeComparator.ts
```diff
@@ -0,0 +1,113 @@
...
+            this.logger.warn(`Plugin for ${plugin.name} failed`, error)
+            return null
+          }
+        }),
+      )
+    ).filter((x) => x != null)
+
+    this.logger.info('Setting timeout...', { timeout: this.config.timeoutMs })
+    // This timeout is needed to make sure our backend indexes latest events
+    await new Promise((resolve) => setTimeout(resolve, this.config.timeoutMs))
```
sz-piotr: Again, I'd have a separate time loop. So one loop for getting the hash pairs and another for checking.
When you find a pair you remember the hashes but also remember the time.
When you check the pairs and you find matching ones you just remove them from the list of remembered.
After checking if there are leftover found pairs with age > 5/10 mins then alert

packages/backend/src/modules/bridges/BridgeComparator.ts
```diff
@@ -0,0 +1,129 @@
...
+    })
+
+    for (const [plugin, { items, type }] of Object.entries(this.pluginData)) {
+      if (items.length === 0) return
+      if (items.length === 0) continue
+
+      const records =
+        type === 'message'
+          ? await this.db.bridgeMessage.getExistingItems(items)
+          : await this.db.bridgeTransfer.getExistingItems(items)
```
sz-piotr: You can simplify this by not looping over plugins. Every plugin just collects items to the same list and tags them with plugin name. Then compare just gets all items.

Furthermore you can simplify this again by modifying the db query from `(from = a and to = b) or (from = c and to = d)` to `from in (a, c) and to in (b, d)`. There should be almost no false positives.

Then instead of doing `find` you can precompute a set from the existing items by concatenating `srcTxHash` and `dstTxHash` together to form the string in the set. After you create the set it's trivial to loop over and you avoid N^2 comparisons.

sz-piotr (approved):

PR #9500, opened 2025-10-02
Improve stages section alignments

maciekzygmunt:
Resolves L2B-11983

### DIFF
--- a/packages/frontend/src/components/projects/sections/StageSection.tsx
+++ b/packages/frontend/src/components/projects/sections/StageSection.tsx
@@ -196,7 +196,7 @@ export function StageSection({
                   {missingForLabel.length === 0 ? (
                     <div className="flex flex-col gap-3 md:flex-row">
                       <div className="flex items-center gap-2 font-bold">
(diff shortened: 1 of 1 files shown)


packages/frontend/src/components/projects/sections/StageSection.tsx
```diff
@@ -6,11 +6,11 @@ import type {
 } from '@l2beat/config'
 import { UnderReviewBadge } from '~/components/badge/UnderReviewBadge'
 import {
-  Accordion,
-  AccordionContent,
-  AccordionItem,
-  AccordionTrigger,
-} from '~/components/core/Accordion'
```
torztomasz, line 13: why change to collapsible?

torztomasz (changes requested):
Image

Code looks good but it still looks misaligned

packages/frontend/src/components/projects/sections/StageSection.tsx
maciekzygmunt: We are chainging design to our collapible component:

PR #9504, opened 2025-10-02
Add plugin name to Bridges DB tables

antooni:
Resolves L2B-11987


packages/backend/src/modules/bridges/plugins/types.ts
```diff
@@ -94,6 +97,7 @@ export function createBridgeEventType<T>(
     type,
     create(ctx: BridgeEventContext, payload: T): BridgeEvent<T> {
       return {
+        plugin: type.split('.')[0],
```
sz-piotr: No. This information should be added later in the thing that calls capture/match not here

sz-piotr (approved):

PR #9516, opened 2025-10-03
Pull only referenced part of discovery 

sdlyy:


packages/config/src/discovery/ProjectDiscovery.ts
```diff
@@ -49,60 +50,33 @@ const paths = getDiscoveryPaths()
...
+      // and incomplete data.
+      // TODO: refactor this whole logic around depenent projects and
+      // references to entrypoints to make it cleaner
+      this.discoveries.forEach((d) => removeReferences(d))
+
+      this.reachableEntries = getReachableEntries(
+        this.discoveries.flatMap((discovery) => discovery.entries),
+        entrypoints,
+      )
+    } catch {
```
adamiak, line 63: We shouldn't silently fail. Why is this necessary?

sdlyy: Legacy behavior - wicked tho, since tests fail all over the place. High time to fix it, I guess. Related to non-existing discoveries when making new instance of PD

PR #9547, opened 2025-10-07
Add Token Database tables (migrations only)

adamiak:
A migrations-only package (as required by CI) to create Token Database tables.


sz-piotr (approved):
LGTM. I am unsure about those typeUtils. I guess they should be used in other repositories as well?

adamiak:
> LGTM. I am unsure about those typeUtils. I guess they should be used in other repositories as well?

I'm trying something more consistent in the TokenDB repos for now. If it works out well, we can migrate other repos too.

PR #9589, opened 2025-10-09
Migrate transfers table to handle Financials

antooni:


packages/database/prisma/schema.prisma
```diff
@@ -399,25 +399,29 @@ model BridgeTransfer {
...
 
-  srcTokenAddress String?  @db.VarChar(66)
-  srcRawAmount    Decimal? @db.Decimal(80, 0)
-  srcSymbol       String?  @db.VarChar(10)
-  srcAmount       Float?   @db.Real
-  srcPrice        Float?   @db.Real
-  srcValueUsd     Float?   @db.Real
+  srcTokenAddress    String?  @db.VarChar(66)
+  srcRawAmount       Decimal? @db.Decimal(80, 0)
+  srcSymbol          String?  @db.VarChar(10)
```
sz-piotr, line 404: we could drop the symbols since we have the ids now

antooni: I have thought about it but let's leave them for now

PR #9620, opened 2025-10-13
chore: remove extra space in the comment

stemlaud:
remove extra space in the comment


radinat:
Thank you for the PR, but for now we do not accept such cosmetic changes.

PR #9667, opened 2025-10-15
Improve the error message for missing actors

mateuszradomski:
Resolves L2B-11400


sekuba (approved):
approving this bc it is beautiful

PR #9670, opened 2025-10-15
Rename bridges to interop

antooni:
Resolves L2B-12016


sz-piotr:
Let's merge this first thing in the morning

PR #9693, opened 2025-10-17
Add in memory cache to Public API

torztomasz:


packages/public-api/src/cache/InMemoryCache.ts
maciekzygmunt, line 1: I would move it to shared as it is exactly the same code that we have in FE

PR #9704, opened 2025-10-20
Use crypto.randomInt for generateIntId to reduce collisions and predictability

sashaodessa:
## Description
- Summary: Replace Math.random-based integer ID generation with crypto.randomInt in generateIntId.
- Motivation: Math.random offers low entropy and is predictable, increasing collision risk and harming reliability.
- Change: Import crypto.randomInt and use it to produce IDs in the 0..999 range.
- Impact: More robust, less predictable IDs; no public API changes; behavior preserved within the same range.


lucadonnoh:
not needed

PR #9726, opened 2025-10-21
Add "ZK Project" to recently added in search bar

maciekzygmunt:
Resolves L2B-12082


packages/frontend/src/components/search-bar/SearchBarDialog.tsx
```diff
@@ -284,7 +284,7 @@ function entryToLabel(entry: AnySearchBarEntry) {
     case 'da':
       return 'DA Layer'
     case 'zkCatalog':
-      return undefined
+      return 'ZK Project'
```
torztomasz, line 287: shouldn't it be ZK Catalog?
all of these entries are projects actually, its like saying Scaling Project instead Scaling

PR #9735, opened 2025-10-22
Add dataSource to activity config

torztomasz:


packages/frontend/src/utils/project/activity/getActivitySection.ts
```diff
@@ -22,5 +24,9 @@ export async function getActivitySection(
 
   return {
     defaultRange: range,
+    dataSource:
+      project.activityConfig.type === 'day'
+        ? project.activityConfig.dataSource
+        : undefined,
```
maciekzygmunt, line 30: Why this?

torztomasz: only type='day' has dataSource field

PR #9751, opened 2025-10-23
Add calldata to InteropEvent context capture

antooni:


packages/backend/src/modules/interop/engine/capture/InteropBlockProcessor.ts
```diff
@@ -88,6 +88,8 @@ function getLogsToDecode(chain: string, block: Block, logs: Log[]) {
         // biome-ignore lint/style/noNonNullAssertion: EVM tx should have it
         value: tx.value!,
         txTo: tx.to !== undefined ? Address32.from(tx.to) : undefined,
+        // biome-ignore lint/style/noNonNullAssertion: EVM tx should have it
+        calldata: tx.data! as string,
```
sz-piotr: Rename to txData. Also rename value to txValue

sz-piotr (approved):

PR #9754, opened 2025-10-23
Simple worker pool for update monitor 

sdlyy:
Runner tests have been removed since the only, single suite was responsible for testing retries. Something we no more care about.


packages/backend/src/config/Config.ts
```diff
@@ -185,6 +185,11 @@ export interface UpdateMonitorConfig {
   readonly disabledProjects: string[]
   readonly discord: DiscordConfig | false
   readonly updateMessagesRetentionPeriodDays: number
+  readonly workerPool: {
+    readonly count: number
```
adamiak: I'd call it `workerCount` for clarity

packages/backend/src/modules/update-monitor/createWorkers.ts
```diff
@@ -0,0 +1,184 @@
...
+      }, settings.timeoutPerRunMs)
+    })
+
+    let nextTaskIndex = 0
+
+    const worker = async (workerId: number) => {
+      logger.debug('Worker started', { workerId })
+
+      while (nextTaskIndex < tasks.length && !runTimeoutOccurred) {
+        const taskIndex = nextTaskIndex++
```
adamiak: This notation confuses me since always - it looks like taskIndex gets assigned incremented nextTaskIndex, while it - probably, I'm not sure ;) - get's assigned nextTaskIndex before the increment. Please break it into something more obviouse, like 

```
const taskIndex = nextTaskIndex
nextTaskIndex++
const task = tasks[taskIndex]
```

adamiak (approved):
Nice job! Approved but left two small comments.

PR #9782, opened 2025-10-27
Use real TokenDb in interop

adamiak:
What this PR does?

- add needed config for TokenDb connection
- extracts DeployedTokenId to separate file, removes AbstractTokenId
- integrates TokenDb into FinancialsLoop, updates test
- implement TokenDb, moves logging about missing tokens there
- renames messageId to transferId in InteropTransferRecord
- instantiates tokenDb client in InteropModule
- adds command `dev-pure` for local DX
- add token-backend dependency to backend
- adds new route to token backend `getByChainAndAddress`

Resolves L2B-12126, L2B-12312

### DIFF
--- a/packages/token-backend/src/trpc/routers/tokens.ts
+++ b/packages/token-backend/src/trpc/routers/tokens.ts
@@ -63,7 +63,7 @@ export const tokensRouter = router({
       })
       return result ?? null
     }),
-  getByChainAndAddress: protectedProcedure
+  getByChainAndAddress: readOnlyProcedure
     .input(v.array(v.object({ chain: v.string(), address: v.string() })))
     .query(async ({ input }) => {
       return await db.deployedToken.getByChainAndAddress(input)


packages/backend/src/modules/interop/engine/financials/TokenDb.ts
```diff
@@ -1,86 +1,68 @@
...
-export class MockTokenDb implements ITokenDb {
-  async getPriceInfo(deployedTokens: DeployedTokenId[]) {
-    await new Promise((r) => setTimeout(r, 1000))
-    const out = new Map()
-    for (const id of deployedTokens) {
-      const priceInfo = map.get(id)
-      if (priceInfo) {
-        out.set(id, priceInfo)
+    for (const d of deployedTokens) {
+      const tokenData = tokens.find(
```
sz-piotr: won't this be returned in the same order? Why do we need to .find?

packages/backend/src/modules/interop/engine/financials/TokenDb.ts
```diff
@@ -1,86 +1,68 @@
...
+    symbol: string
     decimals: number
     coingeckoId: string
   }
 >
 
-export interface ITokenDb {
-  getPriceInfo(deployedTokens: DeployedTokenId[]): Promise<PriceInfo>
-}
+export class TokenDb {
```
sz-piotr: IMO this entire class should go and the InteropFinancials should receive the client directly

packages/backend/src/modules/interop/engine/financials/TokenDb.ts
adamiak: Internally the call goes as a `SELECT` to DB, so there's no guarantee that the order will be the same (or that all tokens will be returned at all). Even if there were, IMHO it would be risky to assume that all future modifications there will preserve the ordering.

OTOH the loop can be done over returned tokens, removing the need to .find altogether. If logging missing tokens in this method is really necessary, it can be done as a separate, a bit more efficient step (e.g. a lookup in a set)

packages/database/src/repositories/DeployedTokenRepository.ts
```diff
@@ -68,6 +70,82 @@ export class DeployedTokenRepository extends BaseRepository {
...
+        'AbstractToken.issuer',
+        'AbstractToken.category',
+        'AbstractToken.iconUrl',
+        'AbstractToken.coingeckoId',
+        'AbstractToken.coingeckoListingTimestamp',
+        'AbstractToken.symbol as abstractSymbol',
+        'AbstractToken.comment as abstractComment',
+        'AbstractToken.reviewed',
+      ])
+      .where((eb) =>
```
adamiak, line 104: Let's always rename columns from joined table so that we don't accidently overwrite data from the other table. We can also use selectAll('DeployedToken') here.
```suggestion
    const result = await this.db
      .selectFrom('DeployedToken')
      .leftJoin(
        'AbstractToken',
        'AbstractToken.id',
        'DeployedToken.abstractTokenId',
      )
      .selectAll('DeployedToken')
      .select([
        'AbstractToken.id as AbstractToken_id',
        'AbstractToken.issuer as AbstractToken_issuer',
        'AbstractToken.category as AbstractToken_category',
        'AbstractToken.iconUrl as AbstractToken_iconUrl',
        'AbstractToken.coingeckoId as AbstractToken_coingeckoId',
        'AbstractToken.coingeckoListingTimestamp as AbstractToken_coingeckoListingTimestamp',
        'AbstractToken.symbol as AbstractToken_symbol',
        'AbstractToken.comment as AbstractToken_comment',
        'AbstractToken.reviewed as AbstractToken_reviewed',
      ])
      .where((eb) =>
```

packages/database/src/repositories/DeployedTokenRepository.ts
```diff
@@ -68,6 +70,82 @@ export class DeployedTokenRepository extends BaseRepository {
...
+              coingeckoId: row.coingeckoId,
+              coingeckoListingTimestamp: toTimestamp(
+                row.coingeckoListingTimestamp,
+              ),
+              comment: row.abstractComment,
+              reviewed: row.reviewed,
+            }
+          : undefined,
+    }))
+  }
```
adamiak, line 145: Don't re-implement mapping to a record here. We already have `toRecord()` for both tables.
```suggestion
    return result.map((row) => ({
      deployedToken: toRecord({
        symbol: row.symbol,
        comment: row.comment,
        chain: row.chain,
        address: row.address,
        abstractTokenId: row.abstractTokenId,
        decimals: row.decimals,
        deploymentTimestamp: row.deploymentTimestamp,
      }),
      abstractToken:
        row.AbstractToken_id === null ||
        row.AbstractToken_symbol === null ||
        row.AbstractToken_category === null ||
        row.AbstractToken_reviewed === null
          ? undefined
          : toAbstractTokenRecord({
              id: row.AbstractToken_id,
              issuer: row.AbstractToken_issuer,
              symbol: row.AbstractToken_symbol,
              category: row.AbstractToken_category,
              iconUrl: row.AbstractToken_iconUrl,
              coingeckoId: row.AbstractToken_coingeckoId,
              coingeckoListingTimestamp:
                row.AbstractToken_coingeckoListingTimestamp,
              comment: row.AbstractToken_comment,
              reviewed: row.AbstractToken_reviewed,
            }),
    }))
  }
```

adamiak:
LGTM

PR #9785, opened 2025-10-27
Change RPCClient param name

antooni:


packages/shared/src/clients/rpc/RpcClient.ts
```diff
@@ -46,7 +47,7 @@ export class RpcClient extends ClientCore implements BlockClient, LogsClient {
   multicallClient?: MulticallV3Client
 
   constructor(private readonly $: Dependencies) {
-    super($)
+    super({ ...$, sourceName: $.chain })
```
torztomasz, line 50: is this intended?

PR #9806, opened 2025-10-28
Add earnm as under review

vincfurc:
Resolves L2B-12294


vincfurc:
> blz run `pnpm tvs:generate earnm` in backend dir

empty, though we are missing EARNM token

sekuba (approved):
discord don't work tho

PR #9808, opened 2025-10-28
Add tooltip for max uops

maciekzygmunt:
Resolves L2B-12208


torztomasz:
Grandpa was supposed to do that

torztomasz (approved):

PR #9847, opened 2025-10-31
fix: set ApeX proofSystem to Validity and add zkCatalogId('stone')

Galoretka:
Updated to use Validity proofs with zkCatalogId: ProjectId('stone'). StarkEx-based systems use ZK validity proofs (not optimistic), and this aligns ApeX with other StarkEx configs (e.g., Immutable X, dYdX v3). The existing config already referenced ZK-based risk/state validation (STATE_ZKP_ST, STARKEX_VALIDITY_PROOFS), so this change resolves the inconsistency.


radinat:
The project is archived and no longer maintained.

PR #9860, opened 2025-11-03
Refactor Interop BigInt handling

antooni:
Resolves L2B-12382


packages/backend/src/modules/interop/engine/capture/InteropArgs.ts
```diff
@@ -0,0 +1,43 @@
+export const InteropArgs = {
+  serialize: serializeInteropArgs,
+  deserialize: deserializeInteropArgs,
+}
+
+function serializeInteropArgs(args: unknown): unknown {
```
sz-piotr: use https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/JSON/stringify#replacer instead

packages/backend/src/modules/interop/engine/capture/InteropArgs.ts
```diff
@@ -0,0 +1,43 @@
...
+    const result: Record<string, unknown> = {}
+    for (const [key, value] of Object.entries(args)) {
+      result[key] = serializeInteropArgs(value)
+    }
+    return result
+  }
+  return args
+}
+
+function deserializeInteropArgs(args: unknown): unknown {
```
sz-piotr: Use https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/JSON/parse#reviver instead

sz-piotr (approved):

PR #9875, opened 2025-11-04
Update error message for union type in validate

maciekzygmunt:
Before:

After:

### DIFF
--- a/packages/config/src/tvs/tvs.test.ts
+++ b/packages/config/src/tvs/tvs.test.ts
@@ -43,9 +43,7 @@ describe('tvs', () => {
       ],
     }
 
-    expect(() => ProjectTvsConfigSchema.parse(mockTvsConfig)).toThrow(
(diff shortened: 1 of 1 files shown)


packages/validate/src/validate.ts
```diff
@@ -485,18 +485,23 @@ function impUnion<
...
         ? element.safeParse(value)
         : element.safeValidate(value)
       if (result.success) {
         return result as Result<Infer<T[number]>>
       }
+
+      const location = result.path ? ` at ${result.path}` : ''
+      errors.push(`Variant ${i}${location}: ${result.message}`)
     }
+    const errorDetails = errors.length > 0 ? ` ${errors.join(' ')}` : ''
```
sz-piotr: Instead of this and the error array and handling the space in front why not just have a string message and say `message += ` Variant ${i}${location}: ${result.message}`. That way we don't need to join and worry about the space in front and check errors.length.

sz-piotr (approved):

PR #9877, opened 2025-11-04
Forknet review

sekuba:
close L2B-12380


packages/config/src/projects/forknet/forknet.ts
```diff
@@ -24,7 +43,113 @@ export const forknet: ScalingProject = upcomingL2({
   ecosystemInfo: {
     id: ProjectId('agglayer'),
   },
-  proofSystem: {
-    type: 'Validity',
+  chainConfig: {
+    name: 'forknet',
+    chainId: 8338,
+    explorerUrl: 'https://forkscan.org',
```
sergeyshemyakov, line 49: In the links above, still the testnet block explorer is given.

PR #9890, opened 2025-11-04
Discovery updates 11/04 basti

sekuba:


packages/config/src/projects/gpt/diffHistory.md
```diff
@@ -1,4 +1,4 @@
-Generated with discovered.json: 0x744a1289e74c127255028ba47b919df5e25c6b5f
```
sergeyshemyakov: Maybe remove this as there is no real diff?

PR #9909, opened 2025-11-05
feat(config): remove dead TemplateVars.fallback

Galoretka:
Removed fallback?: TableReadyValue from TemplateVars and cleaned up the unused import. The AnyTrust template always enforces fallback: DA_LAYERS.ETH_CALLDATA, and no callers provided a custom fallback. This avoids confusion and keeps the template’s type surface aligned with its behavior.


lucadonnoh:
no need

PR #9933, opened 2025-11-06
Update Haust Network: change status to under review and update links

sebastien-byte:
https://x.com/HaustNetwork/status/1983880977056141694 
HaustNetwork has shut down.


radinat:
It was never on mainnet, so it's to be completely removed rather than moved to Under review

sebastien-byte:
> It was never on mainnet, so it's to be completely removed rather than moved to Under review

Understood, I’ve removed it as suggested. @radinat

lucadonnoh:
just testnet

PR #9941, opened 2025-11-06
Add support for read-only access token

adamiak:
Resolves L2B-12313

### DIFF
--- a/packages/token-backend/src/trpc/trpc.ts
+++ b/packages/token-backend/src/trpc/trpc.ts
@@ -1,12 +1,25 @@
+import { TRPCError } from '@trpc/server'
 import { config } from '../config'
-import { generateProcedure, trcpRoot } from './generateProcedure'
+import { protectedProcedure, trcpRoot } from './protectedProcedure'
 
 export const router = trcpRoot.router
 
-export const readOnlyProcedure = generateProcedure(config, {
-  acceptReadOnlyToken: true,
+export const readOnlyProcedure = protectedProcedure(config).use((opts) => {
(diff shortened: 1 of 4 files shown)


packages/token-backend/src/trpc/trpc.ts
```diff
@@ -1,83 +1,25 @@
...
   }
-  const headers = opts.ctx.headers
-  const cookie = headers.get('cookie')
-  if (!cookie) {
-    throw new TRPCError({ code: 'UNAUTHORIZED' })
-  }
-
-  const cookies = parseCookies(cookie)
-  const token = cookies['CF_Authorization']
+  return opts.next()
```
torztomasz, line 14: I'm not sure but maybe we need to pass context here

packages/token-backend/src/trpc/trpc.ts
```diff
@@ -1,83 +1,25 @@
...
+    })
   }
-  const { payload } = decodedToken
-
-  return opts.next({
-    ctx: {
-      email: payload.email as string,
-    },
-  })
+  return opts.next()
```
torztomasz, line 24: same here

packages/token-backend/src/trpc/trpc.ts
adamiak: No needed, context is preserved via `opts.next`

packages/token-backend/src/trpc/trpc.ts
adamiak: No needed, context is preserved via `opts.next`

PR #9950, opened 2025-11-06
refactor(relay): handle empty data state in indexer update

forkfury:
Replace error throwing with returning `from` when no data is available to process.


lucadonnoh:
not the proper solution

PR #9970, opened 2025-11-07
Кeuse lodash chunk in DA project stats

soniseth0:
replace the bespoke `chunkArray` helper in `DaProjectStats` with lodash’s well-tested `chunk` drop the redundant local implementation to reduce maintenance surface while keeping behaviour intact


lucadonnoh:
not needed

PR #10018, opened 2025-11-11
docs: fill totalTimeUnrivaledCache field documentation

DeVikingMark:
Replace TODO comment with complete description of totalTimeUnrivaledCache field.

The description explains how the cached unrivaled time is used for time-based edge confirmation and how it can be updated through timer cache functions.


lucadonnoh:
that's external code. not needed

PR #10037, opened 2025-11-12
Add ability to edit config and templates via GUI

sdlyy:
Resolves L2B-9431

Legacy tags left untouched since we still display severity, category, and handler name.
Config and template control is now centralized and synced across the app - including raw text views.

`ignoreMethods` works only in one direction for now - we can't 'un-ignore' an ignored method since the method in question is not listed in the values panel - will do as a follow-up.

You can modify ANY template and only the MAIN project config via new tiles - we should support config switching based on the selected contract - to be done as a follow-up.

image


packages/protocolbeat/src/apps/discovery/hooks/useTemplateModel.ts
```diff
@@ -0,0 +1,98 @@
...
+  const saveMutation = useMutation({
+    mutationFn: async (content?: string) => {
+      if (!templateId) {
+        return
+      }
+      await writeTemplateFile(templateId, content ? content : templateString)
+    },
+    onSuccess: async () => {
+      await queryClient.invalidateQueries({
+        queryKey: ['template', templateId],
```
mateuszradomski, line 61: Is this correct? In useCurrentTemplate.ts it's `['projects', project, 'template', templateId, selectedAddress]`

packages/protocolbeat/src/apps/discovery/panel-values/Field.tsx
```diff
@@ -1,27 +1,101 @@
...
+                  state={x.isIgnored ? 'enabled' : 'disabled'}
+                />
+              ))}
+            </BadgeWrapper>
+          )}
+        </div>
       </div>
       {field.description && (
-        <div className="-mt-0.5 px-5 pb-1 font-serif italic">
+        <div className="-mt-0.5 word-break-break-word px-5 pb-1 font-serif italic">
```
mateuszradomski: typo?

packages/protocolbeat/src/apps/discovery/hooks/useTemplateModel.ts
sdlyy: Yeah we acutally care only about `'template'` and `templateId` - fixed, now in sync

packages/protocolbeat/src/apps/discovery/panel-values/Field.tsx
sdlyy: removed

PR #10051, opened 2025-11-12
Add groups to Discovery Dashboard

mateuszradomski:
Resolves L2B-11677


packages/backend/src/modules/update-monitor/api/view/groupProjects.ts
```diff
@@ -0,0 +1,115 @@
...
+    predicate: (p: DP) => boolean
+    variant?: GroupVariant
+  }
+
+  const groupConfigs: GroupConfig[] = [
+    // Stacks first
+    {
+      name: 'OP Stack',
+      assignees: ['🐿', '🐱'],
+      predicate: (p: DP) => isStack(p.name, 'OP Stack'),
```
sdlyy: You can remove `DP` type and explicit `(p: DP) => ...` in predicates completely since you're using `GroupConfig[]` literal type check. Just use plain `DashboardProject` inside `GroupConfig` type directly.

packages/backend/src/modules/update-monitor/api/view/DashboardPage.tsx
sdlyy, line 1: Image

Let's fix diff overflows - shouldn't be too hard

sdlyy, line 1: Image

Last two nits: let's try to add some margin on the right side if possible and MAYBE do something with the dropdown chevron - maybe we can remove it completely since we have a text telling you to click it? Up to you

sdlyy (approved):
Few nits, pure fire

sdlyy:
Also feel free to remove `packages/backend/src/modules/update-monitor/api/view/components/Components.tsx` - no longer used.

PR #10067, opened 2025-11-13
Handle BOB, Intmax

vincfurc:
resolves L2B-12495


vincfurc:
bumping this one in case it was forgotten @sekuba

PR #10076, opened 2025-11-13
Better loading states on TokenUI

torztomasz:


packages/token-ui/src/pages/tokens/TokensSummaryPage.tsx
```diff
@@ -35,10 +41,17 @@ export function TokensSummaryPage() {
...
               {isAbstractTokensLoading ? (
-                <LoadingText />
+                <Empty className="h-full">
+                  <EmptyHeader>
+                    <EmptyMedia variant="icon">
+                      <Spinner />
+                    </EmptyMedia>
+                    <EmptyTitle>Loading...</EmptyTitle>
+                  </EmptyHeader>
+                </Empty>
```
maciekzygmunt: It is used multiple times so maybe extract it to one component

maciekzygmunt (changes requested):

packages/token-ui/src/pages/tokens/TokensSummaryPage.tsx
torztomasz: oh right, I was about to do that and then forgot, thanks!

PR #10078, opened 2025-11-13
Add auto fill indicator to DeployedTokenForm

torztomasz:


packages/token-ui/src/utils/languageJoin.ts
maciekzygmunt, line 1: remove

PR #10079, opened 2025-11-13
chore: add Zora documentation URL to the project metadata

forkfury:
Add Zora documentation URL in GitHub to the project metadata


lucadonnoh:
not needed

PR #10084, opened 2025-11-14
Add filtering to the dashboard

mateuszradomski:
Resolves L2B-12507

### DIFF
--- a/packages/backend/src/modules/update-monitor/api/UpdateMonitorRouter.ts
+++ b/packages/backend/src/modules/update-monitor/api/UpdateMonitorRouter.ts
@@ -8,7 +8,19 @@ export function createUpdateMonitorRouter(
   const router = new Router()
 
   router.get('/status/discovery', async (ctx) => {
-    ctx.body = await updateMonitorController.getDiscoveryDashboard()
+    const queryEmoji = ctx.query.emoji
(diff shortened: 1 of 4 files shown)


packages/backend/src/modules/update-monitor/api/UpdateMonitorRouter.ts
```diff
@@ -8,7 +8,19 @@ export function createUpdateMonitorRouter(
   const router = new Router()
 
   router.get('/status/discovery', async (ctx) => {
-    ctx.body = await updateMonitorController.getDiscoveryDashboard()
+    const queryEmoji = ctx.query.emoji
+    const rawEmoji =
+      typeof queryEmoji === 'string'
+        ? queryEmoji
+        : Array.isArray(queryEmoji) && queryEmoji.length > 0
```
sdlyy, line 15: Is it possible?

sdlyy (approved):

packages/backend/src/modules/update-monitor/api/UpdateMonitorRouter.ts
mateuszradomski: You can manually add this in the path, just handle it even though not directly possible

PR #10133, opened 2025-11-19
fix(aptos): set destinationToken risk to WRAPPED

phrwlk:
Per packages/config/src/common/bridgeRiskView.ts, “Canonical” means the bridged asset is considered the canonical token on the destination chain, while “Wrapped” means it is not canonical.
The Aptos (LayerZero) bridge is explicitly described as lock-and-mint: “locks tokens in Ethereum escrow and mints tokens on Aptos,” which produces non-canonical wrapped assets on the destination chain.


sekuba:
please do not suggest changes of this kind as the bridges section is being re-worked right now, thanks!

PR #10138, opened 2025-11-19
Add janction project as upcomingL2

sebastien-byte:


packages/config/src/projects/janction/janction.ts
```diff
@@ -0,0 +1,30 @@
+import { UnixTime } from '@l2beat/shared-pure'
+import type { ScalingProject } from '../../internalTypes'
+import { upcomingL2 } from '../../templates/upcoming'
+
+export const jaction: ScalingProject = upcomingL2({
+  id: 'jaction',
```
torztomasz:
```suggestion
export const janction: ScalingProject = upcomingL2({
  id: 'janction',
```

torztomasz (dismissed):
fix id

sebastien-byte:
> fix id

Thanks for the review! I've updated the id accordingly.

sebastien-byte:
GM@sekuba! I just resolved the conflict.

PR #10152, opened 2025-11-20
Add support for burn/mint to InteropTransfer

antooni:
Resolves L2B-12138


packages/backend/src/modules/interop/plugins/types.ts
```diff
@@ -95,8 +95,8 @@ export interface InteropTransfer {
   plugin: string
   type: string
   events: InteropEvent[]
-  src: TransferSide
-  dst: TransferSide
+  src: TransferSide & { wasBurned?: boolean }
```
sz-piotr: why not change TransferSide?

PR #10140, opened 2025-11-19
Update dead link namechain.ts

strmfos:

PR #10207, opened 2025-11-26
Update dead link namechain.ts

strmfos:
Follow: https://github.com/l2beat/l2beat/pull/10140


radinat:
The new invite link is invalid.

PR #10216, opened 2025-11-26
Update source of smoke test's envs for staging apps

kondziorex7:


.github/workflows/smoke-test-mq.yml
```diff
@@ -11,6 +11,8 @@ jobs:
     env:
       HEROKU_ORGANIZATION: ${{ vars.HEROKU_ORG }}
       HEROKU_API_KEY: ${{ secrets.HEROKU_TOKEN }}
```
torztomasz, line 13: remove

kondziorex7: can't do, we still pull from heroku for production instances

PR #10219, opened 2025-11-26
fix: docs link for avalanche

phrwlk:
https://build.avax.network/docs return 404
https://build.avax.network/ - works


radinat:
The existing link works.

PR #10275, opened 2025-12-04
fix: typos in Wormhole NTT plugin and transaction comments

maximevtush:
Corrected `assosciated` to `associated`
Corrected `serialzed` to `serialized`


radinat:
Thank you, but at this stage we do not accept cosmetic fixes.

PR #10304, opened 2025-12-09
Added new tracked transactions for lens

sergeyshemyakov:
Resolves L2B-12621

### DIFF
--- a/packages/config/src/projects/lens/lens.ts
+++ b/packages/config/src/projects/lens/lens.ts
@@ -115,7 +115,7 @@ export const lens: ScalingProject = zkStackL2({
         firstParameter: EthereumAddress(
           '0xc29d04A93F893700015138E3E334eB828dAC3cef',
         ), // Lens diamond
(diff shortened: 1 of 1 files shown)


packages/config/src/projects/lens/lens.ts
```diff
@@ -105,6 +105,21 @@ export const lens: ScalingProject = zkStackL2({
...
+      },
+    },
+    {
+      uses: [{ type: 'l2costs', subtype: 'batchSubmissions' }],
+      query: {
+        formula: 'sharedBridge',
+        firstParameter: EthereumAddress(
+          '0xc29d04A93F893700015138E3E334eB828dAC3cef',
+        ), // Lens diamond
+        address: EthereumAddress('0x8c0bfc04ada21fd496c55b8c50331f904306f564'),
```
sekuba: this is the old valitimelock, new one is `0x2e5110cF18678Ec99818bFAa849B8C881744b776`

sergeyshemyakov: Ah, good catch, fixed

PR #10312, opened 2025-12-09
xlayer op stack migration

sekuba:
closes L2B-12651


packages/config/src/projects/xlayer/xlayer.ts
```diff
@@ -1,19 +1,40 @@
...
+  0
+
+const sequencer = discovery.getContractValue<ChainSpecificAddress>(
+  'SystemConfig',
+  'batcherHash',
+)
+
+const disputeGameFactory = discovery.getContract('DisputeGameFactory')
 
 export const xlayer: ScalingProject = agglayer({
```
lucadonnoh, line 31: no op stack badge?

sekuba: added

PR #10318, opened 2025-12-10
Dailies Orbit stack 10/12

vincfurc:


packages/config/src/projects/molten/molten.ts
```diff
@@ -16,6 +16,8 @@ export const molten: ScalingProject = orbitStackL3({
   additionalBadges: [BADGES.L3ParentChain.Arbitrum, BADGES.RaaS.Caldera],
   reasonsForBeingOther: [REASON_FOR_BEING_OTHER.CLOSED_PROOFS],
   display: {
+    redWarning:
+      'Critical contracts can be upgraded by an EOA which could result in the loss of all funds.',
```
lucadonnoh, line 20: but the executor cannot really upgrade alone though right?

lucadonnoh (changes requested):
left a comment

packages/config/src/projects/molten/molten.ts
vincfurc: there is no timelock or similar, this executor can directly upgrade calling execute/executeCall (same as Caldera msig)
image

lucadonnoh (approved):

PR #10328, opened 2025-12-10
Celo upgrade - Jello hardfork

vincfurc:


packages/config/src/templates/opStack.ts
```diff
@@ -905,6 +941,37 @@ The Kailua state validation system is primarily optimistically resolved, so no v
...
+    case 'OpSuccinctFDP': {
+      return {
+        categories: [
+          {
+            title: 'Validity proofs',
+            description:
+              'State roots are proposed by whitelisted proposers who create dispute games via the DisputeGameFactory by posting a bond. Once created, the game enters a challenge period during which whitelisted challengers can dispute the proposal by also posting a bond. If challenged, anyone can submit a ZK proof to prove the correct state within the proving period. After the challenge period passes without a successful challenge, or after a valid proof is submitted, anyone can resolve the game and finalize the state root.',
+            references: [
+              {
+                url: 'https://succinctlabs.github.io/op-succinct/architecture.html',
```
sergeyshemyakov: I would link to OP Succinct Lite here: https://succinctlabs.github.io/op-succinct/fault_proofs/fault_proof_architecture.html

packages/config/src/templates/opStack.ts
```diff
@@ -905,6 +941,37 @@ The Kailua state validation system is primarily optimistically resolved, so no v
         ],
       }
     }
+    case 'OpSuccinctFDP': {
+      return {
+        categories: [
+          {
+            title: 'Validity proofs',
+            description:
+              'State roots are proposed by whitelisted proposers who create dispute games via the DisputeGameFactory by posting a bond. Once created, the game enters a challenge period during which whitelisted challengers can dispute the proposal by also posting a bond. If challenged, anyone can submit a ZK proof to prove the correct state within the proving period. After the challenge period passes without a successful challenge, or after a valid proof is submitted, anyone can resolve the game and finalize the state root.',
```
sergeyshemyakov: I would make this explanation more precise by pulling challenge period and proving period from the contract (`maxChallengeDuration` and `maxProveDuration` from `OPSuccinctFaultDisputeGame`). Idk if we need extra steps to make it readable (3.5 days and 1 day).

Also the same about bonds: proposer bond is `initBondGame42` on `DisputeGameFactory` and challenger bond is `challengerBond` on `OPSuccinctFaultDisputeGame` (both currently are 0.01 ETH).

packages/config/src/templates/opStack.ts
```diff
@@ -1078,6 +1145,13 @@ function getRiskViewStateValidation(
         executionDelay: getFinalizationPeriod(templateVars),
       }
     }
+    case 'OpSuccinctFDP': {
+      return {
+        ...RISK_VIEW.STATE_ZKP_ST_SN_WRAP,
```
sergeyshemyakov: I think it should be `STATE_FP_1R_ZK` here, same as with Facet.

packages/config/src/templates/opStack.ts
```diff
@@ -116,7 +127,7 @@ export function EIGENDA_DA_PROVIDER(
         eigenDACertVersion,
         fallback?.value,
       ),
-      bridge: DA_BRIDGES.NONE,
+      bridge,
       badge: BADGES.DA.EigenDA,
```
sergeyshemyakov, line 131: On the FE the EigenDA badge says "EigenDA with no bridge". As I understand, Celo does use EigenDA with a bridge. So should we create a new badge "EigenDA with a bridge"?

packages/config/src/templates/opStack.ts
```diff
@@ -367,11 +378,22 @@ function opStackCommon(
       templateVars.nonTemplateProofSystem ??
```
sergeyshemyakov, line 378: On line 370 it says "Fraud proof system is currently under development. Users need to trust the block proposer to submit correct L1 state roots." which translates into a yellow warning in the risk summary section. I don't quite get what it means, probably it should be removed now with OP Succinct Lite?

packages/config/src/templates/opStack.ts
vincfurc: nice catch, I don't think we have one yet

packages/config/src/templates/opStack.ts
vincfurc: removed it from the template

PR #10330, opened 2025-12-11
chore: replaced broken bridge link

operagxsasha:
networks in eth no longer exist, only the BSC network and native ones remain

Знімок екрана 2025-12-11 о 12 39 44


radinat:
The project is already archived and not maintained.

PR #10348, opened 2025-12-14
Remove broken explorer links that return 404

Ray-Lin92:


radinat:
Seems like the project is no longer active, it can be removed instead

Ray-Lin92:
> Seems like the project is no longer active, it can be removed instead

Yep, agreed. Since the project is no longer active, I’ve already archived it.

PR #10394, opened 2025-12-17
Fix Space and Time badge

vincfurc:
resolves L2B-12688


sekuba:
this one is not posting calldata, but also not posting blobs. it is not live atm but we do not have tracked txs for most zk stacks

possible solutions:
- archive
- add trackedTxs for all zk stack

PR #10413, opened 2025-12-18
Fix icon alignments

lucadonnoh:
Before
image

After
image


lucadonnoh:
why are there so many hardcoded margins wtfffffff

lucadonnoh:
that looks great

PR #10416, opened 2025-12-18
Fix image link in Worldcoin SMTB doc

sashaodessa:
**Description**
- Update SMTB verification diagram to the correct static path


radinat:
The page is not shown and maintained.

PR #10427, opened 2025-12-19
Show count of projects under TVS value

maciekzygmunt:
Resolves L2B-12719


torztomasz (approved):
Maybe it would also be helpful to show project count on the TVS chart?

PR #10429, opened 2025-12-19
Add Megaeth Frontier

vincfurc:
Resolves L2B-12787


sekuba:
unsure how i can check for EigenDA verifier integration?

PR #10500, opened 2026-01-02
Move paradex to others

lucadonnoh:
Resolves L2B-12800


packages/config/src/projects/paradex/paradex.ts
```diff
@@ -341,8 +345,15 @@ export const paradex: ScalingProject = {
     ],
   },
   dataAvailability: {
-    layer: DA_LAYERS.ETH_BLOBS_OR_CALLDATA,
-    bridge: DA_BRIDGES.ENSHRINED,
+    layer: {
+      value: 'Privacy council',
+      sentiment: 'bad',
```
sekuba: sentiment should be 'warning' here

PR #10519, opened 2026-01-05
Add fade to detailed nav

maciekzygmunt:
Resolves L2B-12804

### DIFF
--- a/packages/frontend/src/components/ScrollWithGradient.tsx
+++ b/packages/frontend/src/components/ScrollWithGradient.tsx
@@ -3,6 +3,7 @@ import { cn } from '~/utils/cn'
 export function ScrollWithGradient({
   children,
   className,
+  ref,
   ...props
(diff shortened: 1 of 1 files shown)


packages/frontend/src/components/ScrollWithGradient.tsx
```diff
@@ -35,6 +36,12 @@ export function ScrollWithGradient({
     <div
       className={cn('overflow-y-auto', className)}
       ref={(node) => {
+        // Forward ref to parent
+        if (ref) {
+          if (typeof ref === 'function') ref(node)
```
torztomasz, line 41: when is ref a function?

maciekzygmunt: when you do exactly what is done in this component: 
```
ref={(node) => { ....
```

torztomasz (approved):

PR #10527, opened 2026-01-05
Sort Used in in ZK Catalog by TVS

maciekzygmunt:
Resolves L2B-12811

### DIFF
--- a/packages/frontend/src/components/projects/sections/VerifiersSection.tsx
+++ b/packages/frontend/src/components/projects/sections/VerifiersSection.tsx
@@ -103,10 +103,12 @@ function VerifierCollapsibleWithDetails({
               <p className="font-medium text-label-value-12 text-secondary">
                 Used in
               </p>
-              <ProjectsUsedIn
-                noL2ClassName="text-label-value-12 font-medium text-secondary"
-                usedIn={verifierHash.projectsUsedIn}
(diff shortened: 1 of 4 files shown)


packages/frontend/src/components/projects/sections/VerifiersSection.tsx
```diff
@@ -103,10 +103,12 @@ function VerifierCollapsibleWithDetails({
               <p className="font-medium text-label-value-12 text-secondary">
                 Used in
               </p>
-              <ProjectsUsedIn
-                noL2ClassName="text-label-value-12 font-medium text-secondary"
-                usedIn={verifierHash.projectsUsedIn}
-              />
+              <div onClick={(e) => e.stopPropagation()}>
```
torztomasz, line 106: what is this for?

torztomasz (changes requested):

torztomasz:
maybe instead of sorting tvs everywhere for the used in projects, we should simply sort the config on load?

packages/frontend/src/components/projects/sections/VerifiersSection.tsx
maciekzygmunt: When you click on those icons the whole collapisble opens:

PR #10604, opened 2026-01-09
Create configs for interop protocols

torztomasz:


packages/config/src/types.ts
```diff
@@ -124,6 +127,7 @@ export interface BaseProject {
   isBridge?: true
   isScaling?: true
   isZkCatalog?: true
+  isInteropProtocol?: true
```
maciekzygmunt, line 130: Not a fan of this `protocol` part, is optimsm protocol? Maybe just `isInterop`?

packages/config/src/types.ts
```diff
@@ -1178,6 +1182,29 @@ export interface ProjectDiscoveryInfo {
...
+  name?: string
+  bridgeType: 'canonical' | 'nonMinting' | 'omnichain'
+  plugins: InteropPlugin[]
+}
+
+type InteropPlugin = ByChainPlugin | ByTokenIdPlugin | PlainPlugin
+
+type ByChainPlugin = {
+  filterBy: 'chain'
+  plugin: string
```
maciekzygmunt: We should make it union of names of plugins

packages/config/src/types.ts
torztomasz: Optimism is kinda protocol (OP Canonical is)

packages/config/src/types.ts
torztomasz: I do not think we have it defined

PR #10617, opened 2026-01-11
chore: remove unused L1StandardBridge_soon template

Olexandr88:


Olexandr88:
Removed unused __templates/opstack/L1StandardBridge_soon no references in code/docs, projects use opstack/L1StandardBridge

radinat:
The change is not needed.

PR #10624, opened 2026-01-12
Add the missing bridge link

Jonas-Theis:


radinat:
There is already a link that redirects to this one.

PR #10645, opened 2026-01-13
Enable PR-level smoke tests

sdlyy:
Resolves L2B-12904

> on hold

### DIFF
--- a/.github/workflows/smoke-test.yml
+++ b/.github/workflows/smoke-test.yml
@@ -14,7 +14,7 @@ jobs:
         github.event_name == 'merge_group' ||
         (github.event_name == 'pull_request' && github.event.pull_request.head.repo.fork == false)
       }}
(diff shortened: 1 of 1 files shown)


.github/workflows/smoke-test.yml
```diff
@@ -1,10 +1,16 @@
...
-  merge_group:
+  pull_request:
+  merge_group: {}
 
 jobs:
   config-smoke-test:
+    if: >
+      ${{
+        github.event_name == 'merge_group' ||
+        (github.event_name == 'pull_request' && github.event.pull_request.head.repo.fork == false)
```
kondziorex7, line 15: I've also found a following check that might do the job:
```
github.event.pull_request.head.repo.full_name != github.repository
```
I think it should work the same way as yours, but can you double-check?

I'm actually wondering if your check shouldn't be added to all pull-request workflows reaching to secrets?

PR #10682, opened 2026-01-16
refactor(protocolbeat): remove duplicated withoutUndefinedKeys utility

Aleksandr1732:
Removes duplicated `withoutUndefinedKeys` utility function and replaces it with a clearer, more performant inline implementation.


lucadonnoh:
not needed

PR #10685, opened 2026-01-16
Use real data for Interop

torztomasz:


packages/database/src/repositories/AggregatedInteropTransferRepository.ts
```diff
@@ -105,4 +105,30 @@ export class AggregatedInteropTransferRepository extends BaseRepository {
       .executeTakeFirst()
     return Number(result.numDeletedRows)
   }
+
+  async getByChainsAndLatestTimestamp(
+    srcChains: string[],
+    dstChains: string[],
+  ): Promise<AggregatedInteropTransferRecord[]> {
```
maciekzygmunt, line 112: Add check if those arrays are empty

maciekzygmunt (changes requested):

PR #10689, opened 2026-01-17
fix: remove duplicate hex address in wormhole-ntt comment

aso20455:
Removed duplicate line in the payload example comment.
The hex address 000000000000000000000000bc51f76178a56811fdfe95d3897e6ac2b11dbb62 was listed twice in the sendPayloadToEvm section, kept only one.


lucadonnoh:
not needed

PR #10726, opened 2026-01-20
add debridge and dln to interop page

sekuba:


packages/config/src/projects/debridge/debridge.ts
```diff
@@ -51,6 +51,16 @@ export const debridge: Bridge = {
       },
     ],
   },
+  // this is only the debridge messaging and token bridge
+  interopConfig: {
+    name: 'deBridge Messaging and Token Bridge',
```
torztomasz: does it have to be that long?

sekuba: no

PR #10816, opened 2026-01-26
Modernize Orbit stack plugins

lucadonnoh:
waiting for adrian to add support for calldata when re-syncing


radinat:
Seems like we have the support for tx data
image

PR #11230, opened 2026-03-02
Refactor UIF removal configuration

torztomasz:
Resolves L2B-13421


packages/backend/src/modules/tvs/indexers/TokenValueIndexer.ts
```diff
@@ -120,19 +121,37 @@ export class TokenValueIndexer extends ManagedMultiIndexer<TvsToken> {
   }
 
   override async removeData(configurations: RemovalConfiguration[]) {
-    for (const configuration of configurations) {
+    const [configsWithRange, configsWithoutRange] = partition(
+      configurations,
+      (c) => c.type === 'trim',
+    )
```
maciekzygmunt: Ok, what if we  remove `removeData` method and split into `trimData` and `wipeData`, and then those 2 flags (`configurationsTrimmingDisabled`, `dataWipingAfterDeleteDisabled`) should be removed. 
With current API we need to always remember to handle those cases manually inside one method 
Optional trimData/wipeData methods make it explicit in the interface: missing trimData means trimming is unsupported, missing wipeData means wiping is unsupported.

maciekzygmunt (changes requested):

packages/backend/src/modules/tvs/indexers/TokenValueIndexer.ts
maciekzygmunt: Now that I think about it, maybe both methods should be required. And if we don't want to wipe data, for example, we just don't do it inside the method, just like we do sometimes with invalidate.

torztomasz:
@codex review

torztomasz:
@codex review

torztomasz:
@codex review

maciekzygmunt:
@codex review

packages/backend/src/tools/uif/multi/types.ts
```diff
@@ -16,12 +16,20 @@ export interface SavedConfiguration<T> extends Configuration<T> {
...
+}
+
+export type WipeRemovalConfiguration = {
+  type: 'wipe'
   id: string
-  /** Inclusive */
-  from: number
-  /** Inclusive */
-  to: number
 }
```
maciekzygmunt: I think we dont need RemovalConfiguration, and we can remove `type` property

packages/backend/src/modules/da-beat/DaBeatPricesIndexer.ts
```diff
@@ -74,7 +73,7 @@ export class DaBeatPricesIndexer extends ManagedMultiIndexer<DaBeatPricesConfig>
     }
   }
 
-  override async removeData(configurations: RemovalConfiguration[]) {
+  override async wipeData(configurations: WipeRemovalConfiguration[]) {
```
maciekzygmunt, line 76: I guess here we also need to implement the trim data function, because before trimming disabled was false by default. So now I guess we need to implement this function

packages/backend/src/modules/ecosystems/EcosystemTokenIndexer.ts
```diff
@@ -92,15 +92,18 @@ export class EcosystemTokenIndexer extends ManagedMultiIndexer<EcosystemTokenCon
 
     return () => Promise.resolve(to)
   }
-  override async removeData(
-    configurations: RemovalConfiguration[],
+  override async wipeData(
+    configurations: WipeRemovalConfiguration[],
```
maciekzygmunt, line 96: same i guess, pls check other indexers

maciekzygmunt (changes requested):
check commentes

torztomasz:
@codex review

maciekzygmunt (approved):

adamiak (approved):

PR #11451, opened 2026-03-30
Add `pnpm env:sync` to fetch env vars from Google Sheets

adamiak:
Resolves L2B-13575

The command opens a browser login on every run, reads the spreadsheet tab from `GOOGLE_SHEETS_ENV_URL`, rewrites only the managed block in `.env`, and then revokes the OAuth token.


adamiak:
@codex review again

sdlyy (approved):

adamiak:
Since we switched to using L2BEAT RPC almost exclusively, sharing of env variable is less urgent. Closing for now.

PR #11509, opened 2026-04-07
Implement TvsCleaner

torztomasz:
## Summary

Adds a `TvsCleaner` managed child indexer that deletes archived hourly and six-hourly TVS rows using the same grace-period cutoffs as syncing (`SyncOptimizer`). Retention cutoff helpers move from `Clock` into `SyncOptimizer`; `Clock` stays focused on “now” / tick boundaries.

Repository layers gain `delete*Until`-style methods for prices, amounts, block timestamps, and token values (with tests) to support the cleaner.

Resolves L2B-13651


packages/backend/src/modules/tvs/indexers/TvsCleaner.ts
```diff
@@ -0,0 +1,87 @@
...
+        tokenValueHourlyDeletedRecords,
+        blockTimestampHourlyDeletedRecords,
+        amountHourlyDeletedRecords,
+        priceHourlyDeletedRecords,
+      ] = await Promise.all([
+        this.$.db.tvsTokenValue.deleteHourlyUntil(hourlyRange),
+        this.$.db.tvsBlockTimestamp.deleteHourlyUntil(hourlyRange),
+        this.$.db.tvsAmount.deleteHourlyUntil(hourlyRange),
+        this.$.db.tvsPrice.deleteHourlyUntil(hourlyRange),
+      ])
```
maciekzygmunt: Promise all not needed since we are inside tx

maciekzygmunt (changes requested):

packages/backend/src/modules/tvs/indexers/TvsCleaner.ts
```diff
@@ -2,86 +2,116 @@ import type { Logger } from '@l2beat/backend-tools'
...
     logger: Logger,
   ) {
+    const { repositories, ...rest } = $
+
+    const configurations = repositories.map((name) => ({
+      id: repoNameToConfigId(name),
+      minHeight: 0,
+      maxHeight: null,
+      properties: { name },
+    }))
```
maciekzygmunt: move it to module pls

packages/backend/src/modules/tvs/indexers/TvsCleaner.ts
```diff
@@ -69,18 +69,25 @@ export class TvsCleaner extends ManagedMultiIndexer<TvsCleanerConfig> {
     to: number,
     configurations: Configuration<TvsCleanerConfig>[],
   ) {
-    const adjustedFrom = from !== 0 ? (from - 1) * UnixTime.DAY : undefined
-    const adjustedTo = UnixTime(to * UnixTime.DAY)
+    // If not a midnight, skip the update
+    // On first run, run even not at midnight and round `to` to the previous midnight
+    if (from !== 0 && to !== UnixTime.toStartOf(to, 'day')) {
+      return Promise.resolve(() => Promise.resolve(to))
```
maciekzygmunt, line 44: do you need to wrap it in promise?

torztomasz: unfortunately

maciekzygmunt (approved):

PR #11510, opened 2026-04-07
Improve linking for warnings

vilcht:


maciekzygmunt (changes requested):
something strange is happening on mobile Image

maciekzygmunt (approved):

PR #11519, opened 2026-04-08
Optimize interop event filtering

adamiak:
Resolves L2B-13639

Filters out events sent to .capture method that were not requested by topic0 or address. It's still not exact matching of requested events, but should still significantly lower the amount of events sent to .capture(log).


adamiak:
Unfortunately the filtering of events to send to capture took more time than sending them. Closing for now.

PR #11522, opened 2026-04-09
Implement badges modal

vilcht:


packages/frontend/src/components/projects/sections/BadgesSection.tsx
```diff
@@ -1,35 +1,79 @@
...
+        )}
+        <OverflowWrapper className="-m-4">
+          <div className="flex flex-row gap-1 p-4 lg:flex-wrap">
+            {badges.map((badge, key) =>
+              withDialog ? (
+                <button
+                  key={key}
+                  className="cursor-pointer"
+                  onClick={() => openBadgeDialog(badge.id)}
+                >
```
maciekzygmunt, line 58: lets add prefetch onMouseEnter so we show content faster

packages/frontend/src/components/projects/sections/BadgesSection.tsx
```diff
@@ -1,35 +1,79 @@
...
 }
 
 export function BadgesSection({
   badges,
   hideTitle,
   className,
+  withDialog,
 }: BadgesSectionProps) {
+  const [open, setOpen] = useState(false)
+  const [selectedBadgeId, setSelectedBadgeId] = useState<string>()
```
maciekzygmunt: do we need both states? maybe just selectedBadge is enough and then if it is not undefined we can open dialog

packages/frontend/src/components/projects/sections/BadgesSection.tsx
```diff
@@ -1,35 +1,79 @@
...
+            )}
+          </div>
+        </OverflowWrapper>
+      </div>
+      {withDialog && selectedBadgeId && (
+        <BadgesDialog
+          open={open}
+          onOpenChange={setOpen}
+          data={data}
+          isLoading={isLoading || isFetching}
```
maciekzygmunt: just use isLoading

packages/frontend/src/server/features/projects/badges/getScalingBadgeDialogData.ts
```diff
@@ -0,0 +1,81 @@
...
+      type: project.scalingInfo.type ?? project.scalingInfo.layer,
+    }
+
+    for (const badge of project.display.badges) {
+      const badgeWithParams = getBadgeWithParams(badge)
+      if (!badgeWithParams) continue
+
+      const badgesOfType = badgesByType.get(badge.type) ?? new Map()
+      badgesOfType.set(badge.id, badgeWithParams)
+      badgesByType.set(badge.type, badgesOfType)
```
maciekzygmunt: maybe do seperate maps instead of one nested as it is hard to understand now,
so `badgesById: Map<string, ScalingBadge>` and `badgesByType: Map<string, ScalingBadge[]>`

maciekzygmunt (changes requested):
dont use full height on mobile
Image

maciekzygmunt:
also maybe we should make this badge on top sticky so it stays on top while we scroll:

vilcht:
> also maybe we should make this badge on top sticky so it stays on top while we scroll: 

Linear task description says that the whole modal should scroll as one page

maciekzygmunt (approved):

PR #11532, opened 2026-04-09
Simplify interop transfer dialog

vilcht:


torztomasz (changes requested):
Code looks good. Buuuut if we switched "Source Token" and "Destination Token" to "Tokens" then why didn't "Source Chain" and "Destination Chain" get swapped to "Chains"?

Image

vilcht:
> Code looks good. Buuuut if we switched "Source Token" and "Destination Token" to "Tokens" then why didn't "Source Chain" and "Destination Chain" get swapped to "Chains"?
> 
> Image

I just followed the attatched figma design

torztomasz (approved):

PR #11543, opened 2026-04-10
Migrate financial refresh

sdlyy:
\+ processor statuses, PR was slim enough I could accommodate for one more thing :))


packages/backend/src/modules/interop/engine/dashboard/InteropRouter.ts
```diff
@@ -63,6 +64,7 @@ export function createInteropRouter(
       {
         db,
         getExplorerUrl: config.dashboard.getExplorerUrl,
+        getProcessorStatuses: () => getProcessorsStatus(processors),
```
torztomasz, line 67: weird naming, statuses return status from processors

torztomasz (approved):

PR #11546, opened 2026-04-13
add tooltip for percentage change in tvs

vilcht:


packages/frontend/src/pages/scaling/tvs/components/ScalingTvsStats.tsx
```diff
@@ -100,10 +105,19 @@ function Stat({
...
-            value={change}
-          />
+          <Tooltip>
+            <TooltipTrigger>
+              <span className="inline-flex cursor-default">
+                <PercentChange
+                  textClassName="md:text-heading-20 w-full text-heading-16"
+                  value={change}
+                />
+              </span>
```
torztomasz: do we have to wrap it with span?

torztomasz (changes requested):

torztomasz (approved):

PR #11561, opened 2026-04-14
virtualization example

vilcht:


packages/frontend/src/components/table/VirtualizedBasicTable.tsx
```diff
@@ -0,0 +1,252 @@
...
+              />
+            </TableRow>
+          ) : null}
+        </>
+      )}
+    </BasicTableFrame>
+  )
+}
+
+function VirtualizedBasicTableRow<T extends BasicTableRow>({
```
torztomasz, line 160: it does not support multiple row rows?

packages/frontend/src/components/table/VirtualizedBasicTable.tsx
```diff
@@ -0,0 +1,252 @@
...
+              aria-hidden="true"
+              className="border-0 hover:bg-transparent"
+              highlightId={undefined}
+            >
+              <TableCell
+                colSpan={100}
+                className="h-0 border-0 p-0 first:pl-0 last:pr-0"
+                style={{ height: `${virtualPaddingTop}px` }}
+              />
+            </TableRow>
```
torztomasz, line 121: Why we render it differently if hidden??

torztomasz (changes requested):

PR #11572, opened 2026-04-15
Update range resolution logic to use inclusive comparisons for time ranges in throughput, costs, and liveness features

torztomasz:


packages/frontend/src/utils/range/range.test.ts
```diff
@@ -0,0 +1,55 @@
...
+import { type InstalledClock, install } from '@sinonjs/fake-timers'
+import { expect } from 'earl'
+import { rangeToResolution } from './range'
+
+describe(rangeToResolution.name, () => {
+  let time: InstalledClock
+
+  beforeEach(() => {
+    time = install()
+    time.setSystemTime(new Date('2026-04-15T12:00:00.000Z'))
```
maciekzygmunt, line 11: cant we just do UnixTime.now = new Date(....?

maciekzygmunt (approved):

PR #11585, opened 2026-04-16
Improve search to always show closest matches first

vilcht:


packages/frontend/src/components/search-bar/searchBarResults.ts
```diff
@@ -0,0 +1,117 @@
...
+  return rankSearchResults(searched, options?.limit)
+}
+
+function rankSearchResults<T extends SearchableSearchBarEntry>(
+  entries: SearchResult<T>[],
+  limit?: number,
+): SearchResult<T>[] {
+  const filtered = entries.some((entry) => entry.searchMatchKind === 'direct')
+    ? entries.filter((entry) => entry.searchMatchKind === 'direct')
+    : entries
```
maciekzygmunt: extract those filtered entries to const and then just do arr.length > 0 instead of .some

maciekzygmunt (changes requested):

maciekzygmunt (approved):

PR #11586, opened 2026-04-16
Missing tokens improvements

sdlyy:


packages/backend/src/modules/interop/engine/dashboard/impls/missingTokens.ts
```diff
@@ -0,0 +1,124 @@
...
+  const rows = await db.interopTransfer.getMissingTokensInfo()
+  return mapMissingTokensWithStatus(rows, deps)
+}
+
+export async function mapMissingTokensWithStatus<
+  T extends MissingTokenSelection,
+>(
+  tokens: T[],
+  deps: MissingTokensDeps,
+): Promise<Array<T & { tokenDbStatus: MissingTokenDbStatus }>> {
```
torztomasz: Array<>??

packages/backend/src/modules/interop/engine/dashboard/trpc/routers/missingTokens.ts
```diff
@@ -1,10 +1,61 @@
...
+            return {
+              ...token,
+              outcome:
+                status === 'ready'
+                  ? 'requeued'
+                  : status === 'missing'
+                    ? 'skipped_missing'
+                    : status === 'incomplete'
+                      ? 'skipped_incomplete'
+                      : 'skipped_unsupported',
```
torztomasz: simple Record mapping maybe instead of ternary?
why cant we just return the outcome instead of status?

packages/interop-backoffice-ui/src/pages/missing-tokens/table/columns.tsx
```diff
@@ -67,26 +133,52 @@ export function createMissingTokensColumns(options: {
...
     }),
   ]
 }
+
+export function getMissingTokensSelectionId(row: MissingTokenRow) {
+  return getMissingTokenRowId({
+    chain: row.chain,
+    tokenAddress: row.tokenAddress,
+  })
+}
```
torztomasz, line 197: why does `getMissingTokensSelectionId` even exist?

packages/interop-backoffice-ui/src/pages/missing-tokens/table/MissingTokensTable.tsx
```diff
@@ -1,26 +1,142 @@
...
+          if (!selectableRowIds.has(rowId)) {
+            return
+          }
+
+          setSelectedRowIds((current) => {
+            const next = new Set(current)
+            if (selected) {
+              next.add(rowId)
+            } else {
+              next.delete(rowId)
```
torztomasz: Doesn't Tanstack Table has this logic of selection already built in?

packages/interop-backoffice-ui/src/pages/missing-tokens/table/MissingTokensTable.tsx
```diff
@@ -35,29 +151,122 @@ export function MissingTokensTable({
...
+function writeClickedActionIds(rowIds: string[]) {
+  if (typeof window === 'undefined') {
+    return
+  }
+
+  window.localStorage.setItem(
+    CLICKED_ACTIONS_STORAGE_KEY,
+    JSON.stringify(rowIds),
   )
 }
```
torztomasz: simply use `useLocalStorage` hook

torztomasz (changes requested):

packages/interop-backoffice-ui/src/pages/missing-tokens/table/MissingTokensTable.tsx
sdlyy: yupps, dropped entirely in favor of tanstack

packages/interop-backoffice-ui/src/pages/missing-tokens/table/columns.tsx
sdlyy: just a helper

packages/backend/src/modules/interop/engine/dashboard/trpc/routers/missingTokens.ts
sdlyy: I decided to remove this part completely, wanted to use it later on to show what actually happened per token - but it was too much

packages/backend/src/modules/interop/engine/dashboard/impls/missingTokens.ts
sdlyy: fxd

packages/interop-backoffice-ui/src/pages/missing-tokens/table/MissingTokensTable.tsx
sdlyy: got it, it's there

sdlyy:
@codex review

torztomasz (changes requested):
- packages/interop-backoffice-ui/src/pages/missing-tokens/table/MissingTokensTable.tsx:62 — the onActionVisited callback closes over clickedActionIds, and the columns memo depends on clickedActionIds. Every click rebuilds the column definitions, which resets any per-column internal state TanStack keeps (filter pending, etc.). Consider a ref + functional setClickedActionIds((prev) => …) to stabilize column identity.
  - packages/interop-backoffice-ui/src/hooks/useLocalStorage.ts:30 — setValue's deps list is [key, serialize] but it calls setStoredValue. React's setter identity is stable so this is fine, but deserialize is intentionally omitted from deps yet referenced only in the initializer (also fine). Worth a comment or just drop serialize from deps for symmetry since both helpers are effectively module-stable.
- impls/missingTokens.ts:47 — dedupeMissingTokens(tokens) inside getMissingTokenStatuses is redundant since you build a Map keyed by getMissingTokenKey right after. Dropping it removes one full pass.

torztomasz (approved):

PR #11596, opened 2026-04-17
add interop data to api

vilcht:


packages/public-api/src/routes/interop/interopUtils.test.ts
```diff
@@ -0,0 +1,354 @@
+import type { InteropConfig } from '@l2beat/config'
+import type { AggregatedInteropTransferRecord } from '@l2beat/database'
+import { ProjectId, UnixTime } from '@l2beat/shared-pure'
+import { expect } from 'earl'
+import type { ProjectMetadata } from './utils/getAverageTransferTime'
+import { getInteropChains } from './utils/getInteropChains'
+import { getInteropProtocols } from './utils/getInteropProtocols'
+
+describe('interop utils', () => {
```
torztomasz: do not do this this way. every file should get separate test and it hsould be the same as main file

packages/public-api/src/routes/interop/utils/getInteropChains.ts
```diff
@@ -0,0 +1,142 @@
...
+
+  return [...chains.values()]
+    .map((chain) => ({
+      id: chain.id,
+      name: chain.name,
+      totalVolume: chain.inflowsUsd + chain.outflowsUsd,
+      totalTransferCount: chain.transferCount,
+      inflowsUsd: chain.inflowsUsd,
+      outflowsUsd: chain.outflowsUsd,
+      avgTransferTimeSeconds: getAverageTransferTimeSeconds(chain),
```
torztomasz, line 63: why not return our structure of DurationSplit?

packages/public-api/src/routes/interop/utils/getInteropProtocols.ts
```diff
@@ -0,0 +1,165 @@
...
+function createProtocolAccumulator(
+  id: string,
+  project: ProjectMetadata | undefined,
+): ProtocolAccumulator {
+  return {
+    metadata: {
+      id,
+      slug: project?.slug ?? id,
+      name: project?.interopConfig.name ?? project?.name ?? id,
+    },
```
torztomasz: why do you want to keep it under metadata? chains are not like this

torztomasz (changes requested):

packages/public-api/src/routes/interop/utils/getInteropChains.ts
vilcht: Chain entries aggregate data from multiple projects, and DurationSplit is project specific.
  Different projects can define different split shapes, there is no single valid
  split structure to return.

torztomasz (commented):
Image

Image

Proper const strings for avgTransferTime.type are not shown. We need to handle this. Probably a proper handling of an union in `OpenApi.ts` is required.

packages/public-api/src/routes/interop/utils/getInteropChains.ts
```diff
@@ -0,0 +1,142 @@
...
+  getAverageTransferTimeSeconds,
+  type ProjectMetadata,
+} from './getAverageTransferTime'
+import { mergeTransferTypeStats } from './mergeTransferTypeStats'
+
+type CommonInteropData = {
+  transferCount: number
+  transfersWithDurationCount: number
+  totalDurationSum: number
+  transferTypeStats: AggregatedInteropTransferRecord['transferTypeStats']
```
torztomasz: is this even used anywhere? I think not

packages/public-api/src/routes/interop/utils/getInteropChains.ts
```diff
@@ -0,0 +1,142 @@
...
+  chains: Map<string, ChainAccumulator>,
+  chainId: string,
+  direction: 'inflow' | 'outflow',
+  record: AggregatedInteropTransferRecord,
+  projectsMap: Map<string, ProjectMetadata>,
+) {
+  const metadata = chainMetadata.get(chainId)
+  const chain = chains.get(chainId) ?? {
+    id: chainId,
+    name: metadata?.name ?? chainId,
```
torztomasz: Since records are filtered by INTEROP_CHAINS ids in getLatestAggregatedInteropTransfers, this path is unreachable today, so the fallback is dead code.

packages/public-api/src/routes/interop/utils/getInteropChains.ts
```diff
@@ -0,0 +1,142 @@
...
+  id: string
+  name: string
+  inflowsUsd: number
+  outflowsUsd: number
+  protocols: Map<string, ProtocolBreakdownAccumulator>
+} & CommonInteropData
+
+const chainMetadata = new Map(
+  INTEROP_CHAINS.map((chain) => [chain.id, { id: chain.id, name: chain.name }]),
+)
```
torztomasz: Repeats the same chainMetadata Map already defined in getInteropProtocols.ts — could hoist to utils/chainMetadata.ts

packages/public-api/src/routes/interop/utils/getInteropChains.ts
```diff
@@ -0,0 +1,142 @@
...
+    projects
+      .filter((project) => project.interopConfig.subgroupId !== undefined)
+      .map((project) => project.id.toString()),
+  )
+  const chains = new Map<string, ChainAccumulator>()
+
+  for (const record of records) {
+    if (subgroupProjects.has(record.id)) {
+      continue
+    }
```
torztomasz, line 47: It filters out subgroup records to avoid double-counting (parent records already include children's transfers), but getInteropProtocols intentionally keeps them so each protocol gets its own entry — matching the frontend's getProtocolEntries pattern. This asymmetry is correct but undocumented and untested. A consumer who naively does sum(protocols.totalVolume) will double-count.

Please add a test case in getInteropProtocols.test.ts analogous to the one in getInteropChains.test.ts:104 ('skips subgroup protocol records'), but asserting the opposite — that subgroup children are included as separate entries.

packages/public-api/src/routes/interop/utils/getInteropProtocols.ts
```diff
@@ -0,0 +1,161 @@
...
+  project: ProjectMetadata | undefined
+  totalVolume: number
+  chains: Map<string, ChainBreakdownAccumulator>
+} & CommonInteropData
+
+const chainMetadata = new Map(
+  INTEROP_CHAINS.map((chain) => [chain.id, { id: chain.id, name: chain.name }]),
+)
+
+export function getInteropProtocols(
```
torztomasz, line 36: Right now /v1/interop/protocols returns subgroup children as separate entries with no indication of their parent. Consumers can't tell that e.g. USDT0 rolls up into LayerZero — they just see two sibling protocols with overlapping volume. Please add a subgroupId: v.union([v.string(), v.null()]) (or .optional()) field to InteropProtocolSchema in types.ts, and populate it in getInteropProtocols.ts from project.interopConfig.subgroupId. With that field, consumers can choose to display the hierarchy or filter children out when summing totals.

packages/public-api/src/routes/interop/utils/getInteropChains.ts
torztomasz, line 63: Chain-level avgTransferTimeSeconds mixes duration data from all protocols, including those configured transfersTimeMode === 'unknown'. The per-protocol endpoint correctly gates on hasUnknownTransferTime(project) (getInteropProtocols.ts:89), but the chain endpoint does not. If any unknown-mode protocol emits non-zero transfersWithDurationCount, its numbers leak into the chain average.

torztomasz (changes requested):

packages/public-api/src/routes/interop/utils/getInteropChains.ts
vilcht: It is used.

torztomasz (changes requested):
we need to document somehow what is the subgroupId for
check openapi spec on how to do this and try to implement this

packages/public-api/src/routes/interop/types.ts
```diff
@@ -0,0 +1,73 @@
...
+export const InteropProtocolSchema = v
+  .object({
+    id: v.string(),
+    slug: v.string(),
+    name: v.string(),
+    subgroupId: v
+      .union([v.string(), v.null()])
+      .document(
+        'ID of the aggregate/root interop protocol this protocol belongs to. Null for aggregate/root protocols.',
+      ),
```
torztomasz: why not use already existing `.describe`?

packages/validate/src/validate.ts
```diff
@@ -118,6 +125,11 @@ export class Imp<T> implements Validator<T>, Parser<T> {
     return this
   }
 
+  document(description: string) {
+    this.jsonSchema = { ...this.jsonSchema, description }
+    return this
+  }
```
torztomasz: use `describe`

torztomasz (changes requested):

packages/public-api/src/routes/interop/types.ts
vilcht: .describe() adds validator.description (reusable name for the schema), which is an entirely different thing.

packages/validate/src/validate.ts
vilcht: Our .describe doesn't add an openAPI description field

packages/validate/src/validate.ts
```diff
@@ -118,6 +121,12 @@ export class Imp<T> implements Validator<T>, Parser<T> {
     return this
   }
 
+  meta(metadata: Record<string, unknown>) {
```
torztomasz: just make it Record<string, string>, no need for the asserts later then

torztomasz (changes requested):

torztomasz (approved):

PR #11653, opened 2026-04-22
Align the No transfer text to the bottom

vilcht:


maciekzygmunt (changes requested):
no change:
Image

i think we can just increase padding

maciekzygmunt (approved):

PR #11659, opened 2026-04-23
protocolbeat: speed up large-graph nodes panel

lucadonnoh:
## Summary

The discovery nodes view (`/ui/p/<project>`) lags noticeably for projects with hundreds of nodes: click latency is poor and drag is jittery. The cause is a few hot paths in `NodesAndConnections` and `updateNodePositions` that run O(N²) work per render and re-render every node on every pointermove. This PR fixes them without changing behavior.

Tested against a CCIP-on-Ethereum discovery (~382 nodes / ~5700 fields); dragging a leaf node went from re-rendering all 382 NodeViews per frame to ~10, and selection flips run a single O(N+F) pass instead of O(N²).

### Changes

- **`NodesAndConnections.tsx`**: build a `visibleById` Map once for O(1) target lookup (was O(N) per field), hoist `highlightedIds` out of the per-node `.map` callback (was being recomputed N times), convert `hidden`/`selected` to Sets, wrap computation in `useMemo`, split `hiddenSet`/`selectedSet` into separate memos so their reference stays stable when only `nodes` changes during drag.
- **`NodeView.tsx`**: memoize `NodeView` and `NodeField`. `NodeField` no longer opens two per-field store subscriptions — the parent passes `isHighlighted` / `targetHidden` down as booleans derived from a single shared set.
- **`Connection.tsx`**: memoize.
- **`updateNodePositions.ts`**: only rebuild nodes whose box actually changed (dragged, resized, or height-changed from a hidden-field toggle). For stationary nodes, only rebuild `connection` on fields whose *target* moved; keep `field.box` identity. Preserve node identity otherwise so memoized NodeViews skip re-render. Short-circuit the whole call when nothing moved and layout is already computed (detected via the `connection.nodeId` sentinel).

## Test plan

- [ ] Open `/ui/p/<large-project>` in the discovery UI, confirm edges render on load (the `connection.nodeId` sentinel path).
- [ ] Click a node — selection should be instant, connected nodes highlight, others dim.
- [ ] Drag a node — movement should track the cursor cleanly; only the dragged node and its in/out-edges visibly move.
- [ ] Multi-select + drag — still works, all selected nodes move together.
- [ ] Resize a node (drag right edge) — rebuilds just that node's fields.
- [ ] Toggle hidden fields / hide nodes — heights recompute correctly.
- [ ] Small projects (≤ 30 nodes) — no behavior regression.

### DIFF
--- a/packages/protocolbeat/src/apps/discovery/panel-nodes/store/utils/updateNodePositions.ts
+++ b/packages/protocolbeat/src/apps/discovery/panel-nodes/store/utils/updateNodePositions.ts
@@ -17,73 +17,148 @@ export function updateNodePositions(state: State): State {
     }
   }
 
-  const nodeDimensions: Record<string, Box> = {}
+  // Compute new box for every node. If the node isn't being dragged/resized
+  // (no entry in positionsBeforeMove), we may still end up with the same box
+  // as before — preserve identity when that happens.
+  const newBoxById = new Map<string, Box>()
+  const movedIds = new Set<string>()
   for (const node of state.nodes) {
     const start = state.positionsBeforeMove[node.id]
     const hiddenFieldsHeight =
       node.hiddenFields.length > 0 ? HIDDEN_FIELDS_FOOTER_HEIGHT : 0
-    nodeDimensions[node.id] = {
-      width: node.box.width,
-      height:
-        HEADER_HEIGHT +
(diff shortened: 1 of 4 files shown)


sdlyy:
[Ref](https://github.com/l2beat/l2beat/pull/11757)

PR #11660, opened 2026-04-23
add quantum resistance to statevalidation

vilcht:


packages/frontend/src/components/projects/sections/state-validation/ProverInfo.tsx
```diff
@@ -16,12 +24,26 @@ export function ProverInfo({
...
+              <Tooltip>
+                <TooltipTrigger asChild>
+                  <button type="button" aria-label="Quantum resistance details">
+                    <QuantumResistanceIcon className="size-5" />
+                  </button>
+                </TooltipTrigger>
+                <TooltipPortal>
+                  <TooltipContent>{QUANTUM_RESISTANCE_TOOLTIP}</TooltipContent>
+                </TooltipPortal>
+              </Tooltip>
```
torztomasz, line 44: Let's extract this to a separate component so that it can be used here and in project name cell

torztomasz (changes requested):

packages/frontend/src/components/projects/sections/state-validation/ProverInfo.tsx
vilcht: I wouldnt do that, projectNameCell already renders all the Icons via mimimal helper on mobile, rendering only quantum resistance  tooltip via different component would make the code harder to reason about

torztomasz (approved):

PR #11700, opened 2026-04-28
add deselect tooltip, select largest volumes by default

vilcht:


torztomasz (commented):
Image

Shouldn't we sort the chains too?

torztomasz (approved):

PR #11720, opened 2026-04-29
Add environment switching to interop backoffice UI

torztomasz:
## Summary
- Added an environment provider and selector to switch between local, staging, and production backoffice APIs.
- Routed tRPC requests through environment-specific endpoints and reset the query client when the active environment changes.
- Updated the Vite dev proxy to forward each environment path to the correct backend target.

## Testing
- Not run


packages/interop-backoffice-ui/src/App.tsx
sdlyy, line 1: oppa, old dir copy

sdlyy (approved):
all good

PR #11724, opened 2026-04-29
Update exit window column with second line for emergency exit window

vilcht:


packages/frontend/src/pages/scaling/risk/components/table/ExitWindowCell.tsx
```diff
@@ -0,0 +1,80 @@
...
+        {value.emergency && (
+          <TwoRowCell.Second className="mt-0.5">
+            <SentimentText
+              sentiment={value.emergency.sentiment ?? 'neutral'}
+              className="font-medium"
+            >
+              {`Emergency: ${value.emergency.value}`}
+            </SentimentText>
+          </TwoRowCell.Second>
+        )}
```
maciekzygmunt: not sure if we should accept both second line and emergency. If second line is not used remove it

maciekzygmunt (changes requested):

packages/frontend/src/pages/scaling/risk/components/table/ExitWindowCell.tsx
vilcht: There's code to add second line text if options.multisig is set. But it isn't set anywhere, not sure how to proceed.

radinat:
Leaving here another change to be done: swap the order, so emergency is first (if any) as "Emergency:..." and regular exit is second line "Regular: ...";
if there's no emergency, it's just the regular on first line

packages/frontend/src/pages/scaling/risk/components/table/ExitWindowCell.tsx
maciekzygmunt: Talked with Adi and we can remove this `options.multisig` since it is not used anywhere, and then we can remove second line from component

maciekzygmunt (approved):

sekuba (approved):

PR #11744, opened 2026-04-30
Handle empty synced windows in DA throughput charts

mk0walsk:
This prevents Infinity/-Infinity chart ranges when fully synced throughput buckets are absent, and returns empty results instead of generating invalid timelines.


torztomasz:
What's the issue with current implementation? Was some chart displayed wrong?

mk0walsk:
Hi @torztomasz Thanks for your review. This is more of an edge case: if we only get records from the current unfinished bucket, they’re filtered out and min/max stay as Infinity/-Infinity, which can make chart generation throw.

PR #11747, opened 2026-05-01
Gnosis chain review

sekuba:


packages/config/src/projects/gnosis/gnosis.ts
```diff
@@ -1,32 +1,135 @@
...
+    name: 'Gnosis Chain',
+    shortName: undefined,
+    aliases: ['xDai'],
+    slug: 'gnosis',
+    purposes: ['Universal'],
     description:
-      "Gnosis Chain is a community-owned EVM-based network operated by a diverse set of validators around the world. It is one of Ethereum's first sidechains.",
-    links: {},
-    badges: [],
+      'Gnosis Chain is a community-owned EVM-based sidechain operated by a proof-of-stake validator set aiming to be the first chain in the Ethereum Economic Zone (EEZ). Its canonical Ethereum bridge is validated by dedicated bridge validator multisigs (not the PoS validator set) and supports the native xDAI bridge, AMB messages, and Omnibridge token transfers. This page looks at both the PoS chain and the canonical bridge to Ethereum from an Ethereum-centric perspective.',
```
lucadonnoh: give an hint to what "AMB messages, and Omnibridge token transfers." are. when reading up to this point they don't mean anything to a non-expert reader

lucadonnoh (changes requested):

packages/config/src/common/riskView.ts
```diff
@@ -307,6 +307,13 @@ export const DATA_POS: TableReadyValue = {
   sentiment: 'warning',
 }
 
+export const DATA_POS_NOBRIDGE: TableReadyValue = {
+  value: 'PoS network',
+  description:
+    'Data is guaranteed to be available by an external proof of stake network of validators. There is no DA bridge to Ethereum so this external PoS network is an additional trust assumption.',
```
lucadonnoh: the PoS network is a trust assumption even with the DA bridge (like in Polygon PoS). the consequence of no DA bridge is that Ethereum doesn't know whether messages actually come from attested Gnosis blocks

lucadonnoh (changes requested):

packages/config/src/projects/gnosis/gnosis.ts
```diff
@@ -35,6 +139,192 @@ export const gnosis: BaseProject = {
...
+      risks: [
+        {
+          category: 'Funds can be frozen if',
+          text: 'transaction data is unavailable and bridge validators cannot reconstruct or sign the messages needed to exit.',
+        },
+      ],
+    },
+    otherConsiderations: [
+      {
+        name: 'Hashi validation is not mandatory',
```
lucadonnoh: i have no idea what Hashi is

lucadonnoh (changes requested):

packages/config/src/projects/gnosis/gnosis.ts
lucadonnoh: also it's deprecated: https://forum.gnosis.io/t/deprecation-notice-hashi-on-gnosis-canonical-bridges-ends-maintenance/11467

sekuba: wow ok might just remove this explanation and the disco then
sad bc this was the right direction

packages/config/src/common/riskView.ts
sekuba: i tried to reformulate, pls check if it makes sense now

lucadonnoh (approved):

PR #11779, opened 2026-05-06
Add stages for V&O detection in templates

maciekzygmunt:
Resolves L2B-13815


vincfurc:
I see a lot more Stage 0 in the config diff so there is something wrong with the template resolving logic

packages/config/src/templates/orbitStack.ts
```diff
@@ -1382,48 +1397,93 @@ function extractDAs(daProviders: DAProvider[]): ProjectScalingDa[] {
...
+        callsItselfValidiumOrOptimium: true,
         stateRootsPostedToL1: true,
-        dataAvailabilityOnL1: true,
-        rollupNodeSourceAvailable:
-          templateVars.isNodeAvailable ?? 'UnderReview',
         stateVerificationOnL1: true,
+        daAttestedByIndependentParty:
+          templateVars.daAttestedByIndependentParty ?? null,
+        nodeSourceAvailable: templateVars.isNodeAvailable ?? 'UnderReview',
         fraudProofSystemAtLeast5Outsiders: false,
```
vincfurc: shouldn't this be fetched var rather than hardcoded? it changes Nova from true to false

packages/config/src/templates/orbitStack.ts
```diff
@@ -1382,48 +1397,93 @@ function extractDAs(daProviders: DAProvider[]): ProjectScalingDa[] {
...
       },
       stage1: {
-        principle: false,
-        usersHave7DaysToExit: false,
+        principle: templateVars.stage1Principle ?? null,
         usersCanExitWithoutCooperation: true,
+        usersHave7DaysToExit: false,
```
vincfurc: same as above, it changes Nova from true to false

packages/config/src/templates/orbitStack.ts
```diff
@@ -1382,48 +1397,93 @@ function extractDAs(daProviders: DAProvider[]): ProjectScalingDa[] {
...
         securityCouncilProperlySetUp: false,
```
vincfurc: same as above, it changes Nova from true to false

packages/config/src/templates/zkStack.ts
```diff
@@ -388,9 +404,53 @@ export function zkStackL2(templateVars: ZkStackConfigCommon): ScalingProject {
...
+                daAttestedByIndependentParty:
+                  templateVars.daAttestedByIndependentParty ?? null,
+                nodeSourceAvailable: true,
+                fraudProofSystemAtLeast5Outsiders: null,
+              },
+              stage1: {
+                principle: templateVars.stage1Principle ?? false,
+                usersCanExitWithoutCooperation: false,
+                usersHave7DaysToExit: false,
+                securityCouncilProperlySetUp: true,
```
vincfurc: this on is hardcoded true as well shoul be template var? affects Sophon

vincfurc (approved):

PR #11780, opened 2026-05-06
Extract TVS price indexer into TokenPriceModule

torztomasz:
## Summary

This PR extracts the TVS price indexer into a separate, standalone `TokenPriceModule` that can be enabled independently via the `token-price` feature flag.

### New Module: `TokenPriceModule`
- **`TokenPriceModule.ts`** — Initializes `TokenPriceIndexer` with its own `HourlyIndexer` parent
- **`TokenPriceIndexer.ts`** — Renamed from `TvsPriceIndexer`, uses new identity `token_price_indexer`
- Returns the indexer so TVS `TokenValueIndexer` can use it as a parent

### Separate Config & Feature Flag
- Added `TokenPriceConfig` with its own `getTokenPriceConfig()` function
- Added `token-price` feature flag (independent from `tvs`)
- `TvsConfig` no longer includes `prices`

### Database Migration
- Renamed `TvsPrice` table → `TokenPrice`
- Renamed repository: `TvsPriceRepository` → `TokenPriceRepository`
- Updated `IndexerState.indexerId` and `IndexerConfiguration.indexerId` from `tvs_price_indexer` → `token_price_indexer`
- This ensures the new indexer seamlessly continues from the old one's sync state

### Code Updates
- Updated all backend code to use `db.tokenPrice` and `TokenPriceRecord`
- Updated frontend `getEthPrices.ts` to use `db.tokenPrice`
- Removed old `TVS_PRICE` indexer identity constant

## Verification
- TypeScript type-checking passes
- Linting passes
- All tests pass (1123 passing)


packages/database/src/database.ts
```diff
@@ -116,7 +116,7 @@ export function createDatabase(
     // #endregion
 
     // #region Tvs
-    tvsPrice: new TvsPriceRepository(db),
+    tokenPrice: new TokenPriceRepository(db),
```
maciekzygmunt, line 119: move it out of th is region

packages/backend/src/Application.ts
```diff
@@ -61,14 +62,21 @@ export class Application {
       trpcContributions,
     })
 
+    // Token price module must be initialized before TVS module
+    const tokenPriceModule = initTokenPriceModule(deps)
```
maciekzygmunt, line 66: i think we should unify naming, `initModule` or `createModule`

packages/backend/src/config/features/tvs.ts
```diff
@@ -111,13 +111,12 @@ export async function getTvsConfig(
       'tvsTokenValue',
       'tvsBlockTimestamp',
       'tvsAmount',
-      'tvsPrice',
+      'tokenPrice',
```
maciekzygmunt, line 114: i guess it should happen outside of tvs config

packages/backend/src/config/features/tokenPrice.ts
```diff
@@ -0,0 +1,52 @@
...
+
+  const chainRanges = new Map(
+    (await ps.getProjects({ select: ['chainConfig'] })).map((p) => [
+      p.chainConfig.name,
+      {
+        sinceTimestamp: p.chainConfig.sinceTimestamp,
+        untilTimestamp: p.chainConfig.untilTimestamp,
+      },
+    ]),
+  )
```
maciekzygmunt, line 42: extract it to some function below, getTvsPrices or smth, later we will add similar one for other features, we should also make them uniq then

packages/database/prisma/migrations/20260506115307_rename_tvs_price_to_token_price/migration.sql
```diff
@@ -0,0 +1,14 @@
+-- RenameTable
+ALTER TABLE "TvsPrice" RENAME TO "TokenPrice";
+
```
maciekzygmunt, line 3: not sure if this migration should look like this, it will fire in CI and our BE will work for some time on old code

maciekzygmunt (changes requested):
check comments

PR #11796, opened 2026-05-07
Implement da tracking status page

vilcht:


packages/backoffice/src/utils/formatAge.ts
maciekzygmunt, line 1: cant we use formatSeconds?

maciekzygmunt (changes requested):

maciekzygmunt (approved):

PR #11813, opened 2026-05-08
Improve state val page

lucadonnoh:
Resolves L2B-14000


packages/config/src/types.ts
```diff
@@ -536,10 +536,28 @@ export interface ProjectRiskView {
...
+    /** Whether `executionDelay` is applied on every withdrawal or only when a
+     *  challenge is raised. Treated as 'always' when omitted. */
+    executionDelayMode?: 'always' | 'if-challenged'
     /** @unit seconds */
     challengeDelay?: number
-    /** @unit ETH */
+    /** Defaults to ETH (rendered with Ξ prefix). Set when the bond is paid in
+     *  a non-ETH token; the value is then rendered as "<initialBond> <token>". */
     initialBond?: string
+    initialBondToken?: string
```
maciekzygmunt: lets merge it into one: initalBond:{
    value: string
    token?: string
}

packages/config/src/common/formatDelays.ts
```diff
@@ -16,8 +16,15 @@ export function formatChallengePeriod(period: number | undefined) {
...
+  challengeDelay: number | undefined,
+  executionDelay: number | undefined,
+) {
+  if (challengeDelay === undefined && executionDelay === undefined)
+    return undefined
+  if ((challengeDelay ?? 0) === 0 && (executionDelay ?? 0) === 0)
+    return 'No challenge + execution delay'
+  if ((executionDelay ?? 0) === 0) return formatChallengePeriod(challengeDelay)
+  if ((challengeDelay ?? 0) === 0) return formatExecutionDelay(executionDelay)
+  return `${formatSeconds(challengeDelay!)} challenge period\n+ ${formatSeconds(executionDelay!)} execution delay`
```
maciekzygmunt: remote non-null assertion

maciekzygmunt (changes requested):
How about we make the second line not wrap?
Image

maciekzygmunt (approved):

PR #11814, opened 2026-05-08
Add Lighter bridge interop plugin

lucadonnoh:
## Summary
- New one-sided interop plugin for the Lighter perp DEX bridge (`0x3B4D…`), capturing `Deposit` and `WithdrawPending` events on Ethereum and emitting partial transfers gated on the `lighter` one-sided chain.
- Token address and amount are discovered dynamically from the in-tx ERC20 `Transfer` log; the only static knowledge is `NATIVE_ASSET_INDEX = 1` and `NATIVE_TICK_SIZE = 1e10` for ETH (whose transfer emits no log).
- Position-based attribution anchors on the contract's strict `[NewPriorityRequest, Deposit]` alternation pattern: per-recipient amount = `Transfer.value × baseAmount / sumBase`, multicall-safe and verified against the deployed `AdditionalZkLighter` (`0x22F05515…`) source.
- Adds 19 example txs covering single/batched deposits (up to 14 recipients), multi-claim withdrawals, ETH+ERC20 mixed claims, AAVE/UNI/SKY/LDO/LIT/AZTEC/USDC, plus 22 financial assertions on exact `tokenAmount` values independently verified against `cast receipt`.
- Drive-by: renames `*_BRIDGE_EOA` to `*_BRIDGE_ADDRESS` in `lighter-bridge.ts` and `hyperliquid-bridge.ts` (the variable is a contract address, not an EOA).

## Test plan
- [x] `pnpm interop:example lighter-bridge --simple` — 25 Deposits + 15 WithdrawPending → 40 Transfers, all 23 expectations PASS
- [x] `pnpm interop:example hyperliquid-bridge --simple` still passes after the rename
- [x] Independent verification script that recomputes expected amounts from `cast receipt` — 22/22 assertions match
- [x] `pnpm fix` clean


sekuba (approved):
🔥

PR #11819, opened 2026-05-11
Add flows section for scaling projects with interop data

vilcht:


maciekzygmunt:
It is to dense, try to make "stats" to be one left half, and rest to be one under the other on right half

packages/frontend/src/server/features/scaling/project/getScalingProjectEntry.ts
```diff
@@ -130,6 +134,110 @@ export interface ProjectScalingEntry {
...
+    .slice(0, MAX_SELECTED_CHAINS)
+  const protocols = interopProjects.map((protocol) => ({
+    id: protocol.id,
+    name: protocol.interopConfig.name ?? protocol.name,
+    iconUrl: manifest.getUrl(`/icons/${protocol.slug}.png`),
+  }))
+  const interopFlows = await helpers.interop.flows.fetch({
+    chains: defaultSelectedChains,
+    protocolIds: protocols.map((protocol) => protocol.id),
+  })
```
maciekzygmunt: we only show stats from `defaultSelectedChains` but i guess we want to show from all in summary

maciekzygmunt (changes requested):
check comment

maciekzygmunt:
also I think we can simplify the provider API here. Since we only pin one project chain, I’d replace `lockedChainIds` with a single `pinnedChainId`. I’d also avoid adding `defaultHighlightedChains`. The initial highlight can be derived from `pinnedChainId`, so we keep one source of truth: if a chain is pinned, it is selected, cannot be deselected, and is highlighted by default.

maciekzygmunt (changes requested):
for some reason i dont see this section on arbitrum on preview

packages/frontend/src/pages/interop/components/flows/utils/InteropFlowsContext.tsx
```diff
@@ -186,3 +236,7 @@ function parseIdsParam(value: string, validIds: string[]): string[] {
   const provided = new Set(value.split(','))
   return validIds.filter((id) => provided.has(id))
 }
+
+function uniqueChains(chainIds: string[]) {
+  return [...new Set(chainIds)]
+}
```
maciekzygmunt: no need for this function since it is used in one place

vilcht:
> for some reason i dont see this section on arbitrum on preview

It seems like the preview env doesnt have INTEROP_CHAINS set which now dictates if this is visible. This is the case after  #11834. You can see that normal interop summary also doesnt work.

packages/frontend/src/server/features/scaling/project/getScalingProjectEntry.ts
```diff
@@ -130,6 +134,113 @@ export interface ProjectScalingEntry {
...
+        })),
+        remainingCount: Math.max(
+          (currentChainData?.tokenCount ?? 0) -
+            (currentChainData?.topTokens.length ?? 0),
+          0,
+        ),
+      },
+    },
+  }
+}
```
maciekzygmunt: extract this function and interface to new file

maciekzygmunt (changes requested):
fix

maciekzygmunt (approved):

PR #11822, opened 2026-05-11
ccip research

lucadonnoh:
Resolves L2B-13960


mateuszradomski (approved):
Discovery part

sekuba (changes requested):
only comment i have is that you could allude more specifically to the protocol's complexity, similarly to LayerZero (like how you are showing in the diagram imo). Is there any guarantee that the 'security config' of ccip, e.g. validation logic, is synced across chains?

also i have this [# Monitoring section](https://l2beat.com/interop/protocols/wormhole-wtt#detailed-description) in wormhole and LZ that talks briefly about the offchain monitoring that is available to find weak links in the configs or anomalies in the crosschain transfers, maybe u can add something similar

sekuba (approved):

PR #11823, opened 2026-05-11
Update exit window info on detailed pages and tooltips

vilcht:


torztomasz:
2. Description string can produce ungrammatical output (low)

  riskView.ts:722, :750, :780: when l2TimelockDelay - selfSequencingDelay <= 0, the new code substitutes regular.value, which EXIT_WINDOW formats as "None". The sentence then reads "users have only None to exit". The pre-change version emitted a (negative) formatSeconds result, which was also wrong but in a different
  way. Consider a small branch or guard for the zero/negative case here, or compute the substring separately from regular.value.

  3. Copy is inconsistent across surfaces (low)

  - RiskBanner.tsx:77, 112, 158: "<value> (emergency upgrade path)" / "<value> (regular upgrade path)"
  - RiskValue.tsx:23, 41, BigIndividualRosette.tsx:126, 135, 160, 169, BigPizzaRosette.tsx:119, 128: "Emergency: <value>" / "Regular: <value>"

  Pick one and use it everywhere; or extract a small helper. Otherwise the same project will read "Emergency: 7d" in a rosette tooltip and "7d (emergency upgrade path)" in the banner.

  4. Warning icon/styling dropped when regular is present (low)

  - RiskBanner.tsx:78 hides the RoundedWarningIcon in compact mode when regular exists.
  - BigIndividualRosette.tsx:145 and BigPizzaRosette.tsx:140 replace the WarningBar with a plain <span>/<p>, losing the icon and themed background.

  This may be intentional (visual breathing room with the extra line), but the sentiment cue from warning.sentiment is no longer conveyed visually — only the text remains. Worth confirming this is the intended design rather than an oversight.

  5. Duplicated branching pattern (very minor)

  The {regular ? 'Emergency: ' + value : value} / {regular && ...Regular: ...} / {warning && !regular && WarningBar} / {warning && regular && plainText} pattern is now repeated 3× across rosette components. A small <ExitWindowSentimentLines value warning regular /> component would remove ~40 LOC and make future
  copy/style changes single-source.

  6. regular not propagated for non-L2 surfaces (verify)

  getL2Risks.ts is updated, and getScalingRosetteValues.ts:78 propagates via spread. I didn't find other risk-table projections that drop regular, but worth a quick check on L3 risk lists and the bridge/DA risk paths since they use the same RiskBanner component.

torztomasz (changes requested):
check comments

vilcht:
ad 2. Done
ad 3. It's intentional, tasks states this explicitely and we have different ammount of space available in different places.
ad 4. Riskbanner behaviour is intentional, task stated to remove the icon in the banner. The tooltip regression is fixed.
ad 5. Done
ad 6. Verified

torztomasz (approved):

PR #11841, opened 2026-05-13
Create per token page

vilcht:


maciekzygmunt:
something is wrong

vilcht:
Same thing as in #11819, we now need INTEROP_CHAINS set in .env for them to show. And it seems that it is not set in preview. This is the case after merging #11834.

torztomasz:
image

1. Symbol should not be sortable
2. We should allow user to see full addres
3. Address should not be sortable
4. By default one sort should be selected

torztomasz:
image

fix the spacings here on mobile

torztomasz:
Fix 
image

packages/frontend/src/pages/interop/components/table/transfer-count-cell/columns.tsx
```diff
@@ -20,140 +22,154 @@ export type TransferRow = InteropProtocolTransferDetailsItem & BasicTableRow
...
-      )
-    },
-    meta: {
-      headClassName: 'text-2xs',
-      tooltip:
-        'The source and destination chains for this transfer. External links open the corresponding chain transactions.',
-    },
-  }),
-]
+export const columns = getTransferColumns()
```
torztomasz: remove I guess?

packages/frontend/src/pages/interop/components/flows/FlowsGeneralStats.tsx
```diff
@@ -8,18 +8,32 @@ import { formatPercent } from '~/utils/calculatePercentageChange'
...
 import { TokensDialog } from '../tokens/TokensDialog'
 import { InteropTopItems } from '../top-items/TopItems'
 import { useScaledParticleCounts } from './graph/utils/useScaledParticleCounts'
 import { useInteropFlows } from './utils/InteropFlowsContext'
 
-export function FlowsGeneralStats() {
+export function FlowsGeneralStats({
+  tokenId,
+  title = 'General stats',
+  description = 'For past 24h between the selected chains and protocols',
```
torztomasz, line 19: let's not add defaults

packages/frontend/src/pages/interop/components/protocols/TopProtocolsCell.tsx
```diff
@@ -0,0 +1,42 @@
...
+  const [, setIsOpen] = useState(false)
+
+  const visible = protocols.slice(0, VISIBLE_PROTOCOL_COUNT)
+  const remainingCount = protocols.length - visible.length
+
+  return (
+    <InteropTopItems
+      topItems={{
+        items: visible.map((p) => ({
+          id: String(p.id),
```
torztomasz: .toString()

packages/frontend/src/pages/interop/components/tokens/columns.tsx
```diff
@@ -247,19 +251,36 @@ export const getTopTokensColumns = ({
...
+                Issued by{' '}
+                <span className="capitalize">{ctx.row.original.issuer}</span>
+              </TwoRowCell.Second>
+            )}
+          </>
+        )
+
+        return (
+          <TwoRowCell>
+            {selectedChains && ctx.row.original.id !== 'unknown' ? (
```
torztomasz: I do not like the fact you have to check for `unknown` token id everywhere. I think I would return `undefined` from `getInteropTokenUrl` in this case

packages/frontend/src/pages/interop/components/tokens/columns.tsx
```diff
@@ -346,3 +369,24 @@ export const getTopTokensPairsColumns = ({
...
 ]
+
+function TokenPairSymbol({
+  token,
+  selectedChains,
+}: {
+  token: TokensPairRow['tokenA']
+  selectedChains: InteropSelection | undefined
+}) {
+  if (!selectedChains || token.id === 'unknown') {
```
torztomasz: again

packages/frontend/src/pages/interop/components/widgets/protocols/utils/getProtocolsDataWithOthers.ts
```diff
@@ -32,6 +32,7 @@ export function getProtocolsDataWithOthers(
     {
       name: OTHERS_PROTOCOL_NAME,
       slug: undefined,
+      iconUrl: '',
```
torztomasz: 👀

torztomasz:
@codex review

torztomasz:
Issues & suggestions

  Correctness / consistency

  - Duplicate scopeRecordsToToken with subtly different behavior. There are two near-identical implementations:
    - packages/frontend/src/server/features/scaling/interop/getInteropFlows.ts:262 (preserves only basic fields).
    - packages/frontend/src/server/features/scaling/interop/getInteropTokenData.ts:125 (also preserves transferTypeStats, minTransferValueUsd, maxTransferValueUsd, mintedValueUsd, burnedValueUsd).

  The flows version drops fields that the more complete version keeps, which means token‑scoped flows queries lose those per-token attributes (e.g., net minted) for downstream calculations. Extract a single
  shared helper and pick one canonical shape. If those fields aren't needed in the flows path, add a comment explaining why — otherwise this is a quiet bug.
  - InteropTokenPage.tsx:43 — dead branch. navigationSections is a module-level constant with 5 entries, so isNavigationEmpty is statically false. Either drop the guard or make the sections list dynamic.
  - TopProtocolsCell.tsx:14 — dead state. const [, setIsOpen] = useState(false) — the value is never read and hideDialog={true} means setIsOpen is never called either. Remove it (and probably setIsOpen from
  InteropTopItems when hideDialog).
  - packages/frontend/src/pages/interop/components/table/transfer-count-cell/columns.tsx:191 — leftover compat export. export const columns = getTransferColumns() is kept "just in case" — check git grep "from
  .*transfer-count-cell/columns'" -- '*.ts*'. If nothing imports columns anymore, delete it.
  - packages/frontend/src/pages/interop/components/widgets/protocols/utils/getProtocolsDataWithOthers.ts:34 — iconUrl: '' for the "Others" pseudo-protocol. Empty string will likely render a broken . If the
  protocol entry type now requires iconUrl, make it string | undefined and let consumers branch, or pass a placeholder URL.
  - InteropRouter.ts:179 — incidental whitespace edit. A blank line was removed right before return router. Unrelated to this PR — please drop the noise.

  UX / minor

  - OnchainDeploymentsSection.tsx renders addresses with formatAddress but doesn't link to the chain explorer. Given every other table in the interop UI links txs/chains, this is a noticeable gap.
  - Transfers section uses manual Pagination driven by useInfiniteQuery (TokenTransfersSection.tsx:142), while the analogous transfer table in TransferCountCell.tsx uses infinite scroll. Two UX patterns for the
  same dataset shape is jarring; pick one.
  - getInteropTokenUrl.test.ts:18 — the test name "returns path without query when selection is not one pair" describes behavior that comes from buildInteropUrl, not this function. Since the unit under test is
  essentially path + buildInteropUrl, the second test only exercises buildInteropUrl's behavior. Either rename to clarify, or move it to buildInteropUrl's test.

  Performance

  - getInteropAbstractTokens.ts:36 fetches all abstract tokens (abstractToken.getAll()) on every uncached SSR call to filter against the currently-active token IDs. Fine if abstractToken table is small; worth
  checking that this scales or adding a .getByIds(...) path on the token DB.
  - getInteropTokenData.ts cache key is built from req.params.slug + from + to. Be aware the route accepts any string slug, so a flood of requests with random slugs will populate cache entries that resolve to
  undefined and miss the if (!data) return undefined short-circuit (since the cache stores undefined from getCachedData). Consider validating the slug exists before cache.get.

  Nice-to-haves

  - InteropTokenSummary.tsx:46 does an awkward token?.volume !== null && token?.volume !== undefined check inline; consider extracting a small formatOptionalCurrency helper since the same pattern repeats for
  avgValue.
  - TokenVolumeSection.tsx:33 — defaultSelectedChains slice of 15 is hardcoded; extract to a named constant alongside MIN_SELECTED_CHAINS.

  Risk

  Moderate. The main risk is the divergent scopeRecordsToToken behavior silently producing different aggregates between the flows panel and the token dashboard. Cosmetic issues (dead state, iconUrl: '',
  whitespace) are easy fixes. No security concerns; the new route uses validateRoute with a string slug and the server resolves it via a hashmap lookup.

  Recommendation: request changes to consolidate scopeRecordsToToken and remove the dead code/state before merging. Everything else can ride as small follow-ups.

torztomasz:
image

all should have icon and name

torztomasz:
image

all rows should have explorer link, then you do not need tooltip

torztomasz:
image

data is simply off here or there was no ETH transfer on base

packages/frontend/src/pages/interop/token/InteropTokenPage.tsx
```diff
@@ -0,0 +1,210 @@
...
+  initialSelection: InteropSelection
+}
+
+const navigationSections: ProjectNavigationSection[] = [
+  { id: 'summary', title: 'Summary' },
+  { id: 'interop-volume', title: 'Volume and flows' },
+  { id: 'contracts', title: 'Onchain deployments' },
+  { id: 'interop-tokens', title: 'Top protocols' },
+  { id: 'interop-transfers', title: 'Transfers' },
+]
```
torztomasz: that should have been passed from server as always

packages/frontend/src/pages/interop/token/InteropTokenPage.tsx
```diff
@@ -0,0 +1,210 @@
...
+          </div>
+          <div className="row-start-2">
+            {data === null ? (
+              <InteropEmptyState
+                showResetButton={isDirty}
+                onResetButtonClick={reset}
+              />
+            ) : (
+              <TokenDetails
+                data={data as InteropTokenDashboardData | undefined}
```
torztomasz: no casting please

packages/frontend/src/pages/interop/token/InteropTokenPage.tsx
```diff
@@ -0,0 +1,210 @@
...
+          sectionOrder="2"
+          deployments={data?.deployments}
+          interopChains={interopChains}
+          isLoading={isLoading}
+        />
+        <ProjectSection
+          id="interop-tokens"
+          title="Top protocols"
+          sectionOrder="3"
+        >
```
torztomasz: naaah, its not our pattern

packages/frontend/src/pages/interop/token/InteropTokenPage.tsx
```diff
@@ -0,0 +1,210 @@
...
+      <HighlightableLinkContextProvider>
+        <TokenVolumeSection
+          tokenId={tokenId}
+          data={data}
+          interopChains={interopChains}
+          sectionOrder="1"
+        />
+        <OnchainDeploymentsSection
+          sectionOrder="2"
+          deployments={data?.deployments}
```
torztomasz: this doesnt have to be fetched through query, can be passed from server.

torztomasz (dismissed):
I do not like the whole data fetching layer. It does not use our codebase patterns. Things are fetched dynamically where they do not need to be. Please review and refactor all of that.

torztomasz:
## 🚨 Blockers

  ### 1. `getInteropTokenTransfers.ts` is ~95% copy-paste of `getInteropProtocolTransfers.ts`

  `packages/frontend/src/server/features/scaling/interop/getInteropTokenTransfers.ts`

  The protocol file already accepts a generic matcher (`tokenId` is even plumbed through `toInteropProtocolTransferDetailsItem` — and this branch *adds `srcTokenIssuer`/`dstTokenIssuer`* to it). Yet the new file duplicates:

  - `INTEROP_CHAIN_DETAILS` (L22–31)
  - `getFilteredTransfersPage` → renamed `getFilteredTokenTransfersPage` with one extra `&&` (L93–167)
  - `getPageSize` (L169–180), `toTransferCursor` (L182–189), `DEFAULT_PAGE_SIZE`/`MAX_PAGE_SIZE`/`RAW_BATCH_SIZE`
  - `getMockInteropTokenTransfers` (L191–231) parallels `getMockInteropProtocolTransfers`

  Extract one `getFilteredInteropTransfersPage({ projects?, tokenId?, ... })` and one shared `INTEROP_CHAIN_DETAILS` module. The token version becomes a ~30-line orchestrator that composes the matcher and calls the shared
  helper. Deletes ~180 lines.

  ### 2. `getTopPath` and `InteropTopPathData` duplicated verbatim

  `packages/frontend/src/server/features/scaling/interop/getInteropTokenData.ts` L126–146 is a verbatim copy of `getTopPath` in `getInteropProtocolData.ts`, with the type renamed `InteropTokenTopPathData` (L34–38) but
  structurally identical to `InteropTopPathData`. Move both to `utils/getTopPath.ts` and share. There is no token-specific behavior here.

  ### 3. Dead code: `OnchainDeploymentsSection.tsx` (164 LOC)

  `packages/frontend/src/pages/interop/token/components/OnchainDeploymentsSection.tsx`

  The commit `174b5e1d2 nuke wrong data fetching and hide deployments section` removed the section from the page but **left the file**. It's imported nowhere and not in the section registry. Delete it — git history is fine.

  ---

  ## Strong recommendations

  ### 4. `FlowsGeneralStats.tsx`: props-as-feature-flags

  `packages/frontend/src/pages/interop/components/flows/FlowsGeneralStats.tsx`

  The diff bolted `tokenId?`, `title`, `description`, `hideTokenStats?`, `className?` onto a previously single-purpose component, with `hideTokenStats` gating three separate JSX regions (unique-tokens card, top-token row,
  tokens dialog). Also: `tokensQueryInput` is *not* given `tokenId`, so `hideTokenStats=false` + `tokenId` set is silently inconsistent.

  Either split into `FlowsGeneralStats` (shared shell) + a thin token-aware wrapper, or introduce a discriminated `mode: 'all' | 'token'` so the contract is explicit. Don't keep three independent booleans.

  ### 5. `TokenTopProtocolCard.tsx` — three-layout duplication (269 LOC)

  `packages/frontend/src/pages/interop/token/components/TokenTopProtocolCard.tsx`

  `TopProtocolStatCards` (L138) and `TopProtocolStatRows` (L175) are near-identical: same labels, same tooltip copy duplicated verbatim (L149 vs L186, L161 vs L198), same `topProtocol ? format(...) : <span>{EM_DASH}</span>`
  pattern repeated 4×. `ProtocolStatCard` vs `ProtocolStatRow` (L212–258) differ only in container CSS.

  Drive both layouts from one stats-data array, parametrize the container, and unify the two row/card primitives. Realistic clean size ~150 LOC.

  ### 6. `InteropTokenVolumeSection` duplicates `FlowsGeneralStats` legend

  `components/projects/sections/interop/InteropTokenVolumeSection.tsx` L117–139 recomputes `avgValuePerSecond` and renders the "1 particle ≈ $X" annotation already implemented at
  `pages/interop/components/flows/FlowsGeneralStats.tsx` L206–228. Same formula, same copy. Extract `FlowsParticleLegend`. The magic `25` constant is also repeated at L109 and L147 of the volume section — name it.

  ### 7. `InteropTokenPage.tsx` — `PageLayout` duplicates the protocol page shell

  `pages/interop/token/InteropTokenPage.tsx` L163–215 is structurally the same as the equivalent block in `InteropProtocolPage.tsx` (`SideNavLayout` + `MobileSectionNavigation` + `DesktopProjectNavigation` + grid). Extract
  `<InteropEntityPageLayout>` and reuse.

  ### 8. `scopeRecordsToToken.ts` quietly conflates and drops data

  `server/features/scaling/interop/utils/scopeRecordsToToken.ts`

  - `srcValueUsd: volume, dstValueUsd: volume` (L31–32) collapses directional values — semantic mismatch with `AggregatedInteropTransferWithTokens`. Either add a comment explaining downstream uses don't care, or use a
  narrower output type (e.g. `ScopedInteropTransfer`) so the contract is honest.
  - `transferTypeStats: tokens[0]?.transferTypeStats` (L37) silently picks the first token row. Add an assertion or a comment that `filter(...)` is expected to yield exactly one.
  - `identifiedCount: transferCount` (L34) — same: make the assumption explicit.

  Fix the type, not the comment.

  ### 9. `InteropTokenSummary.tsx` — null-handling repetition

  `pages/interop/token/components/InteropTokenSummary.tsx` L44–47, L78–81

  `token?.X !== null && token?.X !== undefined ? format(X) : EM_DASH` repeats. Replace with one `formatOrDash(value, formatter)` helper.

  ---

  ## Micro-smells

  - `resolveInteropTokenBySlug.ts`: builds a `Map` for a single lookup — use `.find()`.
  - `getInteropTokenUrl.ts` L9: `'unknown'` magic string — promote to a shared sentinel constant; the `id?: string` type is loose since it's only used to detect the sentinel.
  - `getInteropTokenPageData.ts` L59–63 & L111–116: `interopChains` with icons computed twice (once outside, once inside the cache callback). Extract `mapInteropChainsToWithIcons(manifest)`.
  - `getInteropTokenData.ts` L92: magic `flows.slice(0, 2)` — comment or named constant.
  - `InteropTokenTransfersSection.tsx` L38–39: local `selectedFrom`/`selectedTo` initialized from `apiSelection` but never re-synced. If the page-level chain filter changes, this section ignores it — confirm intent or wire
  it up.
  - `getAbstractTokenSlug.ts`: local `slugify` — grep for existing helpers before keeping a private one. The URL is critical and slug-stability matters.

packages/frontend/src/pages/interop/utils/mapInteropChainsToWithIcons.ts
```diff
@@ -0,0 +1,13 @@
+import type { InteropChain } from '@l2beat/config'
+import type { Manifest } from '~/utils/Manifest'
+import type { InteropChainWithIcon } from '../components/chain-selector/types'
+
+export function mapInteropChainsToWithIcons(
```
maciekzygmunt, line 5: lets maybe use it everywhere on interop pages, we do this in many places

packages/frontend/src/server/features/scaling/interop/token/getInteropAbstractTokens.ts
```diff
@@ -0,0 +1,66 @@
...
+export type InteropAbstractToken = Pick<
+  AbstractTokenRecord,
+  'id' | 'symbol' | 'issuer' | 'iconUrl' | 'category'
+>
+
+export async function getInteropAbstractTokens(
+  chainIds: string[],
+): Promise<InteropAbstractToken[]> {
+  if (env.MOCK) {
+    return [
```
maciekzygmunt: move it to some const below, so this function stay clean

packages/frontend/src/server/features/scaling/interop/token/getInteropAbstractTokens.ts
```diff
@@ -0,0 +1,66 @@
...
+  })
+  const tokenIds = unique(
+    records.flatMap((record) =>
+      record.tokens.map((token) => token.abstractTokenId),
+    ),
+  )
+  if (tokenIds.length === 0) return []
+
+  const tokens = await getTokenDb().abstractToken.getAll()
+  const relevantTokenIds = new Set(tokenIds)
```
maciekzygmunt: dont call getAll(), but getByIds

packages/frontend/src/pages/interop/components/widgets/TopTokenWidget.tsx
```diff
@@ -98,27 +107,45 @@ function TopTokenHeading({ hideChainsInfo }: { hideChainsInfo?: boolean }) {
 function TopTokenIdentity({
   topToken,
   isLoading,
+  apiSelection,
 }: {
   topToken: InteropDashboardData['topToken'] | undefined
   isLoading: boolean
+  apiSelection?: InteropSelection
 }) {
+  const selectedChainsContext = useContext(InteropSelectedChainsContext)
```
maciekzygmunt, line 116: why not use `useInteropSelectedChains`?

maciekzygmunt (changes requested):
something is wrong with navigation, i got to USDC page from interop flows view, it take me to usdc page with some params, and after slight moment i can see that those params change

packages/frontend/src/pages/interop/components/widgets/TopTokenWidget.tsx
vilcht: protocol page isn't wrapped in the provider in current implementation.

maciekzygmunt (approved):

PR #11845, opened 2026-05-13
Fix negative-height bug, name collisions and others

sdlyy:
Panel-nodes: fix negative-height bug, name collisions, and related quality issues
While debugging a malformed node bounding box on a CCIP layout (height -66px, click hit-test impossible), surfaced a chain of related bugs in the panel-nodes layout pipeline. Fixes are scoped to that pipeline plus two pre-existing console warnings that surfaced in the same session.

Bugs fixed
1. Negative box height from stale hiddenFields 
When discovery output changed (fields removed/renamed) after a layout was saved, stored hiddenFields accumulated names that no longer existed on the node. The height formula HEADER + (fields.length - hidden.length) * FIELD + padding went negative, producing an invalid inline height style (rendered as auto-fit) and an unclickable hit box (boxContains short-circuits on negative height).
New reconcileHiddenFields(fieldNames, savedHidden) helper: intersects saved hidden names with current field names, dedupes.
Applied in both applyStoredLayout modes and in loadNodes' combinedHiddenFields.
Defensive Math.max(0, …) clamp on visible-field count in updateNodePositions and loadNodes.

2. Duplicate field names from object-key recursion
toNodeFields recursed into [key, value] pairs and used the literal path ${path}.#key for every entry's key — producing N fields all named e.g. poolsMap.#key. Consequence: hide/show was all-or-nothing per object, and the height formula undercounted filtering (one hidden name removed many rendered rows).
Key-side path now nests under the entry: ${entryPath}#key (no leading dot, to avoid splitting on . in buildFieldTree and producing the same property as the sibling value-side field — that earlier produced duplicate React keys in FieldNode).

3. SVG viewBox: Infinity Infinity … when no connections are visible
Bounding-box accumulator started at ±Infinity; if connections was empty (everything hidden), nothing reset them. Now skips the <svg> element entirely when there are no connections.

4. Missing key on Fragment in OutputsBadge
The mapped child was a <>…</> whose first \<li> carried the key. Replaced with <Fragment key={project}> so the iterated element actually carries the key.

Resolves L2B-13875


sdlyy:
@codex review

adamiak (approved):

PR #11847, opened 2026-05-13
Update exit window tooltip

vilcht:


vilcht:
Probably should also update rosette tooltip, but it is changed in a different PR, will wait until it's merged

torztomasz:
LGTM but AI found these. Please review.
```
1. Behavior regression: real warnings silently dropped from the risk table tooltip — ExitWindowCell.tsx:71-77

  The old tooltip rendered value.warning as a WarningBar unconditionally. The new non-regular branch renders only value + description — value.warning is never read on that path.

  Non-regular exit windows with genuine warnings exist:

  - cartesi-prt-honeypot.ts:180 — sentiment bad: "The single hardcoded address can withdraw all funds."
  - katana.ts:185 — "Even though there is a … Timelock for non-emergency upgrades, self-proposing is disabled."

  Both vanish from the risk table tooltip with this PR. And it's inconsistent within this same diff: RosetteTooltipRisk.tsx:41-48 kept the WarningBar for the non-regular case. This looks like an oversight, not a
  design call. Presumptive blocker until either fixed or explicitly justified.

  2. The real code-judo move was missed: warning doubling as "regular description" is the disease — this PR spreads it

  The new code does this in two places (RosetteTooltipRisk.tsx:26, ExitWindowCell.tsx:22):

  description={risk.warning?.value}   // "Regular" section

  This hard-codes an undocumented invariant: when regular exists, warning is not a warning — it's the regular path's description. The config already winks at this coupling — stackExitWindowRisk
  (riskView.ts:945-952) has a comment explaining it must drop warning whenever it drops regular, and RiskValue.tsx:25 carries the props.warning && !props.regular special case for the same reason. This PR takes
  that implicit contract from one renderer to three.

  The judo move is one level down, and it's small: give regular its own description.

  // config/src/types.ts
  export interface ExitWindowRisk extends TableReadyValue {
    regular?: Pick<TableReadyValue, 'value' | 'sentiment' | 'description'>
  }

  Producers (EXIT_WINDOW_ZKSTACK, EXIT_WINDOW_NITRO, EXIT_WINDOW_STARKNET, EXIT_WINDOW_PERMISSIONLESS_BOLD, withRegularExitWindow, plus the hand-rolled zksync.ts:184-198) already build that description string —
  they just stuff it into warning: { value: description, sentiment: 'warning' } today. Move it to regular.description and:

  - the tooltip reads regular.description — the repurposing disappears;
  - stackExitWindowRisk no longer needs to strip warning in lockstep with regular — the comment-documented coupling at riskView.ts:945-948 is deleted;
  - RiskValue.tsx:25's warning && !regular special case collapses to warning;
  - warning becomes a warning again, everywhere, which also makes finding #1 trivially correct: render WarningBar whenever warning exists, no branching on regular.

  Behavior-preserving, deletes an invariant readers must hold in their heads, and removes branches in three files instead of adding them in two. This is the version that feels inevitable in hindsight.

  3. Copy-paste: ExitWindowTooltipSection ≡ RosetteTooltipRiskSection

  RosetteTooltipRisk.tsx:58-83 and ExitWindowCell.tsx:86-105 are the same component written twice — identical prop shape (label, value, sentiment, description), identical structure, diverging only in typography
  (vibrant/text-heading-18 vs plain font-medium). Same for the Emergency/Regular two-section layout wrapped around them (RosetteTooltipRisk.tsx:14-29 vs ExitWindowCell.tsx:56-70).

  Two copies of a brand-new concept in one PR is how these drift apart. Either:

  - extract one shared RiskTooltipSections (or similar) into components/rosette/ and have ExitWindowCell use it, accepting unified typography; or
  - if the table tooltip genuinely needs different type treatment, a single component with a small variant — still one place that knows "regular exit windows render as Emergency/Regular sections."

  Note this also fixes a divergence already present in the diff: required label in one copy, optional in the other.

  4. Minor

  - ExitWindowCell.tsx:22 — const regularDescription = regular ? value.warning?.value : undefined is computed eagerly but only used inside the regular ? branch, where the guard is redundant. Inline
  value.warning?.value at the use site. (Moot if you take #2.)
  - BigIndividualRosette tooltip typography changed as a side effect (description text-xs → text-paragraph-13, value gains text-heading-18 mb-2). Probably intended unification — just confirm it was looked at,
  since the diff reads as a pure prop-removal.
  ```

torztomasz (approved):

PR #11850, opened 2026-05-13
Highlight overlapping nodes

sdlyy:
Adds a visual ghost outline around any node whose bounding box intersects another visible node's, so a small node tucked behind a larger one is no longer invisible. Default on, togglable in the nodes panel settings.

- new computeOverlappingIds (sweep-line on x-axis, O(n log n) typical), wrapped in useMemo keyed on nodes/hidden/preference. Skipped entirely when the preference is off.
- accepts isOverlapping; renders a sibling overlay (pointer-events-none, z-20, 2px dashed border colored with the node's own getColor(node).color). Suppressed for dimmed / grayed-out nodes.
- new userPreferences.highlightOverlapping, default true.
-  new "Highlight overlapping nodes" toggle in the existing settings dialog.

image


sdlyy:
@codex review

adamiak (approved):

PR #11853, opened 2026-05-13
Add underline to show protocols are clickable

vilcht:


vilcht:
I tried with the arrows but it didn't look good so only kept underlines on hover

maciekzygmunt (approved):

PR #11865, opened 2026-05-15
add backoffice stats

vilcht:


torztomasz:
@codex review

packages/backoffice/src/pages/interop/highlights/HighlightsPage.tsx
```diff
@@ -0,0 +1,365 @@
...
+  const { data, error, isError, isLoading, isFetching, refetch } =
+    api.interop.highlights.latest.useQuery()
+
+  const topPath = data?.topPathByVolume ?? null
+  const topChainByInflow = data?.topChainByInflow ?? null
+  const chainIncrease = data?.largestVolumeIncreaseByChain ?? null
+  const tokenIncrease = data?.largestVolumeIncreaseByToken ?? null
+  const protocolIncrease = data?.largestVolumeIncreaseByProtocol ?? null
+  const uopsIncrease = data?.largestUopsIncreaseByChain ?? null
+  const tvsIncrease = data?.largestTvsIncreaseByChain ?? null
```
torztomasz, line 38: why `?? null` cant we use undefined?

packages/backoffice/src/pages/interop/highlights/HighlightsPage.tsx
```diff
@@ -0,0 +1,365 @@
...
+                  entity={tokenIncrease?.abstractTokenId}
+                  primaryLabel="Increase"
+                  primaryValue={
+                    tokenIncrease
+                      ? `+${formatDollars(tokenIncrease.increaseUsd)}`
+                      : undefined
+                  }
+                  secondary={volumeSecondary(tokenIncrease)}
+                  emptyText="No positive token volume delta"
+                />
```
torztomasz, line 235: Image

Show some info about the token. Symbol, icon etc

packages/backoffice/src/pages/interop/highlights/HighlightsPage.tsx
```diff
@@ -0,0 +1,365 @@
...
+    </div>
+  )
+}
+
+function getProtocolHref(protocolId: string | undefined) {
+  if (protocolId === undefined) {
+    return undefined
+  }
+
+  return `${L2BEAT_URL}/interop/protocols/${protocolId}`
```
torztomasz: just hardcode the path, no need for `L2BEAT_URL`

packages/database/src/repositories/ActivityRepository.test.ts
```diff
@@ -50,6 +50,83 @@ describeDatabase(ActivityRepository.name, (db) => {
...
+        )
+
+        expect(result).toEqual({
+          timestamp,
+          projectId: ProjectId('arbitrum'),
+          currentUopsCount: 300,
+          previousUopsCount: 50,
+          increase: 250,
+        })
+      })
```
torztomasz: shouldnt this check percentage increase? I do not care if lighter grows from 1200 to 1250 but I would care if ethereum grew from 20 to 50

packages/database/src/repositories/AggregatedInteropTokenRepository.ts
```diff
@@ -196,4 +204,70 @@ export class AggregatedInteropTokenRepository extends BaseRepository {
...
+      .selectFrom(current)
+      .leftJoin(previous, (join) =>
+        join.onRef(
+          'current.abstract_token_id',
+          '=',
+          'previous.abstract_token_id',
+        ),
+      )
+      .select([
+        sql<string>`"current"."abstract_token_id"`.as('abstract_token_id'),
```
torztomasz:
```suggestion
        'current.abstract_token_id as abstract_token_id',
```

packages/database/src/repositories/AggregatedInteropTokenRepository.ts
```diff
@@ -196,4 +204,70 @@ export class AggregatedInteropTokenRepository extends BaseRepository {
       .executeTakeFirst()
     return Number(result.numDeletedRows)
   }
+
+  async getLargestTokenVolumeIncrease(
```
torztomasz: I do not think this is a good idea to keep such a complex function with business logic inside the repo. I would query data for timestamp and previousTimestamp and calculate all we need in TS.

torztomasz: Same applies for other functions

torztomasz (changes requested):

torztomasz:
@codex review

packages/backend/src/modules/interop/engine/dashboard/impls/highlightsCalculations.ts
```diff
@@ -0,0 +1,543 @@
...
+
+function getTransferVolumeUsd(record: AggregatedInteropTransferRecord) {
+  const { srcValueUsd, dstValueUsd } = record
+
+  if (srcValueUsd !== undefined && dstValueUsd !== undefined) {
+    return Math.max(srcValueUsd, dstValueUsd)
+  }
+
+  return srcValueUsd ?? dstValueUsd ?? 0
+}
```
torztomasz: already exists `packages/shared-pure/src/utils/getInteropTransferValue.ts`

packages/backend/src/modules/interop/engine/dashboard/impls/highlights.ts
```diff
@@ -0,0 +1,348 @@
...
+  windowEnd: number
+  previousWindowStart: number
+  previousWindowEnd: number
+  currentCount: number
+  previousCount: number
+  increase: number
+  increasePercent: number
+}
+
+export async function getInteropHighlights(
```
torztomasz: oh boy, just do early return when latestTimestamp is not present or whatever. look at the code, you are checking it bazzillion times

torztomasz (changes requested):

torztomasz:
@codex review

vilcht:
@codex review

torztomasz:
1. highlightsCalculations.ts:188–355 — three near-identical functions; clear code-judo move (structural regression).
  getLargestSourceChainVolumeIncrease, getLargestProtocolVolumeIncrease, and getLargestTokenVolumeIncrease are byte-for-byte the same algorithm. Each: builds a currentVolumes map, builds a previousVolumes map
  the same way, diffs them, filters increase > 0, and maxBy(increase, tiebreak-by-key). The only differences are the key field (srcChain / id / abstractTokenId) and the value extractor
  (getInteropTransferValue(record) ?? 0 vs record.volume). That's ~150 lines that collapse to one helper plus three one-line call sites:

  function largestVolumeIncrease<T>(
    current: T[], previous: T[],
    keyOf: (r: T) => string | undefined,   // undefined → skip (the src===dst guard)
    valueOf: (r: T) => number,
    timestamp: UnixTime,
  )

  This is the headline issue — please extract it before merge rather than ship three copies.

  2. highlightsCalculations.ts:71–186 and 357–530 — two more duplicated families.
  - getTopPathByVolumeAtTimestamp and getTopDestinationChainByInflowAtTimestamp share the identical "group into {volumeUsd, transferCount, protocolIds:Set}, then maxBy" shape — differ only in grouping key and
  value field. Extract a groupAggregate(records, keyOf, valueOf).
  - getLargestUopsCountIncrease and getLargestTvsIncrease share the whole current/previous/older + isIncluded + projectsWithOlderSnapshot + diff + maxBy skeleton. They can sit on the same helper as #1 with an
  added "must exist in older snapshot" predicate.

  Taken together (#1–#2) this 540-line file is really ~3 algorithms copied 7 times; a principled extraction roughly halves it and makes each highlight a few lines of config.
  
  4. ActivityRepository.ts & TokenValueRepository.ts — dead/speculative methods.
  getLatestTimestamp and getMaxTimestampAtOrBefore (the non-ForProjects variants) were added to both repos but are unused anywhere (confirmed by grep across packages/backend/src). getInteropHighlights only uses
  the ForProjects variant and getByTimestamp. Remove the four unused methods — they're not requested by this feature.

  5. highlights.ts:111–116 — undocumented "magic" timestamp adjustment.
  if (activityTimestamp === UnixTime.toStartOf(latestTimestamp, 'day')) {
    activityTimestamp = activityTimestamp - UnixTime.DAY
  }
  A special-case branch with no comment explaining why (presumably: today's activity row is partial, so step back to the last complete day). At minimum add a comment; this is the kind of incidental conditional
  that becomes unexplained debt.

packages/backend/src/modules/interop/engine/dashboard/impls/highlights.ts
```diff
@@ -0,0 +1,374 @@
...
+  db: Database,
+  tokenDbClient: TokenDbClient,
+  interopProjectIds: readonly string[],
+): Promise<InteropHighlightsData> {
+  const latestTimestamp =
+    await db.aggregatedInteropTransfer.getLatestTimestamp()
+
+  if (latestTimestamp === undefined) {
+    return emptyHighlights()
```
torztomasz: why function not const?

torztomasz (approved):

PR #11869, opened 2026-05-18
Archive few dead projects

lucadonnoh:


maciekzygmunt (changes requested):
add also those same timestamps to `chainConfig` -> `untilTimestamp` property

lucadonnoh:
@maciekzygmunt why? maybe you should make it so that the archive also archives the `chainConfig` automatically no? what's the point of two archives

sekuba (approved):

maciekzygmunt:
@lucadonnoh We need to add it so we don’t fetch data from those chains, since in most cases the RPCs are down. And yeah, I agree that this should be done automatically. We just haven’t had time to refactor it yet

maciekzygmunt (approved):

PR #11875, opened 2026-05-18
Dailies for grvt, myria, lighter, torando cash

sergeyshemyakov:


vincfurc:
lighter proposer failure should be green now with escape hatch?

packages/config/src/projects/lighter/lighter.ts
vincfurc: also this needs updating

vincfurc: this can be removed, reconstructed and exitable

vincfurc (approved):

PR #11879, opened 2026-05-19
add taiko gov, diagrams

sekuba:
close L2B-12949


packages/config/src/projects/taiko/taiko.ts
vincfurc, line 405: is this still true? "The use of SGX (Geth) is mandatory"

vincfurc (approved):

packages/frontend/static/images/upgrades-and-governance/taiko.png
vincfurc, line 1: Can you double check where Taiko Multisig manages the verifiers, I don't see it

PR #11885, opened 2026-05-19
Add updates section

vilcht:


packages/frontend/src/server/features/projects/recent-changes/diffHistory/DiffHistoryParser.ts
```diff
@@ -0,0 +1,196 @@
...
+
+const SECTION_HEADERS: Record<string, DiffHistorySectionKind> = {
+  '## Watched changes': 'watched-changes',
+  '## Initial discovery': 'initial-discovery',
+  '## Source code changes': 'source-code-changes',
+  '## Config/verification related changes': 'config-related-changes',
+  '## Config related changes': 'config-related-changes',
+}
+
+export class DiffHistoryParser {
```
maciekzygmunt: this file is a copy from `l2b` package, lets leave one of them and move it to `shared`

maciekzygmunt: you only add parseWithMap which is only used in one place so it is not needed

packages/frontend/src/server/features/projects/recent-changes/diffHistory/getDiffHistoryPath.ts
```diff
@@ -0,0 +1,70 @@
...
+      }
+      const relativeProjectsDir = config.discovery ?? 'discovery'
+      projectsDir = path.resolve(dir, relativeProjectsDir)
+      return projectsDir
+    }
+    dir = path.dirname(dir)
+  }
+
+  return undefined
+}
```
maciekzygmunt: cant we just hardcode the path?

packages/frontend/src/server/features/projects/recent-changes/diffHistory/getDiffHistoryPath.ts
```diff
@@ -0,0 +1,70 @@
+import { existsSync, readdirSync, readFileSync } from 'fs'
+import path from 'path'
+
+let projectsDir: string | undefined
+let projectsDirResolved = false
+
+export function getDiffHistoryPath(projectId: string): string | undefined {
```
maciekzygmunt: same here, i think we can just hit directly the diffHistroy.md path

packages/frontend/src/server/features/scaling/project/getScalingProjectEntry.ts
```diff
@@ -628,6 +636,8 @@ export async function getScalingProjectEntry(
     })
   }
 
+  appendUpdatesSection(sections, discoveryUpdates)
```
maciekzygmunt: why like this instead of normal way, same as in interop project

packages/frontend/src/utils/diffHistory/diffHistoryMarkdown.ts
```diff
@@ -0,0 +1,81 @@
...
+    spans.push({
+      content: (match[1] ?? '').replace(/\n$/, ''),
+      start: match.index ?? 0,
+      end: (match.index ?? 0) + match[0].length,
+    })
+  }
+  return spans
+}
+
+export function extractDiffBlocks(body: string): string[] {
```
maciekzygmunt: there is no need for this function, also it is expored and not used anywhere

maciekzygmunt (changes requested):
check comments

packages/frontend/src/server/features/scaling/project/getScalingProjectEntry.ts
vilcht: cuz we did add it in 2 places, but yeah, that's an overabstraction

maciekzygmunt (approved):

PR #11909, opened 2026-05-21
fix particle scaling hang on large flow values

torztomasz:
## Summary
- Replaced the incremental loop in `getScaledParticleCounts` with a direct closed-form calculation. The old version bumped `dollarsPerParticle` by `$25` each iteration; with very large input values this took millions of iterations before the caps were satisfied, hanging the page.
- Both per-flow and total-particle caps are linear in `1/dollarsPerParticle`, so the minimum required `dpp` can be computed directly, then rounded up to the next `DOLLARS_PER_PARTICLE_STEP`.
- Renamed a few intermediate variables (`maxBase` → `maxBaseCount`, `minByMax` → `minDppForPerFlowCap`, etc.) for clarity.

## Test plan
- [x] Existing unit tests in `getScaledParticleCounts.test.ts` pass unchanged
- [ ] Verify the interop flows graph renders correctly with large-value flows that previously hung

### DIFF
--- a/packages/frontend/src/pages/interop/components/flows/graph/utils/getScaledParticleCounts.test.ts
+++ b/packages/frontend/src/pages/interop/components/flows/graph/utils/getScaledParticleCounts.test.ts
@@ -31,4 +31,47 @@ describe(getScaledParticleCounts.name, () => {
     expect(result.counts).toEqual([])
     expect(result.dollarsPerParticle).toEqual(50)
   })
+
+  it('returns quickly for very large values without iterating', () => {
+    // The previous implementation looped in $25 increments; with values this
+    // large it would take millions of iterations. Should return in O(n).
+    const start = Date.now()
+    const result = getScaledParticleCounts([1e12, 5e11, 1e10])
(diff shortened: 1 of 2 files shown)


mateuszradomski:
Write tests for this for sanity

packages/frontend/src/pages/interop/components/flows/graph/utils/getScaledParticleCounts.test.ts
```diff
@@ -31,4 +31,47 @@ describe(getScaledParticleCounts.name, () => {
     expect(result.counts).toEqual([])
     expect(result.dollarsPerParticle).toEqual(50)
   })
+
+  it('returns quickly for very large values without iterating', () => {
+    // The previous implementation looped in $25 increments; with values this
+    // large it would take millions of iterations. Should return in O(n).
```
mateuszradomski: These comments are ass, I'd remove them. They are not useful for the future

mateuszradomski (approved):

PR #11916, opened 2026-05-21
Prevent callback memory leaks

sdlyy:
Resolves L2B-14041 as well (should)

We were not disposing some callbacks, causing perf degradation - in the worst-case scenario, these would pile up during long research sessions, causing weird stuff to happen

- unified approach to callbacks
- letting registrars dispose them on their own when unmounting


sdlyy:
@codex review

mateuszradomski (approved):

PR #11917, opened 2026-05-21
DiscoUI hygiene 

sdlyy:
| File | What changed | Why |
|---|---|---|
| `useCodeSettings` | window.history.pushState` → `replaceState` in the URL-sync effect | Line selection changes were creating a new browser history entry each time. Back button stepped through every cursor change instead of leaving the page. |
| `CodePanel` | `files` memo now keyed on `codeResponse.data` + `.isPending` + `.isError` instead of the whole `codeResponse` object. `hasProxy` memo similarly fixed to depend on `codeResponse.data`. | `codeResponse` is a fresh `useQuery` return object every render, so the memo never hit. `files` was a new array reference each time, which churned `EditorView`'s setFile path on every parent re-render — calling Monaco's `saveViewState`/`setModel`/`restoreViewState` for nothing. |
| `debounce.ts` | `useDebouncedCallback` holds the latest callback in a ref and drops `callback` from `useCallback` deps, returning a stable function | Every caller passes an inline arrow, so `callback` was unstable and `useCallback` was effectively a no-op — the returned debounced function changed on every render. Debounce semantics were preserved by the shared `timeoutRef`, but the hook violated its own "stable returned function" contract. |
| `editor.ts` | Added `disposed` flag, early-return in `dispose()` if already disposed | Stray double `dispose()` (e.g., during odd HMR or React reconciliation) would re-run `saveViewState()` on a torn-down Monaco editor and could write garbage into the URI-keyed view-state cache. |
| `CodeEditorComponent.tsx` | Removed the `if (existingEditor) { return cleanup }` branch and dropped `getEditor` from destructure/deps | Dead code that was actively broken: if it ever fired, the component would render an empty `<div>` with no Monaco instance, and on unmount would dispose someone else's editor. Removing surfaces the underlying programmer error (two components with same `editorKey`) instead of silently clobbering. |

### DIFF
--- a/packages/protocolbeat/src/apps/discovery/panel-code/CodePanel.tsx
+++ b/packages/protocolbeat/src/apps/discovery/panel-code/CodePanel.tsx
@@ -46,11 +46,17 @@ export function CodePanel() {
     }
 
     return sources.length > 1
-  }, [codeResponse.isPending, selectedAddress])
+  }, [codeResponse.data])
 
   const files = useMemo(
     () => getCodeFiles(codeResponse, selectedAddress, hasCode),
-    [codeResponse, selectedAddress, hasCode],
+    [
+      codeResponse.data,
+      codeResponse.isPending,
+      codeResponse.isError,
+      selectedAddress,
+      hasCode,
+    ],
   )
 
   const showRediscoverInfo = codeResponse.isError && !IS_READONLY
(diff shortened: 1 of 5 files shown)


sdlyy:
@codex review

mateuszradomski (approved):

PR #11919, opened 2026-05-21
Introduce layout versioning

sdlyy:
Moved legacy layout normalization into migration, where it belongs. In v1 -> v2, legacy OKLCH color objects are now dropped instead of coerced to 0, because 0 is not a neutral placeholder here, it is an explicit “use default chain color” sentinel, so coercing changed restore/merge behavior. With that fixed, applyStoredLayout and loadNodes only consume the normalized current layout shape and no longer need legacy color translation logic.


sdlyy:
@codex review

sdlyy:
@codex review

mateuszradomski (approved):

PR #11936, opened 2026-05-25
Improve chain selector on path-specific pages

maciekzygmunt:
Resolves L2B-14066


packages/frontend/src/pages/interop/components/table/transfer-count-cell/TransferCountCell.tsx
```diff
@@ -55,7 +55,7 @@ export function TransferCountCell({
         protocol={protocol}
         type={type}
         snapshotTimestamp={snapshotTimestamp}
-        selectionForApi={selectionForApi}
+        selectionForApi={selectedChains}
```
torztomasz:
```suggestion
        selectedChains={selectedChains}
```

packages/frontend/src/pages/interop/components/widgets/TopRoutesWidget.tsx
```diff
@@ -0,0 +1,294 @@
...
+  interopChains: InteropChainWithIcon[]
+  isLoading: boolean
+  flows: InteropDashboardData['flows'] | undefined
+  className?: string
+}) {
+  const getChainDetails = (id: string): ChainDetails => {
+    const chain = interopChains.find((c) => c.id === id)
+    assert(chain, `Chain not found: ${id}`)
+    return { id: chain.id, iconUrl: chain.iconUrl, name: chain.name }
+  }
```
torztomasz: can't this be passed directly in flows? looks bad

packages/frontend/src/pages/interop/components/InitialChainSelector.tsx
```diff
@@ -67,13 +66,8 @@ export function InitialChainSelector({
...
               utils.interop.dashboard.prefetch({
-                ...toInteropApiSelection(
-                  {
-                    from: [firstSelectedChainId],
-                    to: [chain.id],
-                  },
-                  'public',
-                ),
+                from: [firstSelectedChainId],
+                to: [chain.id],
```
torztomasz: wrong

torztomasz (changes requested):

torztomasz:
## 1. `flowsMapToSorted` got the wrong signature — most fixable smell

  `packages/frontend/src/server/features/scaling/interop/utils/getFlows.ts` rewrote the signature to:

  ```ts
  export function flowsMapToSorted(
    volumes: Map<string, number>,
    selection: InteropSelectionInput,
    transferCounts: Map<string, number>,
  ): InteropFlowData[]
  ```

  …which forces four callers in `getProtocolEntries.ts` (lines 177, 202, 228, 252) to pass `new Map()` as a placeholder. That's a tell that the abstraction is wrong: one caller (`getFlows`) needs transfer counts, the other
  (bridge-type tables) categorically doesn't.

  Keep `flowsMapToSorted` single-purpose (sort by volume) and enrich post-sort in the *only* caller that needs counts:

  ```ts
  // flowsMapToSorted stays as before — volumes only
  export function getFlows(...) {
    // ...
    const sorted = flowsMapToSorted(volumeMap, selection)
    if (sorted.every(f => f.volume === 0)) return []
    return sorted.map(f => ({
      ...f,
      transferCount: transferMap.get(toFlowKey(f.srcChain, f.dstChain)),
    })).slice(0, 3)
  }
  ```

  This deletes four `new Map()` arguments, removes a misleading optional field from the shared sort helper's contract, and keeps each module single-responsibility.

  Bonus: `InteropFlowData.transferCount?` is misleading — within `getFlows`'s output it's always set; the `?` only exists because bridge-type callers strip it. The post-sort enrichment makes that boundary explicit instead of
   fake-optional.

  ---

  ## 2. `TopRoutesWidget` — the three-layout dispatcher is unearned

  The new file is 294 lines and the bulk of it is layout dispatch by `flowCount`:

  ```tsx
  {!isLoading && flowCount === 0 && <EmptyState />}
  {!isLoading && flowCount === 1 && flows?.[0] && (
    <SinglePairLayout flow={flows[0]} ... />
  )}
  {!isLoading && flowCount === 2 && flows && (
    <TwoPairsLayout flows={flows} ... />
  )}
  {!isLoading && flowCount >= 3 && flows && (
    <TopRoutesList flows={flows} ... />
  )}
  ```

  `SinglePairLayout` is literally `<RouteCard size="lg" ... />`. `TwoPairsLayout` is a 2-col grid of `<RouteCard size="md" ... />`. They're pure thin wrappers. After that, `RouteCard` itself branches all the way down on
  `isLarge` for padding, gap, icon size, font size, etc.

  First reframing — drop the wrappers:

  ```tsx
  {flowCount <= 2 ? (
    <div className={cn('grid gap-3 h-full', flowCount === 2 && 'grid-cols-2')}>
      {flows.map(f => (
        <RouteCard key={f.srcChain + f.dstChain} flow={f} size={flowCount === 1 ? 'lg' : 'md'} />
      ))}
    </div>
  ) : (
    <TopRoutesList flows={flows} ... />
  )}
  ```

  Deeper question: do `RouteCard` (1–2 flows) and `RouteRow` (3+) really need to be separate components carrying the same data with different paddings? Both render `from → to` + volume + transfers. Either way, the
  `SinglePair`/`TwoPairs` wrappers should go — they don't earn the indirection.

  ---

  ## 3. `isDirty` survives the cleanup but has no real consumer

  After unification, the *only* consumer of `isDirty` is `MultiChainSelector.tsx`:

  ```tsx
  const { isDirty, swapPaths } = useInteropSelectedChains()
  // ...
  <div className={cn('flex items-center gap-3 max-md:hidden', !isDirty && !protocols && 'hidden')}>
    {protocols && (<> ... </>)}
  </div>
  ```

  The conditional `!isDirty && !protocols && 'hidden'` is just a roundabout way of saying "hide when there's nothing to render". Drop it:

  ```tsx
  const { swapPaths } = useInteropSelectedChains()
  // ...
  {protocols && (
    <div className="flex items-center gap-3 max-md:hidden">
      <span className="font-medium text-base leading-none">
        Across {protocols.length} protocols
      </span>
      <AllProtocolsDialog protocols={protocols} />
    </div>
  )}
  ```

  Then `isDirty` is unused entirely and can come off the context. The Reset button it used to gate is already deleted. This is leftover scaffolding from the old model — let it go.

  ---

  ## 4. Page-level empty check uses `&&` where the invariant is `||`

  In all four pages (`InteropBurnAndMintPage`, `InteropLockAndMintPage`, `InteropNonMintingPage`, `InteropSummaryPage`):

  ```tsx
  if (selectedChains.from.length === 0 && selectedChains.to.length === 0) {
    return <InitialChainSelector ... />
  }
  ```

  But the per-side error in `MultiChainSelectorButton.tsx` shows `EmptyStateError` whenever `selectedChains[type].length === 0`. These two invariants disagree. The dashboard query also needs both sides populated to return
  anything useful.

  The model is "both sides required to render dashboard". Match it:

  ```tsx
  if (selectedChains.from.length === 0 || selectedChains.to.length === 0) {
    return <InitialChainSelector ... />
  }
  ```

  Otherwise users who deselect all of `from` from inside the dashboard land in a "dashboard with `data === null` and a generic empty card" state when they should be routed back to the entry UI (or shown the per-side error
  consistently). The current `&&`/per-side mismatch is the kind of asymmetric branching that becomes bug reports later.

  ---

  ## 5. `InitialChainSelector` quietly re-introduces the old "pick exactly two" UX into a multi-select world

  `InitialChainSelector.tsx` (lines 67–70 post-PR):

  ```tsx
  utils.interop.dashboard.prefetch({
    from: [firstSelectedChainId],
    to: [chain.id],
    type,
  })
  ```

  …and the click handler ultimately seeds the URL with one chain on each side. That's the *old* public-mode UX (a 1↔1 pair), now re-emerging as the entry path into a multi-select dashboard. It works, but it's a vestige of
  the deleted bifurcation that this PR claims to remove.

  Real product question: in the unified model, should "no selection" route to the multi-select drawer (empty state, user opens, picks several chains, applies) instead of the dedicated `InitialChainSelector` widget? If yes,
  the `InitialChainSelector` file (and the asymmetric onboarding flow it encodes) could go away entirely — that would be the real code-judo finish on this unification.

  If keeping it is intentional UX, that's fine — but please be explicit that "the unified path always starts with exactly one+one" is now the documented entry behavior.

  ---

  ## 6. The three path-specific pages are ~95% the same file

  Diffing `InteropBurnAndMintPage.tsx` vs `InteropLockAndMintPage.tsx`: 5 lines differ out of ~150. `NonMintingPage` is the same shape. This PR replicates its multi-vs-initial fork three times and the
  `Widgets`/`Content`/`Provider` boilerplate three times.

  Not a presumptive blocker for *this* PR (the duplication predates it), but it's exactly where the next round of code-judo lives. A single `<InteropPathPage type="burnAndMint" extraTableProps={...} />` would collapse three
  files into one. Worth filing as the immediate follow-up.

  ---

  ## 7. `buildUrl` is now a one-line wrapper

  `InteropSelectedChainsContext.tsx`:

  ```ts
  const buildUrl = useCallback(
    (path: string) => buildInteropUrl(path, selection),
    [selection],
  )
  ```

  It does have three consumers (`BurnAndMintCard`, `LockAndMintCard`, `NonMintingCard`), so it's not dead — but it's *just* closing over `selection`. Either inline `buildInteropUrl(path, selectedChains)` at the three call
  sites (which already pull from the same context) and drop `buildUrl` from the context surface, or keep it and document it. Right now it reads like surviving scaffolding from when `buildUrl` had real options to encapsulate
  (`{ mode }`).

  Mild call. Lean: inline and shrink the context surface.

packages/frontend/src/pages/interop/components/widgets/TopRoutesWidget.tsx
```diff
@@ -34,13 +39,20 @@ export function TopRoutesWidget({
...
+    return {
+      from: { id: from.id, iconUrl: from.iconUrl, name: from.name },
+      to: { id: to.id, iconUrl: to.iconUrl, name: to.name },
+      volume: flow.volume,
+      transferCount: flow.transferCount,
+    }
+  })
 
-  const flowCount = flows?.length ?? 0
+  const flowCount = enrichedFlows?.length ?? 0
```
torztomasz: move to server

torztomasz (approved):

PR #11946, opened 2026-05-26
Fix Coingecko usage in token ingestion

adamiak:
Three fixes:
- added rate-limiting of coingecko calls
- coingecko errors no longer interrupt ingestion queue processing and instead are added as "Error" state to impacted tokens.
- don't cache abstarct tokens when displaying token history page.


packages/token-backend/src/ingestion/TokenIngestionProcessor.ts
```diff
@@ -860,6 +882,13 @@ function getNewCoingeckoSymbolConflict(
   return `CoinGecko would create abstract token ${newAbstractToken.id}:${newAbstractToken.symbol}, but the deployed token symbol is ${deployedTokenSymbol}.`
 }
 
+function formatError(error: unknown): string {
```
torztomasz: I guess it can be a global fn reused in multiple places

torztomasz (approved):

adamiak:
@codex review

packages/token-backend/src/ingestion/TokenIngestionProcessor.ts
adamiak: thx, I'll move it in another upcoming fix.

adamiak: or actually fixing here :)

adamiak:
@codex review

PR #11962, opened 2026-05-27
Implement Home page

radinat:
https://claude.ai/code/artifact/3e1c49f8-38ad-43d1-871f-df2784e6edf8


vilcht:
image
Duplicate theme switcher

vilcht:
image
1080p screen 100% scaling doesnt have the full size layout but the compressed one, i feel the breakpoints should be different

vilcht:
image
clicking on the interop stretches columns on the left - for example the what's new thumbnail looks awful then, I don't think these layout shifts are good, idk what would good ux be tho

vilcht:
image
this is on the very bottom on mobile and one of the first things on dekstop. Idk if intentional, but weird.

vilcht:
image
Minor nitpick, but i feel like these icons should be a little bit smaller.

vilcht:
image
hover states on the table are misaligned

vilcht:
image
i feel like the whole thing should be gray on hover. Feel free to keep the current behaviour if proposal looks weird.

vilcht:
image
these look really weird with all the data being stacked on the left, maybe experiment with some other designs? Looks ok tho on the full layout where the tiles are a lot smaller. Feel free to keep

packages/frontend/src/layouts/SideNavLayout.tsx
```diff
@@ -1,320 +1,111 @@
...
-          href: '/brand-kit',
-        },
-        {
-          title: 'FAQ',
-          href: '/faq',
-        },
-      ]),
-    [],
+    <TopBanner
+      className={cn('lg:rounded-b-xl 2xl:rounded-br-none', classes.topBanner)}
```
vilcht: why inject classes and not use CVA?

packages/frontend/src/layouts/SideNavLayout.tsx
```diff
@@ -1,320 +1,111 @@
...
           )}
         >
-          <div className="hidden lg:mr-3 lg:block 2xl:mr-0">{topChildren}</div>
           <div
-            style={
-              {
-                '--tablet-content-horizontal-padding': '20px',
-              } as React.CSSProperties
-            }
+            className={cn('hidden lg:block 2xl:mr-0', classes.bannerWrapper)}
```
vilcht: same here and in other places

packages/frontend/src/pages/home/components/HomeInteropCard.tsx
```diff
@@ -0,0 +1,309 @@
...
+          data={data}
+          hasEnoughChains={hasEnoughChains}
+          hasEnoughProtocols={hasEnoughProtocols}
+          isLoading={isLoading}
+          // Center the width-capped graph in the leftover space so the
+          // card doesn't end in a large blank area at xl.
+          className="pb-2"
+        />
+      </div>
+      {/* Always reserve the bottom panel so the graph doesn't jump when a
```
vilcht: it jumps, so the comment is deceptive

packages/frontend/src/pages/home/components/RecentChangesDialog.tsx
```diff
@@ -0,0 +1,128 @@
...
+              {group.name}
+            </h3>
+            <span className="flex shrink-0 items-center gap-1 font-medium text-link text-xs">
+              View updates
+              <ChevronIcon className="-rotate-90 size-2.5 fill-link" />
+            </span>
+          </a>
+          <div className="flex flex-col gap-3">
+            {group.updates.map((update, index) => (
+              <UpdateCard key={`${update.date}-${index}`} update={update} />
```
vilcht: index keys are not stable

packages/frontend/src/pages/home/components/HomeEthereumCard.tsx
```diff
@@ -0,0 +1,116 @@
...
+              value={
+                totalPosted !== undefined ? formatBytes(totalPosted) : undefined
+              }
+              change={dataPostedChange}
+            />
+          }
+          statFooter={
+            rollupShare !== undefined ? (
+              <span className="font-medium text-label-value-12 text-secondary tabular-nums">
+                {formatPercent(rollupShare)} by rollups
```
vilcht: I dont think we count only rollups

packages/frontend/src/pages/home/getHomeData.ts
```diff
@@ -0,0 +1,250 @@
...
+  return {
+    head: {
+      manifest,
+      metadata: getMetadata(manifest, {
+        title: 'Home - L2BEAT',
+        description:
+          'Bird-eye view of the Ethereum scaling ecosystem: total value secured, activity, interoperability, recent additions and what L2BEAT is currently tracking.',
+        url: req.originalUrl,
+        openGraph: {
+          image: '/meta-images/scaling/summary/opengraph-image.png',
```
vilcht: Still points to scaling image

packages/frontend/src/pages/home/components/HomeStatsStrip.tsx
```diff
@@ -0,0 +1,138 @@
...
+      label: 'ZK Catalog',
+      metric: { count: counts.zkCatalog, unit: 'projects' },
+      href: '/zk-catalog',
+      icon: <ZkCatalogIcon className="size-5 stroke-purple-500" />,
+      iconBgClassName: 'bg-purple-500/10',
+    },
+    {
+      label: 'Ecosystems',
+      metric: { count: counts.ecosystems, unit: 'ecosystems' },
+      href: '/ecosystems/agglayer',
```
vilcht, line 68: I know we don't have an ecosystem overview, but clicking ecosystems button and getting to agglayer page is bad UX imo

vilcht (changes requested):

vilcht:
image
we still have two theme switchers

vilcht:
image
this imo doesn't look good at all, the combination of large icons, borders on what on desktop seems to be a list, i just dont like it

vilcht:
image

Same here, i dislike the nav elements being separate cards on mobile, it seems AI-ish. Also would experiment from removing colored background on icons and making them tad bit larger, would clean up the clutter i think.

vilcht:
image

I don't think te what's new in mobile is necessary. Explore can be dropped also imo. Just make the font larger and the gradient go a bit higher because the contrast is not enough.

vilcht:
image

the hover state on medium screens is bad and inconsistent with other card hovers.

vilcht:
image
Border on this is thicker than the rest of borders. Incosistency

packages/frontend/src/hooks/useLocalStorage.ts
```diff
@@ -10,6 +10,21 @@ declare global {
...
+  [K in `whats-new-${string}`]: boolean
+} & {
+  [K in `top-banner-${string}-is-hidden`]: boolean
+} & {
+  'has-finished-legend-onboarding': boolean
+  'last-read-changelog-entry-id': string | undefined
+}
+
+export type LocalStorageKey = keyof LocalStorageSchema
+
```
vilcht, line 27: Registering every callsite of a hook in the hook seems bad to mee

packages/frontend/src/pages/home/components/charts/HomeChart.tsx
```diff
@@ -0,0 +1,243 @@
...
+  height: 18,
+  tickMargin: 3,
+} as const
+
+/**
+ * Sleek single-series line/area chart used inside the home widgets.
+ * Wraps Recharts' AreaChart through `ChartContainer` (size="small") so we get
+ * the shared loader / no-data state, but omits the legend slot. Fills the
+ * parent's height, with a minimal mirrored y-axis and a compact timeline.
+ */
```
vilcht: all these comments just explain written behaviour, unnecessary

packages/frontend/src/pages/home/components/HomeChartSection.tsx
```diff
@@ -0,0 +1,40 @@
...
+    <div className="flex h-full min-w-0 flex-col xl:min-h-0">
+      <div className="flex min-w-0 items-start justify-between gap-3">
+        <div className="flex min-w-0 flex-col gap-1 pt-0.5">
+          <span className="font-medium text-label-value-14">{label}</span>
+          {statFooter}
+        </div>
+        <div className="flex-none">{stat}</div>
+      </div>
+      {/* mt-auto bottom-aligns charts in side-by-side sections, so their
+          x-axes stay level when a header wraps to a second line. */}
```
vilcht: again, unnecessary comments

packages/frontend/src/pages/home/getHomeData.ts
```diff
@@ -0,0 +1,295 @@
...
+          iconUrl: manifest.getUrl(`/icons/${project.slug}.png`),
+          category: 'scaling' as const,
+          scalingCategory: project.scalingInfo.type,
+        }
+      }
+      if (project.daLayer) {
+        return {
+          id: project.id.toString(),
+          name: project.name,
+          href: `/data-availability/projects/${project.slug}/no-bridge`,
```
vilcht, line 266: there could be a situation where theres no /no-bridge page as shouldHaveNoBridgePage can in theory return false

vilcht (commented):
Many unnecessary comments (lot more than those referenced individualy) + few nits, reference individual comments

packages/frontend/src/hooks/useLocalStorage.ts
torztomasz: It's fine imo, we will have the key typesafe. Previously it was possible to use different type values for the same key

vilcht (approved):

PR #11966, opened 2026-05-27
katana upgrade: deposited txs

sekuba:
give +1d because of redwarn


lucadonnoh (changes requested):
image
where are these msigs coming from? they dont have a description nor permissions shown here

lucadonnoh:
also we should remove the yellow exit window for regular upgrade since it's zero anyway (exit window not upgrade delay)

lucadonnoh (approved):

PR #11967, opened 2026-05-27
Base goes multi-proof

vincfurc:


sekuba (changes requested):
add the gametype name in the OptiPortal2 template

Image

sekuba (changes requested):
in the superchainConfig template, this permission is wrong (it gives the incidentResponder full guardian perms)

Image

sekuba (approved):
verify trackedTxs

PR #11977, opened 2026-05-28
Add links to per-token page on Token frameworks

maciekzygmunt:
Resolves L2B-14057

### DIFF
--- a/packages/frontend/src/pages/interop/token-frameworks/components/frameworks-table/FrameworkColumn.tsx
+++ b/packages/frontend/src/pages/interop/token-frameworks/components/frameworks-table/FrameworkColumn.tsx
@@ -61,11 +61,7 @@ export function FrameworkColumn({
                   key={token.id}
                   token={token}
                   framework={framework}
-                  href={
-                    token.isUnknown
-                      ? undefined
-                      : getInteropTokenUrl(token, apiSelection)
-                  }
+                  href={getInteropTokenUrl(token, apiSelection)}
                 />
               ))
             )}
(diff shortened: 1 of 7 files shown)


packages/frontend/src/pages/interop/token-frameworks/components/frameworks-table/Rows.tsx
```diff
@@ -6,28 +6,47 @@ import type {
...
 import { formatInteger } from '~/utils/number-format/formatInteger'
+import { getInteropTokenUrl } from '../../../utils/getInteropTokenUrl'
+import type { InteropSelection } from '../../../utils/types'
 import type { InteropTokenFramework } from '../../getInteropTokenFrameworksData'
 import { TokenFrameworksTransferTrigger } from '../TokenFrameworksTransferTrigger'
 
 export function TokenRow({
   token,
   framework,
+  apiSelection,
```
torztomasz: can't we pass href or simply extract apiSelection from context here?

torztomasz (changes requested):

packages/frontend/src/pages/interop/token-frameworks/components/frameworks-table/FrameworkColumn.tsx
```diff
@@ -53,7 +57,16 @@ export function FrameworkColumn({
...
               entry.tokens.map((token) => (
-                <TokenRow key={token.id} token={token} framework={framework} />
+                <TokenRow
+                  key={token.id}
+                  token={token}
+                  framework={framework}
+                  href={
+                    token.isUnknown
+                      ? undefined
+                      : getInteropTokenUrl(token, apiSelection)
```
torztomasz: maybe getInteropTokenUrl can return undefined if it's undefined so we do not have to make this check everywhere

torztomasz (changes requested):

torztomasz (approved):

PR #11978, opened 2026-05-28
fix(env): preserve separators in Env.key normalization

bravesasha:
Env.Env.key() was removing hyphens instead of converting them to underscores, causing distinct identifiers like service-a and servicea to collapse into the same environment key. This could silently resolve the wrong env var with no exception thrown. Fixed by canonicalizing separators via split(/[-_\s]+/) and added a regression test to prevent future collisions.


torztomasz:
given the fact you changed the test to make it pass - the previous behaviour was expected

PR #11982, opened 2026-05-28
Add tokens to search bar

torztomasz:


maciekzygmunt:
maybe at least lets capitalize issuer:

packages/frontend/src/server/features/projects/search-bar/getSearchBarProjects.ts
```diff
@@ -1,16 +1,33 @@
...
+      category: entry.category,
+      name: entry.name,
+      href: entry.href,
+      type: entry.type,
+      id: entry.id,
+      iconUrl: entry.iconUrl,
+      issuer: entry.issuer,
+      searchMatchKind: entry.searchMatchKind,
+      searchScore: entry.searchScore,
+    }
```
maciekzygmunt, line 27: why not just return entry?

maciekzygmunt (changes requested):

packages/frontend/src/server/features/search-bar/getSearchBarEntries.ts
vilcht, line 1: Should we keep it in features/projects if tokens are also searchable? Maybe searchbar belongs to more generic directory

vilcht (approved):

PR #11995, opened 2026-06-01
Use first data timestamp to generate better charts

torztomasz:
Resolves L2B-14139

## Problem

On the R0ar project page, selecting the **30d** range showed only the last ~3 days. Other projects were unaffected.

## Root cause

`aggregateActivityRecords` intentionally trims leading days where a project's count is `0` (per its test: *"starts aggregating from the first non-Ethereum record with count > 0"*). The chart then derived its x-axis start from the *first surviving data point* rather than the *requested window start*:

```ts
const startTimestamp = Math.min(...Object.keys(aggregatedEntries).map(Number))
```

R0ar is a low-activity chain — its config uses `adjustCount: SubtractOne`, so blocks with only the system tx count as `0`. It had `0` adjusted transactions for most of the last 30 days, so the trim discarded those days and the axis collapsed. Established projects have non-zero counts throughout the window, so they were unaffected.

## Fix

Scoped to single-project charts:

- **Anchor to the selected window start.** `startTimestamp = max(rangeStart, firstProjectTimestamp)`, where `firstProjectTimestamp` is the project's first-ever Activity record (reused from the already-fetched `totalCounts[projectId].sinceTimestamp` — a `MIN(timestamp)` over the whole table, no new query). This keeps the full 30d axis while never rendering days before the project existed. For `max` (null start) we begin at that first record.
- **Fill in-range zero-activity days with `0`** instead of a `null` gap, since every rendered day is now one where the project genuinely existed.

Aggregate (multi-project) charts are unchanged. `aggregateActivityRecords` and its tests are untouched.

## Notes

- Minor behavior change for `max` on single-project pages: it now starts at the project's first record and shows leading idle days as `0` rather than trimming them. Invisible for projects active from day one; only idle-at-launch chains like R0ar look different.
- The public API (`getActivityData`) has the same latent issue but serves both per-project and aggregate routes, so it's left for a follow-up.


torztomasz:
@codex review

sdlyy (approved):

PR #12010, opened 2026-06-02
Facet new rollup contract

vincfurc:


vincfurc:
should we drop the v1 in the name?

sekuba:
so is there no canonical bridge for the new rollup? shouldn't 'assets with additional trust assumptions' be 100%?

sekuba (approved):

PR #12012, opened 2026-06-02
Scroll removes security council (-> Stage 0)

vincfurc:


packages/config/src/projects/scroll/config.jsonc
```diff
@@ -85,11 +87,25 @@
       // Euro Coin Token (EURC) - tracked via L1USDCGateway reference, changes are Circle ops not Scroll
       "ignoreDiscovery": true
     },
-    "eth:0x1a37bF1Ccbf570C92FE2239FefaaAF861c2924DD": {
+    "eth:0xcca54B0916Cee2186b47E9709BEdcb7041A8F761": {
+      "description": "multisig of Scroll team operators that controls the rollup's upgrade and proof-system parameter paths.",
```
lucadonnoh:
```suggestion
      "description": "Multisig of Scroll team operators that controls the rollup's upgrade and proof-system parameter paths.",
```

packages/config/src/projects/scroll/config.jsonc
```diff
@@ -185,13 +201,51 @@
       "ignoreDiscovery": true
     },
     "scr:0x79D83D1518e2eAA64cdc0631df01b06e2762CC14": {
+      "description": "L2 counterpart of TimelockSCSlow.",
```
lucadonnoh: why is this here and not in a template?

lucadonnoh (changes requested):

lucadonnoh (changes requested):
check why the upgrade permission doesnt show up

vincfurc:
> check why the upgrade permission doesnt show up

which one? 
image

lucadonnoh:
but that's an interact permission, it should be an upgrade permission like here 
image

vincfurc:
> but that's an interact permission, it should be an upgrade permission like here image

yeah it is not new it's due to the scroll owner in between, it's not easily fixable tho, Adrian tried a while back

if that's the only thing the pr is mergeable

lucadonnoh (approved):

PR #12020, opened 2026-06-03
Interop detailed pages updates [prototype]

radinat:
Chains that also have an interop configused to appear twice, including duplicate search results. These are now a single page.
  
  What changed:
  - The separate bridge page now redirects to the chain page, and only one entry shows in search.
  - Bridge data lives in the "Volume and flows" section. Added a protocols selector (like the Summary page) so you can filter which bridges are included.
  - Quick "Canonical bridge" shortcut; its URL is shareable
  - Bridge details under the flows: Top tokens, Transfers, and a transfer-size breakdown are shown for the current protocol + chain selection
  - Summary on top keeps the chain stats plus key interop stats (24h volume, transfers, tokens by volume).
  
  Part of L2B-14064


radinat:
Implemented.

PR #12065, opened 2026-06-10
CCIP: FeeOracle v2 upgrade

lucadonnoh:
ignore in watch mode postponed, since there was an actual upgrade the dry run would be too noisy


packages/config/src/projects/_templates/transporter/FeeQuoterV2/template.jsonc
sekuba, line 1: nvm

sekuba (changes requested):

packages/config/src/projects/ccip/layouts/ccip.json
sekuba, line 1: wot is this file?

sekuba (changes requested):

packages/config/src/projects/ccip/layouts/ccip.json
sekuba: disco layout maybe?

lucadonnoh: yes disco layout! trying saving it for the first time. it is intended to be there

sekuba (approved):

PR #12068, opened 2026-06-10
add recent updates button

vilcht:


torztomasz:
image

Spacing between anomalies and recent updates should be smaller

image

Weird wrapping


Probably fixable by putting these two into one div with smaller gap. Then the wrapping should take both of them down and it will look good

torztomasz (approved):

PR #12073, opened 2026-06-11
token-ui: add search to history page

torztomasz:
## What

Adds a search box to the token DB history page so entries can be filtered by token data.

## Details

- **Frontend** (`TokenHistoryPage`): search input in the card header, debounced (300ms) via the existing `useDebouncedValue` hook. Typing resets to page 1, and the empty state reads "No entries match your search" when a search is active.
- **tRPC router** (`tokenDbHistory.getPage`): added an optional `search` input field, passed through to the repository.
- **Repository** (`TokenDbHistoryRepository.getPage`): when `search` is present, both the rows and count queries get a `"command"::text ILIKE '%...%'` filter, so pagination and total count stay consistent with the filtered set. LIKE wildcards (`\`, `%`, `_`) in user input are escaped. Filtering is server-side over the command JSON, so search spans the full history, not just the current page.

Search matches anything inside the command JSON — symbol, chain, address, abstract token id, etc.

## Testing

- Added a repository test covering match-by-symbol, match-by-chain, and no-match cases.
- `pnpm typecheck` passes for database, token-backend, and token-ui; biome clean.

### DIFF
--- a/packages/database/src/repositories/TokenDbHistoryRepository.test.ts
+++ b/packages/database/src/repositories/TokenDbHistoryRepository.test.ts
@@ -115,6 +115,50 @@ describeTokenDatabase(TokenDbHistoryRepository.name, (db) => {
       expect(page.totalCount).toEqual(3)
       expect(page.entries.map((e) => e.userEmail)).toEqual(['second@x.io'])
     })
+
+    it('filters entries by token data in the command via search', async () => {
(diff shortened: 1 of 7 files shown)


mateuszradomski:
@codex, review

mateuszradomski (approved):

PR #12074, opened 2026-06-11
Add RWA categories to abstract token form

torztomasz:
## What

Adds `rwaRestricted` and `rwaPublic` to the category selector in the abstract token form (`AbstractTokenForm.tsx`).

## Why

The `categoryValues` array was missing the two RWA categories that already exist in the canonical `TokenCategory` type (`ether | stablecoin | btc | rwaRestricted | rwaPublic | other`). This array drives both the validation schema (`v.enum(categoryValues)`) and the rendered dropdown options, so both now include the RWA categories.

Note: the category field lives on the abstract token form, not the deployed token form (which has no category field).

### DIFF
--- a/packages/token-ui/src/components/forms/AbstractTokenForm.tsx
+++ b/packages/token-ui/src/components/forms/AbstractTokenForm.tsx
@@ -27,7 +27,14 @@ import { parseDatePaste } from '~/utils/parseDate'
 import { Checkbox } from '../core/Checkbox'
 import { ExternalLink } from '../ExternalLink'
 
-const categoryValues = ['btc', 'ether', 'stablecoin', 'other'] as const
+const categoryValues = [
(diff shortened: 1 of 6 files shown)


mateuszradomski (approved):

packages/token-ui/src/components/forms/AbstractTokenForm.tsx
```diff
@@ -27,7 +28,14 @@ import { parseDatePaste } from '~/utils/parseDate'
 import { Checkbox } from '../core/Checkbox'
 import { ExternalLink } from '../ExternalLink'
 
-const categoryValues = ['btc', 'ether', 'stablecoin', 'other'] as const
+const categoryValues = [
```
mateuszradomski: Why is this duplicated?

PR #12075, opened 2026-06-11
Hyperliquid TVS

sekuba:
do not merge, config change


packages/shared-pure/src/types/ChainSpecificAddress.ts
```diff
@@ -53,7 +53,7 @@ const SHORT_TO_LONG_CHAIN_NAMES = {
   redstone: 'redstone',
   soneium: 'soneium',
   wc: 'worldchain',
-  'hype-evm': 'hyperevm',
```
torztomasz, line 56: why?

torztomasz (changes requested):

packages/shared-pure/src/types/ChainSpecificAddress.ts
sekuba: Just to have the same name globally, and there is no thing called hype-evm

torztomasz (approved):

PR #12111, opened 2026-06-17
IA - aggregate status table

sdlyy:
First slice of the Interop aggregate **promotion gate**. Adds the storage + access layer that later PRs build on. **No behaviour change** — nothing reads this table yet.

## What changed

- **New `InteropAggregateStatus` table** — one row per aggregate snapshot `timestamp`:
  - `status` — `promoted` | `blocked`
  - `promotedBy` — `auto` (engine) or operator email (manual flip)
  - `reasons` — JSON text, populated only when `blocked`
  - `checkedAt` / `updatedAt`
  - Migration is **`CREATE TABLE` only — no backfill** (backfill is a separate, later migration).
- **`InteropAggregateStatusRepository`**
  - `upsertAuto` — engine write, **sticky**: `ON CONFLICT … WHERE promotedBy = 'auto'` (table-qualified to disambiguate from `excluded`), so it never overwrites a manual verdict.
  - `upsert` — manual (operator) write, unconditional.
  - reads used by the eventual cutover: `getLatestPromotedTimestamp`, `getEarliestPromotedTimestampForDay`, `getByTimestamp`, `getRecent`.
  - `deleteOrphaned` — drops status rows whose snapshot no longer exists (keeps the table in lockstep with aggregate retention).
- Registered in the `Database` factory and the record types exported from the package index.

## Notes for the reviewer

- `reasons` is `JSON.stringify`'d on write and parsed on read. The `pg` driver serializes JS **arrays** as Postgres array literals (which `jsonb` rejects), so the column is `Text` and we control the JSON ourselves.
- The sticky `upsertAuto` is the mechanism that lets a later promotion engine re-evaluate the current hour without ever clobbering a human's manual flip.

### DIFF
--- a/packages/database/prisma/migrations/20260617145056_add_interop_aggregate_status/migration.sql
+++ b/packages/database/prisma/migrations/20260617145056_add_interop_aggregate_status/migration.sql
@@ -1,0 +1,11 @@
+-- CreateTable
+CREATE TABLE "InteropAggregateStatus" (
+    "timestamp" TIMESTAMP(6) NOT NULL,
+    "status" VARCHAR(16) NOT NULL,
+    "promotedBy" VARCHAR(255),
+    "reasons" TEXT,
+    "checkedAt" TIMESTAMP(6) NOT NULL,
+    "updatedAt" TIMESTAMP(6) NOT NULL,
+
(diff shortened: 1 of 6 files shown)


sdlyy:
@codex review

adamiak (approved):

PR #12112, opened 2026-06-17
IA - promotion foundation

sdlyy:
Makes the aggregating indexer record promotion status for every new snapshot it writes. **Still zero behaviour change for the frontend** — the read path is untouched, so this only populates status going forward (the display cutover is a later PR).

## What changed

- In `InteropAggregatingIndexer.update`, inside the **existing transaction** and after the aggregate inserts:
  - `interopAggregateStatus.upsertAuto({ timestamp: to, status: 'promoted' })`
  - `interopAggregateStatus.deleteOrphaned()`
- Atomic with the aggregate writes, so a snapshot can never exist without a status row. Runs only when aggregation runs — a skipped (stale-syncers) hour writes nothing.

## Notes for the reviewer

- The hard-coded `'promoted'` is intentional for this stage. The promotion engine (later PR) replaces it with a real verdict via the same `upsertAuto` call — no signature changes needed.
- `deleteOrphaned` is placed after the aggregate retention/overwrite so status stays in lockstep with what aggregate rows actually exist.

### DIFF
--- a/packages/backend/src/modules/interop/engine/aggregation/InteropAggregatingIndexer.test.ts
+++ b/packages/backend/src/modules/interop/engine/aggregation/InteropAggregatingIndexer.test.ts
@@ -171,6 +171,13 @@ describe(InteropAggregatingIndexer.name, () => {
         insertMany: mockFn().resolvesTo(1),
       })
 
+      const interopAggregateStatus = mockObject<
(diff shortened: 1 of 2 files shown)


sdlyy:
@codex review

adamiak (approved):

PR #12113, opened 2026-06-17
IA - backfill past aggregate statuses

sdlyy:
Data migration that marks every pre-existing aggregate snapshot as `promoted`, so that once later logic depends on the status table, **no aggregate is status-blind**. **No behaviour change** — the read path still doesn't consult status until the cutover PR.

## What changed

- New data migration `20260617150000_backfill_interop_aggregate_status_promoted`:

  ```sql
  INSERT INTO "InteropAggregateStatus" ("timestamp", "status", "promotedBy", "checkedAt", "updatedAt")
  SELECT DISTINCT "timestamp", 'promoted', 'backfill', now(), now()
  FROM "AggregatedInteropTransfer"
  ON CONFLICT ("timestamp") DO NOTHING;
  ```

## Notes for the reviewer

- **Order matters:** this ships *after* PR B (the indexer already marks new snapshots). It only fills gaps for snapshots created before the writer was deployed.
- `ON CONFLICT DO NOTHING` means it never overrides a verdict the indexer or an operator already wrote — it only inserts missing rows.
- `SELECT DISTINCT timestamp` collapses the large aggregate table to one row per snapshot (~hundreds of rows), so the insert is small despite the source table size; negligible lock.
- `promotedBy = 'backfill'` distinguishes these rows in audit. Backfilled rows are old snapshots that the indexer never re-evaluates, so the sticky `upsertAuto` rule is irrelevant to them.

### DIFF
--- a/packages/database/prisma/migrations/20260617150000_backfill_interop_aggregate_status_promoted/migration.sql
+++ b/packages/database/prisma/migrations/20260617150000_backfill_interop_aggregate_status_promoted/migration.sql
@@ -5,9 +5,14 @@
 -- gaps for snapshots created before the writer was deployed. `ON CONFLICT DO NOTHING`
 -- guarantees it never overrides a verdict the indexer (or an operator) already wrote.
 --
+-- `promotedBy = 'auto'` (not a dedicated 'backfill' marker): a backfilled row is a system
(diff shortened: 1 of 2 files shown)


adamiak (approved):

sdlyy:
@codex review

PR #12124, opened 2026-06-18
Reject unsafe numbers in Bytes.fromNumber

haoshengzhen:
## Summary

This updates `Bytes.fromNumber` to reject numbers that are outside JavaScript's safe integer range.

Previously, `Bytes.fromNumber` accepted any non-negative integer, including values greater than `Number.MAX_SAFE_INTEGER`. Such values may already be imprecise when represented as a JavaScript `number`, so encoding them into bytes can silently produce an unexpected value.

## Changes

- Use `Number.isSafeInteger` when validating numeric byte inputs
- Add regression coverage for rejecting unsafe integers in `Bytes.fromNumber`

## Testing

- `pnpm -C packages/shared-pure test`
- `pnpm -C packages/shared-pure lint`
- `pnpm -C packages/shared-pure typecheck`
- `pnpm -C packages/shared-pure format`
- `pnpm -C packages/shared-pure build`

### DIFF
--- a/packages/shared-pure/src/types/Bytes.test.ts
+++ b/packages/shared-pure/src/types/Bytes.test.ts
@@ -37,6 +37,9 @@ describe(Bytes.name, () => {
     it('checks constructor arguments', () => {
       expect(() => Bytes.fromNumber(1.5)).toThrow(TypeError)
       expect(() => Bytes.fromNumber(-2)).toThrow(TypeError)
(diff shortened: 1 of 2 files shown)


haoshengzhen:
@sekuba @sergeyshemyakov @mateuszradomski Hi, Could you please review this PR at your convenience? Thank you very much.

mateuszradomski (approved):

mateuszradomski:
@codex, review

PR #12130, opened 2026-06-22
add l2b analyze

vilcht:
also fixes the issue of sending not enough context. Now always sends the whole .flat


vilcht:
@codex review

packages/l2b/src/implementations/analyze/AnalyzeRunner.ts
```diff
@@ -0,0 +1,193 @@
...
+  }
+}
+
+interface LocalCommandResult {
+  stdout: string
+  status: number
+}
+
+function runLocalL2Analyze(args: string[]): LocalCommandResult {
+  const result = spawnSync('l2analyze', args, { encoding: 'utf-8' })
```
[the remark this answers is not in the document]
vilcht: resolved

packages/l2b/src/commands/Analyze.ts
```diff
@@ -0,0 +1,183 @@
...
+export const Analyze = command({
+  name: 'analyze',
+  description: 'Run a source-code analyzer against discovery flat sources.',
+  args: {
+    json: flag({
+      type: boolean,
+      long: 'json',
+      description: 'Print the raw analyzer API response as JSON.',
+    }),
+    cli: flag({
```
mateuszradomski: Wouldn't it be better to have something like --backend cli/api?
If we do it like this, we don't have to verify this at all and we can have an enumeration in the allowed values. Removes the need to have `getRunnerPreference`

packages/l2b/src/commands/Analyze.ts
```diff
@@ -0,0 +1,183 @@
...
+      long: 'api',
+      description: 'Force the remote analyzer API backend.',
+    }),
+    args: restPositionals({
+      displayName: 'args',
+      description: 'Use "list" or "<analyzer-id> <path-inside-.flat>".',
+    }),
+  },
+  handler: async (args) => {
+    dotenv()
```
mateuszradomski: is this needed? i feel like we never had to use this elsewhere

packages/l2b/src/commands/Analyze.ts
```diff
@@ -0,0 +1,183 @@
...
+      type: boolean,
+      long: 'cli',
+      description: 'Force the local l2analyze CLI backend.',
+    }),
+    api: flag({
+      type: boolean,
+      long: 'api',
+      description: 'Force the remote analyzer API backend.',
+    }),
+    args: restPositionals({
```
mateuszradomski: Again, I don't like this. We should be doing more parsing using the library instead of ourselves. This way of choosing what to do is problematic. First of all, we need to remember to check if there are two args if the first arg is not list and more and more edge cases.

The simplest solution to this problem is the following:
- drop args as restPositionals and have two positionals that are optional
- the first positional is analyzer-id
- the second positional is path (we can actually configure the type of this positional such that we will verify the path I think)

If no args are provided: list all the possible analyzers, otherwise run the analyzer that is chosen

packages/l2b/src/commands/Analyze.ts
```diff
@@ -0,0 +1,183 @@
...
+  }
+
+  throw error
+}
+
+function printUsageError() {
+  console.error('Usage:')
+  console.error('  l2b analyze [--cli|--api] list')
+  console.error('  l2b analyze [--cli|--api] <analyzer-id> <path-inside-.flat>')
+}
```
mateuszradomski: This is a mistake, we are using cmd-ts that gives us this for free. We don't need to write this

packages/l2b/src/implementations/analyze/AnalyzeRunner.ts
```diff
@@ -0,0 +1,198 @@
...
+): T {
+  try {
+    return schema.parse(JSON.parse(stdout))
+  } catch {
+    if (stdout) {
+      process.stdout.write(stdout)
+    }
+    throw new AnalyzeRunnerError(`l2analyze ${command} returned invalid JSON`)
+  }
+}
```
mateuszradomski, line 198: This shouldn't be here. Starting the local analyzer server is the responsibility of the user. The code shouldn't be starting it here.

packages/l2b/src/implementations/analyze/AnalyzeRunner.ts
mateuszradomski, line 1: I don't like the entire premise of this approach. First of all, we don't need to have this special syntax sugar of

runWithX((x) => x.usage)

we can just do

const x = getX()
x.usage

Way simpler to later debug in the debugger and stack traces later are way easier to read.

Now, the second thing. We don't need to distinguish between local and remote runners. I don't like the idea of this command having the knowledge of how to start the local server because what if I didn't install all the needed things for the local server. It's just going to fail. Also, if we change the server this command will not need to be changed, which is the way it should be.

To still allow for local and remote running we should just provide one parameter to this entire function/command: url. By default - what 99% of people are going to use - we can just configure it to reach out to the API. We can hardcode the API url in the code, or specify it in the ENV if we don't want to make it public. And local testing can be done by running the analyzer server manually on dev machine and passing

--server-url=localhost:port

packages/l2b/src/implementations/analyze/AnalyzeClient.ts
```diff
@@ -55,11 +68,25 @@ export class AnalyzeClient {
...
+      if (!(error instanceof FetchError)) {
+        throw error
+      }
+
+      throw new AnalyzeClientError(
+        502,
+        `Analyze service request failed: ${error.message}`,
+      )
+    }
+
```
mateuszradomski: Is this really needed? I don't see if we are doing anything with these errors beyond this point. In the end we will exit and print the error

packages/l2b/src/implementations/discovery/getCodePaths.ts
```diff
@@ -0,0 +1,90 @@
...
+function fileNameToOrder(name: string) {
+  const ending = name.match(/\.(\w+)\.sol/)?.[1]
+  if (!ending) {
+    return 1
+  }
+  if (ending === 'p') {
+    return 0
+  }
+  return /^\d+$/.test(ending) ? Number.parseInt(ending) : 2
+}
```
mateuszradomski, line 113: Please don't use regexes for very simple things. This can be just .split() and some ifs. Way easier to read. Also document what is the format of file names we are parsing here

mateuszradomski (changes requested):

mateuszradomski (approved):

PR #12140, opened 2026-06-23
Add CCIPHome discovery

lucadonnoh:
## Summary

Adds **CCIPHome** (and its **CapabilitiesRegistry**) to the `ccip` discovery, so the home-chain DON configuration that drives every lane's OCR digest is finally legible instead of showing up only as opaque digests and transmitter reorders on the remote OffRamps.

Doing that well needed a few **general-purpose additions to the `discovery` package**, which are the main point of this PR and are reusable by any project. Each is independent and unit-tested.

### 1. `sort` blip edit op (`blip/`)

A new edit operation: `["sort"]` sorts array elements directly, `["sort", keyExpr]` sorts by a computed key. Ascending, stable, deterministic.

Use it to normalize **set-like arrays** so that a pure reordering produces no diff. Onchain, OCR signer/transmitter and DON node lists are sets whose order carries no semantic meaning but reshuffles on every config push, which previously flooded diffs. Added to `BlipRuntime`, `validateBlip`, and the `BlipSexp` type.

### 2. Nested field access in `EventHandler` (`handlers/user/EventHandler.ts`)

`extractKey` now understands dot-notation, so a handler's `groupBy` / `select` can reach into nested fields, e.g. `groupBy: "config.chainSelector"`. This makes it possible to group/select an event by a field that lives inside a struct parameter (previously only top-level event params worked).

Caveat documented in code: discovery decodes structs as **positional arrays** (`toContractValue` drops the ABI field names), so a nested struct field is addressed by its index (`config.1`), not its name. Plain (non-dotted) keys behave exactly as before.

### 3. `BytesToString` type-caster (`type-casters/`)

Decodes a variable-length `bytes` hex value into its UTF-8 string. Use for onchain `bytes` fields that hold human-readable text (e.g. JSON config blobs) so they are stored decoded rather than as opaque hex. Returns the input unchanged if it is not valid UTF-8.

### 4. `Ocr3OffchainConfig` type-caster (`type-casters/`)

Fully decodes the libOCR OCR3 `OffchainConfig` protobuf into a named object: the scheduling/timing params (`deltaProgress`, `deltaResend`, …), the transmission schedule, the offchain public keys, the p2p peer ids, and the embedded `reportingPluginConfig` (parsed as JSON for CCIP). It is a small self-contained protobuf reader (no new dependency); the field map is taken from libocr's generated `offchainreporting3_offchain_config.pb.go`. It also strips `null`s recursively, since `ContractValue` has no `null` and the plugin JSON contains them (e.g. `tokenDataObservers: null` on non-CCTP lanes).

## `config` changes (ccip)

- **CCIPHome**: tracks owner/registry/typeAndVersion, the per-chain `chainConfigurations` (event-sourced from `ChainConfigSet`, config JSON decoded via `BytesToString`), and the per-DON `executionConfigs`/`commitConfigs` (the configDigest preimage: version, F, the sorted DON nodes, and the fully decoded `offchainConfig`). `ignoreRelative` keeps the operator addresses from being crawled as relatives.
- **CapabilitiesRegistry**: tracked with the large, frequently-rotating `getNodes`/`getDONs` `ignoreMethods`-d, keeping node operators, capabilities, and the DON counter.
- **`OfframpV3` template**: OCR signers and transmitters are now `sort`-ed, so a pure DON reorder no longer churns the OffRamp diffs (the `## Watched changes` section now shows only real `configDigest` rotations).
- Regenerated `discovered.json` / `diffHistory.md` account for most of the additions.

## Testing

New unit tests for the `sort` op, nested `groupBy`/`select`, `BytesToString`, and `Ocr3OffchainConfig`; the full `discovery` suite passes.


packages/discovery/src/blip/validateBlip.ts
mateuszradomski, line 1: Write tests for this in `validateBlip.test.ts`

packages/discovery/src/blip/BlipRuntime.ts
```diff
@@ -66,6 +66,26 @@ export class BlipRuntime {
         assert(Array.isArray(v), 'map requires an array input')
         return v.map((item) => this.executeBlip(item, fn))
       }
+      case 'sort': {
+        // ['sort'] sorts elements directly; ['sort', keyExpr] sorts by a
```
mateuszradomski, line 70: Write docs for this function in discovery README.md

packages/discovery/src/discovery/handlers/user/EventHandler.ts
```diff
@@ -330,12 +330,25 @@ function extractKeys(
   return result
 }
 
+// Supports dot-notation for reaching into nested values, e.g.
```
mateuszradomski, line 333: Document this feature in discovery README.md in the Event Handler section

mateuszradomski (approved):

sekuba (approved):

sekuba:
rebirth of the transporter disco, the endgame of disco

PR #12146, opened 2026-06-23
Adjust deduplication logic for $pastUpgrades

sdlyy:
In pastUpgrades.ts, dedupeByTransaction collapsed every $pastUpgrades entry that shared a transaction hash, keeping only the last one. That rule was written for diamonds (one DiamondCut event per cut, all resolving to the same cumulative facet set within a tx).

But OP Stack single-implementation proxies like the hashkey SuperchainConfig (0xfd12…B40A) upgrade in the pattern proxy → StorageSetter → new impl within one transaction, emitting two distinct Upgraded events.

Changed the dedupe key from txhash alone to txhash + implementation set:
Diamonds — duplicate-txhash entries are identical impl sets → still collapse to one row (verified: abstract/wonder/adi diamonds all have identical dups).
Single-impl proxies — entries with the same tx but different impls are kept → the StorageSetter step is no longer hidden, and the displayed row count matches $upgradeCount.

### DIFF
--- a/packages/protocolbeat/src/apps/discovery/panel-values/pastUpgrades.test.ts
+++ b/packages/protocolbeat/src/apps/discovery/panel-values/pastUpgrades.test.ts
@@ -140,10 +140,11 @@ describe('buildPastUpgradeRows', () => {
     expect(rows[0]?.cells[2]?.diffUrl).toEqual(`/diff/${E}/${D}`)
   })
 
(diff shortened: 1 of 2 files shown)


sdlyy:
@codex review

sdlyy:
@codex review

mateuszradomski (approved):

PR #12160, opened 2026-06-29
Align ZK catalog milestones with the TVS bump they explain

torztomasz:
Resolves L2B-14223

## Problem

On the ZK catalog TVS chart, a project's "joined" milestone marker sat one datapoint *before* the TVS bump it's meant to explain. The bump first appears at the datapoint satisfying the `>= sinceTimestamp` gate, while the milestone was rounded down to the **start of the day**. So on the daily chart the dot landed on the 26th while the bump showed on the 27th — and the mismatch grew worse at sub-day zoom (6h/hourly), where a start-of-day milestone could sit hours away from its bump.

## Fix

Round **both** sides down to the active resolution so the marker and the bump land on the same datapoint at every zoom level.

- **Query** (`getDetailedTvsChartWithProjectsRanges.ts`, ZK-catalog-only): floor each project's `sinceTimestamp` to the active resolution before the DB gate and `firstProjectTimestamp`, moving the bump onto the start-of-period datapoint. `untilTimestamp` is left untouched — no milestone aligns with a project *leaving* a system.
- **Milestones** (`ChartMilestones.tsx`, global): snap each milestone to the largest datapoint at or before its timestamp. The datapoint grid is resolution-aligned, so this floors to the active resolution with no new props. Tooltip/drawer dates are unaffected — they still read the true `milestone.date`.

## Tradeoff

This counts a project's TVS as ZK-secured up to one resolution-period before it actually joined — bounded by the granularity the user can perceive (≤1h on the hourly chart, ≤1 day only on the long daily chart). We chose this over moving just the marker, which would have shown the marker on the 27th while the tooltip said the 26th.

## Scope of the global change

Making milestone snapping resolution-aware affects every chart, but only at sub-day zoom and only as a correctness improvement: markers move from midnight to the event's true bucket, and same-day events stop merging when they're genuinely at different x-positions. Daily (long-range) charts are unchanged.

## Tests

New `ChartMilestones.test.ts` covers `getTimestampedMilestones`:
- daily-grid snap to start of day
- hourly-grid snap to the event's hour (not midnight)
- milestones before the first datapoint are dropped
- same-day grouping preserved on the daily grid
- same-day events separated on the hourly grid

## Manual check worth doing before merge

Verify on a live ZK catalog project with a recent join (so the 7-day hourly view shows the bump) that the dot and bump visually coincide — the unit tests cover the snapping logic but not the end-to-end render.

### DIFF
--- a/packages/frontend/src/components/chart/tvs/stacked/zk-catalog/ZkCatalogProjectsTvsChart.tsx
+++ b/packages/frontend/src/components/chart/tvs/stacked/zk-catalog/ZkCatalogProjectsTvsChart.tsx
@@ -1,5 +1,5 @@
 import type { Milestone } from '@l2beat/config'
-import type { ProjectId, UnixTime } from '@l2beat/shared-pure'
+import { type ProjectId, UnixTime } from '@l2beat/shared-pure'
 import { useQuery } from '@tanstack/react-query'
 import { useMemo } from 'react'
 import { Area, AreaChart } from 'recharts'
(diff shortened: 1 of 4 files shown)


mateuszradomski:
@codex, review

mateuszradomski (approved):

PR #12161, opened 2026-06-29
Grouping in DiscoUI

mateuszradomski:
Resolves L2B-14254

### DIFF
--- a/packages/l2b/src/implementations/discovery-ui/types.ts
+++ b/packages/l2b/src/implementations/discovery-ui/types.ts
@@ -56,6 +56,7 @@ export type ApiAddressType =
   | 'Timelock'
   | 'Untemplatized'
   | 'Contract'
+  | 'Group'
   | 'Unknown'
 
 export interface ApiAddressEntry {
--- a/packages/protocolbeat/src/apps/discovery/multi-view/BottomBar/BottomBar.tsx
+++ b/packages/protocolbeat/src/apps/discovery/multi-view/BottomBar/BottomBar.tsx
@@ -2,6 +2,7 @@ import { useEffect, useState } from 'react'
 import { useParams } from 'react-router-dom'
 import { IS_READONLY } from '../../../../config/readonly'
 import { useIsomorphicKeys } from '../../hooks/useIsomorphicKeys'
+import { useStore as useNodesStore } from '../../panel-nodes/store/store'
 import { useDiscoveryCommand } from '../../panel-terminal/useDiscoveryCommand'
 import { useSearchStore } from '../../search/store'
 import { useGlobalSettingsStore } from '../../store/global-settings-store'
@@ -56,6 +57,13 @@ export function BottomBar() {
       }
       if (e.code === 'KeyK' && e.altKey) {
         killCommand()
+      }
+      if (e.code === 'KeyG' && e.altKey) {
+        if (e.shiftKey) {
+          useNodesStore.getState().ungroupSelected()
+        } else {
+          useNodesStore.getState().groupSelected()
+        }
       }
     }
 
@@ -113,6 +121,13 @@ export function BottomBar() {
               </li>
               <li>
                 <Keys keys={[altKey, 'Enter']} /> - Add panel
+              </li>
+              <hr className="my-1" />
+              <li>
+                <Keys keys={[altKey, 'G']} /> - Group nodes
+              </li>
+              <li>
+                <Keys keys={[altKey, 'Shift', 'G']} /> - Ungroup nodes
               </li>
               <hr className="my-1" />
               <li>
--- a/packages/protocolbeat/src/apps/discovery/multi-view/BottomBar/BottomBar.tsx
+++ b/packages/protocolbeat/src/apps/discovery/multi-view/BottomBar/BottomBar.tsx
@@ -2,7 +2,7 @@ import { useEffect, useState } from 'react'
 import { useParams } from 'react-router-dom'
 import { IS_READONLY } from '../../../../config/readonly'
 import { useIsomorphicKeys } from '../../hooks/useIsomorphicKeys'
-import { useStore as useNodesStore } from '../../panel-nodes/store/store'
+import { useStore } from '../../panel-nodes/store/store'
 import { useDiscoveryCommand } from '../../panel-terminal/useDiscoveryCommand'
 import { useSearchStore } from '../../search/store'
 import { useGlobalSettingsStore } from '../../store/global-settings-store'
@@ -60,9 +60,9 @@ export function BottomBar() {
       }
       if (e.code === 'KeyG' && e.altKey) {
         if (e.shiftKey) {
-          useNodesStore.getState().ungroupSelected()
+          useStore.getState().ungroupSelected()
         } else {
-          useNodesStore.getState().groupSelected()
+          useStore.getState().groupSelected()
         }
       }
     }
--- a/packages/protocolbeat/src/apps/discovery/multi-view/BottomBar/BottomBar.tsx
+++ b/packages/protocolbeat/src/apps/discovery/multi-view/BottomBar/BottomBar.tsx
@@ -20,6 +20,8 @@ export function BottomBar() {
   const { altKey } = useIsomorphicKeys()
   const setOpen = useSearchStore((state) => state.setOpen)
   const maxReachableDepth = useGlobalSettingsStore((s) => s.maxReachableDepth)
+  const groupSelected = useStore((s) => s.groupSelected)
+  const ungroupSelected = useStore((s) => s.ungroupSelected)
 
   const depthSpecified = maxReachableDepth !== null
 
@@ -60,9 +62,9 @@ export function BottomBar() {
       }
       if (e.code === 'KeyG' && e.altKey) {
         if (e.shiftKey) {
-          useStore.getState().ungroupSelected()
+          ungroupSelected()
         } else {
-          useStore.getState().groupSelected()
+          groupSelected()
         }
       }
     }
--- a/packages/protocolbeat/src/apps/discovery/panel-nodes/controls/GroupControls.tsx
+++ b/packages/protocolbeat/src/apps/discovery/panel-nodes/controls/GroupControls.tsx
@@ -1,0 +1,48 @@
+import { IconGroup } from '../../../../icons/IconGroup'
+import { IconUngroup } from '../../../../icons/IconUngroup'
+import { useStore } from '../store/store'
+import { ControlButton } from './ControlButton'
+import { ControlGroup } from './ControlGroup'
+
+const MIN_NODES = 2
+
+export function GroupControls() {
+  const selectedCount = useStore((state) => state.selected.length)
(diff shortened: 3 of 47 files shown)


packages/protocolbeat/src/apps/discovery/panel-nodes/store/actions/group.ts
```diff
@@ -0,0 +1,108 @@
...
+  internal: Set<string>,
+): Field[] {
+  const outgoing: Field[] = []
+  const seen = new Set<string>()
+  for (const node of subnodes) {
+    for (const field of node.fields) {
+      if (internal.has(field.target) || seen.has(node.name)) {
+        continue
+      }
+      seen.add(node.name)
```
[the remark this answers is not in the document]
mateuszradomski: I want it working like this

mateuszradomski:
@codex, review

mateuszradomski:
@codex, review

packages/protocolbeat/src/apps/discovery/panel-nodes/store/actions/group.ts
```diff
@@ -0,0 +1,108 @@
...
+
+export function collectOutgoingFields(
+  subnodes: Node[],
+  internal: Set<string>,
+): Field[] {
+  const outgoing: Field[] = []
+  const seen = new Set<string>()
+  for (const node of subnodes) {
+    for (const field of node.fields) {
+      if (internal.has(field.target) || seen.has(node.name)) {
```
torztomasz: Bug: `seen` is keyed by the **member name** and added on the member's first external field, so each member contributes only its *first* outgoing edge — any further external targets are dropped. Two members sharing a name (e.g. two `owner` EOAs) makes the second member's edges vanish entirely. A collapsed group therefore under-draws its connections depending on field order. If "one representative edge per member" is intentional it needs a comment and should dedup by `field.target`, not `node.name` (and `node.name` shouldn't double as both dedup key and the rewritten field name).

packages/protocolbeat/src/apps/discovery/panel-nodes/store/store.ts
```diff
@@ -95,6 +104,7 @@ export const useStore = create<StoreState>()(
       onMouseDown: wrapHistoryStartAction(set, onMouseDown),
       onMouseUp: wrapHistoryEndAction(set, onMouseUp),
       onMouseMove: wrapAction(set, onMouseMove),
+      onDoubleClick: wrapHistoryStartAction(set, onDoubleClick),
```
torztomasz: `wrapHistoryStartAction` only sets `history.pending` when an action transitions `mouseMoveAction` into an undoable drag. `onDoubleClick` only flips `opened` and never touches `mouseMoveAction`, so this wrapper is a no-op here — toggling open/close is not undoable, yet it's dressed in drag-start semantics. Use `wrapAction` if open/close should stay out of undo (reasonable); otherwise the undo support is missing. Right now it only "works" by accident.

packages/protocolbeat/src/apps/discovery/panel-nodes/store/actions/onMouseMove.ts
```diff
@@ -115,15 +116,14 @@ export function onMouseMove(
...
+        // reason. Only the selection rectangle and selected ids change. The
+        // render graph is used so members of opened groups are selectable too.
         const previousSelected =
           state.mouseMoveAction === 'select-add'
             ? new Set(state.selected)
             : undefined
         const selected: string[] = []
-        for (const node of state.nodes) {
-          if (hiddenSet.has(node.id)) continue
+        for (const node of buildRenderGraph(state.nodes, state.hidden).nodes) {
```
torztomasz: Perf: the comment just above explains you deliberately skip `updateNodePositions` to avoid rebuilding nodes+connections per move — but `buildRenderGraph` does exactly that work (`expand` + `indexDescendants` + `layoutFields` recomputing every field box and `processConnection`) whenever a group is open, on every mousemove. Select-box hit-testing only needs boxes. No-op when nothing is open (early return), but it bites the multi-select-into-a-group path the feature exists for. Expose a boxes-only path or hoist the graph out of the per-move loop.

packages/l2b/src/implementations/discovery-ui/types.ts
```diff
@@ -56,6 +56,7 @@ export type ApiAddressType =
   | 'Timelock'
   | 'Untemplatized'
   | 'Contract'
+  | 'Group'
```
torztomasz, line 59: Boundary leak: `'Group'` is a client-only construct (phantom nodes; the backend never emits them). The protocolbeat copy needs it for the icon atlas/`AddressIcon` typing, but this server-side `ApiAddressType` should not advertise an address type it can never produce. Keep `'Group'` on a client-side node-type union (or a client extension), not in the shared server contract.

packages/protocolbeat/src/apps/discovery/panel-nodes/store/actions/loadNodes.ts
```diff
@@ -145,6 +152,100 @@ function combinedHiddenFields(
...
+    }
+    const node = byId.get(group.id)
+    if (node) {
+      result.push(node)
+    }
+  }
+  return result
+}
+
+function makeGroupNode(group: StoredGroup, members: Node[]): Node {
```
torztomasz: `makeGroupNode` and `createPhantomNode` (group.ts) build the same 16-field Group node, differing only in where `color`/`opened`/`box` come from. Extract one `makeGroupNode({id, name, color, opened, box, subnodes})` factory and call it from both — otherwise a new field on the Group shape must be edited in two places.

packages/protocolbeat/src/apps/discovery/panel-nodes/store/State.ts
```diff
@@ -58,6 +58,8 @@ export interface Node {
   readonly hueShift: number
   readonly data: unknown
   readonly isReachable: boolean
+  readonly opened: boolean
+  readonly subnodes: Node[]
```
torztomasz: Minor: contents are mutable (`Node[]`) while sibling arrays like `fields` are `readonly`. Since nodes are treated as immutable and shared into history snapshots by reference, make this `readonly Node[]` to prevent accidental in-place mutation.

packages/protocolbeat/src/apps/discovery/panel-nodes/store/utils/subnodes.ts
```diff
@@ -0,0 +1,21 @@
+import type { Node } from '../State'
+
+export function topLevelByDescendant(
```
torztomasz, line 3: This is one of ~14 bespoke tree-walks the nested model forces (see the PR-level comment). `topLevelByDescendant`, `collectIds`, `collectAllIds`, `selectedLeafIds`, `collectDragPositions` and the storage walk are all the same traversal with different accumulators — a shared `walkTree(nodes, visit)` would collapse them.

torztomasz (commented):
Thermo-nuclear code-quality pass. Two correctness bugs (collectOutgoingFields, onDoubleClick wrapper), one hot-path perf regression, a boundary leak, and one structural theme worth deciding on before this calcifies. Details inline.

**Structural theme (no single line):** a group is a `Node` with nested `subnodes`, but nearly every consumer needs a *flattened* view, so the diff hand-rolls ~14 different recursions over the same tree (`selectedLeafIds`, `collectIds`, `collectAllIds`, `topLevelByDescendant`, `collectDragPositions`, `shiftSubnodes`, `indexSubnodes`, `toggleOpened`, `translateNode`, `expand`, `indexDescendants`, the walks in `buildStoredNodeLayout`/`reconstructGroups`, `snapshotNodesEqual`). The persisted model is already flat (`StoredGroup.members: string[]`) and the rendered model is flat — only the in-memory model is nested, and it's the sole source of all this recursion. Minimum: extract `walkTree`/`mapTree`/`flattenLeaves` primitives and express the walkers in terms of them. Worth considering: a flat `nodes` + `groups` map so drag/selection/storage reuse the existing flat machinery.

packages/protocolbeat/src/apps/discovery/multi-view/BottomBar/BottomBar.tsx
```diff
@@ -2,6 +2,7 @@ import { useEffect, useState } from 'react'
 import { useParams } from 'react-router-dom'
 import { IS_READONLY } from '../../../../config/readonly'
 import { useIsomorphicKeys } from '../../hooks/useIsomorphicKeys'
+import { useStore as useNodesStore } from '../../panel-nodes/store/store'
```
torztomasz: do we need this 'as' casting?

packages/protocolbeat/src/apps/discovery/multi-view/BottomBar/BottomBar.tsx
```diff
@@ -57,6 +58,13 @@ export function BottomBar() {
       if (e.code === 'KeyK' && e.altKey) {
         killCommand()
       }
+      if (e.code === 'KeyG' && e.altKey) {
+        if (e.shiftKey) {
+          useNodesStore.getState().ungroupSelected()
+        } else {
+          useNodesStore.getState().groupSelected()
+        }
+      }
```
torztomasz, line 69:
```suggestion
      if (e.code === 'KeyG' && e.altKey) {
        const state =  useNodesStore.getState()
        if (e.shiftKey) {
          state.ungroupSelected()
        } else {
          state.groupSelected()
        }
      }
```

packages/protocolbeat/src/apps/discovery/panel-nodes/store/actions/group.ts
```diff
@@ -0,0 +1,108 @@
...
+    subnodes,
+  }
+}
+
+export function collectIds(node: Node, into: Set<string>): void {
+  into.add(node.id)
+  for (const subnode of node.subnodes) {
+    collectIds(subnode, into)
+  }
+}
```
torztomasz: why do we need to use set everywhere? does it mean that some subnodes might have the same id or what?

packages/protocolbeat/src/apps/discovery/panel-nodes/controls/GroupControls.tsx
```diff
@@ -0,0 +1,89 @@
...
+      onBlur={commit}
+      onKeyDown={(event) => {
+        if (event.key === 'Enter') {
+          commit()
+          event.currentTarget.blur()
+        }
+      }}
+      className="w-32 self-stretch rounded-lg border border-coffee-600 bg-coffee-800 px-3 text-coffee-100 text-xs outline-none focus:border-coffee-400"
+    />
+  )
```
torztomasz, line 94: https://github.com/user-attachments/assets/bba2049f-3a64-4828-a1b4-d389a9dbbe3e

if easy we should handle this blur too

torztomasz:
https://github.com/l2beat/l2beat/pull/12161

opening then clicking outside makes it impossible to focus out

torztomasz (changes requested):

mateuszradomski:
@codex, review

torztomasz (approved):

PR #12168, opened 2026-06-29
Remove no-op JSON.stringify tRPC transformer

torztomasz:
## What

Removes the `{ serialize: JSON.stringify, deserialize: JSON.parse }` data transformer from every tRPC client and server in the repo.

## Why

A tRPC data transformer exists to preserve types JSON can't represent — `Date`, `BigInt`, `Map`, `Set`, `undefined` (what `superjson` does). The `JSON.stringify`/`JSON.parse` pair preserved **none** of those: tRPC already JSON-encodes the wire payload itself, so this transformer just **double-encoded** everything — extra encode/decode work per request with zero fidelity benefit. (A `Date` was already serialized to an ISO string regardless.)

## Changes

- Removed the transformer from every client/server pair so the wire formats stay in sync:
  - `frontend` (`src/server/trpc/trpc.ts`, `src/trpc/React.tsx`)
  - `backend` (`src/trpc/init.ts`)
  - `backoffice` (`src/react-query/trpc.tsx`)
  - `token-backend` (`src/trpc/trpc.ts`, `src/client.ts`)
  - `token-ui` (`src/react-query/trpc.tsx`)
- Dropped the now-stable `unstable_` prefix on `httpBatchStreamLink`.
- Fixed the prefetch e2e helper `parseCalls`, which decoded the old double-stringified wire format. With no transformer the batch `input` map holds raw values directly, so the per-index `JSON.parse` is removed — it would otherwise corrupt JSON-looking string inputs (e.g. `"123"` → `123`).

## Notes

`grep` confirms zero `serialize: JSON.stringify` transformers remain, so no client/server pair is left mismatched.

### DIFF
--- a/packages/backend/src/trpc/init.ts
+++ b/packages/backend/src/trpc/init.ts
@@ -2,10 +2,6 @@ import { initTRPC } from '@trpc/server'
 import type { BaseContext } from './context'
 
 const t = initTRPC.context<BaseContext>().create({
-  transformer: {
-    serialize: JSON.stringify,
-    deserialize: JSON.parse,
-  },
   errorFormatter({ shape, error }) {
     return {
       ...shape,
--- a/packages/backoffice/src/react-query/trpc.tsx
+++ b/packages/backoffice/src/react-query/trpc.tsx
@@ -110,10 +110,6 @@ function buildClient(url: string) {
           (op.direction === 'down' && op.result instanceof Error),
       }),
       httpBatchLink({
-        transformer: {
-          serialize: JSON.stringify,
-          deserialize: JSON.parse,
-        },
         url,
         headers: () => {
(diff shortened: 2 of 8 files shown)


mateuszradomski (approved):

torztomasz:
Abandoning: the JSON.stringify transformer is load-bearing for types (it makes tRPC skip its Serialize<T> output transform). Removing it breaks ~90 type sites across frontend + token-ui. Keeping the transformer everywhere.

torztomasz:
Closing: the transformer isn't a no-op, it's load-bearing for types. Removing it makes tRPC apply Serialize<T> (`field: string | undefined` -> `field?`), breaking ~90 type sites in frontend + token-ui. Not worth the refactor.

PR #12195, opened 2026-07-02
Optimize Tokens Ingestion memory usage

adamiak:
Resolves L2B-14309

## Problem

`TokenIngestionProcessor.refreshInteropTransferIndex()` loaded the entire interop transfer table into memory (`getAll()`) every minute to build the transfer index. This was fine with 24h retention, but #12152 extended retention to 7 days and the loop started erroring with Out-Of-Memory - peak usage is ~3× the table (old index + raw rows + new index) on every refresh.

## Fix

Aggregate in SQL instead of loading rows: new `getTokenRoutes()` groups transfers by `(src token, dst token, bridge-type evidence)` and returns one row per group with a transfer count and a sample transfer id. The in-memory index is built from these aggregates, so its size scales with distinct bridged token pairs, not transfer volume.

The only place that needs a full transfer row - the `non-swapping-transfer` assignment proof - now fetches the group's sample transfer by primary key, only when a plan actually resolves from transfer evidence.

Behavior is unchanged otherwise: same index interface, same refresh-before-plan freshness guarantee, full 7-day evidence window, and trace evidence counts still reflect transfer counts. Also updates stale "~24h retention" mentions in docs/comments.

(Implemented with Fable)


adamiak:
@codex review

torztomasz (approved):

PR #12201, opened 2026-07-03
add transfer time to flows stats

vilcht:


vilcht:
@codex review

packages/frontend/src/server/features/scaling/interop/getInteropFlows.ts
```diff
@@ -284,6 +321,10 @@ function computeChainsData(
   chainTopProtocols: Map<string, FlowProtocol[]>,
   chainTokenCounts: Map<string, number>,
   chainProtocolCounts: Map<string, number>,
+  chainDurations: Map<string, DurationData>,
+  resolveAvgDuration: (
```
torztomasz: can't we just not create the function in the local scope and do it in the file? then no need to pass the fn through params

torztomasz (approved):

PR #12202, opened 2026-07-03
im on nixos btw

sekuba:
my tests fail with the oldschool shebang so i added a newschool shebang that should work on all unixes?


kondziorex7 (approved):
LGTM, thanks :)

PR #12217, opened 2026-07-07
Implement Token Relations in TokenDB

adamiak:
Implemented token relations end-to-end across TokenDB, token-backend ingestion, and token-ui.

This replaces the unused `TokenConnection` model with `TokenRelation`, shaped around interop transfer evidence: source token, destination token, plugin, burn/mint flags, bridge type, and a stored sample transfer JSON. Added repository support, schema exports, migrations, CRUD commands/intents/planning, commit routing, and history coverage.

Automatic token ingestion now detects missing relations from interop transfers while processing queued token addresses. It can create relation-only writes when deployed tokens are already stable, without introducing a separate queue.

The deployed token page in token-ui now has a Relations tab showing outgoing and incoming token relations, including the related token, plugin, flags, bridge type, and transfer evidence.

Docs and tests were updated to cover the new relation model, ingestion behavior, and intent/plan/execute support.


adamiak:
All is done on a child branch (different PR)

PR #12222, opened 2026-07-08
Show if plugin x chain delay will block the aggregation

sdlyy:
Resolves L2B-14361
@codex review


packages/backoffice/src/pages/interop/status/table/columns.tsx
```diff
@@ -61,6 +62,35 @@ export const pluginStatusColumns: TableOptions<PluginStatusRow>['columns'] = [
           : 'n/a',
     },
   }),
+  columnHelper.accessor(
+    (row) =>
+      row.blocksAggregation === undefined
```
torztomasz: why can it be `undefined`?

torztomasz (approved):

packages/backoffice/src/pages/interop/status/table/columns.tsx
sdlyy: good point, resolved

PR #12234, opened 2026-07-09
increase default and add configurable celestia timeout

vilcht:


packages/backend/src/config/Config.ts
```diff
@@ -356,6 +356,7 @@ export type BlockLayerDaTrackingConfig = {
   name: string
   url: string
   callsPerMinute: number
+  timeout?: number
```
torztomasz, line 359: why make it optional if env has default param

torztomasz (approved):

PR #12244, opened 2026-07-10
Interop financial filters

sdlyy:
@codex review

Allow for selective refresh


packages/backoffice/src/App.tsx
```diff
@@ -64,9 +64,10 @@ function EnvironmentScopedApp() {
               element={<EventDetailsPage />}
             />
 
+            <Route path="/interop/financials" element={<FinancialsPage />} />
             <Route
               path="/interop/financials/actions"
-              element={<FinancialActionsPage />}
+              element={<Navigate to="/interop/financials" replace />}
```
torztomasz: do we need this one?

torztomasz (approved):

packages/backoffice/src/App.tsx
sdlyy: Wanted to keep the old urls in tact, but I guess it was not heavily used that much - so removed

PR #12246, opened 2026-07-10
change 1day percentage to 7day percentage

vilcht:


packages/frontend/src/pages/scaling/project/tvs-breakdown/components/tables/columns.tsx
```diff
@@ -167,3 +170,30 @@ export const columns = [
     },
   }),
 ]
+
+function PercentChangeNotAvailable() {
+  return (
+    <span className="relative text-secondary">
+      <span className="relative inline-block w-[52px] text-right text-xs">
+        %N/A
```
torztomasz: i do not think that's how we handle not available % change

torztomasz (changes requested):

packages/frontend/src/pages/scaling/project/tvs-breakdown/components/tables/columns.tsx
vilcht: undefined percentages are normally omitted, but this task explicitly calls for %N/A

torztomasz (approved):

PR #12249, opened 2026-07-10
Token relations

adamiak:
This is a big PR that originally was supposed to be a few smaller PRs, but with frontier AI I could reliably implement all features much faster, and refactor multiple times seeing where they don't work well together.

The features implemented here are:
- A separate Interop Transfer ingestion process: each "unique" transfer route (including plugin that was used to create it) becomes a TokenRelation in TokenDB.
- Display of TokenRelations in history and on DeployedToken pages (can be treated as "minters" view since it shows the plugin name that was responsible for each route)
- TokenRelations graph - rudimentary first version using force directed graph via d3.js. Show conflicting relations (different AbstractTokens in from and to) as red.


packages/database/src/repositories/TokenRelationRepository.ts
```diff
@@ -0,0 +1,193 @@
...
+    keys: TokenRelationPrimaryKey[],
+  ): Promise<TokenRelationRecord[]> {
+    if (keys.length === 0) return []
+
+    const rows = await this.db
+      .selectFrom('TokenRelation')
+      .selectAll()
+      .where((eb) =>
+        eb.or(
+          keys.map((key) =>
```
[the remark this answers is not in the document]
adamiak: fixed

packages/token-ui/src/components/PlanConfirmationDialog.tsx
```diff
@@ -52,6 +52,9 @@ export function PlanConfirmationDialog({
     queryClient.invalidateQueries(
       trpc.deployedTokens.getByChainAndAddress.queryFilter(),
     )
+    queryClient.invalidateQueries(
+      trpc.deployedTokens.getRelations.queryFilter(),
+    )
```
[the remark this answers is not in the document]
adamiak: fixed

adamiak:
@codex review

packages/token-backend/src/schemas/TokenRelation.ts
```diff
@@ -0,0 +1,49 @@
...
+
+type _ = Expect<Equal<TokenRelationRecord, DbTokenRelationRecord>>
+export type TokenRelationRecord = v.infer<typeof TokenRelationRecord>
+export const TokenRelationRecord = v.object({
+  tokenFromChain: v.string(),
+  tokenFromAddress: v.string(),
+  tokenToChain: v.string(),
+  tokenToAddress: v.string(),
+  plugin: v.string(),
+  bridgeType: v.enum(InteropBridgeTypeValues),
```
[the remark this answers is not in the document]
adamiak: Since there is no ability to add token relations via UI, we should not complicate the code further. We can also assume that people adding relations manually understand what they're doing.

packages/token-backend/src/trpc/routers/deployedTokens/index.ts
```diff
@@ -31,6 +31,128 @@ export const deployedTokensRouter = (deps: DeployedTokensRouterDeps) =>
         ctx.tokenDb.deployedToken.getByChainAndAddress(input),
       ),
 
+    getRelations: readOnlyProcedure
+      .input(v.object({ chain: v.string(), address: v.string() }))
+      .query(async ({ ctx, input }) => {
+        const [outgoing, incoming] = await Promise.all([
+          ctx.tokenDb.tokenRelation.getRelationsFrom(input),
+          ctx.tokenDb.tokenRelation.getRelationsTo(input),
+        ])
```
torztomasz, line 40: Can't we have one query for `from` and `to`?

packages/token-backend/src/ingestion/TokenRelationIngestion.ts
```diff
@@ -0,0 +1,185 @@
...
+    const startedAt = Date.now()
+    let scanned = 0
+    let inserted = 0
+
+    const setting = await this.tokenDb.tokenDbSettings.get(
+      TOKEN_RELATIONS_LAST_SERIAL_ID_KEY,
+    )
+    let lastSerialId = setting?.value ?? '0'
+
+    while (true) {
```
torztomasz: [P1] Relation ingestion can starve token ingestion

`runOnce()` drains every transfer page before returning. Because `TokenIngestionLoop` awaits it before enqueueing or draining tokens, an initial backfill or long outage can monopolize the only ingestion tick; only memory per page is bounded, not runtime or total work. Error containment does not prevent this starvation. Give relation ingestion an independent loop/lifecycle, or enforce a page/time budget per tick and resume from the cursor.

packages/token-ui/src/pages/tokens/TokenHistoryPage.tsx
```diff
@@ -107,6 +111,25 @@ export function TokenHistoryPage() {
   )
 
   const entries = historyPage?.entries ?? []
+  const relationTokenKeys = useMemo(
+    () => getRelationTokenKeys(entries),
+    [entries],
+  )
+  const { data: relationTokens } = useQuery(
```
torztomasz, line 118: [P2] Move history enrichment behind a typed backend boundary

This adds a client-side join by inspecting raw history command JSON, extracting up to 200 relation endpoints, and issuing another request; it then requires the global tRPC transport to switch to POST. The same change pushes this page from roughly 549 to 756 lines with more `Record<string, unknown>` fallback plumbing. Return a typed relation display projection from the history read model instead, eliminating `getRelationTokenKeys`, most of `relationInfo`, and the transport-wide workaround.

packages/token-ui/src/components/PlanConfirmationDialog.tsx
```diff
@@ -274,6 +290,59 @@ function CommandItem({ command }: { command: Command }) {
           </Tooltip>
         </li>
       )
+    case 'AddTokenRelationCommand':
```
torztomasz, line 293: [P2] Collapse the copy-pasted command rendering matrix

These three branches duplicate the existing add/update/delete JSX almost verbatim, while the success handler adds another three near-identical cases. Every new entity now multiplies branches across two switches. Introduce shared add/update/delete renderers driven by an entity label and typed payload descriptor, and group intents with identical invalidation behavior. This should delete the branch family rather than extend it.

packages/token-ui/TOKEN_RELATIONS_GRAPH.md
```diff
@@ -0,0 +1,24 @@
+# Token relations graph
```
torztomasz: should this live here?

torztomasz (changes requested):

torztomasz (approved):

packages/token-backend/src/trpc/routers/deployedTokens/index.ts
adamiak: We could merge them into one OR-query, but we'd then re-split rows into outgoing/incoming in JS anyway (and handle a self-loop landing in both). Two indexed point lookups in parallel cost the same and keep the split explicit - keeping as is

packages/token-ui/src/pages/tokens/TokenHistoryPage.tsx
adamiak: Agreed the history read model returning typed display projections would be nicer, but the page already renders every entity from raw command JSON - relations just follow that pattern. Typing the whole read model is a real refactor and I'd rather do it in a dedicated follow-up than special-case relations here.

packages/token-ui/src/components/PlanConfirmationDialog.tsx
adamiak: The matrix predates this PR - I followed the existing pattern rather than mixing a refactor of the abstract/deployed branches into this diff.

packages/token-ui/TOKEN_RELATIONS_GRAPH.md
adamiak: Moved to an mdbook

packages/token-backend/src/ingestion/TokenRelationIngestion.ts
adamiak: ok, add fix to enforce page budget.

PR #12274, opened 2026-07-14
show total uops count

vilcht:


packages/database/src/repositories/ActivityRepository.ts
```diff
@@ -238,13 +255,22 @@ export class ActivityRepository extends BaseRepository {
...
-      ]),
+          sinceTimestamp,
+        }
+
+        if (hasCompleteUopsHistory) {
+          totals.uopsCount = Number(row.total_uops_count)
+        }
+
+        return [ProjectId(row.projectId), totals]
+      }),
```
torztomasz: let's do this on frontend side, database should return simple data

torztomasz (changes requested):

torztomasz (approved):

PR #12277, opened 2026-07-14
Sergey dailies 13 07

sergeyshemyakov:
Resolves L2B-14359


sekuba (approved):

sekuba:
@sergeyshemyakov plx review the SOTA slop

sergeyshemyakov:
Config hash should not be in progHashes. Maybe we could review config hashes as well at some point in future, but IMO this is not necessary. Exploits with malicious config hashes are much more limited (e.g. you could replay messages from other chain IDs or forge DA encryption keys if there are any, but not realistically rug people).

Also, if we look into config hashes, there are many more for other chains that we should look into as well.

sekuba (approved):
great dailies

PR #12278, opened 2026-07-14
Increase contrast in bridge type badges

vilcht:


torztomasz (changes requested):
I guess there is not usage of yellow-900 right now. Delete it

torztomasz (approved):

PR #12303, opened 2026-07-17
support deduplication by arg for aztec liveness

sekuba:
for liveness, aztec state updates should be deduplicated by their target epoch. if not we show low average intervals like 3minutes instead of 38 minutes
for costs, we should count them all and not dedup
this PR is fully configured to support aztec liveness
https://l2beat.com/scaling/projects/aztecnetwork#liveness


sekuba:
rejected for its beauty

PR #12304, opened 2026-07-17
da tracking fix when theres duplicate config

vilcht:


torztomasz:
the issue was somewhere else

PR #12324, opened 2026-07-21
allow users to link to updates

vilcht:


packages/frontend/src/server/features/projects/recent-changes/getDiscoveryUpdates.ts
```diff
@@ -129,17 +130,32 @@ function toPublicDiscoveryUpdate(
...
     description: entry.description,
     isHighSeverity: bodies.some((body) => isHighSeverityDiffBody(body)),
     changeCount: bodies.reduce((sum, body) => sum + countDiffChanges(body), 0),
     sections,
   }
 }
 
+function getUpdateId(date: string, timestamp: number | null): string {
+  if (timestamp !== null) {
+    return `update-${timestamp}`
```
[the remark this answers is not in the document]
vilcht: fixed

packages/frontend/src/pages/scaling/ScalingRouter.ts
```diff
@@ -130,7 +131,22 @@ export function createScalingRouter(
...
+          ssr: {
+            ...data.ssr,
+            props: {
+              ...data.ssr.props,
+              selectedUpdateId: getSelectedUpdateId(req.query),
+            },
+          },
+        },
+        req.originalUrl,
+      )
```
torztomasz: this should live in getData

packages/frontend/src/pages/interop/InteropRouter.ts
```diff
@@ -122,7 +123,22 @@ export function createInteropRouter(
...
+          ssr: {
+            ...data.ssr,
+            props: {
+              ...data.ssr.props,
+              selectedUpdateId: getSelectedUpdateId(req.query),
+            },
+          },
+        },
+        req.originalUrl,
+      )
```
torztomasz: same

packages/frontend/src/components/projects/sections/UpdatesSection.tsx
```diff
@@ -205,6 +240,11 @@ function UpdateCard({ update }: { update: DiscoveryUpdate }) {
   )
 }
 
+function getUpdateHref(updateId: string): string {
+  const encodedId = encodeURIComponent(updateId)
+  return `?update=${encodedId}#${encodedId}`
```
torztomasz: why repeat the id?

torztomasz (changes requested):

torztomasz (commented):
> 🤖 This review was AI-generated (Claude Code), requested by @torztomasz.

Solid, well-tested change. Things that are done well:

- `getSelectedUpdateId` treats `req.query` as `unknown` and rejects non-strings, so `?update[x]=y` or repeated params degrade gracefully. Reflected-input XSS is also covered since SSR props go through `jsonForInlineScript`, which escapes `<`.
- `selectedUpdateId` is layered onto the page data *after* `cache.get`, so the in-memory cache stays keyed by slug only and isn't polluted with query-specific props.
- The `useState` pagination initializer computes the same page on server and client, so there's no hydration mismatch and the `#fragment` scroll works even before hydration (and with JS disabled).
- The hardcoded `https://l2beat.com` base matches the existing convention in `GlossaryItem.tsx`.

One consideration outside the diff: the same path now renders different HTML per `?update=` value. If any CDN/edge layer caches these routes while normalizing away query strings, users could get a variant with the wrong card pre-expanded — worth a quick check of the CDN config.

Inline comments below for the few substantive points.

packages/frontend/src/server/features/projects/recent-changes/getDiscoveryUpdates.ts
```diff
@@ -129,17 +130,39 @@ function toPublicDiscoveryUpdate(
...
+  const dateSlug = entry.date
+    .toLowerCase()
+    .replace(/[^a-z0-9]+/g, '-')
+    .replace(/^-|-$/g, '')
+
+  if (entry.discoveryHash !== null) {
+    return `update-${timestamp ?? dateSlug}-${entry.discoveryHash}`
+  }
+
+  if (timestamp !== null) {
```
torztomasz: > 🤖 AI-generated comment (Claude Code)

Duplicate ids are still possible here: two entries with the same timestamp and no `discoveryHash` (or two legacy entries with the same date string) produce identical ids. That yields duplicate React keys — the old `${update.date}-${index}` key was collision-proof, the new `key={update.id}` is not — plus duplicate DOM ids as anchor targets.

Dates have second granularity so this is unlikely in practice, but a positional tiebreaker (e.g. appending the entry index only on collision) would make it airtight.

packages/frontend/src/components/projects/sections/UpdatesSection.tsx
```diff
@@ -129,9 +156,17 @@ function UpdateCard({ update }: { update: DiscoveryUpdate }) {
...
           <div className="flex min-w-0 flex-1 flex-wrap items-center gap-x-2 gap-y-1">
-            <span className="shrink-0 font-medium text-primary leading-snug">
-              {formatUpdateDate(update)}
-            </span>
+            <div className="flex shrink-0 items-center gap-1">
+              <span className="font-medium text-primary leading-snug">
+                {formatUpdateDate(update)}
+              </span>
+              <CopyButton
+                toCopy={`https://l2beat.com${pathname}${getUpdateHref(update.id)}`}
```
torztomasz: > 🤖 AI-generated comment (Claude Code)

Minor: the copied URL is built from `pathname` + `?update=…`, so any other query params present on the current URL are dropped. Nothing relevant exists on these pages today, so just a note for the future.

Also, `encodeURIComponent` is redundant given the id alphabet is `[a-z0-9x-]` by construction — harmless either way.

packages/frontend/src/pages/scaling/ScalingRouter.ts
```diff
@@ -130,7 +131,22 @@ export function createScalingRouter(
         res.status(404).send('Not found')
         return
       }
-      const html = await render(data, req.originalUrl)
+      if (data.ssr.page !== 'ScalingProjectPage') {
```
torztomasz: > 🤖 AI-generated comment (Claude Code)

Non-blocking: this guard is unreachable in practice (`getScalingProjectData` only ever returns `ScalingProjectPage`), so it reads as a type-narrowing device. A short comment saying so would clarify intent.

The same injection block is duplicated in `InteropRouter.ts` — fine at two occurrences, but worth extracting a helper if a third page ever gets updates.

packages/frontend/src/components/projects/sections/UpdatesSection.tsx
```diff
@@ -129,9 +161,17 @@ function UpdateCard({ update }: { update: DiscoveryUpdate }) {
...
```
torztomasz: make it work on previews too, extract the full url from window

packages/frontend/src/components/projects/sections/UpdatesSection.tsx
```diff
@@ -205,6 +245,10 @@ function UpdateCard({ update }: { update: DiscoveryUpdate }) {
   )
 }
 
+function getUpdateHref(updateId: string): string {
+  return `?update=${updateId}`
```
torztomasz: no need for this fn

packages/frontend/src/pages/scaling/project/getScalingProjectData.ts
```diff
@@ -6,12 +8,36 @@ import { getProjectMetadataDescription } from '~/ssr/head/getProjectMetadataDesc
...
+    ...data,
+    ssr: {
+      ...data.ssr,
+      props: {
+        ...data.ssr.props,
+        selectedUpdateId: getSelectedUpdateId(req.query),
+      },
+    },
+  }
+}
```
torztomasz, line 37: this is some weird data flow change
cache should live outside this function, selectedUpdateId should alredy be inside data.ssr.props

getSelectedUpdateId seems like really redundant, just use correct type and validation on request

packages/frontend/src/server/features/projects/recent-changes/getDiscoveryUpdates.ts
```diff
@@ -129,17 +131,43 @@ function toPublicDiscoveryUpdate(
...
+  const fingerprint = hashJson([
+    entry.date,
+    entry.current?.kind ?? null,
+    entry.current?.value ?? null,
+    entry.description,
+    entry.sections.flatMap((section) => [section.kind, section.body]),
+  ])
+
+  return `update-${timestamp ?? dateSlug}-${fingerprint}`
+}
```
torztomasz: can't this be just a simple hashJson.slice(0,8) for better UX? If you have any better idea - please suggest

torztomasz (approved):

PR #12333, opened 2026-07-21
Aztec network activity

sdlyy:
@codex review

Aztec activity tracking by counting `body.txEffects` from `node_getBlocks(includeTransactions: true)`, which includes both public and private mined transactions.  
The change is layered: configure the Aztec RPC and block activity → add an Aztec RPC client/provider with range validation and timestamp lookup → normalize it through ActivityBlockProvider → reuse the existing block activity aggregation/indexer. Standard chains still use their existing per-block fetch and Uops analyzers; Aztec reports uopsCount: null because transaction effects do not expose a reliable execution/call structure for uops.
Tests cover Aztec RPC parsing/errors, provider range validation and timestamp lookup, Aztec effect-to-activity mapping, and preservation of standard transaction/uops behavior.
Follow-up before production: Aztec RPC permits at most 50 blocks per request, while the block indexer’s range is inclusive. The checked-in Aztec batch size should be 49 (or the batching semantics should be corrected);
(stresstestedlocally)

Also if you mind why I made changes to core wiring:

Aztec does not fit the existing normalized `BlockClient` contract, so this adds a separate `aztec-rpc` client stack (Clients → AztecBlockProviders) with the same retry, rate-limit, multi-endpoint fallback, and block-timestamp capabilities as other block providers.
`ActivityBlockProviders` is the narrow compatibility layer above those stacks. It exposes one activity-shaped block source to the activity module: standard providers return transaction arrays plus optional uops; Aztec returns `txEffectsCount` plus `null uops`. This keeps protocol-specific RPC handling out of `ActivityModule` and `BlockTxsCountService`, while preserving existing standard-chain behavior peeeace


packages/config/src/projects/aztecnetwork/aztecnetwork.ts
```diff
@@ -332,6 +332,11 @@ export const aztecnetwork: ScalingProject = {
   },
   config: {
     associatedTokens: ['AZTEC'],
+    activityConfig: {
+      type: 'block',
+      startBlock: 1,
+      batchSize: 50,
```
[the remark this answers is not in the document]
sdlyy: fxd

sdlyy:
@codex review once again please

packages/shared/src/clients/rpc-aztec/AztecRpcClient.ts
```diff
@@ -0,0 +1,92 @@
...
+        response: JSON.stringify(response),
+      })
+      throw new Error(`${method}: Error during parsing`)
+    }
+
+    return parsedResponse.data.result
+  }
+
+  async getBlocks(start: number, limit: number): Promise<AztecBlock[]> {
+    const method = 'node_getBlocks'
```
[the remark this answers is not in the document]
sdlyy: Ok this is so cool and terrifying at the same time - what do they feed you with?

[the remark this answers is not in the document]
sdlyy: fixd

sdlyy:
@codex review once again

packages/shared/src/clients/rpc-aztec/AztecRpcClient.ts
```diff
@@ -0,0 +1,92 @@
...
+}
+
+export class AztecRpcClient extends ClientCore implements AztecBlockClient {
+  constructor(private readonly $: Dependencies) {
+    super($)
+  }
+
+  async getLatestBlockNumber(): Promise<number> {
+    const method = 'aztec_getBlockNumber'
+    const response = await this.query(method, [])
```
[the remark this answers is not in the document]
sdlyy: We have one-hour long safety net behind the tip

sdlyy:
@codex review

packages/shared/src/clients/rpc-aztec/AztecRpcClient.ts
```diff
@@ -0,0 +1,92 @@
...
+        response: JSON.stringify(response),
+      })
+      throw new Error(`${method}: Error during parsing`)
+    }
+
+    return parsedResponse.data.result
+  }
+
+  async getBlocks(start: number, limit: number): Promise<AztecBlock[]> {
+    const method = 'aztec_getBlocks'
```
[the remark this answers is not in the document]
sdlyy: na

sdlyy:
@codex review - please focus on the key parts - not the outline

sdlyy:
Codex's CD is stuck

packages/shared/src/providers/block/AztecBlockProvider.ts
```diff
@@ -0,0 +1,72 @@
...
+    private readonly clients: AztecBlockClient[],
+  ) {
+    assert(clients.length > 0, 'Clients cannot be empty')
+  }
+
+  async getLatestBlockNumber(): Promise<number> {
+    return await this.withClient((client) => client.getLatestBlockNumber())
+  }
+
+  async getBlocks(start: number, limit: number): Promise<AztecBlock[]> {
```
torztomasz, line 21: **Main risk:** the 50-block cap is enforced only by config convention — nothing here or in `AztecRpcClient` knows about it. If `batchSize` is ever raised, or omitted (`ActivityModule` defaults to 100), every `update()` gets a permanent `-32701 "Too big: expected number to be <=50"` (verified against the live endpoint), which `validateResponse` treats as retryable, and `BlockActivityIndexer` uses `getInfiniteRetryStrategy()` — the indexer retries a permanent error forever.

Either of these would fix it:
- chunk requests of ≤50 here in `getBlocks` (makes config `batchSize` a pacing knob, not a correctness knob), or
- `assert(limit <= AZTEC_MAX_BLOCKS_PER_REQUEST)` in the client with a named constant.

packages/shared/src/providers/block/AztecBlockProvider.ts
```diff
@@ -0,0 +1,72 @@
...
+        assert(
+          block.number === start + index,
+          `Expected block ${start + index}, got ${block.number}`,
+        )
+      }
+      return blocks
+    })
+  }
+
+  async getBlockNumberAtOrBefore(
```
torztomasz, line 49: Nit (perf): the binary search runs on every `BlockTargetIndexer` tick (~20 calls per tick at current chain height, growing with `log2(n)`), and each call is `getBlocks(number, 1)` with `includeTransactions: true`. The param exists precisely to skip bodies — a lightweight variant (`includeTransactions: false` with a header-only schema) would meaningfully cut payload on this hot path. Not a blocker.

packages/shared/src/providers/block/AztecBlockProvider.ts
```diff
@@ -0,0 +1,72 @@
...
+  ): Promise<T> {
+    for (const [index, client] of this.clients.entries()) {
+      try {
+        return await callback(client)
+      } catch (error) {
+        if (index === this.clients.length - 1) throw error
+      }
+    }
+
+    throw new Error(`Missing Aztec RPC client for ${this.chain}`)
```
torztomasz, line 85: Nit: this throw is unreachable (the constructor asserts non-empty clients). Consistent with `BlockProvider`'s copy of the same pattern, so fine — just noting the fallback loop is now duplicated across three provider classes.

packages/config/src/projects/aztecnetwork/aztecnetwork.ts
```diff
@@ -332,6 +332,11 @@ export const aztecnetwork: ScalingProject = {
   },
   config: {
     associatedTokens: ['AZTEC'],
+    activityConfig: {
+      type: 'block',
+      startBlock: 1,
+      batchSize: 49,
```
torztomasz, line 339: `49` is a magic number whose justification currently lives only in the PR description. Worth a comment here: the indexer range is inclusive, so `batchSize: 49` → exactly 50 blocks per request, which is the Aztec RPC cap. (Or make it structurally enforced in the provider/client and keep this as a pacing knob.)

torztomasz (commented):
Solid, well-layered PR — the `ActivityBlockProvider` cut keeps protocol-specific RPC handling out of `ActivityModule`/`BlockTxsCountService`, and `AztecRpcClient` faithfully follows the `SvmRpcClient` conventions.

I verified the RPC contract against the live endpoint (`https://aztec-mainnet.drpc.org`): top-level `number`, `body.txEffects` array, and `timestamp` as a decimal string all match the `AztecBlock` validator exactly, and `aztec_getBlocks` with limit 51 returns error `-32701 "Too big: expected number to be <=50"`. The batch math also checks out: `BlockActivityIndexer` fetches `from..from+batchSize` **inclusive**, so `batchSize: 49` → exactly 50 blocks per request, right at the cap — so the follow-up from the description is effectively already resolved, it just needs structural enforcement (see inline comment).

Test coverage is good on the happy paths. Gaps worth a small follow-up: `AztecBlockProvider.getBlockNumberAtOrBefore` (binary-search adapter, the `start >= end` reset) and multi-client fallback are untested; same for the duplicate-chain assert in `ActivityBlockProviders`.

Verdict: LGTM once the 50-block cap comment is addressed; the rest are nits. Details inline.

packages/config/src/projects/aztecnetwork/aztecnetwork.ts
```diff
@@ -375,8 +380,14 @@ export const aztecnetwork: ScalingProject = {
   chainConfig: {
     gasTokens: ['AZTEC'],
     name: 'aztecnetwork',
-    chainId: 677868, // on chainlist, but aztec is not evm
-    apis: [], // TODO: add
+    chainId: 677868, // TODO: verify
```
torztomasz: Nit: the comment regressed from the informative `// on chainlist, but aztec is not evm` to `// TODO: verify` — keep the original or resolve the TODO before merge.

packages/backend/src/modules/activity/services/txs/types.ts
```diff
@@ -0,0 +1,11 @@
+export interface ActivityBlock {
+  number: number
+  timestamp: number
```
torztomasz: Nit: `aggregatePerDay` expects `timestamp: UnixTime` and `BlockActivityIndexer` treats `latestTimestamp` as `UnixTime`. It works because `UnixTime` is number-compatible, but typing this (and `AztecBlock.timestamp` in `packages/shared/src/clients/types.ts`) as `UnixTime` would match codebase convention and keep the contract honest.

packages/backend/src/modules/activity/services/txs/types.ts
```diff
@@ -0,0 +1,11 @@
...
+  txsCount: number
+  uopsCount: number | null
+}
+
+export interface ActivityBlockProvider {
+  chain: string
+  getBlocks(from: number, to: number): Promise<ActivityBlock[]>
```
torztomasz, line 13: Nit: the ascending-order contract is implicit — `BlockTxsCountService` takes `.at(-1)` as `latestTimestamp`. Both implementations honor it, but a one-line doc comment here ("returns blocks in ascending block-number order") would prevent a future provider from breaking it silently.

packages/shared/src/clients/rpc-aztec/AztecRpcClient.test.ts
```diff
@@ -0,0 +1,102 @@
...
+
+      const result = await client.getBlocks(100, 1)
+
+      expect(result).toEqual([
+        { number: 100, timestamp: 1234567890, txEffectsCount: 2 },
+      ])
+    })
+  })
+
+  describe(AztecRpcClient.prototype.query.name, () => {
```
torztomasz: Nit: this `describe` block is labeled `query` but actually tests `getBlocks`' request shape — mislabeled.

packages/shared/src/providers/block/AztecBlockProvider.ts
sdlyy: Good catch. I moved the cap into `AztecBlockProvider`: oversized ranges are now fetched sequentially in <=50-block chunks. `batchSize: 49` remains the one-request pacing default and is documented in config.

packages/shared/src/providers/block/AztecBlockProvider.ts
sdlyy: Agreed. Added `getBlockHeaders`, which calls `aztec_getBlocks` with `includeTransactions: false`; timestamp search now uses that header-only path. Added coverage for it.

packages/shared/src/providers/block/AztecBlockProvider.ts
sdlyy: Agreed. This mirrors the existing `BlockProvider` fallback pattern. I am leaving it as-is to avoid an unrelated provider abstraction in this PR.

packages/config/src/projects/aztecnetwork/aztecnetwork.ts
sdlyy: Added a config comment explaining the inclusive range semantics. The provider now also chunks requests structurally, so this value is no longer correctness-critical.

packages/config/src/projects/aztecnetwork/aztecnetwork.ts
sdlyy: Restored the original Chainlist/non-EVM comment.

packages/backend/src/modules/activity/services/txs/types.ts
sdlyy: Updated Aztec and activity block timestamps to `UnixTime`.

packages/backend/src/modules/activity/services/txs/types.ts
sdlyy: Documented the ascending block-number ordering contract on `ActivityBlockProvider`.

packages/shared/src/clients/rpc-aztec/AztecRpcClient.test.ts
sdlyy: Renamed the test group to `getBlocks`.

torztomasz (approved):

PR #12334, opened 2026-07-21
Deduplicate home chart series merging and category counts shape

torztomasz:
Follow-up to #11962 (targets the `layer2s-rename` branch), addressing the remaining parts of review finding 3 — the divergent change-computation half was already fixed by `move calculations to server side` (`computeSeriesChange`).

## Change

- `getHomeLayer2sCharts.ts`: the two verbatim-identical ~20-line `byTimestamp` merge blocks (TVS + activity) collapse into one `mergeSeriesByTimestamp` helper, covered by unit tests.
- `HomeLayer2sCategoryCounts` is now a single exported type (defined next to the card that renders it) instead of three inline `{rollups, validiumsAndOptimiums, others}` copies in HomePage, HomeLayer2sCard and CountsLine.
- The `others` count — computed and serialized into page props but never rendered anywhere — is dropped.
- Tiny consistency cleanup: `pastDayActivityUops` uses `findLast` like the adjacent `latestTvs` instead of `[...arr].reverse().find`.

## Behavior preservation

- **Fuzz equivalence**: the new helper vs a verbatim copy of the old inline block over 10,000 randomized series (duplicate timestamps, nulls, empty inputs) — byte-identical output on every case.
- **Unit tests**: 5 new cases for `mergeSeriesByTimestamp`; full suite passes (449 tests).
- **Screenshots**: home page before/after at 5 viewports (375–2560px) in mock mode — only the known interop particle-graph animation noise differs; the Layer 2s card region (counts line, sparklines, stat values, change badges) is pixel-identical. SSR props confirmed to no longer serialize `others`.
- `tsc --noEmit` and biome pass.

### DIFF
--- a/packages/frontend/src/pages/home/HomePage.tsx
+++ b/packages/frontend/src/pages/home/HomePage.tsx
@@ -15,6 +15,7 @@ import type { InteropFlowsProtocol } from '../interop/components/flows/utils/Int
 import { HomeAnomaliesTile } from './components/HomeAnomaliesTile'
 import { HomeEthereumCard } from './components/HomeEthereumCard'
 import { HomeInteropCard } from './components/HomeInteropCard'
+import type { HomeLayer2sCategoryCounts } from './components/HomeLayer2sCard'
 import { HomeLayer2sCard } from './components/HomeLayer2sCard'
 import { HomeRecentChangesTile } from './components/HomeRecentChangesTile'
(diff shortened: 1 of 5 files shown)


torztomasz:
Replaced by the summary-page-based port of the same change — see the new PR referencing this one. Branch left in place.

PR #12336, opened 2026-07-22
show better unverified contracts

vilcht:


packages/frontend/src/server/features/utils/getIsProjectVerified.ts
```diff
@@ -1,59 +1,66 @@
...
+        ? project.contracts?.programHashesDescription
+        : undefined,
+    },
+    unverifiedContracts,
   }
 }
 
-function areContractsVerified(
-  becameVerifiedContracts: ChainSpecificAddress[],
+function getUnresolvedUnverifiedContracts(
```
torztomasz: what does unresolved unverified contract mean?

torztomasz: this function looks like it just returns unverfied contracts

packages/frontend/src/server/features/utils/getIsProjectVerified.ts
```diff
@@ -1,59 +1,66 @@
...
-  contracts: string | undefined
-  programHashes: string | undefined
-  programHashesDescription: string | undefined
-}
-
 export function getProjectVerificationWarnings(
   project: Project<'statuses', 'contracts'>,
   changes: ProjectChanges | undefined,
 ): ProjectVerificationWarnings {
+  return getProjectVerification(project, changes).warnings
```
torztomasz: naaah, bad pattern, just use `getProjectVerification`

packages/frontend/src/utils/project/risk-summary/getScalingRiskSummary.ts
```diff
@@ -9,7 +9,10 @@ import { groupRisks } from './groupRisks'
 export function getScalingRiskSummarySection(
   project: Project<'scalingTechnology', 'contracts'>,
   verificationWarnings: ProjectVerificationWarnings,
-): Omit<RiskSummarySectionProps, keyof ProjectSectionProps> {
+): Omit<
+  RiskSummarySectionProps,
+  keyof ProjectSectionProps | 'unverifiedContracts'
```
torztomasz: why omitted here?

torztomasz: seems like scaling risk summary needs them

torztomasz: yep, we are removing it from return and then doing

```tsx
const riskSummary = getScalingRiskSummarySection(
    project,
    projectVerificationWarnings,
  )
  if (riskSummary.riskGroups.length > 0) {
    sections.push({
      type: 'RiskSummarySection',
      props: {
        ...riskSummary,
        id: 'risk-summary',
        title: 'Risk summary',
        hostChainWarning: hostChainWarningWithRiskCount,
        isUnderReview: !!project.statuses.reviewStatus,
        unverifiedContracts,
      },
    })
  }
```

packages/frontend/src/components/projects/sections/ContractEntry.tsx
```diff
@@ -110,20 +117,29 @@ export function ContractEntry({
...
                 addresses={contract.addresses}
+                chain={contract.chain}
+                type={type}
                 className="basis-full"
               />
             ) : (
               entries.map((address, i) => (
                 <HighlightableLink
                   key={i}
+                  id={getContractAddressAnchor(
```
torztomasz: This assigns an anchor id to **every** rendered address — `entries` is `[...contract.addresses, ...contract.admins]`, so implementations and admins get one too. Those repeat across entries, so the same `id` lands on many elements and fragment navigation resolves to the *first* match.

Measured against all 236 projects in `packages/config/build/db.sqlite`:

- **180 of 236 projects** emit duplicate anchor ids (413 distinct duplicated ids). `eigenda` alone has one id repeated **13×**.
- **3 unverified-contract links land on a duplicated id.** Clearest case is **zkswap2**: `eth:0x0DCCe462…` is the `UpgradeGatekeeper` contract, but the first element with `id="contracts-eth:0x0DCC…"` is the *Admin* row inside the `ZkSync` entry — the banner says "UpgradeGatekeeper" and scrolls you to `ZkSync`.
- `syndicate` `eth:0x481D2904…` has 9 elements with that id and no contract entry of its own (it's purely an admin address).

Duplicate `id` attributes are also invalid HTML, independent of the navigation bug.

**Suggestion:** the anchor is only ever *targeted* at a contract's own address or a permission account, so only emit it there. Instead of reconstructing the id in the component, set an `anchorId` on `TechnologyContractAddress` in `getContractsSection` / `getPermissionsSection` (first occurrence wins), and here just `id={address.anchorId}`.

That also lets you drop the `type` / `chain` props and the `ChainSpecificAddress.fromLong` round-trip. `fromLong` throws for chain names outside the hard-coded `SHORT_TO_LONG_CHAIN_NAMES` map, and `ProjectContract.chain` is a free-form `string` — so a single odd config would crash SSR for the whole project page, and a `chain` that disagrees with the address prefix (e.g. `polygon` vs `polygonpos`, both valid keys) silently produces a non-matching id. Both are clean today (0 throws, 0 mismatches across all projects), but passing the already-known `ChainSpecificAddress` down avoids the class entirely.

packages/frontend/src/utils/project/contracts-and-permissions/getUnverifiedContractEntries.ts
```diff
@@ -0,0 +1,62 @@
...
+        id: getContractAddressAnchor('permissions', account.address),
+        label: getDisplayLabel(
+          permission.name || account.name,
+          account.address,
+        ),
+      })
+    }
+  }
+
+  const projectContracts = Object.values(contracts?.addresses ?? {}).flat()
```
torztomasz: `getProjectUnverifiedContracts` in config sources addresses from four places: `contracts.addresses`, `permissions`, **`config.escrows`**, and **DA-bridge contracts/permissions belonging to another project**. Only the first two are indexed here, so the other two fall through to `target: undefined` and render as a bare, unlabelled `0x…` in the list.

Zero occurrences today — I checked all 236 projects and every one of the 52 unverified addresses resolves to a target — so this is latent, not live. Still worth hardening:

- Index `contracts.escrows` as well. Unmatched escrows **are** rendered in the contracts section with a `contracts-` anchor (`escrowToProjectContract` in `getContractsSection.ts`), so the link would work.
- Note that `getUnverifiedEscrows` does **not** filter `isHistorical`, while `getContractsSection` renders only `!isHistorical` escrows — a historical unverified escrow can therefore never have an anchor.
- DA-bridge addresses belong to a different project and can never resolve on this page.

A more robust shape overall: resolve targets against the **built** `ContractsSection` / `PermissionSection` rather than raw config. Then "is this address rendered, and under what heading" is answered by construction, and the escrow / historical / DA-bridge cases all fall out correctly for free.

(Re: the Codex bot's related comment above — the count itself stays truthful either way. This maps over *all* of `unverifiedContracts`; nothing is silently dropped, only the link is missing.)

packages/frontend/src/utils/project/contracts-and-permissions/getUnverifiedContractEntries.ts
```diff
@@ -0,0 +1,62 @@
...
+
+  const projectPermissions = Object.values(permissions ?? {}).flatMap(
+    ({ roles = [], actors = [] }) => roles.concat(actors),
+  )
+  for (const permission of projectPermissions) {
+    for (const account of permission.accounts) {
+      targetByAddress.set(account.address, {
+        id: getContractAddressAnchor('permissions', account.address),
+        label: getDisplayLabel(
+          permission.name || account.name,
```
torztomasz: Preferring `permission.name` over `account.name` can disagree with the heading you actually land on.

In `getPermissionsSection.toTechnologyContract`, an actor with more than one account is grouped and rendered under a generic heading — `` `${permission.accounts.length} ${allEoas ? 'EOAs' : 'actors'}` `` — with the per-account name (`displayName ?? account.name`) on the address row. So the banner can say "Sequencers" while the entry it scrolls to is titled "5 EOAs".

For grouped actors it's the account-level name that identifies the address; `permission.name` is only the heading in the non-grouped case.

torztomasz (changes requested):

torztomasz:
@codex review

torztomasz:
@codex review

torztomasz (approved):

PR #12344, opened 2026-07-22
Add Uniswap v3 discovery 

vincfurc:
resolves L2B-14444


lucadonnoh:
made some changes because many descriptions were hardcoding fields, while descriptions should be generic and only about the impl / shape. now lgtm and waiting for dev review

vincfurc:
continued at https://github.com/l2beat/l2beat/pull/12486

PR #12346, opened 2026-07-23
add datracking for aztec

sekuba:
close L2B-14494

### DIFF
--- a/packages/config/src/projects/aztecnetwork/aztecnetwork.ts
+++ b/packages/config/src/projects/aztecnetwork/aztecnetwork.ts
@@ -1,5 +1,6 @@
 import {
   ChainSpecificAddress,
+  EthereumAddress,
   formatLargeNumber,
   formatSeconds,
   ProjectId,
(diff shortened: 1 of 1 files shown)


lucadonnoh:
have you checked for false positives?

sekuba:
> have you checked for false positives?

latest 50k ethereum blocks are clean, but filtering by emitter would be nice ofc

packages/config/src/projects/aztecnetwork/aztecnetwork.ts
```diff
@@ -344,6 +345,17 @@ export const aztecnetwork: ScalingProject = {
         tokens: ['AZTEC'],
       }),
     ],
+    daTracking: [
+      {
+        type: 'ethereum',
+        daLayer: ProjectId('ethereum'),
+        sinceBlock: v5ActivationTimestamp,
```
vilcht: this expects block and not timestamp

packages/config/src/projects/aztecnetwork/aztecnetwork.ts
```diff
@@ -344,6 +345,17 @@ export const aztecnetwork: ScalingProject = {
...
+        inbox: EthereumAddress.ZERO, // Event-only tracking
```
vilcht, line 354: We never zero the address, it's made explicitely for a reason. This would work but I'd make it optional, and if it's not supplied topics would have to be supplied. That way the shape makes sense and doesnt look hacky but it's a nit.

packages/config/src/projects/aztecnetwork/aztecnetwork.ts
sekuba: how should i have ever known that :'''((((
vitaminsun

vilcht (approved):

PR #12348, opened 2026-07-23
Change ApeX Pro to ApeX Omni, update disco and project page

sergeyshemyakov:
Resolves L2B-14487
I will look into the proof system of zkLink a bit later, it seems to be a non-trivial task. I was not able to find the circuits repo (seems to be private now).


sekuba (approved):

sekuba:
disabled automerge for the rename

PR #12349, opened 2026-07-23
Aztec liveness grouping

vilcht:
resolves L2B-14450

Depends on #12410 for database support.

## Summary

- add function-call parameter grouping to tracked liveness configuration and results
- configure Aztec state updates to group proofs by epoch
- propagate grouping keys through batch and real-time liveness processing
- ignore malformed grouped calldata in real-time liveness matching

## Why

Multiple Aztec proofs can belong to the same epoch. Grouping them keeps liveness calculations aligned with epochs while preserving cost records per transaction.

## Validation

- backend tests: 1,253 passing
- config tests: 23,706 passing
- shared tests: 390 passing
- backend, config, and shared typechecks


packages/backend/src/modules/anomalies/RealTimeLivenessProcessor.ts
```diff
@@ -118,12 +121,24 @@ export class RealTimeLivenessProcessor implements BlockProcessor {
...
+          configurationId: config.id,
+        }
+
+        if (hasLivenessGrouping(config)) {
+          record.groupingKey = getLivenessGroupingKey(
+            tx.data as string,
+            config.params,
+            config.groupBy,
+          )
+        }
```
torztomasz, line 149: `getLivenessGroupingKey` throws on calldata that doesn't ABI-decode, and this path matches raw block txs by `to` + 4-byte selector only — no receipt/success check. Anyone can send a cheap reverting tx to the Rollup address with selector `0x069d1525` and garbage calldata: the decode throws, `matchLivenessTransactions` rejects, and `BlockIndexer` catches it at the processor level — so the **whole block is skipped** for `RealTimeLivenessProcessor`: all projects' real-time liveness matching *and* the `checkForAnomalies` call for that block are silently lost (just a `Processor failed` log). Repeatable for ~21k gas per block.

The batch path is safe (the Dune query filters `tr.success = true`, so the contract itself already decoded any calldata we see), but here please wrap the key extraction in a try/catch and drop the record on failure — a call the contract couldn't decode isn't a real proof, so skipping it is also semantically correct.

The `key.length <= 255` assert in `getLivenessGroupingKey` has the same throw-on-adversarial-input shape — unreachable for Aztec's `uint256` (≤ 78 chars), but the function is generic, so it deserves the same treatment.

packages/backend/src/modules/tracked-txs/modules/liveness/LivenessUpdater.ts
```diff
@@ -31,11 +31,22 @@ export class LivenessUpdater implements TxUpdaterInterface<'liveness'> {
...
+        configurationId: transaction.id,
+        txHash: transaction.hash,
+      }
+
+      if (
+        'groupingKey' in transaction &&
+        transaction.groupingKey !== undefined
+      ) {
+        record.groupingKey = transaction.groupingKey
+      }
```
torztomasz, line 48: We avoid probing union members with the `in` operator — and thanks to your own `model.ts` split the discriminants are right there:

```ts
if (
  transaction.formula === 'functionCall' &&
  transaction.type === 'liveness' &&
  transaction.groupingKey !== undefined
) {
  record.groupingKey = transaction.groupingKey
}
```

(The same `in`-probe pattern also appears in `getProjects.test.ts` and `transformFunctionCallsQueryResult.test.ts` — worth cleaning up there too while at it.)

torztomasz (changes requested):
Solid implementation overall — I verified the Aztec config end-to-end (the configured signature hashes to `0x069d1525` and `path: [0, 0]` decodes to `start`, which the Rollup contract guarantees is unique per epoch), and applied the three migrations to a scratch DB via `prisma migrate deploy`: the `CREATE UNIQUE INDEX CONCURRENTLY` single-statement trick works and both partial indexes come out valid. The keep-earliest upsert semantics are also nicely robust to out-of-order arrival.

Requesting changes for the two inline comments below — mainly the unhandled decode failure in the real-time path, which is externally triggerable. (Separately, `forbid-mixed-changes` will require splitting the migrations into their own PR before this can land.)

packages/backend/src/modules/tracked-txs/utils/getLivenessGroupingKey.ts
```diff
@@ -0,0 +1,47 @@
...
+): string {
+  let value: unknown = decodeFunctionCallInput(config.signature, input)
+
+  for (const index of grouping.path) {
+    assert(Number.isInteger(index) && index >= 0, 'Invalid parameter path')
+    assert(
+      typeof value === 'object' && value !== null,
+      'Parameter path does not exist',
+    )
+    value = Reflect.get(value, index)
```
torztomasz: do we need to use `Reflect` here? wont using value[index] just work?

packages/config/src/internalTypes.ts
```diff
@@ -187,18 +187,40 @@ export interface ProjectScalingRiskView extends ProjectRiskView {
...
+  | {
+      type: 'l2costs'
+      subtype: TrackedTxsConfigSubtype
+    }
+  | {
+      type: 'liveness'
+      subtype: TrackedTxsConfigSubtype
+      /** Liveness grouping is only supported for functionCall queries. */
+      groupBy?: undefined
+    }
```
torztomasz: Honestly I can not understand these types

torztomasz (approved):

torztomasz:
## Main issue: unguarded decode in the batch path

In `transformFunctionCallsQueryResult.ts`, `getLivenessGroupingKey` is called with no try/catch, while `RealTimeLivenessProcessor` deliberately catches and skips malformed calldata. `getLivenessGroupingKey` throws on: undecodable input, a path that doesn't resolve, a non-scalar target, or a key > 255 chars.

- If any Dune-returned row for a grouped config fails decoding, the throw propagates out of `TrackedTxsClient.getData` and stalls the **entire** tracked-txs update — liveness *and* l2costs for all projects in that batch — until someone intervenes, and it will re-throw on every retry since the offending tx stays in range.
- The Dune query filters `tr.success = true`, which makes malformed args unlikely for Aztec's `submit` (a successful call almost certainly had well-formed calldata), so this is a tail risk rather than a live bug. But the failure mode is severe and asymmetric with the real-time path. Recommend either mirroring the real-time skip (with a `logger.warn` — silent skip would hide real problems) or at minimum a comment stating that batch is intentionally fail-loud.
- Related: the real-time `catch {}` swallows *all* four assert reasons silently, including the 255-char limit. A `logger.warn` there would make config mistakes (wrong `path`) observable instead of just producing zero grouped records.

torztomasz:
## Things I verified are correct

- **Mixed `getFullInput` for identical params** (the exact Aztec case: grouped liveness + ungrouped l2costs on the same address/selector): `getFunctionCallQuery` dedups configs by address+selector and decides full input per *address* (`fullInputAddresses`), so there's no duplicate-row or truncated-input hazard — the liveness config's `getFullInput: true` wins for the shared address. The transform then matches both configs against the full-input row and only the liveness result gets a key. The new transform test covers exactly this.
- **ID change → resync**: `createTrackedTxId` now hashes `groupBy`, so Aztec's `stateUpdates` liveness gets a new configuration id, wiping and re-syncing that config's records from `sinceTimestamp`. This is necessary (old records lack keys) and appears intended, but it's worth calling out in the PR description as an operational effect: old records are deleted and a full Dune backfill runs on deploy. Confirm the anomalies/real-time side tolerates the window where the new config has no history.
- **Path traversal safety**: `Reflect.get` with integer-only indices, null/undefined/array rejection, and `String(BigNumber)` producing decimal all check out; indexing "into" a BigNumber dead-ends at `undefined` → proper assert.
- **`decodeFunctionCallInput` cache**: bounded by static config signatures (also now shared with `isFistParameterMatching`), so the unbounded `Map` is fine; nice incidental perf win for the shared-bridge hot path, which previously built a new `Interface` per call.
- **Defense in depth**: the runtime assert in `RealTimeLivenessProcessor.updateConfigurations` re-checks what `topics?: never` and the `getProjects` assert already guarantee — redundant across layers, which is consistent with how this codebase treats class vs. config boundaries.

PR #12353, opened 2026-07-23
fix(discovery): use Blockscout v2 for contract creation lookups

vincfurc:
Blockscout's Etherscan-compatible v1 API for getcontractcreation returns blockNumber/contractCreator/contractFactory/creationBytecode but omits txHash, causing our schema parse to always fail and the caller to fall back to EthereumAddress.ZERO deployer.

Switch getContractDeploymentTx() to Blockscout v2
(/api/v2/addresses/{address}) which returns creation_transaction_hash directly. Drop the defensive 'unsupported: { getContractCreation }' flag on robinhood now that discovery matches UpdateMonitor's output.


vincfurc:
@codex

PR #12380, opened 2026-07-29
DA tracking: shared identity hash + guard test against silent data wipes

torztomasz:
## Problem

The backend keys indexed DA data by a hash of (type, daLayer, inbox, sorted sequencers) — `createDaTrackingId` in `packages/backend/src/config/features/da.ts` — and `ManagedMultiIndexer` wipes all data for configuration ids that disappear. Since stack templates derive daTracking from current discovery state, routine discovery dailies have been silently re-keying configs and wiping indexed history (zeronetwork: 5 identities / 3+ wipes since Apr 2025; abstract: 7 / 6+).

This PR makes any identity disappearance an explicit, reviewed decision instead of a silent side effect.

## Changes

1. **`createDaTrackingId` moved to `@l2beat/shared`** (next to `createTrackedTxId`) — structurally typed so shared doesn't depend on config. Tests pin ids computed with the original backend implementation against real production configs (zeronetwork eras, zksync2's `eth:`-prefixed inbox, sophon appIds, …), proving the move is behavior-preserving. Backend imports the shared function; no behavior change.
2. **DA identity guard test** in `packages/config`: all configuration ids for every project (incl. sovereign projects tracked via a DA layer's `sovereignProjectsTrackingConfig`) are pinned in a committed snapshot `src/snapshots/daTracking/snapshot.json` (186 ids across 166 projects). Two directions, both fail CI:
   - an id in the snapshot missing from `getProjects()` output → error stating the backend **will wipe** that data on deploy;
   - an id in the output missing from the snapshot → snapshot is stale.

   `pnpm snapshots:generate` regenerates it — the snapshot diff in a PR is the explicit sign-off for any identity change.

## Verification

- `packages/shared`: 389 tests green (10 new pinned-hash tests).
- `packages/backend`: typechecks.
- `packages/config`: 23751 tests green (332 new guard tests); negative-tested both directions (deleted a snapshot entry → wipe warning; added a fake one → staleness failure).

Follow-up PR #12379 builds on this: a `zkStackDaTracking` discovery handler that resolves full era histories so discovery updates stop re-keying configs in the first place.


packages/config/src/snapshots/guard.test.ts
```diff
@@ -0,0 +1,77 @@
...
+        it(projectId, () => {
+          const currentIds = new Set(
+            (current[projectId] ?? []).map((e) => e.id),
+          )
+          const missing = identities.filter((e) => !currentIds.has(e.id))
+          if (missing.length > 0) {
+            throw new Error(
+              `${domain.name} identities disappeared for ${projectId}:\n` +
+                missing.map((e) => `- ${e.id} (${e.label})`).join('\n') +
+                `\n${domain.wipeWarning}\n${REGENERATE_HINT}`,
```
adamiak: For removals, please add a specific comment for AI to delegate this to human instead of figuring it out itself, e.g.:

"This error must be addressed by a human. If you're an AI, don't address this error and pass it over to a human".

packages/config/src/snapshots/guard.test.ts
```diff
@@ -0,0 +1,77 @@
...
+    describe('snapshot is up to date', () => {
+      for (const [projectId, identities] of Object.entries(current)) {
+        it(projectId, () => {
+          const snapshotIds = new Set(
+            (snapshot[projectId] ?? []).map((e) => e.id),
+          )
+          const unknown = identities.filter((e) => !snapshotIds.has(e.id))
+          if (unknown.length > 0) {
+            throw new Error(
+              `${domain.name} identities not present in the snapshot for ${projectId}:\n` +
```
adamiak: This information is too cryptic. I would not understand what it means and why it's a problem. Let's add more info, e.g.:"This usually is not a problem and means that you have added a new configuration for data tracking."

adamiak (approved):

packages/config/src/snapshots/guard.test.ts
torztomasz: Done in 49345bbafc — the removal error now ends with: "This error must be addressed by a human: verify on-chain that the removal is intentional and the data loss acceptable, then run 'pnpm snapshots:generate' in packages/config and commit the updated snapshot as the explicit sign-off. If you're an AI, don't address this error yourself - pass it over to a human."

packages/config/src/snapshots/guard.test.ts
torztomasz: Done in 49345bbafc — the message now reads: "New da-tracking identities are not yet in the snapshot for <project> … This is usually not a problem - it means a new data tracking configuration was added for this project (or an existing one was re-keyed; if this error appears together with a 'disappeared' error for the same project, resolve that one first). To register the new identities, run 'pnpm snapshots:generate' in packages/config and commit the updated snapshot."

PR #12401, opened 2026-07-30
Fix TokenRelations Part 2 -  code changes

adamiak:
Orient token relations by role instead of transfer direction

A relation says two token addresses are the same asset bridged by a given
mechanism. That is not directional: users bridge both ways over the same route,
and each transfer is evidence of the same relation. What a lockAndMint pair does
have is an asymmetry — one token is escrowed, the other is its minted
representation — and because the roles are complementary, naming one names both.

The endpoints were stored in the order the sample transfer happened to travel,
so every graph arrow and every role in the Relations tab was a coin flip, and
one route observed in both directions became two rows nothing could tell apart.

Endpoints now go into the A/B slots in lexicographic order, so the primary key
admits exactly one row per pair, and `lockedToken` ('A' or 'B') carries the role.
`InteropTransferClassifier.inferLockedTransferSide` is the single place transfer
flags become a role, applied once at ingestion; nothing downstream opens the
evidence JSON to recover roles. A relation whose flags identify no side keeps
`lockedToken` NULL, which sits outside the primary key so a later observation
resolves it with a plain update.

This release also renames the record fields to match the columns the previous
release introduced, dropping the transitional translation in the repository.
`TokenDbHistory` snapshots are immutable, so the history page keeps reading the
old field names too.


packages/database/src/repositories/TokenRelationRepository.ts
```diff
@@ -46,31 +61,78 @@ export type TokenRelationUpdateable = Omit<
...
+
+// Same order as the table's `CHECK` constraint, which compares the endpoints as
+// `(chain, address)` rows under the `C` collation — byte order, like JavaScript.
+function isOrdered(
+  a: { chain: string; address: string },
+  b: { chain: string; address: string },
+): boolean {
+  if (a.chain !== b.chain) return a.chain < b.chain
+  return a.address <= b.address
+}
```
torztomasz, line 116: I would move it to bottom

packages/token-backend/src/ingestion/TokenRelationIngestion.ts
```diff
@@ -97,24 +103,35 @@ export class TokenRelationIngestion {
...
     }
 
     const existing = await this.tokenDb.tokenRelation.getByPrimaryKeys(
       Array.from(candidates.values(), (candidate) => candidate.route),
     )
+    // An existing relation whose locked endpoint was never identified is
+    // upgraded as soon as an observation identifies one — possible only because
+    // `lockedToken` sits outside the primary key. A known locked endpoint is
+    // never overwritten.
+    const resolutions: TokenRelationRecord[] = []
```
torztomasz, line 116:
```suggestion
    const relations: TokenRelationRecord[] = []
```

torztomasz (approved):

packages/database/src/repositories/TokenRelationRepository.ts
adamiak: Will do in a follow-up PR.

packages/token-backend/src/ingestion/TokenRelationIngestion.ts
adamiak: Will do in a follow-up PR

PR #12410, opened 2026-07-30
Add liveness grouping database support

vilcht:
## Summary

- add nullable `groupingKey` columns to liveness tables
- add partial unique indexes for grouped liveness records
- update the Prisma schema to match the migrations
- persist grouped records while keeping the earliest proof per configuration and grouping key

## Why

This is the database prerequisite for #12349. It is split from the application changes so database migrations and the code that starts producing grouping keys land separately and satisfy the migration-isolation check.

The repository changes are included here because Prisma generation exposes the new nullable columns to database row types; keeping the mapping and repository tests with the schema makes this prerequisite independently deployable and testable.

## Validation

- `pnpm --filter @l2beat/database typecheck`
- `pnpm --filter @l2beat/database test -- --reporter dot` (19 passing, 53 database-backed tests pending locally)
- `pnpm --filter @l2beat/database exec prisma validate --schema prisma/schema.prisma`
- Biome on changed repository files

### DIFF
--- a/packages/database/src/repositories/LivenessRepository.ts
+++ b/packages/database/src/repositories/LivenessRepository.ts
@@ -3,24 +3,37 @@ import { assert, UnixTime } from '@l2beat/shared-pure'
 import type { Insertable, Selectable } from 'kysely'
 import { BaseRepository } from '../BaseRepository'
 import type { Liveness } from '../kysely/generated/types'
+import {
+  insertGroupedKeepingEarliest,
+  splitLivenessRecords,
+} from './utils/livenessGrouping'
 
 export interface LivenessRecord {
   timestamp: UnixTime
   blockNumber: number
   txHash: string
   configurationId: TrackedTxId
+  groupingKey?: string
 }
 
 export function toRecord(row: Selectable<Liveness>): LivenessRecord {
-  return {
-    ...row,
+  const { groupingKey, ...rest } = row
+  const record: LivenessRecord = {
+    ...rest,
     timestamp: UnixTime.fromDate(row.timestamp),
   }
+
+  if (groupingKey !== null) {
+    record.groupingKey = groupingKey
+  }
+
+  return record
 }
 
 export function toRow(record: LivenessRecord): Insertable<Liveness> {
   return {
     ...record,
+    groupingKey: record.groupingKey ?? null,
     timestamp: UnixTime.toDate(record.timestamp),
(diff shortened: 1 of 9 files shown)


packages/database/src/repositories/LivenessRepository.ts
```diff
@@ -3,24 +3,37 @@ import { assert, UnixTime } from '@l2beat/shared-pure'
...
+  const record: LivenessRecord = {
+    ...rest,
     timestamp: UnixTime.fromDate(row.timestamp),
   }
+
+  if (groupingKey !== null) {
+    record.groupingKey = groupingKey
+  }
+
+  return record
```
torztomasz: weird pattern of this record.groupingKey=groupingKey, check other places in repo how we handle this

packages/database/src/repositories/RealTimeLivenessRepository.ts
```diff
@@ -3,28 +3,41 @@ import { UnixTime } from '@l2beat/shared-pure'
...
+  const record: RealTimeLivenessRecord = {
+    ...rest,
     timestamp: UnixTime.fromDate(row.timestamp),
   }
+
+  if (groupingKey !== null) {
+    record.groupingKey = groupingKey
+  }
+
+  return record
```
torztomasz: same

torztomasz (approved):

packages/database/src/repositories/utils/livenessGrouping.ts
```diff
@@ -0,0 +1,88 @@
...
+  table: LivenessTable,
+  rows: Insertable<Liveness | RealTimeLiveness>[],
+): Promise<void> {
+  await db
+    .insertInto(table)
+    .values(rows)
+    .onConflict((cb) =>
+      cb
+        .columns(['configurationId', 'groupingKey'])
+        .where('groupingKey', 'is not', null)
```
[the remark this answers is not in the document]
vilcht: The PostgreSQL conflict behavior is correct, but the migration replay scenario doesn’t apply here. Adding `groupBy` changes the tracked transaction configuration ID, so existing ungrouped rows and new grouped rows have different `(configurationId, txHash)` primary keys. Grouped producers also don’t persist rows without a grouping key. This could be treated as defensive repository hardening, but it isn’t a reachable P1 in the current flow.

PR #12430, opened 2026-07-31
research centralized sequencing and add section / table

sekuba:
close L2B-14200
- [x] append the chart to the decentralized sequencing table that is above, to show that they belong together and are separate from the centralized sequencing below
- [ ] frontend code review

research: during the ai review it claimed that the op stack calc omits the 'Standard clock extension' `CLOCK_EXTENSION` from the cumulative path extension value in State validation → Challenges . the fix replaces 10d with 10d3h for op mainnet

old string:
> The maximum clock extension that a top level claim can get is therefore 10d.

new:
> Along a full-depth path, cumulative extensions can therefore add up to 10d 3h.


lucadonnoh (approved):

sekuba:
merged

PR #12447, opened 2026-08-03
Add STRK20 privacy pool stats & TVS

sdlyy:
## Summary

Adds end-to-end Starknet support for STRK-20 privacy-pool statistics and TVS.

- Introduces Starknet RPC client/event support and a Starknet privacy-flow indexer.
- Adds `starknetBalanceOf` TVS formula support, including Cairo `u256` balance decoding.
- Configures STRK-20 pool tokens, TVS balances, privacy flows, prices, and pool timestamps.
- Extends frontend TVS breakdowns to render the new balance formula with token and pool explorer links.
- Regenerates the TVS JSON schema.

## Reviewer focus

1. **Starknet event extraction**
   - Validate the STRK-20 deposit/withdrawal event selectors and token-address extraction.
   - Confirm pagination/range handling in `StarknetClient.getEvents()` and `StarknetPrivacyFlowIndexer`.

2. **TVS correctness**
   - `starknetBalanceOf` calls the token’s `balance_of(pool)` at the historical Starknet block.
   - Check Cairo `u256` decoding, especially the high limb.
   - Confirm provider failures follow the existing TVS behavior.

3. **Timestamp semantics**
   - Project `addedAt` remains Jun 17.
   - Pool tracking starts at Apr 20 deployment.
   - `strkBTC` starts May 4, when its token contract exists.
   - Privacy token price timestamps match their flow timestamps, so flow indexing cannot request a price before it is collected.

4. **Indexer dependencies**
   - Privacy flow indexing depends on privacy prices and Starknet block timestamps.
   - TVS token values depend on the Starknet amount indexer.

## Validation

- Config typecheck/build
- Backend typecheck and focused TVS/privacy tests
- Frontend typecheck and formula address-extraction test

image
image
image
image

### DIFF
--- a/packages/backend/src/modules/privacy/indexers/StarknetPrivacyFlowIndexer.ts
+++ b/packages/backend/src/modules/privacy/indexers/StarknetPrivacyFlowIndexer.ts
@@ -1,0 +1,247 @@
+import type { Logger } from '@l2beat/backend-tools'
+import type { Database, PrivacyFlowEventRecord } from '@l2beat/database'
+import type { BlockProvider, StarknetClient } from '@l2beat/shared'
+import { assert, UnixTime } from '@l2beat/shared-pure'
+import { Indexer } from '@l2beat/uif'
+import { createPrivacyConfigurationId } from '../../../config/features/privacy'
+import { INDEXER_NAMES } from '../../../tools/uif/indexerIdentity'
+import { ManagedMultiIndexer } from '../../../tools/uif/multi/ManagedMultiIndexer'
+import type {
+  Configuration,
+  ManagedMultiIndexerOptions,
+  TrimRemovalConfiguration,
+  WipeRemovalConfiguration,
+} from '../../../tools/uif/multi/types'
+import type {
+  StarknetPrivacyEvent,
+  StarknetPrivacyFlowIndexerConfig,
+} from '../types'
+import { extractStarknetPrivacyFlow } from '../utils/extractStarknetPrivacyFlow'
+
+interface StarknetPrivacyFlowIndexerDeps
+  extends Omit<
+    ManagedMultiIndexerOptions<StarknetPrivacyFlowIndexerConfig>,
+    'name' | 'logger'
+  > {
+  chain: string
+  blockProvider: BlockProvider
+  starknetClient: StarknetClient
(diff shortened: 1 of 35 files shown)


sdlyy:
@codex review

packages/backend/src/modules/tvs/tools/extractPricesAndAmounts.ts
```diff
@@ -265,13 +267,17 @@ export function createAmountConfig(
         ...formula,
       }
     case 'starknetTotalSupply':
+    case 'starknetBalanceOf':
```
torztomasz, line 270: lets just add it as a separate case

packages/backend/src/modules/privacy/indexers/StarknetPrivacyFlowIndexer.ts
```diff
@@ -0,0 +1,247 @@
...
+    ManagedMultiIndexerOptions<StarknetPrivacyFlowIndexerConfig>,
+    'name' | 'logger'
+  > {
+  chain: string
+  blockProvider: BlockProvider
+  starknetClient: StarknetClient
+  db: Database
+}
+
+export class StarknetPrivacyFlowIndexer extends ManagedMultiIndexer<StarknetPrivacyFlowIndexerConfig> {
```
torztomasz, line 32: I wouldn't create a separate indexer for different config. We can simply handle starknet configs inside the `PrivacyFlowIndexer`

torztomasz (changes requested):

torztomasz (approved):

packages/backend/src/modules/privacy/indexers/StarknetPrivacyFlowIndexer.ts
sdlyy: ok rn we only have two implementations, and their provider/event semantics differ enough that a generic adapter layer would add types and indirection without much immediate payoff. The duplicated code is straightforward and testable.

I’ll do the unification myself when either:
- a third event provider or non-EVM chain is added
- shared pricing/persistence logic starts changing for some reason,
- fixes repeatedly need to be applied to both indexers,
- the two implementations begin drifting behaviorally

PR #12449, opened 2026-08-04
Aztec liveness fix

vilcht:
## Summary

This PR re-enables Aztec liveness tracking and introduces two changes to reduce Dune data consumption:

1. **Project grouping arguments in Dune**

   For grouped function calls, Dune now extracts the configured fixed-width argument into `grouping_value`. The backend no longer needs the complete calldata just to calculate the grouping key.

2. **Scope full-calldata requests by function selector**

   Consumers that genuinely require complete calldata, such as SHARP and shared bridge tracking, now match both the contract address and function selector. This prevents unrelated calls to the same contract from exporting their full input.

## Implementation

- Return the 4-byte function selector as `input` for ordinary function-call tracking.
- Return the selected argument bytes as `grouping_value` when grouping is configured.
- Decode `grouping_value` in the backend using its ABI type.
- Preserve full calldata for SHARP and shared bridge consumers.
- Compile and normalize the function-call query plan once, then share it between SQL generation and result transformation.
- Reject conflicting grouping projections for the same address and selector.
- Re-enable Aztec liveness tracking.

## Impact

For the tested Aztec transaction, the exported calldata-related payload decreased from approximately 26,756 bytes to 36 bytes:

- 4-byte function selector
- 32-byte grouping value

This represents approximately a 99.87% reduction in the calldata-related payload for that row. Other result columns and Dune execution overhead are unchanged.


packages/config/src/processing/getProjects.test.ts
```diff
@@ -556,23 +556,22 @@ describe('getProjects', () => {
...
+      const liveness = stateUpdates?.find(
+        (config) => config.type === 'liveness',
+      )
+
+      assert(liveness !== undefined)
+      expect(liveness.groupBy).toEqual({
+        type: 'functionCallParameter',
+        path: [0, 0],
+      })
+    })
```
torztomasz: remove

packages/backend/src/modules/tracked-txs/TrackedTxsClient.ts
```diff
@@ -175,13 +176,28 @@ function combineCalls(
...
-      ...c.properties.params,
-      getFullInput: hasLivenessGrouping(c.properties),
+    ...functionCallsConfig.map((config) => ({
+      address: config.properties.params.address,
+      selector: config.properties.params.selector,
+      inputBytes: hasLivenessGrouping(config.properties)
+        ? (getFunctionCallParameterPrefix(
+            config.properties.params.signature,
+            config.properties.groupBy.path,
+          ) ?? ('full' as const))
```
torztomasz: let's not fallback to full, this will cost us a lot

packages/backend/src/modules/tracked-txs/TrackedTxsClient.ts
```diff
@@ -175,13 +176,28 @@ function combineCalls(
...
+        : SELECTOR_BYTES,
```
torztomasz, line 189: why ask for SELECTOR_BYTES if we do not have grouping?

packages/backend/src/modules/tracked-txs/TrackedTxsClient.ts
torztomasz: I think this should not return undefined

torztomasz (changes requested):

packages/backend/src/modules/tracked-txs/TrackedTxsClient.ts
```diff
@@ -175,13 +176,28 @@ function combineCalls(
...
-  // TODO: unique
   return [
-    ...functionCallsConfig.map((c) => ({
-      ...c.properties.params,
-      getFullInput: hasLivenessGrouping(c.properties),
+    ...functionCallsConfig.map((config) => ({
+      address: config.properties.params.address,
+      selector: config.properties.params.selector,
+      inputBytes: hasLivenessGrouping(config.properties)
+        ? (getFunctionCallParameterPrefix(
```
torztomasz: **correctness:** `combineCalls` calls `getFunctionCallParameterPrefix` unguarded at query-build time, so one misconfigured `groupBy` path or unparsable signature aborts the entire tracked-txs sync for all projects instead of degrading per-row.

**Failure scenario:** a project ships `groupBy` path `[0, 12]` against a signature whose tuple has fewer members (`getProjects.ts:273-286` validates only formula and topics, never path-vs-signature). `locateScalar` asserts `'Parameter path does not exist'` inside `combineCalls`, `getFunctionCalls` rejects, the `Promise.all` in `getData` (`TrackedTxsClient.ts:79`) rejects, and `TrackedTxsIndexer.multiUpdate` retries forever — liveness, l2costs AND transfers stop syncing for every project. Pre-PR, the same bad path only threw per-row inside the try/catch at `transformFunctionCallsQueryResult.ts:111` (warn + drop one row). The only mitigation is a unit test in another package (the `functionCallParameter.test.ts` loop), not a code guard; wrapping the prefix computation to warn and fall back to `'full'` would restore the old blast radius.

packages/backend/src/modules/tracked-txs/utils/functionCallParameter.ts
```diff
@@ -0,0 +1,212 @@
...
+): string | undefined {
+  const { inputs, selector } = parseSignature(signature)
+  assert(
+    input.toLowerCase().startsWith(selector),
+    'Input does not match the function selector',
+  )
+
+  const data = Buffer.from(input.slice(2 + SELECTOR_BYTES * 2), 'hex')
+  const target = locateScalar(inputs, path, (containerStart, slot) => {
+    return containerStart + readWord(data, slot)
```
torztomasz, line 63: **correctness:** the Dune prefix is sized assuming canonical ABI layout (`canonicalTupleOffset`) but extraction follows the actual pointer in calldata, so a legal non-canonical encoding permanently and silently drops the liveness datapoint — and the batch pipeline diverges from the realtime one.

**Failure scenario:** a prover submits `submitEpochRootProof` via a custom encoder with tuple offset `0x40` instead of `0x20` (legal ABI, accepted on-chain). Dune only fetched 68 bytes, `extractFixedWidthScalar` resolves a position past the prefix, `readBytes` throws `'Unexpected end of function call input'`, and `transformFunctionCallsQueryResult.ts:111` catches it and returns `[]` — the record is dropped, and resync re-fetches the same 68 bytes so the loss is unrecoverable, producing a false Aztec liveness gap/anomaly. `RealTimeLivenessProcessor.ts:135` passes full `tx.data` and succeeds on the same tx, so the realtime and Dune liveness stores disagree. An ungrouped-liveness fallback shape already exists (`model.ts:48` `groupingKey?: string`; `transformFunctionCallsQueryResult.ts:124` emits it), so the catch could return `[{...common, type: 'liveness'}]` instead of dropping.

packages/backend/src/modules/tracked-txs/TrackedTxsClient.ts
torztomasz: **observability:** the `?? ('full' as const)` fallback silently reverts a grouped config to full-calldata export with no log, metric, or guard — the exact Dune billing regression this PR exists to fix.

**Failure scenario:** a future config edit puts the grouped parameter behind a dynamic sibling (`canonicalTupleOffset` returns `undefined` the moment any preceding parameter is dynamic); `getFunctionCallParameterPrefix` returns `undefined`, the `??` swallows it, and Dune starts exporting full `submitEpochRootProof` proof blobs again (~53 credits/day scale). `combineCalls` has no logger parameter, `DuneQueryService` only emits `apiExportCostCredits` as a discarded info-log field with no threshold, and the new test discards non-Aztec prefixes — so the first signal is the Dune bill. At minimum, log a warning when the prefix computation falls back to `'full'`.

torztomasz (commented):
Automated multi-agent review of this PR. Pipeline: 8 finder angles produced ~35 candidates; after dedup, 6 contested candidates went to adversarial verifiers (3 plausible, 3 refuted — the Trino INTEGER/BIGINT typing concern, the selector/signature-mismatch concern, and the SELECT DISTINCT prefix-collapse concern were all refuted with cited guards or pre-existing behavior), and 5 cleanup candidates were verified directly against the source.

**Key takeaway:** the prefix math itself is correct — independently re-verified against real Aztec calldata (prefix 68, extraction from the truncated prefix matches full-input decoding, and selector `0x069d1525` matches the signature). The three surviving correctness/operational findings all concern failure handling around the new machinery, not the machinery itself:

1. an unguarded config-time throw that widens blast radius from one bad row to the whole tracked-txs sync,
2. permanent (unrecoverable-by-resync) datapoint drops for legal non-canonical ABI encodings, where an ungrouped-liveness fallback already exists in the type model,
3. a silent full-calldata cost fallback with no log/metric — the exact Dune billing regression this PR exists to fix.

The remaining five inline comments are cleanup/maintainability. Details on each finding are inline.

packages/backend/src/modules/tracked-txs/utils/functionCallParameter.test.ts
```diff
@@ -0,0 +1,120 @@
...
+        if (
+          project.id === 'aztecnetwork' &&
+          config.subtype === 'stateUpdates'
+        ) {
+          aztecPrefix = prefix
+        }
+      }
+    }
+
+    expect(aztecPrefix).toEqual(68)
```
torztomasz, line 47: **test-coverage:** the cross-project prefix test computes a prefix for every configured grouping path but asserts only the Aztec constant, so any other project silently falling back to full-input passes green.

**Failure scenario:** a second project adds a grouped liveness config whose valid path yields `undefined` (dynamic sibling → silent `'full'` billing, see the `TrackedTxsClient.ts:187` comment): the loop assigns `prefix` at line 34, discards it for non-Aztec projects, and `expect(aztecPrefix).toEqual(68)` still passes — no test failure while production requests full calldata. Asserting inside the loop that every configured path yields a defined prefix (or is explicitly allow-listed as `'full'`) would catch both the billing fallback and shrink the blast radius of the unguarded-throw finding.

packages/backend/src/modules/tracked-txs/utils/functionCallParameter.ts
```diff
@@ -0,0 +1,212 @@
...
+
+const SELECTOR_BYTES = 4
+const WORD_BYTES = 32
+
+type ParsedSignature = {
+  inputs: readonly utils.ParamType[]
+  selector: string
+}
+
+const signatures = new Map<`function ${string}`, ParsedSignature>()
```
torztomasz, line 12: **simplification:** the new module-level `signatures` cache duplicates the signature-to-`ethers.Interface` cache that `decodeFunctionCallInput.ts` already maintains, so grouped signatures are parsed and retained twice.

Every grouped config signature that reaches `getFunctionCallParameter` hits both caches: `parseSignature` builds and stores an `Interface` in `signatures` while the full-decode fallback path builds a second identical `Interface` in `decodeFunctionCallInput`'s `interfaces` map. Two module-level caches keyed identically must both be reasoned about for growth/consistency. Exporting a `getInterface(signature)` helper from `decodeFunctionCallInput.ts` and deriving inputs/selector from it in `parseSignature` removes the duplicate.

packages/backend/src/modules/tracked-txs/TrackedTxsClient.ts
```diff
@@ -175,13 +176,28 @@ function combineCalls(
...
+      address: config.address,
+      selector: config.selector,
+      inputBytes: 'full' as const,
     })),
-    ...sharpSubmissionsConfig.map((c) => ({ ...c, getFullInput: true })),
-    ...sharedBridgesConfig.map((c) => ({ ...c, getFullInput: true })),
   ]
 }
+
+const SELECTOR_BYTES = 4
```
torztomasz: **simplification:** `SELECTOR_BYTES = 4` is declared here and again in `functionCallParameter.ts`, with the same `4` hard-coded a third time in `getFunctionCallQuery.ts` (the `>= 4` assert in `mergeCalls` and `substr(tr.input, 1, 4)` in the SQL).

The three sites are semantically coupled: prefixes are computed as `SELECTOR_BYTES + position + WORD_BYTES`, the client passes its own `SELECTOR_BYTES` as non-grouped `inputBytes`, and the SQL builder independently asserts `>= 4`. A change to any one silently desynchronises the SQL truncation from the decoder's expectation. The module already has `utils/const.ts` for shared constants — define `SELECTOR_BYTES` once there (or export it from `functionCallParameter.ts`) and import it everywhere.

packages/backend/src/modules/tracked-txs/utils/functionCallParameter.ts
```diff
@@ -0,0 +1,212 @@
...
+export function getFunctionCallParameter(
+  signature: `function ${string}`,
+  input: string,
+  path: readonly [number, ...number[]],
+): string {
+  const extracted = extractFixedWidthScalar(signature, input, path)
+  if (extracted !== undefined) return extracted
+
+  let value: unknown = decodeFunctionCallInput(signature, input)
+  for (const index of path) {
```
torztomasz: **simplification:** the full-decoder fallback in `getFunctionCallParameter` re-implements path validation (`'Invalid parameter path'` / `'Parameter path does not exist'`) that `locateScalar` in the same file already performs.

The same two invariants exist as two independently worded loops (lines 46-53 and inside `locateScalar`), with different reachability: fixed-width paths hit the `locateScalar` copy, dynamic paths hit the loop copy. Any future change to path rules (depth cap, array support) must be made in both places or validation becomes inconsistent between the prefix and fallback paths — which would produce different grouping keys for the same transaction. Hoist a single `assertValidPath(path)` called once at the top of `getFunctionCallParameter`.

packages/backend/src/modules/tracked-txs/utils/sql/getFunctionCallQuery.test.ts
```diff
@@ -103,75 +101,39 @@ describe(getFunctionCallQuery.name, () => {
...
+            inputBytes: 'full' as const,
+          },
+        ],
+        FROM,
+        TO,
+      ),
+    )
   })
 
   it('handles empty helper tables', () => {
```
torztomasz, line 139: **test-maintainability:** `EXPECTED_SQL` was extracted to deduplicate two tests, but this 'handles empty helper tables' test still inlines a ~65-line near-copy of the same query (at line 142, just below) differing only in the `allowed_calls` VALUES row.

Any future change to the query body (new column, whitespace tweak) must be hand-applied to two nearly identical 65-line string literals, and drift between them is unreadable in diffs. Turning the constant into a function parameterised by the VALUES row lets this test assert only on `(CAST(NULL AS varbinary), CAST(NULL AS varbinary), CAST(NULL AS bigint))`.

packages/backend/src/modules/tracked-txs/TrackedTxsClient.ts
vilcht: The selector is still required even without grouping because function-call queries are batched and the transformer matches each result back to its configuration using `(address, selector)`. Address alone is not unique; Aztec currently has both `propose` and `submitEpochRootProof` tracked. These four bytes are identification metadata, not grouping calldata.

packages/backend/src/modules/tracked-txs/TrackedTxsClient.ts
vilcht: We have a [test](https://github.com/l2beat/l2beat/blob/edec9ed578efcc193917d08ca50856e190402b38/packages/backend/src/modules/tracked-txs/utils/functionCallParameter.test.ts#L23-L48) that loads every config and runs the prefix function against it, it's caught at the CI level.

packages/backend/src/modules/tracked-txs/utils/functionCallParameter.ts
vilcht: Falling back to ungrouped liveness would be incorrect because it would bypass epoch deduplication and create false liveness events. The current code logs and drops only that liveness record instead of deriving wrong data; l2costs is still preserved. I also tested 30,789 real Aztec submissions and all used the expected layout.

packages/backend/src/modules/tracked-txs/utils/functionCallParameter.test.ts
vilcht: This is resolved by removing the `undefined` return. Every configured path now either produces a prefix or throws, so this test fails for any unsupported configuration.

packages/backend/src/modules/tracked-txs/utils/functionCallParameter.ts
vilcht: These paths validate different representations: `locateScalar` walks the ABI type structure, while the fallback walks ethers’ decoded value. Shared helper would just add more abstarction and make it harder to reason about.

packages/backend/src/modules/tracked-txs/utils/functionCallParameter.ts
vilcht: Instead of sharing the caches, I removed the full-decoding fallback. Unsupported paths are already rejected when building the Dune query, so valid grouped calls could never use that fallback. Grouped signatures now use only one cache.

torztomasz (approved):

PR #12456, opened 2026-08-04
Rewrite db-restore scripts in TypeScript

torztomasz:
## Description

Rewrites `db-restore.sh` and `table-restore.sh` as TypeScript scripts run via `tsx`, sharing a `common.ts` (the shell versions duplicated their helper functions).

- Exposed as `pnpm db:restore <FEATURE> [SINCE]` and `pnpm db:restore-table <TABLE>...` from `packages/database`; the backend `restore-table` alias now delegates to the latter.
- All DB access still goes through `psql`/`pg_dump`/`pg_restore` for exact behavior parity (auth, sslmode, `.pgpass` behave identically). The bash `DO $$` blocks became plain query-then-execute logic, and argv-array spawning replaces the shell quoting gymnastics with small `quoteIdent`/`quoteLiteral` helpers.
- `.env` is loaded via `dotenv`, so missing env vars now fail with a clear message instead of the silent-empty-var behavior in `table-restore.sh`.
- Behavior kept: SINCE validation before wiping local data (connection errors distinguished from invalid timestamps), `InteropConfig` full-copy override, remote column list on both sides of `\copy`, and sequence syncing after row-by-row copies.
- `packages/database/tsconfig.json` now typechecks `scripts/` too; `tsconfig.build.json` overrides back to `src`-only so the `dist/` layout is unchanged.

## Testing

- `pnpm typecheck`, `pnpm build`, `pnpm lint`, `pnpm format` pass in `packages/database`.
- Smoke-tested the usage/unknown-feature/no-args paths (each exits 1 before touching any database). The restore paths were not run against a live DB.


mateuszradomski:
@codex, review

mateuszradomski:
@codex, review

mateuszradomski (approved):

PR #12462, opened 2026-08-04
Allow to select data time range in the interop backoffice

sdlyy:
@codex review

Part of L2B-14596

### DIFF
--- a/packages/backend/src/modules/interop/engine/dashboard/trpc/transferDataRange.ts
+++ b/packages/backend/src/modules/interop/engine/dashboard/trpc/transferDataRange.ts
@@ -1,0 +1,45 @@
+import type { Database, InteropTransferTimeRange } from '@l2beat/database'
+import { UnixTime } from '@l2beat/shared-pure'
+import { v } from '@l2beat/validate'
+import { TRPCError } from '@trpc/server'
+
+export const INTEROP_TRANSFER_DATA_RANGES = [
+  'last24h',
+  'lastPromoted',
+  'all',
+] as const
+
+export const InteropTransferDataRange = v.enum(INTEROP_TRANSFER_DATA_RANGES)
+
+export type InteropTransferDataRange =
+  (typeof INTEROP_TRANSFER_DATA_RANGES)[number]
+
+export const InteropTransferDataRangeRequest = v
+  .object({
+    range: InteropTransferDataRange.optional(),
+  })
+  .optional()
+
+export async function resolveInteropTransferTimeRange(
+  db: Database,
+  range: InteropTransferDataRange | undefined,
+): Promise<InteropTransferTimeRange | undefined> {
+  switch (range ?? 'last24h') {
+    case 'all':
+      return undefined
+    case 'last24h': {
+      const to = UnixTime.now()
+      return { from: to - UnixTime.DAY, to }
+    }
+    case 'lastPromoted': {
+      const to = await db.interopAggregateStatus.getLatestPromotedTimestamp()
+      if (to === undefined) {
+        throw new TRPCError({
+          code: 'NOT_FOUND',
+          message: 'No promoted aggregate snapshot is available.',
+        })
+      }
+      return { from: to - UnixTime.DAY, to }
+    }
+  }
+}
--- a/packages/backend/src/modules/interop/engine/dashboard/trpc/transferDataRange.ts
+++ b/packages/backend/src/modules/interop/engine/dashboard/trpc/transferDataRange.ts
@@ -14,17 +14,15 @@ export const InteropTransferDataRange = v.enum(INTEROP_TRANSFER_DATA_RANGES)
 export type InteropTransferDataRange =
   (typeof INTEROP_TRANSFER_DATA_RANGES)[number]
 
-export const InteropTransferDataRangeRequest = v
-  .object({
-    range: InteropTransferDataRange.optional(),
-  })
-  .optional()
+export const InteropTransferDataRangeRequest = v.object({
+  range: InteropTransferDataRange,
+})
 
 export async function resolveInteropTransferTimeRange(
   db: Database,
-  range: InteropTransferDataRange | undefined,
+  range: InteropTransferDataRange,
 ): Promise<InteropTransferTimeRange | undefined> {
-  switch (range ?? 'last24h') {
+  switch (range) {
     case 'all':
       return undefined
     case 'last24h': {
--- a/packages/backoffice/src/pages/interop/TransferDataRangeSelect.tsx
+++ b/packages/backoffice/src/pages/interop/TransferDataRangeSelect.tsx
@@ -1,0 +1,47 @@
+import { Label } from '~/components/core/Label'
+import {
+  Select,
+  SelectContent,
+  SelectItem,
+  SelectTrigger,
+  SelectValue,
+} from '~/components/core/Select'
+import type { InteropTransferDataRange } from './transferDataRange'
+
+interface TransferDataRangeSelectProps {
+  value: InteropTransferDataRange
+  onValueChange: (value: InteropTransferDataRange) => void
+  disabled?: boolean
+}
+
+export function TransferDataRangeSelect({
+  value,
+  onValueChange,
+  disabled = false,
+}: TransferDataRangeSelectProps) {
+  return (
+    <div className="flex items-center gap-2">
+      <Label htmlFor="interop-transfer-data-range" className="sr-only">
+        Transfer data range
+      </Label>
+      <Select
+        value={value}
+        onValueChange={(nextValue) =>
+          onValueChange(nextValue as InteropTransferDataRange)
+        }
+        disabled={disabled}
+      >
+        <SelectTrigger id="interop-transfer-data-range" size="sm">
+          <SelectValue />
+        </SelectTrigger>
+        <SelectContent>
+          <SelectItem value="last24h">Last 24 hours</SelectItem>
+          <SelectItem value="lastPromoted">
+            Last promoted aggregate (24h)
+          </SelectItem>
+          <SelectItem value="all">All retained data</SelectItem>
+        </SelectContent>
+      </Select>
+    </div>
+  )
+}
--- a/packages/backoffice/src/pages/interop/TransferDataRangeSelect.tsx
+++ b/packages/backoffice/src/pages/interop/TransferDataRangeSelect.tsx
@@ -6,7 +6,10 @@ import {
   SelectTrigger,
   SelectValue,
 } from '~/components/core/Select'
-import type { InteropTransferDataRange } from './transferDataRange'
+import {
+  type InteropTransferDataRange,
+  parseInteropTransferDataRange,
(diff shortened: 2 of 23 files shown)


packages/backoffice/src/pages/interop/transferDataRange.ts
```diff
@@ -0,0 +1,20 @@
+export const INTEROP_TRANSFER_DATA_RANGES = [
+  'last24h',
+  'lastPromoted',
+  'all',
+] as const
+
+export type InteropTransferDataRange =
+  (typeof INTEROP_TRANSFER_DATA_RANGES)[number]
```
torztomasz: This enum + type is a character-for-character copy of the backend's `transferDataRange.ts`, with nothing tying them together — if the backend adds `last7d`, this copy silently keeps parsing it out of URLs as `last24h`.

The backoffice already solves exactly this via `BackendRouterInputs` (see `transfers/types.ts`). Keep the runtime array (needed for URL parsing), but derive the type and pin the array to it:

```ts
export type InteropTransferDataRange = NonNullable<
  NonNullable<BackendRouterInputs['interop']['transfers']['stats']>['range']
>

export const INTEROP_TRANSFER_DATA_RANGES = [
  'last24h', 'lastPromoted', 'all',
] as const satisfies readonly InteropTransferDataRange[]
```

Then a renamed/removed backend variant is a type error here instead of a silent behavior change. Related: `'last24h'` as the default is also encoded twice (here and `range ?? 'last24h'` in the backend resolver) — see the comment on the request schema for how to delete the backend copy.

packages/backoffice/src/pages/interop/activity/SuspiciousTransfersPage.tsx
```diff
@@ -16,6 +22,9 @@ import type {
 
 export function SuspiciousTransfersPage() {
   const trpc = useBackendTrpc()
+  const [range, setRange] = useState<InteropTransferDataRange>(
```
torztomasz: Four pages now share the same range control but use two different state models: TransfersPage/TransferDetailsPage keep it in the URL (`?range=`), this page and MissingTokensPage use `useState`. So the selection survives reload and is linkable on two pages, and silently resets on the other two — that reads as incidental (only the transfers flow needed URL propagation for the details link), not a deliberate product distinction.

One small hook unifies all four and deletes the duplicated wiring:

```ts
export function useTransferDataRange() {
  const [searchParams, setSearchParams] = useSearchParams()
  const range = parseInteropTransferDataRange(searchParams.get('range'))
  const setRange = (next: InteropTransferDataRange) => {
    const params = new URLSearchParams(searchParams)
    params.set('range', next)
    setSearchParams(params)
  }
  return [range, setRange] as const
}
```

It also makes the correct merge-not-replace `setSearchParams` behavior the only path (see comment on TransfersPage).

packages/backoffice/src/pages/interop/missing-tokens/MissingTokensPage.tsx
```diff
@@ -1,27 +1,36 @@
...
+  type InteropTransferDataRange,
+} from '../transferDataRange'
 import { MissingTokenStatusBadge } from './MissingTokenStatusBadge'
 import { MissingTokenStatusGuide } from './MissingTokenStatusGuide'
 import { MissingTokensTable } from './table/MissingTokensTable'
 import type { ChainMetadata, MissingTokenRow } from './types'
 
 export function MissingTokensPage() {
   const trpc = useBackendTrpc()
+  const [range, setRange] = useState<InteropTransferDataRange>(
```
torztomasz: Same as SuspiciousTransfersPage — this `useState` gives the range different persistence behavior than the transfers pages (resets on reload, not linkable). A shared `useTransferDataRange()` hook (see comment there) replaces this state + default-import wiring on all four pages.

torztomasz (commented):
Code-quality pass. The overall design is right (single backend resolver, explicit NOT_FOUND instead of a silent unbounded fallback, targeted tests) — comments below are about deleting duplication and unifying state handling, all behavior-preserving.

packages/backoffice/src/pages/interop/transfers/TransfersPage.tsx
```diff
@@ -30,15 +35,24 @@ export function TransfersPage() {
...
-              disabled={isFetching}
-            >
-              <RefreshCwIcon className={isFetching ? 'animate-spin' : ''} />
-              Refresh
-            </Button>
+            <>
+              <TransferDataRangeSelect
+                value={range}
+                onValueChange={(nextRange) =>
+                  setSearchParams({ range: nextRange })
```
torztomasz: `setSearchParams({ range: nextRange })` **replaces** all search params instead of merging — TransferDetailsPage does this correctly by copying `searchParams` first. Harmless today only because this page has no other params, but it's a trap for the next one added. The shared hook proposed on SuspiciousTransfersPage makes the merge behavior the only path.

packages/backend/src/modules/interop/engine/dashboard/trpc/transferDataRange.ts
```diff
@@ -0,0 +1,45 @@
...
+export const InteropTransferDataRange = v.enum(INTEROP_TRANSFER_DATA_RANGES)
+
+export type InteropTransferDataRange =
+  (typeof INTEROP_TRANSFER_DATA_RANGES)[number]
+
+export const InteropTransferDataRangeRequest = v
+  .object({
+    range: InteropTransferDataRange.optional(),
+  })
+  .optional()
```
torztomasz: Double optionality: an optional object containing an optional field gives four input shapes (`undefined`, `{}`, `{range: undefined}`, `{range: x}`) that collapse to two meanings, and forces `input?.range` at every router call site plus the `?? 'last24h'` fallback in the resolver. But every real client now always sends `{ range }` — the frontend parse already guarantees a value.

Make it `v.object({ range: InteropTransferDataRange })`, non-optional. Call sites become `input.range`, the resolver drops its `| undefined` param, and the `?? 'last24h'` disappears — which also deletes the second copy of the default (the frontend already owns it). If you want to keep omitted-input compat for ad-hoc callers, keep exactly one optional level, not two.

packages/database/src/repositories/InteropTransferRepository.ts
```diff
@@ -462,6 +475,7 @@ export class InteropTransferRepository extends BaseRepository {
   async getValueMismatchTransfers(
     valueDifferencePercentThreshold: number,
     minimumSideValueUsdThreshold = 0,
+    timeRange?: InteropTransferTimeRange,
```
torztomasz: This is now the third trailing optional positional param — the next addition forces callers to pass `undefined` placeholders. Fold `minimumSideValueUsdThreshold` and `timeRange` into an options object, like the neighboring `getByType` already does.

packages/backoffice/src/pages/interop/TransferDataRangeSelect.tsx
```diff
@@ -0,0 +1,47 @@
...
+}: TransferDataRangeSelectProps) {
+  return (
+    <div className="flex items-center gap-2">
+      <Label htmlFor="interop-transfer-data-range" className="sr-only">
+        Transfer data range
+      </Label>
+      <Select
+        value={value}
+        onValueChange={(nextValue) =>
+          onValueChange(nextValue as InteropTransferDataRange)
```
torztomasz: `parseInteropTransferDataRange` is one import away — `onValueChange(parseInteropTransferDataRange(nextValue))` and the cast goes away.

packages/backoffice/src/pages/interop/transfers/table/columns.tsx
```diff
@@ -9,73 +10,78 @@ import {
...
 const columnHelper = createColumnHelper<TransferStatsRow>()
 
-export const transferStatsColumns: TableOptions<TransferStatsRow>['columns'] = [
-  columnHelper.accessor('plugin', {
-    header: 'Plugin',
-    meta: {
-      csvHeader: 'Plugin',
-      filter: {
-        kind: 'select',
+export function transferStatsColumns(
```
torztomasz, line 13: nit: 142 lines of factory churn for one URL param — only the `count` column's link uses `range`. Not blocking (factory + `useMemo` is a fine pattern), but worth knowing the whole rewrite exists for one link; if we ever adopt TanStack's table `meta` for context injection, this file is the first candidate to shrink back to a static array.

packages/backoffice/src/pages/interop/transferDataRange.ts
sdlyy: Done. The frontend range type now derives from BackendRouterInputs, and the runtime array is checked with satisfies. The backend input is now required, so the default lives only in the frontend URL parser.

packages/backoffice/src/pages/interop/activity/SuspiciousTransfersPage.tsx
sdlyy: Done. Added useTransferDataRange() and moved this page to the URL-backed range state. It merges existing search params and is shared by all range controls.

packages/backoffice/src/pages/interop/missing-tokens/MissingTokensPage.tsx
sdlyy: Done. Missing Tokens now uses the shared URL-backed range hook, so the selection survives reloads and is linkable.

packages/backoffice/src/pages/interop/transfers/TransfersPage.tsx
sdlyy: Done. TransfersPage now uses the shared hook, whose setter copies existing search params before setting range.

packages/backend/src/modules/interop/engine/dashboard/trpc/transferDataRange.ts
sdlyy: Done. The request object and range are required; router call sites use input.range and the resolver no longer contains a fallback default.

packages/database/src/repositories/InteropTransferRepository.ts
sdlyy: Done. minimumSideValueUsdThreshold and timeRange are now in one optional options object.

packages/backoffice/src/pages/interop/TransferDataRangeSelect.tsx
sdlyy: Done. The select now uses parseInteropTransferDataRange directly, removing the cast.

packages/backoffice/src/pages/interop/transfers/table/columns.tsx
sdlyy: Acknowledged. I kept the existing factory: adding TanStack meta just to avoid it would introduce broader table plumbing for one link. The current factory and useMemo stay contained; this is a good candidate if we later add shared table context.

sdlyy:
@codex review

torztomasz (approved):

PR #12475, opened 2026-08-05
 add relayer stats for privacy projects

vilcht:
This resolves only Privacy Pools and tornadocash. Railgun is an entirely different pipeline and will be done in a separate pr

### DIFF
--- a/packages/backend/src/modules/privacy/indexers/PrivacyRelayerActivityIndexer.ts
+++ b/packages/backend/src/modules/privacy/indexers/PrivacyRelayerActivityIndexer.ts
@@ -1,7 +1,7 @@
 import type { Logger } from '@l2beat/backend-tools'
 import type { Database, PrivacyRelayerActivityRecord } from '@l2beat/database'
 import type { BlockProvider, LogsProvider } from '@l2beat/shared'
-import { assert, UnixTime } from '@l2beat/shared-pure'
+import { UnixTime } from '@l2beat/shared-pure'
 import { Indexer } from '@l2beat/uif'
 import { createPrivacyConfigurationId } from '../../../config/features/privacy'
 import { INDEXER_NAMES } from '../../../tools/uif/indexerIdentity'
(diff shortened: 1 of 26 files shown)


packages/backend/src/modules/privacy/indexers/PrivacyRelayerActivityIndexer.ts
```diff
@@ -0,0 +1,203 @@
...
+  }
+
+  private async fetchRecords(
+    configurations: Configuration<PrivacyRelayerActivityIndexerConfig>[],
+    from: number,
+    to: number,
+  ): Promise<PrivacyRelayerActivityRecord[]> {
+    if (configurations.length === 0) return []
+
+    const { blockFrom, blockTo } = await resolvePrivacyBlockRange(
```
sdlyy: We extracted the individual log-processing helpers, but this indexer and `PrivacyFlowIndexer` still repeat the full orchestration: resolve the range, build filters, fetch logs, resolve timestamps, build the configuration map, and match logs.

Would it make sense for `privacyLogIndexerUtils` to expose one cohesive operation returning flat `{ log, timestamp, configuration }[]` matches? Each indexer would then own only its extraction and persistence logic.

I would still keep the two indexers and their cursor/failure state separate—this is only about sharing the acquisition pipeline, not combining their lifecycle.

packages/backend/src/modules/privacy/utils/extractPrivacyRelayerActivity.ts
```diff
@@ -0,0 +1,84 @@
...
+    return toRelayerActivity(
+      String(parsedLog.args._relayer),
+      String(parsedLog.args._recipient),
+    )
+  },
+}
+
+const tornadoCashWithdrawal: RelayerExtractorDefinition = {
+  event: tornadoCashInterface.getEventTopic('Withdrawal'),
+  extract: (log) => {
```
sdlyy, line 40: The Tornado Cash metric described in L2B-14413 kept only pool `Withdrawal` logs whose transaction called `TornadoRouter.withdraw(...)`, then intersected the extracted relayer with the current `registeredRelayers` set and excluded self-relays.

This extractor only applies the nonzero/non-self checks, so direct withdrawals and nonregistered addresses are included. That makes `Active Relayers 30D` a different—and potentially larger—population than the researched metric.

Can we confirm with research whether production intentionally tracks all observed relayer fields? Otherwise, this needs the Router and registry filtering before these addresses are counted as active relayers.

sdlyy (commented):
One metric-definition concern.

packages/backend/src/modules/privacy/utils/extractPrivacyRelayerActivity.ts
vilcht: I confirmed with Sergey that the current behaviour is correct and the linear desc is outdated

sdlyy (approved):

PR #12483, opened 2026-08-06
Fix lighter on robinhood risks after validating desert verifier

sergeyshemyakov:
Resolves L2B-14627


packages/config/src/projects/lighter-robinhood/lighter-robinhood.ts
vincfurc: outdated?

vincfurc: outdated it can be added

vincfurc: outdated

vincfurc: outdated

vincfurc (approved):

PR #12492, opened 2026-08-07
Zksync starknet gov profiles

sergeyshemyakov:
Resolves L2B-14601
Resolves L2B-14602


sekuba (changes requested):
Image

mention guardians here (and maybe the team ms, but guardians are main check against SC capturing the protocol)

sekuba:
this is called LegalVetoPeriod (see the diagram)
image
maybe also put 'on L1' somewhere there

sekuba:
i would call this on-chain payload just like in arbitrum
image
'commit to a spec hash' sounds like confusing detail, and the govops path exists in other systems like arbitrum too.

the important message is that protocol upgrades and token upgrades are completely auditable

sekuba:
above was zksyc, this is starknet:
image
don't we need to know the members list at least (without the mapping)? the link found on the security council is dead: https://governance.starknet.io/learn/security_council

sekuba:
image
clarify what is actually onchain-enforced (see arbitrum for an example)

same for the minor path that comes after

sekuba:
image
suggest to remove this and include the info in the appropriate lines as i said above

imo we should try to establish a 'standard table' and only diverge if there is something really special, like maybe the guardians in zksync. but creating 'Offchain enforcement' while lots of other projects have the same mechanism will confuse people and makes it harder to compare. in general i would suggest one more pass over the two tables and see if you can normalize the line headers to the ones used in op, arb, taiko

sekuba:
image
maybe call this 'permissioned' instead of manual?

sekuba (approved):

PR #12534, opened 2026-08-11
Skip tokens the backend cannot price

mateuszradomski:
`AbstractToken.coingeckoId` is nullable, both in the schema (`packages/database/src/kysely/generated/types.ts:13`) and in `AbstractTokenRepository`'s record type, and `abstractTokens.getAllWithDeployedTokens` returns every abstract token including unpriced ones. The validator here declared `coingeckoId: v.string()`, so a single `null` would reject the whole batch response, `fetchTokens()` would throw, and every TVL request would return 500.

The field is now parsed as nullable and records that cannot be priced are dropped, so one unpriceable token costs us that token instead of the entire list. `BackendToken.coingeckoId` stays a plain `string`, so nothing downstream has to think about the nullable case.

Parsing moved out of `fetchTokens` into an exported `parseTokens`, which gives the behaviour a test seam that does not need `fetch` stubbed. Four tests cover the flattening, the skip, the absent icon, and a malformed body.

## Status

Latent, not currently firing: **0 of 5,964** live abstract tokens have a null `coingeckoId` today, which is why the panel works. It breaks the moment an unpriced token is added upstream.

Verified against the real payload with a null injected into a token that has deployed tokens:

```
unmodified payload parses to 9644 tokens
nulling "ICCM" which has 1 deployed tokens
with a null id it parses to 9643 tokens
```

One token dropped rather than the whole list rejected.

### DIFF
--- a/packages/l2b/src/implementations/discovery-ui/tokenBackend.test.ts
+++ b/packages/l2b/src/implementations/discovery-ui/tokenBackend.test.ts
@@ -1,0 +1,85 @@
+import { expect } from 'earl'
+import { parseTokens } from './tokenBackend'
+
+interface TestAbstractToken {
+  symbol: string
+  iconUrl: string | null
+  coingeckoId: string | null
+  deployedTokens: {
+    chain: string
+    address: string
+    symbol: string
+    decimals: number
+  }[]
+}
+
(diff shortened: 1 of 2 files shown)


mateuszradomski:
Folded into discoui-tvl directly as 01ff0c0a38; this was opened by mistake.

PR #12542, opened 2026-08-11
Show minting bridges on interop token deployments

radinat:
Add a Minters column to "Onchain deployments" table


radinat:
Implemented

PR #12544, opened 2026-08-11
Add railgun relayer stats and backend

vilcht:


packages/backend/src/modules/privacy/indexers/PrivacyRelayerSampleIndexer.ts
```diff
@@ -0,0 +1,164 @@
...
+      logger,
+    )
+  }
+
+  override async multiUpdate(
+    _from: number,
+    to: number,
+    configurations: Configuration<PrivacyRelayerSampleIndexerConfig>[],
+  ) {
+    // Observations measure the current network state - a sample can only be
```
sdlyy: This source is explicitly non-replayable, but `ManagedMultiIndexer` gives it historical cursor, trim, and wipe semantics. Here we skip historical days while still advancing the cursor, and configuration reconciliation can delete samples that cannot be reconstructed.

Could we model this as a scheduled sampler instead, using the existing clock/task-queue pattern? It should sample the current day, skip an existing healthy sample, and retry on a later tick. That would make the lifecycle match what the source can actually provide.

packages/backend/src/modules/privacy/railgun/RailgunBroadcasterProvider.ts
```diff
@@ -0,0 +1,138 @@
...
+  }
+
+  async observe(
+    options: RailgunObserveOptions,
+  ): Promise<Map<number, RailgunObservationResult>> {
+    const chainIds = Array.from(new Set(options.chainIds))
+    const { durationMs } = options
+    if (chainIds.length === 0) return new Map()
+
+    // js-waku is ESM-only, hence the dynamic import from this CJS build
```
sdlyy, line 50: Could we check whether the official `WakuBroadcasterClient` can replace this low-level implementation and, if so, put it behind a narrow adapter?

The official client already owns peer discovery, fee parsing, authorization, compatibility policy, and caching. Reimplementing the network constants, topics, parsing, and eligibility rules means upstream changes can silently make this metric incorrect.

This still needs to be confirmed with a small integration spike: whether its public API exposes the broadcaster enumeration we need, how `trustedFeeSigner` should be configured, and whether the ESM package works in the backend production build. If any of those block adoption, we should document that explicitly before maintaining the custom implementation.

packages/database/src/repositories/PrivacyRelayerSampleRepository.ts
```diff
@@ -0,0 +1,133 @@
...
+  }
+
+  async getAverageRelayerCount(
+    projectId: string,
+    fromInclusive: UnixTime,
+    toExclusive: UnixTime,
+  ): Promise<number | undefined> {
+    const row = await this.db
+      .selectFrom('PrivacyRelayerSample')
+      .select(sql<number | null>`AVG("relayerCount")`.as('averageCount'))
```
sdlyy: `AVG` only considers rows that exist, so one daily sample in this range can be presented as a 30-day average. Missing observation days disappear from the denominator.

The intended semantics should first be confirmed against the research: does “30D average” mean an average across every expected calendar day, or only across healthy observations that happened to be collected? The current implementation silently chooses the latter.

Once confirmed, could this return a coverage-aware result such as `{ average, observedDays, expectedDays }`? The caller could then expose partial coverage or hide the statistic until the required window is available.

packages/frontend/src/server/features/privacy/getPrivacyProjectDetails.ts
```diff
@@ -273,29 +278,58 @@ export async function getPrivacyProjectDetails(
...
   to: UnixTime,
-): Promise<number | undefined> {
-  if (!project.privacyInfo.relayerTracking?.length) {
+): Promise<PrivacyRelayerStat | undefined> {
+  const tracking = project.privacyInfo.relayerTracking
+  if (!tracking) {
     return undefined
   }
 
+  const kind = relayerStatKind(tracking)
```
sdlyy: `tracking.type` is already the exhaustive domain discriminator, but here we translate it into `kind` and then use `kind` to select the repository. This creates a second discriminator, and anything other than `avgDailyRelayers` implicitly falls into the onchain query.

Could we switch exhaustively on `tracking.type` once and return the complete `{ kind, value }` result from each case? That keeps the source type, database query, and response kind coupled, and a future tracking type cannot accidentally reach the wrong repository.

packages/backend/src/config/features/privacy.ts
```diff
@@ -48,8 +56,16 @@ export async function getPrivacyConfig(
     return false
   }
 
+  const chainNames = new Map(
```
sdlyy: `makeConfig` already loads the chain configurations and passes them into other feature-config builders. Could we pass those into `getPrivacyConfig` as well instead of issuing another `ProjectService` query solely to build this lookup?

This would also let us filter or assert defined chain IDs at that boundary and pass `ReadonlyMap<number, string>` into `toRelayerSampleConfig`, rather than allowing `undefined` as a map key even though the Railgun lookup requires a concrete chain ID.

packages/backend/src/modules/privacy/railgun/RailgunBroadcasterProvider.ts
vilcht: I think we shouldnt replace my impl,

- it peer-depends on the full `@railgun-community/wallet` engine - we'd be pulling the whole wallet stack just to count advertisers
- it runs a `patch-package` postinstall that patches its `@waku` deps - pnpm blocks that unless we allowlist it, otherwise it silently runs unpatched
- its enumeration reads a wallet-policy fee cache ("who can relay for me now"), not "who advertised in this window", and it exposes no message counters

packages/database/src/repositories/PrivacyRelayerSampleRepository.ts
vilcht: Confirmed with the reaserch and implemented. 30D counted only from succesful observation with minimal 7D to display the avg

packages/config/src/projects/railgun/railgun.ts
```diff
@@ -225,6 +225,14 @@ export const railgun: BaseProject = {
   },
   privacyInfo: {
     tokens: privacyTokens,
+    relayerTracking: {
+      type: 'railgunWaku',
+      chainId: 1,
+      // Sampling observes the current network state, so there is no history
+      // to backfill before the first deployment of the sampler.
+      // TODO(L2B-14560): bump to the actual deployment date before merging.
+      sinceTimestamp: UnixTime(1785974400), // 2026-08-06T00:00:00Z
```
sdlyy: Optional: This is still marked as a pre-merge TODO, and `2026-08-06` predates the sampler’s production deployment. Because these Waku observations cannot be backfilled, retaining this value makes the configuration claim tracking started before the first possible observation.

Could we replace it with the actual planned production start date and remove the TODO before merging?

packages/backend/src/modules/privacy/PrivacyRelayerSampler.ts
```diff
@@ -0,0 +1,131 @@
...
+    this.taskQueue = new TaskQueue(
+      async () => this.sampleToday(),
+      this.logger.for('taskQueue'),
+      { metricsId: PrivacyRelayerSampler.name },
+    )
+  }
+
+  start() {
+    this.logger.info('Started')
+    this.$.clock.onNewHour(() => this.taskQueue.addIfEmpty())
```
sdlyy: Optional: `Clock` is injected and emits the scheduling timestamp, but the callback discards it and `sampleToday()` derives the day from global `UnixTime.now()`. This splits time control across two sources and forces the tests to use the real current date, making midnight-boundary behavior nondeterministic.

Could we pass the clock timestamp into the queued task and derive the day from it, using `clock.getLastHour()` for the startup run? This is not blocking, but it would complete the intended clock abstraction and make the tests deterministic.

sdlyy (commented):
Two optional cleanup comments.

packages/config/src/projects/railgun/railgun.ts
vilcht: Updated in 9d5e35bfc9: tracking now starts on the planned production deployment date, 2026-08-31, and the pre-merge TODO is removed.

packages/backend/src/modules/privacy/PrivacyRelayerSampler.ts
vilcht: Fixed in 196d70ba1b. The queued task now carries the `Clock` timestamp, startup uses `clock.getLastHour()`, and sampling derives its day from that timestamp. The sampler tests now use a fixed timestamp instead of wall-clock time.

sdlyy (approved):

PR #12545, opened 2026-08-11
Add privacy relayer activity migration

vilcht:
## Summary

- Add the `PrivacyRelayerActivity` table.
- Add indexes for project relayer queries and configuration time-range queries.

## Why

This migration is isolated so the database change can land before the application code that starts using the table.

## Validation

- Applied the complete migration chain successfully to a fresh PostgreSQL 17 database.
- Verified the privacy relayer repositories against that database.

### DIFF
--- a/packages/database/prisma/migrations/20260805120000_add_privacy_relayer_activity/migration.sql
+++ b/packages/database/prisma/migrations/20260805120000_add_privacy_relayer_activity/migration.sql
@@ -8,7 +8,6 @@ CREATE TABLE "PrivacyRelayerActivity" (
     "txHash" VARCHAR(66) NOT NULL,
     "logIndex" INTEGER NOT NULL,
     "relayerAddress" VARCHAR(42) NOT NULL,
-    "recipientAddress" VARCHAR(42) NOT NULL,
 
     CONSTRAINT "PrivacyRelayerActivity_pkey" PRIMARY KEY ("configurationId", "txHash", "logIndex")
 );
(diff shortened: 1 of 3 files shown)


packages/database/prisma/migrations/20260805120000_add_privacy_relayer_activity/migration.sql
```diff
@@ -0,0 +1,20 @@
...
+CREATE TABLE "PrivacyRelayerActivity" (
+    "configurationId" CHAR(12) NOT NULL,
+    "projectId" VARCHAR(255) NOT NULL,
+    "chain" VARCHAR(32) NOT NULL,
+    "timestamp" TIMESTAMP(6) NOT NULL,
+    "blockNumber" INTEGER NOT NULL,
+    "txHash" VARCHAR(66) NOT NULL,
+    "logIndex" INTEGER NOT NULL,
+    "relayerAddress" VARCHAR(42) NOT NULL,
+    "recipientAddress" VARCHAR(42) NOT NULL,
```
sdlyy: The stacked consumer only counts distinct relayer addresses by project and time range. `recipientAddress` is used to reject self-relays before insertion, but is then persisted together with chain and block metadata even though none of those fields participate in the product query.

Could we confirm whether retaining a permanent raw-event history is an explicit audit or future-chart requirement? If not, I would keep only the minimal projection needed for the rolling statistic and idempotency—configuration/project, timestamp, relayer, and stable event identity—and define bounded retention slightly beyond the query window. Otherwise this becomes an unbounded event table for one 30-day statistic.

vilcht: I agree that we can remove the recipient. When it comes to to the rest i think we should keep them. This is the same shape as the rest of privacy tables and we need tx+log to not duplicate rows. Nothing else in privacy i think bounds to 30days, so i'd keep it unbounded.

sdlyy (approved):

PR #12546, opened 2026-08-11
Add privacy relayer sample migration

vilcht:
## Summary

- Add the `PrivacyRelayerSample` table.
- Add an index for project timestamp queries.

## Why

This migration is isolated so the database change can land before the Railgun sampling code that starts using the table.

## Validation

- Applied the complete migration chain successfully to a fresh PostgreSQL 17 database.
- Verified the privacy relayer repositories against that database.


packages/database/prisma/migrations/20260807120000_add_privacy_relayer_sample/migration.sql
```diff
@@ -0,0 +1,13 @@
+-- CreateTable
+CREATE TABLE "PrivacyRelayerSample" (
+    "configurationId" CHAR(12) NOT NULL,
+    "projectId" VARCHAR(255) NOT NULL,
+    "chain" VARCHAR(32) NOT NULL,
+    "timestamp" TIMESTAMP(6) NOT NULL,
+    "relayerCount" INTEGER NOT NULL,
```
sdlyy, line 7: The stacked implementation calculates `messagesReceived`, `messagesParsed`, and `messagesAccepted`, but this table persists only the final count. Once the process exits, a legitimate zero, a transport failure, and all messages being rejected by policy become indistinguishable.

Since this migration is still unmerged, could we store the observation state as part of the row—for example the observation window, counters, a small status field, nullable `relayerCount`, and the policy version? Public aggregation could then include only healthy observations while retaining enough evidence to explain missing or zero results.

vilcht: Indexer already asserts messagesReceived and messagesPardes are both >0. So legitimate zero is distinguishable from the transport failiure. I don't think any policy version exists, and observation duration is a constant. Will add the rest of the columns for observability.

packages/database/prisma/migrations/20260807120000_add_privacy_relayer_sample/migration.sql
```diff
@@ -0,0 +1,16 @@
+-- CreateTable
+CREATE TABLE "PrivacyRelayerSample" (
```
sdlyy, line 2: This migration adds `PrivacyRelayerSample`, but the table is not registered under `FEATURES.privacy` in `packages/database/scripts/db-restore/db-restore.ts`.

As a result, `pnpm db:restore privacy` will neither clear an existing local sample table nor restore its production rows, leaving the restored privacy dataset inconsistent. Could we add `PrivacyRelayerSample` to that list alongside `PrivacyRelayerActivity`?

sdlyy (commented):
One database-restore integration concern.

packages/database/prisma/migrations/20260807120000_add_privacy_relayer_sample/migration.sql
vilcht: Fixed in b6a5096fcc — `PrivacyRelayerSample` is now included in `FEATURES.privacy` alongside the other privacy tables.

sdlyy (approved):

PR #12551, opened 2026-08-12
Update CCIP RMN discovery and add CCIP 2.0 interop support

lucadonnoh:
## Summary

- refresh CCIP discovery for the Ethereum RMN 2.1 migration, the new curse-governance stack, route curses, router lanes, and the GHO bridge-limit increase
- add the RMN discovery template and update ARM proxy, legacy CommitStore, and OffRamp descriptions for the new RMN semantics
- normalize CCIPHome OCR3 oracle identities as correlated p2pId-sorted tuples so node, offchain key, peer ID, and shared-secret encryption positions stay aligned
- preserve the genuine Aptos oracle rotation while reducing Avalanche and Monad to their version/config-digest updates
- retain v1.6 and v2.0 ramp addresses simultaneously and capture the new CCIPMessageSent and ExecutionStateChanged event ABIs
- decode and integrity-check packed MessageV1 token transfers, with a sealed real-transaction fixture covering burn/mint, lock/mint, and data-only messaging

## Review notes

- `diffHistory.md` has a large one-time config/verification section because all existing OCR3 identities are migrated to the canonical representation; the watched section is substantially smaller and future dailies should not contain permutation-only array churn
- CCIP 2.0 is deployed alongside v1.6, so the backend config intentionally keeps separate `onRampV2` and `offRampV2` fields instead of overwriting the existing ramp addresses
- most of the backend byte-size increase is the compressed, sealed eight-transaction replay input (`ccip-v2-inputs.json.gz`)

## Validation

- `pnpm --filter @l2beat/discovery test -- --grep "Ocr3OffchainConfig|CCIPOcr3Config"`
- `pnpm --filter @l2beat/discovery lint`
- `pnpm --filter @l2beat/discovery typecheck`
- `pnpm --filter @l2beat/config lint`
- `pnpm --filter @l2beat/config typecheck`
- `pnpm --filter @l2beat/l2b typecheck`
- two successful local-worktree CCIP discovery runs
- `pnpm --filter @l2beat/backend typecheck`
- live and sealed `ccip-v2` example: 15/15 expectations passing
- sealed interop snapshot replay suite: 4 passing


packages/backend/src/modules/interop/plugins/ccip/ccip.v2.test.ts
```diff
@@ -0,0 +1,111 @@
+import { EthereumAddress } from '@l2beat/shared-pure'
```
sekuba: tests intended?

lucadonnoh: removed

sekuba (approved):

PR #12553, opened 2026-08-12
Fix Hyperlane native warp route classification

lucadonnoh:
## Summary

- load Hyperlane aggregated warp-route configuration into a map keyed by chain and router address
- classify `EvmHypNative` routes from registry data instead of transaction-wide ERC-20 transfer heuristics
- retain previously observed routers so historical transfers remain classifiable
- add the Arbitrum to Base native ETH transaction pair that previously misclassified WETH as minted

## Testing

- `pnpm typecheck`
- Biome check on changed files
- all examples loading the Hyperlane config: `hyperlane`, `hyperlane-merkly`, `hyperlane-eco`, `hyperlane-mix`, `hwr`, `hwr-native-eth-utb`, `hwr-xerc20-lockbox`, `hwr-onesided`, and `hwr-batch`

### DIFF
--- a/packages/backend/src/modules/interop/examples/examples.jsonc
+++ b/packages/backend/src/modules/interop/examples/examples.jsonc
@@ -5211,6 +5211,33 @@
       ]
     }
   },
+  "hwr-native-eth-utb": {
+    // UTB unwraps Arbitrum WETH, bridges native ETH through HypNative,
+    // then wraps the released native ETH into Base WETH. The Hyperlane leg is
+    // native ETH on both sides and does not mint or burn either bridged token.
+    "loadConfigs": ["hyperlane"],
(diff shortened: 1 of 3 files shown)


packages/backend/src/modules/interop/plugins/hyperlane.config.test.ts
```diff
@@ -0,0 +1,55 @@
+import { expect } from 'earl'
```
sekuba: are we doing tests in interop plugins?

lucadonnoh: not really but i wanted to do a manual test. removed it

sekuba (approved):

PR #12567, opened 2026-08-13
add defi page

vilcht:


packages/frontend/src/server/features/defi/project/getDefiProjectEntry.ts
```diff
@@ -0,0 +1,251 @@
...
+
+  const defaultChartRange = optionToRange('1y')
+  const icon = manifest.getUrl(`/icons/${project.slug}.png`)
+  const [
+    contractUtils,
+    allProjectsWithContracts,
+    zkCatalogProjects,
+    dependencyProjects,
+  ] = await Promise.all([
+    getContractUtils(),
```
torztomasz: **Dead links in "Used in" badges.** `getContractUtils()` builds the "Used in" hrefs via `getProjectUrl`, which has no `defiInfo` branch (and `getContractUtils` doesn't select `defiInfo`), so DeFi projects fall through to `/scaling/projects/<slug>`.

liquityv2 and chainlink share 86 contract/permission addresses (e.g. `eth:0x536218f9E9Eb48863970252233c8F271f554C2d0` `Chainlink_rETH_ETH`), so on `/defi/projects/liquityv2` the "Used in: Chainlink" badge links to `/scaling/projects/chainlink#…` → 404 (no `scalingInfo`), and vice-versa on the chainlink page.

Fix: add a `defiInfo → /defi/projects/<slug>` branch to `getProjectUrl.ts` and add `'defiInfo'` to the `optional` select in `getContractUtils.ts`.

packages/frontend/src/server/features/defi/getDefiSummaryEntries.ts
```diff
@@ -0,0 +1,165 @@
...
+  if (env.MOCK) {
+    return new Map(
+      trackedIds.map((projectId) => [
+        projectId,
+        Math.random() * 1_000_000_000 + 1,
+      ]),
+    )
+  }
+
+  const tokenValues = await getDb().tvsTokenValue.getLastNonZeroValueByProjects(
```
torztomasz: Table TVL sums each token's **last non-zero** `valueForProject` (`getLastNonZeroValueByProjects` filters `value > 0` per `(project, token)`), while the project-page chart (`getTvsChartByProjects` → `getSummedByProjectForRanges`) sums all tokens at each timestamp. The two can disagree: if a collateral pool is fully drained, the newest rETH row is `0` (zero rows are persisted by `TokenValueIndexer`) and gets excluded, so the table keeps adding the stale last non-zero value while the chart's latest point already excludes it — and the table mixes tokens from different timestamps (see `TokenValueRepository.test.ts:487-509`).

Same convention as `getPrivacySummaryEntries`, so pre-existing, but if table and chart should agree, summing the latest per-project timestamp via `getSummedByProjectForRanges` (as `get7dTvsBreakdown` does) would match.

packages/frontend/src/server/features/defi/project/getDefiProjectEntry.ts
```diff
@@ -0,0 +1,251 @@
...
+
+  const isUnderReview = !!project.statuses.reviewStatus
+  const permissionsSection = getPermissionsSection(
+    {
+      id: project.id,
+      permissions: project.permissions,
+      isUnderReview,
+    },
+    contractUtils,
+    EMPTY_PROJECTS_CHANGE_REPORT,
```
torztomasz: All three DeFi projects are discovery-driven (`discoveryInfo` set, `discovered.json`/`diffHistory.md` present), so they *do* appear in `getProjectsChangeReport()` (it filters only by `discoveryInfo`). Passing `EMPTY_PROJECTS_CHANGE_REPORT` here and at L110 means implementation-change / high-severity / became-verified markers never render on DeFi contracts & permissions — on the page whose purpose is contract review. DA and interop entry builders fetch and pass the real report; suggest doing the same here.

packages/frontend/src/server/features/defi/resolveDefiDependencies.ts
```diff
@@ -0,0 +1,51 @@
...
+        description: dependency.description,
+        reviewed: false,
+      }
+    }
+
+    const project = projectsById.get(dependency.projectId)
+    if (!project) {
+      return {
+        name: dependency.projectId,
+        icon: TOKEN_PLACEHOLDER_ICON_URL,
```
torztomasz, line 74: A `tracked` dependency whose `projectId` doesn't resolve is silently rendered as a **reviewed** dependency named by the raw id with the token placeholder icon. `ps.getProjects({ ids })` silently omits unknown ids and nothing in config validates `tracked.projectId`, so a typo like `ProjectId('chainlnk')` ships with all tests green (the unit test even asserts this fallback).

Suggest an existence check in `packages/config` `getProjects.test.ts` (as done for DA bridge / zkCatalog references), or at least `reviewed: false` here.

packages/frontend/src/server/features/defi/resolveDefiDependencies.ts
```diff
@@ -0,0 +1,51 @@
...
+        description: dependency.description,
+        reviewed: true,
+      }
+    }
+
+    return {
+      name: project.name,
+      icon: manifest.getUrl(`/icons/${project.slug}.png`),
+      description: dependency.description,
+      ...(project.isDefi ? { href: `/defi/projects/${project.slug}` } : {}),
```
torztomasz: Links are emitted only for `isDefi` projects, so a tracked dependency that is a bridge / DA layer / privacy project renders as plain text even though it has a project page (the test "shows tracked non-DeFi projects without a DeFi link" locks this in). The config type says only "An L2BEAT project this project depends on" with no DeFi restriction. Not hit today (only tracked dep is chainlink), but `getProjectUrl` already resolves cross-vertical URLs if the callers select the other vertical info fields — is the DeFi-only link intentional?

packages/frontend/src/server/features/defi/project/getDefiProjectEntry.ts
```diff
@@ -0,0 +1,251 @@
...
+    href: `https://disco.l2beat.com/ui/p/${project.id}`,
+    images: {
+      desktop: manifest.getUrl('/images/disco-ui-desktop.png'),
+      mobile: manifest.getUrl('/images/disco-ui-mobile.png'),
+    },
+  }
+
+  const sections: ProjectDetailsSection[] = []
+
+  if (project.display.detailedDescription) {
```
torztomasz: `display.references` are only rendered inside this section, which is pushed only when `detailedDescription` is truthy. They're independent optionals on `ProjectDisplay`, so a DeFi config with `references` but no `detailedDescription` (or an empty markdown file) silently drops the references — this is the only consumer of `display.references` in the frontend. Not triggered by the three current projects, but easy to trip later.

packages/frontend/src/components/projects/sections/ExternalDependenciesSection.tsx
```diff
@@ -0,0 +1,62 @@
...
+      </ProjectSection>
+    )
+  }
+
+  return (
+    <ProjectSection {...sectionProps}>
+      <ul className="flex flex-col gap-4">
+        {dependencies.map((dependency) => (
+          <li
+            key={`${dependency.name}:${dependency.icon}`}
```
torztomasz: `${name}:${icon}` isn't unique per entry — two tracked entries for the same `projectId` with different descriptions (e.g. one Chainlink entry per feed), or two not-tracked entries sharing name+icon, collide. `ProjectExternalDependency` has no uniqueness constraint. Same pattern in `DefiDependenciesCell.tsx:33,39`. Using the array index (or including `description`) avoids the duplicate-key warning.

packages/frontend/src/components/projects/sections/defi/DefiTvlSection.tsx
```diff
@@ -0,0 +1,61 @@
...
+import type { ChartRange } from '~/utils/range/range'
+import { ProjectSection } from '../ProjectSection'
+import type { ProjectSectionProps } from '../types'
+
+export interface DefiTvlSectionProps extends ProjectSectionProps {
+  defaultRange: ChartRange
+  project: ChartProject
+}
+
+export function DefiTvlSection({
```
torztomasz: This is a verbatim copy of `PrivacyTvlSection` differing only in the tRPC path and the range-controls component, and `defi.tvlChart` is a byte-identical second registration of `privacy.tvlChart` (both `procedure.input(TvsChartByProjectsParams).query(getTvsChartByProjects)`). The PR already unified the server function and `TvsValueChart` — collapsing one layer further into a single `TvsValueSection` (range controls as a prop; `TvsChartRangeControls` vs `PrivacyFlowsChartRangeControls` differ only by the `180d` option) and one procedure (e.g. `tvs.chartByProjects`) would remove two section types, two `ProjectDetails` cases and one router.

packages/frontend/src/server/features/defi/project/getDefiProjectEntry.ts
```diff
@@ -0,0 +1,251 @@
...
+    warnings: {
+      yellow: project.statuses.yellowWarning,
+      red: project.statuses.redWarning,
+      emergency: project.statuses.emergencyWarning,
+    },
+    sections,
+  }
+}
+
+async function getTrackedDependencyProjects(
```
torztomasz: This (+ the inline `Map` construction around L164) re-implements `getDependencyProjectsById` from `getDefiSummaryEntries.ts:72` — same Set-dedupe of tracked ids, same `ps.getProjects({ ids, optional: ['defiInfo'] })`, same `{ name, slug, isDefi }` mapping — with a different return shape. Both feed `resolveDefiDependencies`, so a single exported `getDefiDependencyProjectsById(dependencies, seed?)` next to it would serve both. Also `return await` at L247 is redundant.

packages/frontend/src/server/features/defi/project/getDefiProjectEntry.ts
```diff
@@ -0,0 +1,251 @@
...
+  const icon = manifest.getUrl(`/icons/${project.slug}.png`)
+  const [
+    contractUtils,
+    allProjectsWithContracts,
+    zkCatalogProjects,
+    dependencyProjects,
+  ] = await Promise.all([
+    getContractUtils(),
+    ps.getProjects({
+      select: ['contracts'],
```
torztomasz: Both `select: ['contracts']` (all projects' contracts, multi-MB JSON parse, unmemoized) and `select: ['zkCatalogInfo']` are fetched on every per-slug cache miss, but they're consumed only by `getProgramHashes` inside `getContractsSection`, which returns `[]` unless `project.contracts.programHashes` is set — and no DeFi project has `programHashes`. Suggest gating on `project.contracts?.programHashes?.length` (privacy has the same pattern, so a shared fix pays twice).

torztomasz (changes requested):

packages/frontend/src/components/projects/sections/tvs/TvsValueSection.tsx
```diff
@@ -44,13 +47,18 @@ export function PrivacyTvlSection({
     [chartData],
   )
 
+  const RangeControls =
+    rangeControls === 'tvs'
+      ? TvsChartRangeControls
+      : PrivacyFlowsChartRangeControls
```
torztomasz, line 53: just use TvsChartRangeControls and remove PrivacyFlowsChartRangeControls

packages/frontend/src/components/projects/sections/ExternalDependenciesSection.tsx
```diff
@@ -0,0 +1,62 @@
...
+      </ProjectSection>
+    )
+  }
+
+  return (
+    <ProjectSection {...sectionProps}>
+      <ul className="flex flex-col gap-4">
+        {dependencies.map((dependency, index) => (
+          <li
+            key={`${index}:${dependency.name}:${dependency.icon}`}
```
torztomasz: do we need icon in key?

packages/frontend/src/components/projects/sections/types.ts
```diff
@@ -268,9 +271,9 @@ interface ProjectDetailsUpdatesSection {
   props: ProjectDetailsProps<UpdatesSectionProps>
 }
 
-interface ProjectDetailsPrivacyTvlSection {
-  type: 'PrivacyTvlSection'
-  props: ProjectDetailsProps<PrivacyTvlSectionProps>
+interface ProjectDetailsTvsValueSection {
+  type: 'TvsValueSection'
```
torztomasz, line 275: How is it different from L2TvsSection and ZkCatalogTvsSection? Maybe we can merge more

torztomasz (approved):

PR #12618, opened 2026-08-19
Guard DA tracking ranges with a freeze recipe in the error

torztomasz:
## Summary

- Extend the DA tracking snapshot to record each configuration's range (`since`/`until`) alongside its `id` and `label`, so range moves are visible in review.
- Add a shared `snapshots/compare.ts` holding every guard failure message: removal, range change, addition and duplicate-id, each ending with an AI guard-rail telling agents to hand the error to a human.
- Fail the guard when a range moves: the id hashes identity fields only, so a changed `since`/`until` keeps the id but makes the backend re-sync (a full wipe and re-index without `trimData`). The message prints old and new range plus a resolution recipe.
- Disappeared-identity errors carry a step-by-step freeze recipe (turn the old entry into literals, close it with `untilBlock`, add the new entry starting where the old one ended) instead of inviting a blind snapshot regeneration — the #12464 failure mode.
- Document the guard, the freeze recipe and the range-change recipe in `docs/da-tracking.md`.
- Regenerate `daTracking/snapshot.json` with the new range fields (format only — same identities).

A coverage-gap check was part of an earlier revision and was dropped: a project can legitimately stop posting and resume later, so holes between consecutive entries are left to human review.

## Testing

- `pnpm test` in `packages/config` — unit tests for `findRangeChanges` and the message builders; guard passes against the regenerated snapshot (23 951 tests green).
- Red-test verified: tampering a snapshot `since` fails with the old → new range and the full recipe; deleting an id fails with the freeze recipe.

Fixes L2B-14746


torztomasz:
@codex review

adamiak (approved):

PR #12619, opened 2026-08-19
DA block indexer trims data instead of wiping on range edits

torztomasz:
## Motivation

`sinceBlock` / `untilBlock` are excluded from the daTracking configuration id so ranges can be edited in place, but `DaIndexer` did not implement `trimData`, so `mergeConfigurations` turned every range edit into a full wipe of that configuration's history (and a re-index from scratch). Closing a historical entry with `untilBlock` should be safe.

Fixes L2B-14741. Part of the [daTracking config store](https://linear.app/l2beat/project/datracking-config-store-59c4257b58d0) project.

## Changes

- `DaIndexer.trimData` (ethereum / celestia / avail): maps the trimmed block range to timestamps through a new `DaBlobProvider.getBlockTimestamp` and deletes every hour it touches. Raising `sinceBlock` deletes only the records before the new start, setting or lowering `untilBlock` only the records after the new end. Lowering `sinceBlock` still wipes and re-indexes (no gaps).
- **Boundary bucket:** records are hourly and the hour straddling an edge holds blobs from both sides, so it is deleted together with the out-of-range hours. At most an hour of in-range data is lost per edited edge; nothing is ever counted twice (keeping it would double count if indexing resumed inside that hour). Documented on `trimData` and in `docs/da-tracking.md`.
- `DataAvailabilityRepository.deleteByConfigInTimeRange` — same shape as the liveness / l2 costs repositories.
- `DaBlobProvider.getBlockTimestamp` on the three providers (ethereum: rpc `getBlock`, celestia: existing client method, avail: block-time formula).
- `mergeConfigurations` and the rest of `tools/uif` are untouched.

## Testing

- `DaIndexer.test.ts`: `trimData` deletes every touched hour (both edge hours included); range edits through `initialize()` with the real `mergeConfigurations` — setting `untilBlock` trims, raising `sinceBlock` trims, lowering `sinceBlock` wipes.
- `DataAvailabilityRepository.test.ts`: DB test that only the configuration's rows inside the range go, other configurations untouched.
- Provider tests for `getBlockTimestamp` on ethereum, celestia, avail.
- Suites run locally: backend 1345 (incl. DB-backed), database 800 (real Postgres), shared 402; typecheck / lint clean.


mateuszradomski:
@codex, review

torztomasz:
@codex review

mateuszradomski (approved):

PR #12627, opened 2026-08-19
Move daTracking into explicit per-project arrays; templates stop deriving it

torztomasz:
Stacked on #12618 (range-aware snapshot guard) — merge that first, then retarget this to `main`.

## Summary

Every scaling project's DA tracking history is now an explicit array in its own `.ts` (the shape mint/ink/zksync2 already used), and the stack templates no longer produce daTracking behind the project's back.

- **Templates** (`opStack`, `orbitStack`, `zkStack`, `agglayer`): the runtime daTracking derivation and the `celestiaDa` / `availDa` vars are gone; `nonTemplateDaTracking` is renamed to `daTracking` and passed through as-is. Typecheck enforces it (the old vars are excess properties now). `zkStack` also drops `usesEthereumBlobs`, which only fed the removed derivation.
- **Helpers** `getOpStackDaTracking(discovery, { sinceBlock, untilBlock? })`, `getOrbitStackDaTracking(...)`, `getZkStackDaTracking(...)` are exported next to each template and called explicitly as the open (last) array element. Identity fields (inbox/batcher, SequencerInbox/batchPosters, ValidatorTimelock/validatorsVTL) stay discovery-driven inside the helper — that is the rotation detection (rotation → id changes → guard fails with the freeze recipe). Ranges are literal arguments only, never read from discovery.
- **Migration** of all 86 tracked projects, generated from today's runtime output:
  - 28 template-derived ethereum entries (23 OP Stack, 4 Orbit, 1 ZK Stack) → helper call with today's `sinceBlock` as a literal
  - 18 `celestiaDa` + 2 `availDa` → literal entries (comments preserved)
  - 9 existing explicit arrays whose open entry was byte-identical to the helper output (aevo, base, bobanetwork, celo, hashkey, ink, mantle, optimism, zeronetwork) → helper call, so they get rotation detection too
  - discovery-driven `sinceBlock`s (aevo, celo, mantle, soon) pinned to their current literal values
  - the rest (`nonTemplateDaTracking` → `daTracking` rename only)
- **Guard recipes + docs**: `identities.ts` recipe text and `docs/da-tracking.md` describe the new shape ("Declaring daTracking in a project") and add the new-project checklist (adding daTracking is an explicit step; accepted gap — no test can force it).

Left as literals on purpose (open entry differs from what the helper would derive today — a human call, not a migration): **soon** (SystemConfig.sequencerInbox is zero in discovery), **swell** (tracked batcher `0x3780…` ≠ discovery batcherHash `0xa008…` — possibly a pending rotation worth a look), **abstract** (custom Gateway setup), and archived projects whose entries are all closed.

- **One guard verdict per project** (`compareProject`): a rotation used to fail two tests with conflicting advice (disappeared → freeze vs. appeared → regenerate); now disappearances, appearances (labelled as the likely new era) and range moves are one combined failure with the recipes that apply. Plain additions keep the routine regenerate note.
- **`pnpm snapshots:generate` is append-only**: it registers new identities but keeps every committed entry whose id disappeared or whose range moved (closing an entry included), printing how many it kept and pointing at `--overwrite` — the explicit sign-off that accepts the wipe/re-sync.

## No-op assertions

- Runtime `daTrackingConfig` for every project, dumped via `getProjects()` before and after: deep-equal.
- `pnpm snapshots:generate` → `src/snapshots/daTracking/snapshot.json` unchanged (203 identities / 166 projects, ids **and** ranges), i.e. the deploy is a no-op.

## Simulated rotation

Edited `blast/discovered.json` locally (`SystemConfig.batcherHash` → `0x1111…`):

1. guard fails: `da-tracking identities disappeared for blast: - 576440a9a827 (… since 19300387) [19300387 -> open]` + the freeze recipe (+ the "new identity" message for the new id)
2. following the recipe — old entry frozen to literals with `untilBlock: 24541879` (`usedBlockNumbers.ethereum`), `getOpStackDaTracking(discovery, { sinceBlock: 24541879 })` appended — the guard now reports the intended range change; `pnpm snapshots:generate` then makes `guard.test.ts` + `gaps.test.ts` green (only the discovery-hash tests still fail, as expected for a hand-edited `discovered.json`)
3. everything reverted; nothing from the simulation is in this PR

## Testing

- `pnpm typecheck`, `pnpm lint`, `pnpm format`, `pnpm test` in `packages/config` (23960 passing)

Fixes L2B-14749


torztomasz:
@codex review

adamiak (approved):

PR #12647, opened 2026-08-21
add option to sort by the change in columns

vilcht:


packages/frontend/src/components/table/controls/ColumnsControls.tsx
```diff
@@ -32,6 +32,10 @@ export function ColumnsControls<T>({ columns }: Props<T>) {
     column.columns.length > 0 ? column.columns : [column],
   )
 
+  const pickerColumns = columnsWithoutGroups.filter((column) =>
+    column.getCanHide(),
+  )
```
torztomasz:
```suggestion
const pickerColumns = columns
	.flatMap((column) => column.columns.length > 0 ? column.columns : [column])
	.filter((column) => column.getCanHide())
```

packages/frontend/src/hooks/useTable.ts
```diff
@@ -4,9 +4,10 @@ import { useReactTable } from '@tanstack/react-table'
 export function useTable<TData extends RowData>(options: TableOptions<TData>) {
   return useReactTable({
     enableSortingRemoval: false,
+    ...options,
     initialState: {
       sorting: [{ id: '#', desc: false }],
+      ...options.initialState,
     },
```
torztomasz, line 11: that may break behavior of some tables. verify it does not

packages/frontend/src/components/table/sorting/changeSortColumn.ts
```diff
@@ -0,0 +1,48 @@
...
+ * Pairs a value column with a TanStack companion used only to sort by
+ * percentage change. BasicTable does not render the companion;
+ * enableHiding: false keeps it out of the column picker.
+ */
+export function withChangeSort<TData, TValue>(
+  columnHelper: ColumnHelper<TData>,
+  column: ColumnDef<TData, TValue>,
+  opts: {
+    getChange: (row: TData) => number | undefined
+    period: PercentageChangePeriod
```
torztomasz: `period` is hard-coded at every call site while the cell reads `changePeriod` from the row data. They agree today ('7D' / 'last30d' / '1D'), but nothing ties them together, so the header label and the actual period can drift silently. Consider deriving the label from the row's `changePeriod` (or asserting they match) so there's one source of truth.

packages/frontend/src/pages/layer2s/activity/components/table/columns.tsx
```diff
@@ -91,34 +92,42 @@ export const getL2ActivityColumns = (
...
+              >
+                {formatActivityCount(data.pastDayCount.value)}
+              </ValueWithPercentageChange>
+            </SyncStatusWrapper>
+          )
+        },
+        sortUndefined: 'last',
+        meta: {
+          align: 'right',
+          headClassName: 'max-w-[110px]',
```
torztomasz, line 121: `max-w-[110px]` is kept on the header, but the header now also renders the "7D" sort control next to "Past day UOPS/TPS". Please check visually that it doesn't wrap/clip.

packages/frontend/src/components/table/sorting/ValueAndChangeSortingHeader.tsx
```diff
@@ -0,0 +1,59 @@
...
+  const changeColumn = changeSortColumnId
+    ? header.getContext().table.getColumn(changeSortColumnId)
+    : undefined
+
+  if (!changeColumn?.getCanSort()) {
+    return valueSortingArrows
+  }
+
+  const changeHeader =
+    typeof changeColumn.columnDef.header === 'string'
```
torztomasz: This `typeof ... === 'string'` guard is dead — `withChangeSort` always sets a string header on the companion. The fallback would silently drop the change control instead of surfacing a wiring mistake. Suggest removing the guard (or asserting).

packages/frontend/src/components/table/utils/renderedTableColumns.ts
```diff
@@ -0,0 +1,27 @@
+import type { Cell, Header } from '@tanstack/react-table'
+
+/** TanStack columns that BasicTable does not render. */
+export function isSortOnlyColumn(column: {
```
torztomasz: One concept is spelled three ways: `isChangeSortColumn` (meta), `isSortOnlyColumn` (util), `withChangeSort` (factory). Pick one term ("change sort" or "sort-only") and use it everywhere.

packages/frontend/src/pages/layer2s/tvs/components/table/columns.tsx
```diff
@@ -104,27 +105,65 @@ export const getL2TvsColumns = (
...
-      header: 'Canonically bridged',
+function tvsBreakdownColumn<
+  T extends
+    | TvsToken['category']
+    | Exclude<TvsToken['source'], 'custom-canonical'>,
+>(opts: {
+  id: string
+  header: string
+  dataKey: T
+  type: 'bridgingType' | 'category'
```
torztomasz: nit: `dataKey` and `type` are independent here, so nothing prevents e.g. `dataKey: 'ether', type: 'bridgingType'`. Mirrors the existing `BreakdownCell` typing so fine to leave, but a discriminated union would make the pairing explicit.

torztomasz (commented):
Overall nice code. Some findings below. Also I would change the period to have the % next to it so that it will be 7D%

Image

torztomasz (approved):

torztomasz:
@codex review

torztomasz (approved):

PR #12652, opened 2026-08-24
Throw instead of storing 0 on TVS balance/totalSupply RPC failures

torztomasz:
## Problem

`BalanceProvider`, `TotalSupplyProvider`, `StarknetBalanceProvider` and `StarknetTotalSupplyProvider` silently converted RPC failures into `0n` (log warn + return 0). The `OnchainAmountIndexer` then persisted these zeros as legitimate amounts.

For `diff`-based TVS configs this is poison: katana-ETH computes `const(2^128-1) - balanceOfEscrow(native)`, so a spurious zero balance produced amounts of `~3.4e20 ETH` (`~8e23 USD`) in `TokenValue`.

## Fix

- **RPC-level errors now propagate** instead of returning `0n` — the provider falls back to the next configured RPC client, and throws once all are exhausted. The amount indexer uses an infinite retry strategy, so a flaky/pruned RPC means "retry later" instead of stored garbage. This also fixes a missing `await` on `Promise.all` in both EVM providers that let rejections escape the client-fallback loop entirely.
- **Multicall responses now carry raw `tryAggregate` results**: `decode()` previously collapsed "reverted" and "succeeded with empty returndata" into `success: false`. Now `success` is the raw chain flag and consumers derive meaning: a **revert** gives no number, so the providers throw; **empty returndata** means the target is not deployed at the queried block, so `0n` is legitimate (throwing there would stall whole chains on configs whose `since` slightly predates deployment). Existing consumers that relied on the collapsed flag were updated (`getElasticChainTokens`, wormhole's single-call fallback); the rest already check data emptiness themselves.
- **Native `getEthBalance` always throws on failure** — it can neither revert nor return empty, so either is an RPC fault.
- Starknet providers log and rethrow; starknet errors don't distinguish "contract not found" from transient failure, so they fail loud.

## Tests

Replaced the assertions expecting `0n` from errors with explicit rejection tests, added revert→throw and next-client-fallback tests for both EVM providers, and updated the `decode` tests to the raw semantics (empty returndata keeps `success: true`).

Resolves L2B-14352

### DIFF
--- a/packages/shared/src/clients/rpc/multicall/MulticallV3Client.ts
+++ b/packages/shared/src/clients/rpc/multicall/MulticallV3Client.ts
@@ -4,6 +4,9 @@ import type { CallParameters } from '../types'
 
 export interface MulticallV3Response {
   success: boolean
+  // distinguishes a revert from a successful call with empty returndata
+  // (e.g. target not deployed at the queried block) - both have success=false
+  reverted: boolean
   data: Bytes
 }
 
@@ -48,6 +51,7 @@ export class MulticallV3Client {
       const bytes = Bytes.fromHex(data)
       return {
         success: success && bytes.length !== 0,
+        reverted: !success,
         data: bytes,
       }
     })
--- a/packages/shared/src/clients/rpc/multicall/MulticallV3Client.ts
+++ b/packages/shared/src/clients/rpc/multicall/MulticallV3Client.ts
@@ -3,8 +3,6 @@ import { utils } from 'ethers'
 import type { CallParameters } from '../types'
 
 export interface MulticallV3Response {
-  // raw tryAggregate flag: false = the call reverted. A successful call can
-  // still return empty data, e.g. target not deployed at the queried block
   success: boolean
   data: Bytes
 }
--- a/packages/shared/src/clients/rpc/multicall/MulticallV3Client.ts
+++ b/packages/shared/src/clients/rpc/multicall/MulticallV3Client.ts
@@ -3,10 +3,9 @@ import { utils } from 'ethers'
 import type { CallParameters } from '../types'
 
 export interface MulticallV3Response {
+  // raw tryAggregate flag: false = the call reverted. A successful call can
+  // still return empty data, e.g. target not deployed at the queried block
   success: boolean
-  // distinguishes a revert from a successful call with empty returndata
-  // (e.g. target not deployed at the queried block) - both have success=false
-  reverted: boolean
   data: Bytes
 }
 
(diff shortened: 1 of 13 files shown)


adamiak:
@codex review

sdlyy (approved):

packages/shared/src/clients/rpc/multicall/MulticallV3Client.ts
```diff
@@ -3,6 +3,8 @@ import { utils } from 'ethers'
 import type { CallParameters } from '../types'
 
 export interface MulticallV3Response {
+  // raw tryAggregate flag: false = the call reverted. A successful call can
+  // still return empty data, e.g. target not deployed at the queried block
```
torztomasz:
```suggestion
```

packages/shared/src/providers/totalSupply/TotalSupplyProvider.ts
```diff
@@ -27,7 +27,14 @@ export class TotalSupplyProvider {
         if (client.isMulticallDeployed(blockNumber)) {
           const res = await client.multicall(calls, blockNumber)
           return res.map((r, i) => {
-            if (r.success === false) {
+            // a revert gives no number - throw instead of storing a poisoned 0
```
torztomasz:
```suggestion
```

packages/shared/src/providers/totalSupply/TotalSupplyProvider.ts
```diff
@@ -27,7 +27,14 @@ export class TotalSupplyProvider {
...
+            if (!r.success) {
+              throw new Error(
+                `Failed to fetch totalSupply of ${tokens[i]} at block ${blockNumber}`,
+              )
+            }
+            if (r.data.length === 0) {
+              // token not deployed at this block
```
torztomasz, line 36:
```suggestion
```

packages/shared/src/providers/balance/BalanceProvider.ts
```diff
@@ -35,7 +35,18 @@ export class BalanceProvider {
...
+            // a revert gives no number - throw instead of storing a
+            // poisoned 0. getEthBalance always returns data, so empty
+            // data for native is a fault too
+            if (reverted || (isDataEmpty && queries[i].token === 'native')) {
+              throw new Error(
+                `Failed to fetch balance of ${queries[i].token} for ${queries[i].holder} at block ${blockNumber}`,
+              )
+            }
+            if (isDataEmpty) {
+              // token not deployed at this block
```
torztomasz:
```suggestion
            // a revert gives no number - throw instead of storing a
            // poisoned 0. getEthBalance always returns data, so empty
            // data for native is a fault too
            if (!success || (r.data.length === 0 && queries[i].token === 'native')) {
              throw new Error(
                `Failed to fetch balance of ${queries[i].token} for ${queries[i].holder} at block ${blockNumber}`,
              )
            }
            if (isDataEmpty) {
              // token not deployed at this block
```

PR #12653, opened 2026-08-24
Fix interop flows graph particle lag

torztomasz:
## Problem

The flows graph was laggy in two places, for two different reasons:

- **Firefox, every page** — the CSS `offset-path: url(#id)` motion introduced in #12636 is never run on the compositor by Gecko, so each particle became per-frame main-thread work.
- **Home page, Chromium** — the interop card is a CSS size container. Inside one, Blink applies the SMIL `<animate attributeName="opacity">` through style per particle per frame, and every such update forces a layout: ~11k layouts/s, 70% of the main thread in Layout, 41 FPS.

## What changed

`ParticleLayer.tsx` only; the home card and `MAX_TOTAL_PARTICLES` (700) are untouched.

- Particles are SMIL in every engine again (no CSS `offset-path`, no engine sniffing).
- `<animateMotion>` is the only animation. Instead of an opacity animation, particles are hidden by position: each flow's path is relative to its source and the flow group is translated there, so a particle that hasn't started sits at the source bubble's center and an idle one parks at the path end, the destination's center.
- The layer is clipped with a static even-odd `clipPath` that punches out the bubble discs, so parked particles never paint (some icons are transparent in the middle) and particles vanish/appear at the bubble rim.
- The positive start delay is kept, so particles still stream out of the sources over the first cycle.

## Results (dev, 403 particles on the home page)

| | Layouts/s | Main thread | FPS |
|---|---|---|---|
| Chromium before | 11,168 | 96% | 41 |
| Chromium after | 120 | 28% | 120 |
| Firefox after (container, 60 Hz cap) | — | — | 60 |

`useCssSupports` and the `interop-particle-*` keyframes are now unused; can go in a follow-up.


packages/frontend/src/pages/interop/components/flows/graph/ParticleLayer.tsx
```diff
@@ -74,9 +64,19 @@ export function ParticleLayer({
   baseDollarsPerParticle,
 }: Props) {
   const { highlightedChains } = useInteropFlows()
-  const supportsCssMotion = useCssSupports(CSS_MOTION_CONDITION)
-  const pathIdPrefix = useId()
   const particleRadius = isSmallScreen ? 1.5 : 2
+  const clipId = `particles-clip-${useId().replace(/\W/g, '')}`
+
+  // Even-odd: a huge rect with one circle subpath per bubble punched out
```
torztomasz:
```suggestion
  // This exists to hide dots below the project bubbles
```

vilcht (approved):

PR #12658, opened 2026-08-25
Add support for BlockScoutV2

sdlyy:
Fixes issue with Robinhood - Blockscout V1 on RH is heavily rate-limited, disco falls back to Sourcify (see combined client) - Sourcify returns 404, thus the whole combined-client call returns unverified (yet contract is verified on BS, but API is barely working - also UM cached verified response) - that's where Blockscout V2 kicks is - it works just fine for some reason, so why not just add it

@codex review


mateuszradomski:
No longer a problem.

PR #12661, opened 2026-08-25
Return partial Relay requests instead of failing the update

mateuszradomski:
From daily checks

A limit of 500 is fetched as 10 sequential pages, so one slow page threw and syncedTo was never written, leaving the indexer reissuing the same request. Pages are sorted by updatedAt ascending, so a prefix is contiguous and syncedTo just lands earlier in the window. A first page failure still throws.

### DIFF
--- a/packages/backend/src/modules/interop/plugins/relay/RelayApiClient.ts
+++ b/packages/backend/src/modules/interop/plugins/relay/RelayApiClient.ts
@@ -1,3 +1,4 @@
+import type { Logger } from '@l2beat/backend-tools'
 import type { HttpClient } from '@l2beat/shared'
 import { v } from '@l2beat/validate'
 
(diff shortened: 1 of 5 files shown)


sdlyy:
@codex review

sdlyy (approved):

packages/backend/src/modules/interop/plugins/relay/RelayApiClient.ts
```diff
@@ -196,11 +196,19 @@ export class RelayApiClient {
...
+        res = await this.getRequests({
+          ...options,
+          limit: Math.min(limit, 50),
+          continuation,
+        })
+      } catch (error) {
+        if (result.requests.length === 0) {
+          throw error
+        }
+        break
```
adamiak, line 229: We should still log the error, so that we know we're constantly hitting some problem.

adamiak, line 229: This only works properly if the **caller** passes: sortBy: 'updatedAt' and sortDirection: 'asc' . But there's no guarantee the caller does that.  For this function to not lose entries, these sort values must be forced here (which will look a bit weird though)

adamiak, line 229: Here it's  necessary to set `result.continuation = continuation`

mateuszradomski:
@codex, review

adamiak (approved):

PR #12665, opened 2026-08-25
Add token relation manually via UI

adamiak:
L2B-14721


packages/token-ui/src/utils/manualRelationEvidence.ts
```diff
@@ -0,0 +1,43 @@
...
+  }
+}
+
+function isRecord(value: unknown): value is Record<string, unknown> {
+  return typeof value === 'object' && value !== null && !Array.isArray(value)
+}
+
+function asString(value: unknown): string | undefined {
+  return typeof value === 'string' ? value : undefined
+}
```
torztomasz: broooooo

torztomasz (approved):

packages/token-backend/src/planning.ts
```diff
@@ -427,24 +457,72 @@ async function planAddTokenRelation(
...
+ * plugin is the `manual` sentinel the evidence must be a manual entry, and its
+ * `user` is stamped here — at plan time, like deployed-token assignment
+ * proofs — so the confirmation diff already shows the record exactly as it
+ * will be stored, and the client never asserts who is writing. Evidence of
+ * relations with any other plugin passes through untouched.
+ *
+ * No timestamp is stamped: executePlan regenerates the plan and requires it
+ * to deep-equal the confirmed one, and the history row carries the time.
+ */
+function stampManualRelationEvidence(
```
torztomasz, line 523: This helper is where the PR's central invariant — *a `manual` relation's evidence is a `ManualRelationEvidenceInput`* — ends up living, and it lives as a runtime `if` + `safeParse` + hand-written error, called from two places (the second wrapped in another `transfer === undefined` conditional at L478). That's three branches for one fact the intent schema could state directly:

```ts
export const TokenRelationInput = v.union([
  v.object({ ...endpoints, plugin: v.literal(MANUAL_RELATION_PLUGIN),
             bridgeType: KnownInteropBridgeType, transfer: ManualRelationEvidenceInput }),
  v.object({ ...endpoints, plugin: v.string().check((p) => p !== MANUAL_RELATION_PLUGIN || '…'),
             bridgeType: InteropBridgeType, transfer: JsonValue }),
])
```

With that on `AddTokenRelationIntent.record`, tRPC input validation rejects malformed manual evidence (and `unknown` for manual rows) before the planner runs, and this function collapses to a pure map — `plugin === 'manual' ? { ...transfer, user } : transfer` — with no `PlanningError` and no "must carry manual-entry evidence" test. The `Expect<Equal<TokenRelationRecord, DbTokenRelationRecord>>` guard is unaffected: the input schema is a new type that narrows to the record. (A dedicated `AddManualTokenRelationIntent` would work too — only two sites switch on intent type — but the union keeps the intent surface as-is.)

packages/token-backend/src/schemas/TokenRelation.ts
```diff
@@ -55,3 +55,49 @@ export const TokenRelationUpdateable = v.object({
...
+
+/**
+ * What the `transfer` evidence column holds for a manually added relation —
+ * a human attestation instead of a sample interop transfer. The `kind`
+ * discriminator tells the two evidence shapes apart inside the JSON; the
+ * row-level `plugin = 'manual'` sentinel tells them apart without opening it.
+ * See docs/mdbook/specs/l2b_specs/token_db/token_relations.md.
+ */
+export type ManualRelationEvidence = v.infer<typeof ManualRelationEvidence>
+export const ManualRelationEvidence = v.object({
```
torztomasz: This schema is never parsed anywhere — it only serves as a `satisfies` target in the planner. Three exported schemas for one concept is two too many; `type ManualRelationEvidence = ManualRelationEvidenceInput & { user: string }` is the whole thing. (If a runtime validator for the stored shape *is* wanted, it's for the UI reader — see the thread on `manualRelationEvidence.ts` — and then it belongs in `shared-pure`, not here.)

packages/token-ui/src/utils/manualRelationEvidence.ts
```diff
@@ -0,0 +1,43 @@
...
+    | undefined
+}
+
+/**
+ * Lenient reader for the `{ kind: 'manual', ... }` evidence stored on a
+ * manually added relation, telling it apart from sample-transfer evidence by
+ * the `kind` discriminator. Display-only: tolerates missing or malformed
+ * fields instead of throwing, since the raw JSON is always shown alongside.
+ */
+export function readManualRelationEvidence(
```
torztomasz: This is a 43-line hand-rolled duck-typing reader (with its own `View` type where every field is `| undefined`) for a shape that already exists as a validator in token-backend. token-ui already depends on `@l2beat/validate`, and the schema is pure data with no backend deps — move `ManualRelationEvidence` next to `MANUAL_RELATION_PLUGIN` in `shared-pure` (the file that already documents the shape) and have the panel do `ManualRelationEvidence.safeParse(relation.transfer)`. This file, the `View` interface, `isRecord` and `asString` all go away.

The "lenient because the raw JSON is shown alongside" rationale also doesn't hold up next to its sibling: `readTransferEvidence` in the same panel *throws* on malformed evidence. Two evidence readers with opposite philosophies in one component is exactly the kind of thing that trips the next reader.

packages/token-ui/src/pages/tokens/TokenRelationsGraphDetailsPanel.tsx
```diff
@@ -547,6 +548,38 @@ function RelationEvidence({
   relation: RelationDetails
   chains: Chain[]
 }) {
+  const manualEvidence = readManualRelationEvidence(relation.transfer)
```
torztomasz: The PR's own spec and prisma comment establish `plugin === 'manual'` as *the* row-level discriminator ("any reader can select or exclude manual relations on the `plugin` column alone, without opening the evidence JSON") — and then the first reader sniffs `transfer.kind` instead. Branch on `relation.plugin === MANUAL_RELATION_PLUGIN` and let the parse be the assertion, so the UI follows the rule the PR sets.

packages/token-ui/src/pages/tokens/AddTokenRelationDialog.tsx
```diff
@@ -0,0 +1,499 @@
...
+  ) => void
+}) {
+  form.setError('otherAddress', {
+    type: 'validation',
+    message:
+      'No deployed token with this chain and address — relations can only be added between catalogued tokens',
+  })
+}
+
+function ChainCombobox({
```
torztomasz, line 444: This is a near-verbatim lift of the Popover+Command chain picker in `DeployedTokenForm.tsx:216-270` (and the same shape again in `AbstractTokenPage.tsx:406`). Extracting a `ChainCombobox` was the right instinct — but as a private function in this file it's a fourth copy, not the canonical one. Promote it to `components/ChainCombobox.tsx` and make `DeployedTokenForm` use it in this PR; otherwise this is the copy that gets copied next.

packages/token-ui/src/pages/tokens/AddTokenRelationDialog.tsx
```diff
@@ -0,0 +1,499 @@
...
+          form.reset()
+          setPlan(data.plan)
+        } else {
+          toast.error(data.error)
+        }
+      },
+    }),
+  )
+
+  function onSubmit(values: AddTokenRelationSchema) {
```
torztomasz, line 149: `onSubmit` re-implements validation the planner already owns: the same-token check reproduces the planner's message verbatim, and the "no deployed token" message appears twice — inline JSX ~20 lines below and again via `setOtherTokenMissingError` (L430), a helper called once with a structural `form` type invented just to call `setError`. The planner already returns friendly `PlanningError`s that this dialog toasts.

Keep the live "no such token" hint (good UX), delete the wrapper and the duplicated same-token rule, and let the planner be the single validation authority. If inline field errors are non-negotiable, put the cross-field rules in the schema so `validateResolver` owns them instead of an imperative `setError` ladder.

torztomasz (commented):
Deep code-quality pass. Behavior is solid — CI green, planner tests thorough, spec updated — and approved on that basis, but a handful of structural points came out of it that I'd like addressed, here or in a fast follow-up. In order of weight:

1. The `plugin ⇔ evidence` invariant lives in a runtime `if` in the planner (`stampManualRelationEvidence`) instead of the intent schema.
2. `readManualRelationEvidence` re-implements a validator that already exists, and dispatches on `transfer.kind` instead of the `plugin` sentinel the PR itself defines as canonical.
3. `ChainCombobox` is a fourth copy of the chain picker.
4. `onSubmit` duplicates planner validation — and `ManualRelationBridge` is looser than it needs to be, which is what creates that validation.
5. The manual skip in `resolveMinters` is a feature check in a shared public-site path; it belongs at the data boundary.
6. "Mirrors the table's CHECK constraints" isn't true for `unknown`.

Details inline. Smaller things not worth their own thread:

- Spec ~L555: "they render … in the graph" is false for `nonMinting` (`isGraphRelation` drops it) — Codex already flagged this.
- The sentinel's rationale is written out in four places (shared-pure, prisma comment, schema, spec). Keep it in the spec and link from code.
- `requiredCheck` is `minLengthCheck(1)` with a better message — add a message param to the canonical helper instead.
- The `lockedSide` → `lockedToken` nested ternary reads better as a `Record<LockedSide, 'A' | 'B' | null>` lookup.
- Nothing crosses 1k lines; `TokenRelationsGraphDetailsPanel.tsx` (~830) is the one to watch — splitting `RelationEvidence` into two section components would keep it flat.

packages/token-backend/src/schemas/TokenRelation.ts
```diff
@@ -55,3 +55,49 @@ export const TokenRelationUpdateable = v.object({
...
   transfer: JsonValue.optional(),
 })
+
+/**
+ * The bridge a manually added relation claims as its mechanism, e.g. the WETH
+ * contract for the ETH <-> WETH pair. Display-only, like all relation
+ * evidence: no read path derives roles or identity from it.
+ */
+export type ManualRelationBridge = v.infer<typeof ManualRelationBridge>
+export const ManualRelationBridge = v.object({
```
torztomasz: `{ name, chain | null, address | null }` allows a bridge with a name and no contract, a chain with no address, an address with no chain, etc. That looseness is exactly what forces the "name the bridge or clear its chain and address" rule in the dialog's `onSubmit` *and* the `bridge.chain && bridge.address` render guard in the details panel. `bridge: { name: string; contract: { chain: string; address: string } | null } | null` makes both structural — the rule and the guard disappear.

packages/frontend/src/server/features/layer2s/interop/token/getInteropTokenEntry.ts
```diff
@@ -81,6 +81,12 @@ function resolveMinters(
 ): InteropTokenOnchainDeploymentsRow['minters'] {
   const projects = deployment.mintingPlugins.flatMap(
     ({ plugin, bridgeType, relatedChain }) => {
+      // A manually added relation names no interop plugin, so it can never
+      // resolve to an interop project. Skipped deliberately — not warned
+      // about — until the public site decides how to present manual bridges.
+      if (plugin === MANUAL_RELATION_PLUGIN) {
```
torztomasz, line 87: This is a feature check bolted into a shared path: `resolveMinters` is about interop-project resolution, and now it also knows about a token-DB sentinel (which is why `shared-pure` grew an export the public site otherwise has no reason to import). The spec advertises the sentinel precisely so readers can `WHERE plugin != 'manual'` — do that in `getMintingPluginsForMany` (its only caller is this page), or at minimum filter once where `mintingPlugins` is assembled in `getInteropTokenOnchainDeployments.ts`. `resolveMinters` and its test then stay untouched.

packages/token-backend/src/planning.ts
```diff
@@ -401,12 +404,39 @@ function extractAbstractTokenId(displayId: string): string {
 async function planAddTokenRelation(
   db: TokenDatabase,
   intent: AddTokenRelationIntent,
+  opts: PlanOptions,
 ): Promise<Command[]> {
-  await assertRelationEndpointsExist(db, intent.record)
+  // These mirror the table's CHECK constraints, so a violation surfaces here
```
torztomasz: Not quite: there is no CHECK for `bridgeType != 'unknown'` — the migration only defines `endpoints_ordered`, `lockedToken_valid` and `lockedToken_only_for_lock_and_mint`. The `unknown` rejection is a planner policy (the spec repeats the same claim). Say so — or better, make it a type (`KnownInteropBridgeType` on the manual input branch) so it isn't a runtime throw at all.

packages/token-backend/src/planning.ts
```diff
@@ -427,24 +457,72 @@ async function planAddTokenRelation(
...
+    )
+  }
   const existing = await db.tokenRelation.findByPrimaryKey(intent.pk)
   if (existing === undefined) {
     throw new PlanningError(
       `TokenRelation ${formatTokenRelationPrimaryKey(intent.pk)} doesn't exist`,
     )
   }
 
+  const update =
```
torztomasz, line 491: No client sends `UpdateTokenRelationIntent` with `transfer` — the only relation update the UI issues is `lockedToken`. This branch (plus its test and the "stays editable through the update intent" sentence in the spec) guards a path that doesn't exist yet. If evidence editing isn't a requirement now, drop it; if it is, it needs a UI in this PR or a ticket.

packages/token-backend/src/schemas/TokenRelation.ts
adamiak: Keeping the flat shape. This JSON is display-only by design — like sample transfers, no read path derives anything from it — so the looseness costs at most a hidden detail row. And partial knowledge is a real state here: "I know the bridge and its chain but haven't dug out the contract address" is worth storing as entered rather than forcing to null. Since nothing parses evidence to act on it, tightening later (if it ever starts to matter) is a one-off update of a display-only column, not a migration.

packages/token-ui/src/utils/manualRelationEvidence.ts
adamiak: Fixed — the whole file is gone (the panel now safeParses the shared-pure schema) 🫡

packages/token-backend/src/planning.ts
adamiak: Fixed in 15a529b99d: the comment and the spec paragraph now say the bridgeType rule is planner policy — only the lockedToken and endpoint rules mirror actual `CHECK` constraints. And the policy got stricter while I was at it: `nonMinting` is now rejected too (a relation asserts same-asset endpoints; a nonMinting route may swap — the same rule ingestion applies), with the dialog no longer offering it. Didn't take the make-it-a-type route since that hangs on the intent-schema union, which I'm not taking (see that thread).

docs/mdbook/specs/l2b_specs/token_db/token_relations.md
```diff
@@ -495,6 +508,58 @@ moves with the endpoints when they are swapped, so the role a human stated is
...
+requires it to deep-equal the confirmed one, and the history row carries the
+time. Relations with any other plugin pass their evidence through untouched —
+the backend deliberately keeps accepting arbitrary plugin values on the
+intent (it is an authenticated internal tool and every write lands in
+`TokenDbHistory`); only relations claiming the `manual` sentinel must carry
+manual-entry evidence.
+
+Manual relations are ordinary rows everywhere else: they render in the
+Relations tab and the graph (edge label `manual`), `getMintingPluginsFor`
+counts them, and the delete intent removes them. Same-chain pairs are legal —
```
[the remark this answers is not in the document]
adamiak: Fixed in 15a529b99d — and more strictly than a doc qualification: manual adds no longer allow `nonMinting` at all. A relation asserts its endpoints are the same asset, and a nonMinting route may swap assets — the same reason ingestion keeps only the non-swapping types — so the planner now rejects it and the dialog doesn't offer it. The spec now states that `nonMinting` is not supported in the table (writer policy; the graph/minter readers stay defensive), which makes the quoted sentence unconditionally true: every manual row is a graph- and minter-qualifying type.

packages/token-ui/src/pages/tokens/AddTokenRelationDialog.tsx
adamiak: Agreed on the end state, but not inside this PR: the duplication pre-dates it (DeployedTokenForm and AbstractTokenPage already carry copies), so promoting this one and converting a single call site would leave three variants plus a "canonical" one — and reworking a stable, unrelated form here is regression surface with no behavior change. Consolidating the chain picker is a separate mechanical PR that converts all call sites at once.

packages/token-ui/src/pages/tokens/AddTokenRelationDialog.tsx
adamiak: Keeping the pre-submit checks — they're UX, not a second authority. The error lands on the field it belongs to without a server round-trip, and the planner still re-validates everything at both generate and execute, so behavior can't drift; the worst case is message copy diverging in one dialog. The two "no deployed token" surfaces also do different jobs: the inline hint reacts while typing, the submit-time field error is what actually blocks a submit racing the async check.

packages/token-ui/src/pages/tokens/TokenRelationsGraphDetailsPanel.tsx
adamiak: Fixed in 15a529b99d — the branch is now `relation.plugin === MANUAL_RELATION_PLUGIN` and the parse is the assertion, with the raw-JSON block as the rendering fallback if it ever fails.

packages/token-ui/src/utils/manualRelationEvidence.ts
adamiak: Fixed in 15a529b99d: the file is deleted. The panel branches on the plugin sentinel and safeParses the shared-pure `ManualRelationEvidence`; a failed parse falls back to the raw-JSON block instead of throwing, so display robustness is kept without a second reader philosophy.

packages/token-backend/src/planning.ts
adamiak: Keeping this in the planner, for three reasons:

- The planner can't stop being a validation layer regardless — endpoint existence, the duplicate-PK check, and the CHECK mirrors all need the DB — so the union wouldn't consolidate validation, it would split the manual invariant away from the rest of the rules.
- The update path needs the runtime check either way: `TokenRelationUpdateable.transfer` must stay arbitrary JSON for non-manual rows, and an input schema can't tie `pk.plugin` to the shape of `update.transfer` without a second union over the whole intent. So `stampManualRelationEvidence` survives the refactor; only one of its call sites gets thinner.
- A `PlanningError` surfaces as the friendly outcome-error toast; a tRPC input-validation failure surfaces raw. Keeping every user-facing rejection on one path is deliberate.

Nothing bypasses the check — both generate and execute run the planner. If the intent surface grows more per-plugin rules, a discriminated record type is worth revisiting then.

packages/token-backend/src/schemas/TokenRelation.ts
adamiak: Fixed in 15a529b99d, taking your parenthetical: `ManualRelationEvidence` (with the bridge schema) moved to shared-pure next to the sentinel, and it is now genuinely parsed — the graph panel safeParses it for display. token-backend keeps only `ManualRelationEvidenceInput`, reusing the shared bridge schema.

packages/frontend/src/server/features/layer2s/interop/token/getInteropTokenEntry.ts
adamiak: Keeping the skip here. Filtering in the repo would fork semantics between `getMintingPluginsFor` and `...ForMany` — token-ui's reader must keep counting manual rows (the Relations tab and graph show them), so the repo can't own the rule. Filtering at assembly in `getInteropTokenOnchainDeployments` would be behaviorally identical, just farther from the symptom: this line is exactly where the unresolvable-plugin warning would otherwise fire, and it's the single marked spot to delete once we decide how to present manual bridges publicly. Keeping the data flowing to the frontend was deliberate — that future decision should be a display change, not re-plumbing. (shared-pure hosts the sentinel precisely because both UIs consume it.)

packages/token-backend/src/planning.ts
adamiak: The path does exist: `transfer` is part of `TokenRelationUpdateable` on main — the update endpoint accepted evidence updates before this PR; the UI just doesn't send them yet. Given that surface, this branch is what stops an update intent from overwriting a manual row's evidence with arbitrary JSON and silently shedding the user stamp — dropping it would open the exact hole the add path closes. Narrowing the updateable schema itself would be its own discussion, out of scope here.

adamiak:
Thanks for the deep pass @torztomasz — every inline thread is answered and resolved; the ones that landed are in 15a529b99d. Answering the smaller notes from the review summary here since they have no thread:

- **Spec ~L555 / nonMinting in the graph**: fixed, and more strictly than a doc qualification — manual adds no longer allow `nonMinting` at all. A relation asserts its endpoints are the same asset and a nonMinting route may swap, so the planner rejects it (same rule ingestion applies), the dialog doesn't offer it, and the spec now states nonMinting is not supported in the table.
- **Sentinel rationale in four places**: keeping them — each is a couple of lines serving its own reader (someone at the prisma schema, someone importing the constant), and they all point at the spec section that holds the full story.
- **`requiredCheck` / the `lockedSide` ternary**: cosmetic as charged, but not worth touching the shared checks helper or reshaping working code in this PR — both fold naturally into the chain-picker consolidation when that PR happens.
- **Panel size**: agreed it's the one to watch; splitting `RelationEvidence` into two section components is the natural first cut when it next grows.

PR #12666, opened 2026-08-25
Show tokens without relations on the internal graph

adamiak:
L2B-14628

This is only for the internal TokenDB UI graph view.

### DIFF
--- a/packages/token-backend/src/trpc/routers/deployedTokens/index.ts
+++ b/packages/token-backend/src/trpc/routers/deployedTokens/index.ts
@@ -173,17 +173,51 @@ export const deployedTokensRouter = (deps: DeployedTokensRouterDeps) =>
         }
       })
 
+      const endpointNodes = tokenKeys.map((key): RelationsGraphNode => {
+        const token = tokenMap.get(tokenKey(key))
+        return {
+          id: tokenKey(key),
+          chain: key.chain,
+          address: key.address,
+          symbol: token?.symbol ?? null,
+          abstractTokenId: token?.abstractTokenId ?? null,
+          isDeployed: token !== undefined,
+          hasRelations: true,
+        }
+      })
+
+      // Deployed tokens manually assigned to an abstract token but observed
+      // in no relation. The graph places each one beside the cluster whose
+      // most common abstract token matches its assignment, so only tokens
+      // whose abstract token appears among the endpoints are shipped — one
+      // whose abstract token has no cluster at all has nowhere to go.
+      const endpointAbstractTokenIds = new Set(
+        endpointNodes.flatMap((node) => node.abstractTokenId ?? []),
+      )
+      const endpointIds = new Set(endpointNodes.map((node) => node.id))
+      const nodesWithoutRelations = (await ctx.tokenDb.deployedToken.getAll())
+        .filter(
+          (token) =>
+            !token.ignored &&
+            token.abstractTokenId !== null &&
+            endpointAbstractTokenIds.has(token.abstractTokenId) &&
+            !endpointIds.has(tokenKey(token)),
+        )
+        .map(
+          (token): RelationsGraphNode => ({
+            id: tokenKey(token),
+            chain: token.chain,
+            address: token.address,
+            symbol: token.symbol,
+            abstractTokenId: token.abstractTokenId,
+            isDeployed: true,
+            hasRelations: false,
+          }),
+        )
+        .sort((a, b) => a.id.localeCompare(b.id))
+
       return {
-        nodes: tokenKeys.map((key) => {
-          const token = tokenMap.get(tokenKey(key))
-          return {
-            id: tokenKey(key),
-            chain: key.chain,
-            address: key.address,
-            symbol: token?.symbol ?? null,
-            isDeployed: token !== undefined,
-          }
-        }),
+        nodes: [...endpointNodes, ...nodesWithoutRelations],
         relations: graphRelations,
       }
     }),
@@ -231,6 +265,22 @@ function sortRelations<
         ].join(':'),
       ),
   )
+}
+
+export interface RelationsGraphNode {
+  id: string
+  chain: string
+  address: string
+  symbol: string | null
+  abstractTokenId: string | null
+  isDeployed: boolean
+  /**
+   * False for a node appended because of its abstract token assignment
+   * rather than an observed relation. The UI draws such nodes hollow and
+   * attaches them to the cluster of their abstract token with a layout-only
+   * link.
+   */
+  hasRelations: boolean
 }
 
 function isGraphRelation(relation: { bridgeType: string }): boolean {
--- a/packages/token-ui/src/pages/tokens/TokenRelationsGraph.tsx
+++ b/packages/token-ui/src/pages/tokens/TokenRelationsGraph.tsx
@@ -60,7 +60,6 @@ export function TokenRelationsGraph({
   highlightAnomalies,
   deletedRelationIds,
   onSelectionChange,
-  onLayoutComplete,
 }: {
   graph: RelationGraph
   selection: RelationGraphSelection | undefined
@@ -68,8 +67,6 @@ export function TokenRelationsGraph({
   highlightAnomalies: boolean
   deletedRelationIds: ReadonlySet<string>
   onSelectionChange: (selection: RelationGraphSelection | undefined) => void
-  /** Called after the blocking scene build whenever `graph` changes. */
-  onLayoutComplete?: () => void
 }) {
   const canvasRef = useRef<HTMLCanvasElement>(null)
   const tooltipRef = useRef<HTMLDivElement>(null)
@@ -84,6 +81,12 @@ export function TokenRelationsGraph({
     () => getRelationGraphFocus(graph, selection, deletedRelationIds),
     [graph, selection, deletedRelationIds],
   )
+  // Runs the force layout to completion, blocking the main thread for
+  // seconds on large graphs. It lives in render rather than in the lifecycle
+  // effect below so the page can pass `graph` derived from a deferred value:
+  // the urgent render (switched radio, loading overlay) paints first, and
+  // the build then happens inside React's deferred re-render.
+  const scene = useMemo(() => buildRelationGraphScene(graph), [graph])
   // The draw loop reads the latest style inputs through a ref so that the
   // canvas lifecycle effect does not have to re-run on every style change.
   const styleStateRef = useRef({
@@ -93,15 +96,9 @@ export function TokenRelationsGraph({
     selection,
   })
 
-  const onLayoutCompleteRef = useRef(onLayoutComplete)
-
   useEffect(() => {
     onSelectionChangeRef.current = onSelectionChange
   }, [onSelectionChange])
-
-  useEffect(() => {
-    onLayoutCompleteRef.current = onLayoutComplete
-  }, [onLayoutComplete])
 
   useEffect(() => {
     styleStateRef.current = {
@@ -122,7 +119,6 @@ export function TokenRelationsGraph({
     const canvas = canvasElement
     const context = contextOrNull
 
-    const scene = buildRelationGraphScene(graph)
     const theme = resolveTheme(canvas)
     let cameraTransform = d3.zoomIdentity
     let cameraInitialized = false
@@ -415,10 +411,6 @@ export function TokenRelationsGraph({
     })
     observer.observe(canvas)
 
-    // The scene build above is the multi-second blocking part; from here on
-    // the graph only waits for cheap resize/draw callbacks.
-    onLayoutCompleteRef.current?.()
-
     return () => {
       if (frame !== undefined) cancelAnimationFrame(frame)
       observer.disconnect()
@@ -428,7 +420,7 @@ export function TokenRelationsGraph({
       zoomToNodeRef.current = undefined
       setTooltip(undefined)
     }
-  }, [graph])
+  }, [scene])
 
   useEffect(() => {
     if (zoomTarget === undefined) return
--- a/packages/token-ui/src/pages/tokens/TokenRelationsGraph.tsx
+++ b/packages/token-ui/src/pages/tokens/TokenRelationsGraph.tsx
@@ -60,6 +60,7 @@ export function TokenRelationsGraph({
   highlightAnomalies,
   deletedRelationIds,
   onSelectionChange,
+  onLayoutComplete,
 }: {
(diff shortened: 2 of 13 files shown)


torztomasz (approved):

packages/token-ui/src/pages/tokens/TokenRelationsGraphPage.tsx
```diff
@@ -69,6 +98,35 @@ export function TokenRelationsGraphPage() {
...
+    const outerFrame = requestAnimationFrame(() => {
+      innerFrame = requestAnimationFrame(() =>
+        setAppliedWithoutRelationsMode(withoutRelationsMode),
+      )
+    })
+    return () => {
+      cancelAnimationFrame(outerFrame)
+      if (innerFrame !== undefined) cancelAnimationFrame(innerFrame)
+    }
+  }, [appliedWithoutRelationsMode, withoutRelationsMode])
```
torztomasz: **Blocker (structural): this whole choreography can be deleted.**

To get the radio + overlay painted before the multi-second layout, the PR adds a pending/applied mode pair, an `isRelayouting` flag, this double-`requestAnimationFrame` effect, an `onLayoutComplete` callback (+ ref + sync effect in `TokenRelationsGraph`), and the "no graph mounted, so nobody reports completion" special case above — ~70 lines and three concepts, plus ~30 lines of comments explaining frame timing.

The root cause is that the blocking work lives in a `useEffect`, which React flushes synchronously after a click. Move the scene build into render and this collapses to the idiomatic primitive:

```tsx
const [withoutRelationsMode, setWithoutRelationsMode] =
  useState<TokensWithoutRelationsDisplayMode>('supported')
// Deferred so the switched radio and the overlay paint before the
// layout rebuild blocks the main thread.
const appliedWithoutRelationsMode = useDeferredValue(withoutRelationsMode)
const isRelayouting = withoutRelationsMode !== appliedWithoutRelationsMode
```

(with `buildRelationGraphScene` hoisted into a `useMemo` in `TokenRelationsGraph` — see the comment there). The urgent render commits and paints radio + overlay; the deferred render then does the blocking build; the commit that clears the overlay is the same one that shows the new graph. Gone: `appliedWithoutRelationsMode`, the `isRelayouting` state, this effect, `changeWithoutRelationsMode`'s empty-graph branch, `onLayoutComplete` + ref + effect, and the "two animation frames" paragraph in `token_relations.md`. Initial load improves too: the spinner stays visible instead of a blank frozen canvas.

One caveat worth a comment: if an urgent update lands after the memo finishes but before commit, React restarts the deferred render and rebuilds once more. The overlay covers the canvas so this is rare, and it's only a perf cost, never wrong state.

packages/token-ui/src/pages/tokens/TokenRelationsGraph.tsx
```diff
@@ -406,6 +415,10 @@ export function TokenRelationsGraph({
     })
     observer.observe(canvas)
 
+    // The scene build above is the multi-second blocking part; from here on
+    // the graph only waits for cheap resize/draw callbacks.
+    onLayoutCompleteRef.current?.()
```
torztomasz: Graph-side half of the `useDeferredValue` move (see the page comment): hoist the build out of the effect.

```tsx
const scene = useMemo(() => buildRelationGraphScene(graph), [graph])
useEffect(() => { /* unchanged, reads only `scene` */ }, [scene])
```

Nothing else in this effect reads `graph`, so the dependency swap is mechanical. Then `onLayoutComplete`, `onLayoutCompleteRef`, its sync effect, and this call all go away — the blocking work happens inside the deferred render, so "layout done" is simply "the deferred value caught up".

torztomasz (commented):
## Code quality review

Behavior looks right, and the hollow-node drawing plus the `mostCommon` extraction are clean. But two pieces of hand-rolled machinery land where React / the layout module already have the primitive, and the backend scans the whole table. In order of weight:

1. **Blocker — relayout choreography** (`TokenRelationsGraphPage.tsx`, `TokenRelationsGraph.tsx`): pending/applied mode pair + `isRelayouting` + double-rAF effect + `onLayoutComplete` callback/ref plumbing + an empty-graph special case, all replaceable by `useDeferredValue` once the scene build moves into a `useMemo`. ~70 lines and three concepts deleted.
2. **Duplicated clustering** (`relationGraphScene.ts`): `connectedComponents` / `getOrCreate` / `clusterSize` re-implement what `relationGraphLayout.ts#buildClusters` already computes, including the largest-first order.
3. **Backend `getAll()`** (`deployedTokens/index.ts`): full-table load filtered in memory; a batched `getByAbstractTokenIds` mirroring `getByPrimaryKeys` keeps it proportional to the graph.
4. Minor: `getNodeIdsOutsideRelationChains` is a predicate exported for its test; `isDeployed` / `hasRelations` admit an impossible state.

The three-way picker is explicitly decision-support scaffolding ("may be simplified once the chain-support decision is made") — fine for an internal tool, and after (1) it's cheap to delete, which is what matters. No file crosses 1k lines.

CI: `config-diff` fails; looks unrelated to these files.

packages/token-ui/src/pages/tokens/relationGraphScene.ts
```diff
@@ -134,6 +136,95 @@ function buildSimulationLinks<Node extends LayoutNode>(
...
+      const current = pending.pop()
+      if (current === undefined || visited.has(current.id)) continue
+      visited.add(current.id)
+      component.push(current)
+      pending.push(...(neighbors.get(current.id) ?? []))
+    }
+    components.push(component)
+  }
+  return components
+}
```
torztomasz, line 217: **Duplicated canonical helper.** `relationGraphLayout.ts#buildClusters` already computes connected components, sorts each cluster's nodes by id, and sorts clusters largest-first with a deterministic tie-break. `buildVirtualLinks` re-derives all three: its own BFS here, `.filter().sort().at(0)` for the anchor, and `clusterSize` bookkeeping for "largest wins".

Export `buildClusters` (e.g. as `groupIntoClusters`) and this becomes:

```ts
const layout = layoutRelationGraph(workingNodes, [
  ...simulationLinks,
  ...buildVirtualLinks(groupIntoClusters(workingNodes, simulationLinks)),
])

function buildVirtualLinks(clusters) {
  // Clusters arrive largest first, so the first to claim an abstract token
  // is the largest cluster representing it.
  const anchors = new Map<string, Node>()
  for (const cluster of clusters) {
    if (cluster.links.length === 0) continue // a lone node represents nothing
    const id = mostCommonAbstractTokenId(cluster.nodes.map((n) => n.data))
    if (id === undefined || anchors.has(id)) continue
    // Nodes are id-sorted, so .find() is the deterministic pick.
    anchors.set(id, requireNodeOf(cluster, id))
  }
  return clusters
    .flatMap((cluster) => cluster.nodes)
    .flatMap((node) => { /* same lookup as now */ })
}
```

~40 lines deleted (`connectedComponents`, `getOrCreate`, the size bookkeeping) and one owner for graph traversal. The only semantic change is the tie-break between equal-size clusters (smallest first node id instead of payload order) — both deterministic.

packages/token-backend/src/trpc/routers/deployedTokens/index.ts
```diff
@@ -173,17 +173,51 @@ export const deployedTokensRouter = (deps: DeployedTokensRouterDeps) =>
...
+      )
+      const endpointIds = new Set(endpointNodes.map((node) => node.id))
+      const nodesWithoutRelations = (await ctx.tokenDb.deployedToken.getAll())
+        .filter(
+          (token) =>
+            !token.ignored &&
+            token.abstractTokenId !== null &&
+            endpointAbstractTokenIds.has(token.abstractTokenId) &&
+            !endpointIds.has(tokenKey(token)),
+        )
```
torztomasz, line 205: **Loads the whole `DeployedToken` table to pick a handful of rows.** `getAll()` followed by an in-memory `abstractTokenId ∈ endpointAbstractTokenIds` filter grows with the table, not with the graph. The repository already has `getByAbstractTokenId(id)` and the batched-`IN` pattern in `getByPrimaryKeys`; a `getByAbstractTokenIds(ids)` is a ~10-line addition that lets the DB do the filter and leaves only `!ignored && !endpointIds.has(key)` here. Four other routers calling `getAll()` is precedent, not justification.

While here: the two `RelationsGraphNode` literals (endpoint nodes above, these below) are the same shape with `hasRelations` flipped — one `graphNode(key, token, hasRelations)` helper.

packages/token-ui/src/pages/tokens/relationGraphModel.ts
```diff
@@ -158,16 +158,93 @@ export function nodeLabel(node: RelationGraphNode) {
...
+      relation.tokenAChain,
+      relation.tokenBChain,
+    ]),
+  )
+  return new Set(
+    graph.nodes
+      .filter((node) => !node.hasRelations && !relationChains.has(node.chain))
+      .map((node) => node.id),
+  )
+}
```
torztomasz, line 205: This builds a `Set` of ids only so `filterTokensWithoutRelations` can filter against it, and is exported only for its test. It's a predicate:

```ts
const relationChains = new Set(
  graph.relations.flatMap((r) => [r.tokenAChain, r.tokenBChain]),
)
nodes: graph.nodes.filter(
  (node) => node.hasRelations || relationChains.has(node.chain),
)
```

Inline it into `filterTokensWithoutRelations`; the `filterTokensWithoutRelations` tests already cover the behavior.

packages/token-ui/src/pages/tokens/TokenRelationsGraph.tsx
```diff
@@ -464,7 +477,11 @@ function nodeTooltip(node: SceneNode) {
   return [
     `${node.label} on ${data.chain}`,
     data.address,
-    data.isDeployed ? 'Deployed token exists' : 'Missing deployed token',
+    !data.isDeployed
+      ? 'Missing deployed token'
+      : data.hasRelations
+        ? 'Deployed token exists'
+        : 'No observed relations — placed by its abstract token assignment',
```
torztomasz, line 476: Minor: `isDeployed` + `hasRelations` admit an impossible state (`false`/`false`), and this nested ternary is the symptom. Not worth a migration now, but if this node type grows again, make it a `kind: 'deployed' | 'missing' | 'unlinked'` discriminant.

packages/token-ui/src/pages/tokens/TokenRelationsGraphPage.tsx
adamiak: Agreed — fixed as suggested; coming in the next push. `useDeferredValue(withoutRelationsMode)` with `isRelayouting` derived from the two copies differing, and the scene build hoisted into a `useMemo` in `TokenRelationsGraph`. The pending/applied pair, the double-rAF effect, the empty-graph branch and the `onLayoutComplete` plumbing are all gone (~60 lines net), and since the busy flag is now derived it can no longer get stuck. Also added the restart caveat as a comment.

packages/token-ui/src/pages/tokens/TokenRelationsGraph.tsx
adamiak: Done as part of the `useDeferredValue` fix — `const scene = useMemo(() => buildRelationGraphScene(graph), [graph])`, the lifecycle effect depends on `scene` (its only `graph` read was the build), and `onLayoutComplete` + ref + sync effect are deleted. Coming in the next push.

packages/token-ui/src/pages/tokens/relationGraphScene.ts
adamiak: Leaving this one as is. The virtual-links mechanism is decision-support scaffolding — it may well get deleted once we decide what to do with tokens on chains interop doesn't cover — so I'd rather not export/reshape the layout module's internals to serve possibly-temporary code. The duplication is contained in `buildVirtualLinks`, deterministic, and covered by tests. If the feature survives the decision, consolidating on a shared `groupIntoClusters` is the follow-up I'd make.

packages/token-backend/src/trpc/routers/deployedTokens/index.ts
adamiak: Not changing this in this PR: it's perf-only, on an internal endpoint, and the tokens-without-relations payload may be short-lived depending on the chain-support decision. If it survives and graph load time starts to matter, `getByAbstractTokenIds` mirroring the `getByPrimaryKeys` batching is exactly the fix I'd make. Same for the `graphNode` literal dedup — will fold it in next time this code is touched.

packages/token-ui/src/pages/tokens/relationGraphModel.ts
adamiak: Keeping the named helper for now — the name documents what the set means, and it's three lines of surface on a feature that may not outlive the chain-support decision, so I'd rather not churn it. If the picker stays, I'll inline it on the next touch.

packages/token-ui/src/pages/tokens/TokenRelationsGraph.tsx
adamiak: Agreed — if this node type grows a third flag, a `kind: 'deployed' | 'missing' | 'unlinked'` discriminant is the right move. Leaving as is for now, as you suggest.

packages/token-backend/src/trpc/routers/deployedTokens/index.ts
```diff
@@ -173,17 +173,51 @@ export const deployedTokensRouter = (deps: DeployedTokensRouterDeps) =>
...
+        endpointNodes.flatMap((node) => node.abstractTokenId ?? []),
+      )
+      const endpointIds = new Set(endpointNodes.map((node) => node.id))
+      const nodesWithoutRelations = (await ctx.tokenDb.deployedToken.getAll())
+        .filter(
+          (token) =>
+            !token.ignored &&
+            token.abstractTokenId !== null &&
+            endpointAbstractTokenIds.has(token.abstractTokenId) &&
+            !endpointIds.has(tokenKey(token)),
```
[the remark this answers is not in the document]
adamiak: Intentional. `nonMinting` relations aren't something the graph (or, currently, the TokenRelations table) is meant to cover — they're usually swapping relations, and for now we only want non-swapping relations there. The graph deliberately draws `burnAndMint`/`lockAndMint` edges only, so a token whose relations are all `nonMinting` has no edge to draw; treating it as "without relations" and placing it by its abstract token assignment is exactly what we want, whereas excluding it would remove it from the graph entirely.

PR #12674, opened 2026-08-27
Extract withClient fallback loop in BlockProvider

torztomasz:
## Summary
- Pure refactor: the four `BlockProvider` methods repeated the same try-the-next-client loop; it now lives in one private `withClient` helper.
- First of a stack that cuts backend `eth_getBlockByNumber` usage (~96% of all backend RPC calls, 8.5M/day): next are `getBlockHeader` support, the block-sync tip poll and the timestamp search.

## Testing
- `packages/shared`: typecheck, lint, tests (410 passing, unchanged)


torztomasz:
Folded into #12675 (first commit) so the optimizations do not stack on a refactor.

PR #12675, opened 2026-08-27
Add getBlockHeader to block clients and BlockProvider

torztomasz:
## Summary
- `BlockClient` now **requires** `getBlockHeader(number | 'latest')` — a block without transaction bodies (number, hash, timestamp) — so callers that only need those don't download full blocks. Implemented for every remaining block client:
  - `RpcClient` / `RpcClientCompat`: `eth_getBlockByNumber(n, false)`
  - `FuelClient`: the block GraphQL query without `transactionIds`
  - `StarknetClient`: `starknet_getBlockWithTxHashes` (also now backs `getBlockTimestamps`, which previously pulled full `getBlockWithTxs` blocks); adds `block_hash` to the block schema
  - `CelestiaRpcClient`, `PolkadotRpcClient`: the header of their existing block query
- `BlockProvider.getBlockHeader` with the usual client fallback; since the capability is required there is no full-block fallback path.
- The first commit extracts the try-the-next-client loop that every `BlockProvider` method repeated into one `withClient` helper.
- **Based on #12683** (delete archived Loopring/DeGate/zkSync Lite clients) so the required method never has to be implemented on clients that are being removed. Merge #12683 first.
- Additive to callers (no behavior change). Prerequisite of #12676 (tip poll) and #12677 (timestamp search), which are independent of each other. #12681 doesn't need it.

## Testing
- `packages/shared`: typecheck, lint, tests (405 passing; new getBlockHeader tests for RPC, Fuel, Starknet, Celestia, Polkadot)
- `packages/config`, `packages/backend`: typecheck, tests (1349 passing)

### DIFF
--- a/packages/backend/src/modules/interop/examples/snapshot/replay.ts
+++ b/packages/backend/src/modules/interop/examples/snapshot/replay.ts
@@ -70,6 +70,10 @@ export class RpcReplay implements Omit<RpcClientCompat, 'ethRpcClient'> {
     throw new ReplayError(key)
   }
 
+  getBlockHeader(blockNumber: number | 'latest'): Promise<EVMBlock> {
+    return this.getBlock(blockNumber, false)
+  }
+
   getBlockParentBeaconRoot(blockNumber: number): Promise<string> {
     const key = this.buildSnapshotKey([
       'blockParentBeaconRoot',
--- a/packages/shared/src/clients/fuel/FuelClient.test.ts
+++ b/packages/shared/src/clients/fuel/FuelClient.test.ts
@@ -39,6 +39,44 @@ describe(FuelClient.name, () => {
     })
   })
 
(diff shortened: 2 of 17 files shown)


mateuszradomski:
@codex, review

mateuszradomski:
@codex, review

mateuszradomski (approved):

torztomasz:
Closing: the RPC proxy caches `eth_getBlockByNumber` per full params, so `(n, false)` and `(n, true)` are separate entries — header requests would be billed misses wherever the full block is already cached, and would not pre-warm block-sync's full-block fetches. Without a proxy change to serve headers from cached full blocks, the header capability has no consumer worth shipping. Decision: keep full-block fetches; the billed volume is dominated by staging-only traffic and one fetch per block per synced chain (details in the #12676/#12677 threads).

PR #12678, opened 2026-08-27
Seed BlockProvider timestamp search with the previous result

torztomasz:
## Summary
- Consumers ask for increasing hourly timestamps, but every search started from block 1, where interpolation is poor (early block times are irregular): ~8–12 probes per hourly tick per chain.
- The provider remembers the last found header and uses it as the lower bound whenever it is not after the requested timestamp; interpolating between an hour-old block and `latest` converges in ~2–3 probes. Stacked on #12677.

## Testing
- Two new tests: the seed is used for a later timestamp and ignored for an earlier one (417 passing)


torztomasz:
Folded into #12677 so the timestamp search is one independent PR.

PR #12679, opened 2026-08-27
Add block observers to block sync

torztomasz:
## Summary
- `ModuleDependencies.blockObservers`: `BlockProcessor`-shaped listeners that receive the consistent blocks of the chains block sync already fetches for its processors. They never cause a chain to be synced (chains still come from `blockProcessors` only) and run in a silent loop: no per-block "Processor finished" log line, failures are logged and skipped.
- Nothing registers an observer yet; the activity module will, to stop re-downloading blocks that block sync already has. Independent of the `getBlockHeader` stack.

## Testing
- `BlockIndexer` test: observers get the consistent blocks with their logs, an observer error does not stop processing
- `packages/backend`: typecheck, lint, tests (1350 passing)

### DIFF
--- a/packages/backend/src/Application.ts
+++ b/packages/backend/src/Application.ts
@@ -48,6 +48,7 @@ export class Application {
       providers,
       db,
       blockProcessors: [],
+      blockObservers: [],
     }
 
     // Modules with TRPC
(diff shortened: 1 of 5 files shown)


torztomasz:
Folded into #12681 so the activity change is one standalone PR.

PR #12680, opened 2026-08-27
Add ActivityBlockCache

torztomasz:
## Summary
- Per-chain ring of activity summaries (`number`, `timestamp`, `txsCount`, `uopsCount`) keyed by `number % capacity`, kept in typed arrays: 100k slots ≈ 2.4 MB, allocated lazily on the first write.
- Not used yet; the next PR fills it from block sync and reads it before the RPC. Stacked on #12679.

## Testing
- Unit tests: roundtrip, null uops count, slot eviction, empty cache (1354 passing)


torztomasz:
Folded into #12681 so the activity change is one standalone PR.

PR #12691, opened 2026-08-27
Dedupe tracked-tx records per transaction (Aztec batched epoch proofs)

vilcht:
## Stack

1. 👉 **this PR** — backend dedupe for multi-call transactions
2. #12695 — database transition for the `'none'` sentinel
3. #12696 — liveness primary key per grouping key

## What happened

On Aug 26 `TrackedTxsIndexer` got stuck, endlessly retrying an insert failing with `duplicate key value violates unique constraint "L2Cost_pkey"`. The story in steps:

1. An Aztec transaction always carried exactly one `submitEpochRootProof` call — the prover node called the Rollup directly, and our tracking was built on that shape.
2. Aztec v5.1 (2026-07-22) added `PROVER_NODE_PROOF_SUBMISSION_TARGET_ADDRESS`, letting provers submit through a wrapper contract. On Aug 26 a prover started using one (`0x9ea4…ee8d`) that batches **two epoch proofs into a single Ethereum transaction**.
3. Our Dune query works per internal call, so it correctly returned two rows for such a transaction. Before the liveness `groupBy` change (#12349) only the 4-byte selector was selected, so the two rows looked identical and `SELECT DISTINCT` collapsed them *by accident*. With the grouping parameter now selected, both rows survive — the query is right; it exposed a dormant assumption downstream.
4. `L2CostsUpdater` turned both rows into records sharing the `(configurationId, txHash)` primary key. The insert failed, the transaction rolled back, and the indexer could not advance.

## The fix

- `TrackedTxsIndexer` dedupes the **l2costs** stream per `(configuration, transaction)`: a transaction has one gas bill, and both rows carried the same transaction-level values, so this is pure dedupe with zero information loss.
- The **liveness** stream is intentionally *not* deduped there: one transaction can legitimately carry several grouped records (one per epoch, each with its own grouping key). `LivenessUpdater` dedupes by storage identity `(configuration, transaction, groupingKey)` instead, collapsing only byte-identical duplicates (possible for configs that select full calldata, e.g. SHARP).
- Repository inserts stay strict — no `ON CONFLICT DO NOTHING` — so a real conflict with committed data still fails loudly instead of concealing future transform bugs.
- The `transform`-level dedupe test moved to `TrackedTxsIndexer` together with the behavior it covers; all other existing tests are untouched.

## What this PR does NOT fix

The `Liveness` primary key is still `(configurationId, txHash)`, so the table cannot yet hold two epochs proven by one transaction — a wrapper transaction that is the *first* submission for both of its epochs would crash liveness the same way (it hasn't happened yet only because every observed batch contained one already-known epoch). That is what the rest of the stack (#12695, #12696) fixes.

### DIFF
--- a/packages/backend/src/modules/tracked-txs/modules/liveness/LivenessUpdater.ts
+++ b/packages/backend/src/modules/tracked-txs/modules/liveness/LivenessUpdater.ts
@@ -22,8 +22,16 @@ export class LivenessUpdater implements TxUpdaterInterface<'liveness'> {
     }
 
     const transformedTransactions = this.transformTransactions(transactions)
+    const deduplicated = transactions.length - transformedTransactions.length
+    if (deduplicated > 0) {
+      this.logger.info('Deduplicated liveness transactions', {
+        count: deduplicated,
+      })
+    }
     await this.db.liveness.insertMany(transformedTransactions)
-    this.logger.info('Updated liveness', { count: transactions.length })
+    this.logger.info('Updated liveness', {
+      count: transformedTransactions.length,
+    })
   }
 
   async deleteFromById(id: TrackedTxId, fromInclusive: UnixTime) {
@@ -31,7 +39,9 @@ export class LivenessUpdater implements TxUpdaterInterface<'liveness'> {
   }
 
   transformTransactions(transactions: TrackedTxResult[]): LivenessRecord[] {
-    return transactions.map((transaction) => {
+    const records = new Map<string, LivenessRecord>()
+
+    for (const transaction of transactions) {
       const record: LivenessRecord = {
         timestamp: transaction.blockTimestamp,
         blockNumber: transaction.blockNumber,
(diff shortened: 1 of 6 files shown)


packages/backend/src/modules/tracked-txs/modules/liveness/LivenessUpdater.ts
```diff
@@ -47,7 +57,12 @@ export class LivenessUpdater implements TxUpdaterInterface<'liveness'> {
...
 
-      return record
-    })
+      const key = `${record.configurationId}-${record.txHash}-${record.groupingKey ?? ''}`
+      if (!records.has(key)) {
+        records.set(key, record)
+      }
+    }
+
+    return [...records.values()]
```
torztomasz: I guess we do not need this at all anymore?

torztomasz (changes requested):

torztomasz (approved):

PR #12695, opened 2026-08-28
Prepare liveness grouping for the 'none' sentinel

vilcht:
## Stack

1. #12691 — backend dedupe for multi-call transactions
2. 👉 **this PR** — database transition for the `'none'` sentinel
3. #12696 — liveness primary key per grouping key (stacked on this PR)

## Why

One Ethereum transaction can now prove several Aztec epochs (see #12691 for the full story). To store one liveness record per epoch, the follow-up PR changes the `Liveness` primary key to `(configurationId, txHash, groupingKey)`, with `'none'` marking ungrouped records. This PR makes the database package work with **both** the current and the future schema, so the migration later deploys with no failure window.

## Changes

- New partial unique grouping index `(configurationId, groupingKey) WHERE groupingKey <> 'none'`, built `CONCURRENTLY`. Today it is functionally identical to the existing `IS NOT NULL` index; it is the only one that can survive the sentinel backfill.
- Ungrouped inserts **omit** `groupingKey` entirely, letting the column default decide the stored value: `NULL` on today's schema, `'none'` after the follow-up migrations.
- Reads map both `NULL` and `'none'` to `undefined` — the public `LivenessRecord` is unchanged for all consumers.
- The grouped upsert's `ON CONFLICT` target becomes `groupingKey <> 'none'`, verified against PostgreSQL to match **both** the old and the new index (`x <> 'none'` implies `x IS NOT NULL`).
- `RealTimeLiveness` deliberately keeps the nullable convention (its writer only inspects top-level calldata and cannot produce multi-epoch records), hence the per-table conflict target in `livenessGrouping.ts`.

## Deploy notes

- Stacked on #12691 (stack #12697) — merges after it.
- Must be **deployed** before the follow-up PR's migrations run.

### DIFF
--- a/packages/database/prisma/migrations/20260827130000_create_liveness_grouping_index_excluding_none/migration.sql
+++ b/packages/database/prisma/migrations/20260827130000_create_liveness_grouping_index_excluding_none/migration.sql
@@ -1,0 +1,8 @@
+-- Keep this as a single statement so PostgreSQL can build the index
+-- concurrently, outside an implicit multi-statement transaction.
+-- Replaces the IS NOT NULL grouping index: ungrouped records are about to
+-- receive the 'none' sentinel and must stay out of the unique grouping
+-- constraint.
+CREATE UNIQUE INDEX CONCURRENTLY "Liveness_configurationId_groupingKey_idx"
+ON "Liveness"("configurationId", "groupingKey")
+WHERE "groupingKey" <> 'none';
--- a/packages/database/prisma/migrations/20260827130000_create_liveness_grouping_index_excluding_none/migration.sql
+++ b/packages/database/prisma/migrations/20260827130000_create_liveness_grouping_index_excluding_none/migration.sql
@@ -5,4 +5,4 @@
 -- constraint.
 CREATE UNIQUE INDEX CONCURRENTLY "Liveness_configurationId_groupingKey_idx"
 ON "Liveness"("configurationId", "groupingKey")
-WHERE "groupingKey" <> 'none';
+WHERE "groupingKey" != 'none';
--- a/packages/database/src/repositories/utils/livenessGrouping.ts
+++ b/packages/database/src/repositories/utils/livenessGrouping.ts
@@ -63,7 +63,7 @@ export async function insertGroupedKeepingEarliest(
       // The conflict target must match each table's partial grouping index.
       const guarded =
         table === 'Liveness'
-          ? target.where('groupingKey', '<>', UNGROUPED_GROUPING_KEY)
+          ? target.where('groupingKey', '!=', UNGROUPED_GROUPING_KEY)
           : target.where('groupingKey', 'is not', null)
       return guarded
         .doUpdateSet((eb) => ({
--- a/packages/database/src/repositories/utils/livenessGrouping.ts
+++ b/packages/database/src/repositories/utils/livenessGrouping.ts
@@ -4,6 +4,12 @@ import type { Liveness, RealTimeLiveness } from '../../kysely/generated/types'
 
 type LivenessTable = 'Liveness' | 'RealTimeLiveness'
 
+/**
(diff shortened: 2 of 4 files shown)


packages/database/prisma/migrations/20260827130000_create_liveness_grouping_index_excluding_none/migration.sql
```diff
@@ -0,0 +1,8 @@
+-- Keep this as a single statement so PostgreSQL can build the index
+-- concurrently, outside an implicit multi-statement transaction.
+-- Replaces the IS NOT NULL grouping index: ungrouped records are about to
+-- receive the 'none' sentinel and must stay out of the unique grouping
+-- constraint.
+CREATE UNIQUE INDEX CONCURRENTLY "Liveness_configurationId_groupingKey_idx"
```
torztomasz, line 6: do we really need that unique index?

packages/database/prisma/migrations/20260827130000_create_liveness_grouping_index_excluding_none/migration.sql
```diff
@@ -0,0 +1,8 @@
...
+ON "Liveness"("configurationId", "groupingKey")
+WHERE "groupingKey" <> 'none';
```
torztomasz: can we use `!=` instead of `<>`

packages/database/src/repositories/utils/livenessGrouping.ts
```diff
@@ -52,17 +58,21 @@ export async function insertGroupedKeepingEarliest(
...
-    .onConflict((cb) =>
-      cb
-        .columns(['configurationId', 'groupingKey'])
-        .where('groupingKey', 'is not', null)
+    .onConflict((cb) => {
+      const target = cb.columns(['configurationId', 'groupingKey'])
+      // The conflict target must match each table's partial grouping index.
+      const guarded =
+        table === 'Liveness'
+          ? target.where('groupingKey', '<>', UNGROUPED_GROUPING_KEY)
```
torztomasz: same

packages/database/src/repositories/utils/livenessGrouping.ts
```diff
@@ -52,17 +58,21 @@ export async function insertGroupedKeepingEarliest(
...
     .values(rows)
-    .onConflict((cb) =>
-      cb
-        .columns(['configurationId', 'groupingKey'])
-        .where('groupingKey', 'is not', null)
+    .onConflict((cb) => {
+      const target = cb.columns(['configurationId', 'groupingKey'])
+      // The conflict target must match each table's partial grouping index.
+      const guarded =
+        table === 'Liveness'
```
torztomasz, line 65: don't we also need to handle RealTimeLiveness?

torztomasz (changes requested):

packages/database/prisma/migrations/20260827130000_create_liveness_grouping_index_excluding_none/migration.sql
vilcht: Yep, we need it, the "do nothing / upsert" earlier tx for the epoch depends on ON CONFLICT

packages/database/src/repositories/utils/livenessGrouping.ts
vilcht: RealTimeLiveness can't hit the same problem as the processor only matches top level call data, not the wrapper one. This raises a related problem, the RTL records nothing for Aztec's wrapper based proofs. This could raise a false anomaly, but other provers still submit directly (~75 RTL records/48h), so a false anomaly would need the wrapper prover to become the sole submitter for a ≥ 15 gap  and even that only produces a potential anomaly to be approved by research not bad public data. I haven't found a proper fix thats not very expensive so i would just leave it as is.

torztomasz (approved):

PR #12698, opened 2026-08-28
Add payy

sergeyshemyakov:
Resolves L2B-14783


packages/config/src/projects/payy/detailedDescription.md
```diff
@@ -0,0 +1,19 @@
+Payy is a payments-focused ZK network, designed around private stablecoin transfers. Ethereum receives only state roots and the public deposit and withdrawal messages of each state update. Balances live offchain as UTXO-style notes whose commitments form a sparse Merkle tree maintained by the Payy validators.
```
sekuba, line 1: description from payy.ts gets prepended here and makes this partly redundant (see preview)

packages/config/src/projects/payy/payy.ts
```diff
@@ -0,0 +1,342 @@
...
+        passed: false,
+        reason:
+          'Withdrawals must be included in a proven state update by the permissioned Payy operators. There is no mechanism for users to exit on Ethereum if the operators halt.',
+      },
+    },
+    reproducibility: {
+      value: 'Reproducible',
+      sentiment: 'good',
+      description:
+        'The contracts, Noir circuits and node software are published in the Payy repository and can be built and run locally.',
```
sekuba: DA is not on ethereum and there is no DA verifier, so you cannot count on reproducing state, might want to add this here too because users will not go to the scaling page i think. also: what about frontend? (looks very closed) and api

packages/config/src/projects/payy/payy.ts
```diff
@@ -0,0 +1,342 @@
...
+    },
+    privacy: {
+      value: 'Linkable notes',
+      sentiment: 'warning',
+      description:
+        'Transactions within the Payy network do not hide the links between token transferred: anyone with Payy block data can reconstruct the spend graph. Only the contents of each note (owner address, amount and asset) are hidden.',
```
sekuba: so like zama? i wonder if both this and zama should be red, the practical privacy is quasi nothing (see also https://x.com/unhappyben/status/2090825255476445519)

packages/config/src/projects/payy/payy.ts
```diff
@@ -0,0 +1,342 @@
...
+      risks: [
+        'The Payy-operated note registry, note lookup and wallet backup services learn which addresses receive and hold which notes and can link senders to recipients. The same operator holds KYC data for users of the Payy card and fiat ramps, which can tie those addresses to real identities.',
+        'A compromise of a recipient key or of the encryption scheme exposes all note data ever stored in the encrypted registry.',
+      ],
+    },
+    attributes: [
+      PRIVACY_ATTRIBUTES.zk,
+      PRIVACY_ATTRIBUTES.transfers,
```
sekuba: use private amounts instead of transfers like for zama?

packages/config/src/projects/payy/payy.ts
```diff
@@ -0,0 +1,342 @@
...
+    },
+    privacy: {
+      value: 'Linkable notes',
+      sentiment: 'warning',
+      description:
+        'Transactions within the Payy network do not hide the links between token transferred: anyone with Payy block data can reconstruct the spend graph. Only the contents of each note (owner address, amount and asset) are hidden.',
+    },
+    noteDiscovery: {
+      description:
+        "Funds are held as UTXO-like notes and the L2 chain state stores only their commitments, so a recipient cannot reconstruct an incoming note from the chain alone. Payy runs an encrypted registry, where the sender submits full note data, encrypted to the receiver.\n\nThis registry also stores the recipient's public key in the clear, and clients poll it for their own address (`GET /registry/notes/{public_key}`). So the centralized operator learns which addresses receive notes and when, and who transacted with whom, while note values stay encrypted.\n\nThe protocol itself allows direct peer-to-peer note communication that bypasses the registry.",
```
sekuba: tldr centralized note discovery, operator learns transaction graph?
can you make this less verbose

packages/config/src/projects/payy/riskSummary.md
```diff
@@ -0,0 +1,12 @@
+## Funds can be stolen if
+1. the proof system is broken and a malicious verifier is registered, allowing forged spends or withdrawals.
```
sekuba: replace and with or (also wouldn't we have a standard sentence for this?)

sekuba (changes requested):

packages/config/src/projects/payy/payy.ts
sergeyshemyakov: - Frontend sources are not available, I was confused initially by this package: https://github.com/polybase/payy/tree/main/app
- About DA: I still need to dig a bit deeper. In principle, their data is on celestia, which is good enough for reproducibility on privacy dashboard. In practice, my fable was not able to find their namespace on celestia, so I could not independently check whether they are really posting anything. I will not update this for now.

packages/config/src/projects/payy/riskSummary.md
sergeyshemyakov: I meant validator, not verifier. You need both zk + the mighty 2/3 consensus on 1 validator to settle on L1

packages/config/src/projects/payy/payy.ts
sergeyshemyakov: Screenshot 2026-08-28 at 17 03 17

sergeyshemyakov: Yeah, I do think they should be red. In practice you can't move your tokens through Payy / Zama to a reasonably unlinkable new address, IMO this should be a minimal criterion for yellow

sekuba:
i think you can make zama privacy red too in this pr

packages/config/src/projects/payy/payy.ts
```diff
@@ -0,0 +1,342 @@
...
+import { BADGES } from '../../common/badges'
+import { PRIVACY_ATTRIBUTES } from '../../common/privacyAttributes'
+import { ProjectDiscovery } from '../../discovery/ProjectDiscovery'
+import type { ScalingProject } from '../../internalTypes'
+import { getDiscoveryInfo } from '../../templates/getDiscoveryInfo'
+import { readProjectMarkdown } from '../../utils/readMarkdown'
+
+const discovery = new ProjectDiscovery('payy')
+
+const upgradesAndGovernanceContent = `All privileged functions are ultimately controlled by the PayyMultisig (${discovery.getMultisigStats('PayyMultisig')}). It owns the ProxyAdmin, which can instantly upgrade the Rollup proxy, and it is the owner of the Rollup contract, allowing it to manage provers, validators, ZK verifiers, supported tokens and burn substitutors, and to directly overwrite the state root via setRoot(). There is no timelock or exit window on any of these actions.`
```
torztomasz: Could you extract `upgradesAndGovernanceContent` to a separate `upgradesAndGovernance.md` and load it with `readProjectMarkdown`? That matches the convention (e.g. railgun) and keeps the sibling `riskSummary.md` company.

packages/config/src/projects/payy/payy.ts
```diff
@@ -0,0 +1,342 @@
...
+
+export const payy: ScalingProject = {
+  type: 'layer2',
+  id: ProjectId('payy'),
+  capability: 'appchain',
+  addedAt: UnixTime(1787616000), // 2026-08-25T00:00:00Z
+  badges: [BADGES.VM.AppChain, BADGES.DA.Celestia, BADGES.Other.Privacy],
```
torztomasz: `BADGES.DA.Celestia` here is inconsistent with `dataAvailability.layer: DA_LAYERS.NONE` below (and the prose says the Celestia commitment check is an empty placeholder). The badge implies DA that the rest of the config denies — worth resolving one way or the other.

packages/config/src/projects/payy/payy.ts
```diff
@@ -0,0 +1,342 @@
...
+      orderHint: 0,
+      description:
+        'The PayyMultisig can instantly upgrade the Rollup contract, replace ZK verifiers and directly overwrite the state root, so users have no time to exit before a malicious change takes effect.',
+      walkawayTest: {
+        passed: false,
+        reason:
+          'Withdrawals must be included in a proven state update by the permissioned Payy operators. There is no mechanism for users to exit on Ethereum if the operators halt.',
+      },
+    },
+    reproducibility: {
```
torztomasz, line 270: `value: 'Reproducible'` seems to overstate what its own description supports — wallet/UI sources are unpublished and the repo has no license. Maybe a more qualified value would fit better here?

packages/config/src/projects/payy/payy.ts
```diff
@@ -0,0 +1,342 @@
...
+  },
+  dataAvailability: {
+    layer: DA_LAYERS.NONE,
+    bridge: DA_BRIDGES.NONE,
+    mode: DA_MODES.TRANSACTION_DATA,
+  },
+  chainConfig: {
+    name: 'payy',
+    chainId: undefined,
+    apis: [
```
torztomasz: The `payy` API config is dead code at this point: the `ChainBasicApi<'payy'>` union member, the no-op `case 'payy'` in `getChainConfig.ts` and this `apis` entry have no consumer — unlike lighter, whose API type is consumed by `LighterDayProvider` (`DayProviders.ts:42`). Suggest dropping all three (`apis: []` is valid) and parking the validator URL in the `TODO: activityConfig` comment, then reintroducing them together with `activityConfig` once the Payy day provider exists — they are one atomic feature.

torztomasz:
Pushed 3d8df12 to fix the detailedDescription problem: privacy now owns its detailed description via a new `ProjectPrivacyInfo.detailedDescription` field (falling back to `display.detailedDescription`, so the 8 existing standalone privacy projects need no changes). This mirrors how scaling, bridges and interop each own theirs (`scalingTechnology`/`bridgeTechnology`/`interopConfig.detailedDescription`) — the privacy page reading the generic `display` one was the outlier, and for processed scaling projects that field never reaches `BaseProject.display` anyway (it's mapped into `scalingTechnology` in `layer2Or3ToProject`). Payy's privacy-flavored `detailedDescription.md` moved from `display` to `privacyInfo`, so it now renders on the privacy page; the L2 page has no detailed description until someone writes a scaling-flavored one (just set `display.detailedDescription`). Config + frontend typecheck clean, all config tests pass.

sekuba (approved):

PR #12700, opened 2026-08-28
Bound Relay updates by time window instead of row count

mateuszradomski:
The watermark was inferred from the last fetched row, so progress depended on data density. A burst of 619 requests sharing updatedAt second 1787583059 overran the 500 row budget, every fetched row floored to that same second, syncedTo computed to from - 1, and the indexer threw "No entries" on every attempt for four days while 619 real transfers sat in the window.

Bounding by time instead makes syncedTo known before the fetch, so progress no longer depends on how many rows a second holds, and an empty window advances normally. This matches BlockActivityIndexer and BlobIndexer. endTimestamp is exclusive, hence syncedTo + 1. A window that does not fit in one update throws rather than advancing, since committing a partial window would skip entries.

### DIFF
--- a/packages/backend/src/modules/interop/plugins/relay/relay.indexer.ts
+++ b/packages/backend/src/modules/interop/plugins/relay/relay.indexer.ts
@@ -13,9 +13,20 @@ import { buildRelayBootstrapChainNamesById, RelayConfig } from './relay.config'
 type RelayMetadata = GetRequestsResponse['requests'][number]['data']['metadata']
 type RelayCurrency = NonNullable<RelayMetadata>['currencyIn']
 
-export const SAFE_TIME_OFFSET = 10
+export interface RelayIndexerConfig {
+  batchSize: number
+  maxRequestsPerUpdate: number
+  safeTimeOffset: number
+}
 
 export class RelayRootIndexer extends RootIndexer {
+  constructor(
+    logger: Logger,
+    private readonly safeTimeOffset: number,
+  ) {
+    super(logger)
+  }
+
   override initialize() {
     setInterval(() => this.requestTick(), 1_000)
     this.requestTick()
@@ -23,7 +34,7 @@ export class RelayRootIndexer extends RootIndexer {
   }
 
   tick(): Promise<number> {
-    return Promise.resolve(UnixTime.now() - SAFE_TIME_OFFSET)
+    return Promise.resolve(UnixTime.now() - this.safeTimeOffset)
   }
 }
 
@@ -51,9 +62,6 @@ export const TokenReceived = createInteropEventType<TokenReceivedArgs>(
   { direction: 'incoming' },
 )
 
-export const BATCH_SIZE = 60
-export const MAX_REQUESTS_PER_UPDATE = 10_000
-
 export class RelayIndexer extends ManagedChildIndexer {
   private sentIds = new Set<string>()
   private receivedIds = new Set<string>()
@@ -63,6 +71,7 @@ export class RelayIndexer extends ManagedChildIndexer {
     chains: { id: number; name: string }[],
     private configs: InteropConfigStore,
     private trackedChains: string[],
+    private readonly relayConfig: RelayIndexerConfig,
     private relayApiClient: RelayApiClient,
     private db: Database,
     private interopEventStore: InteropEventStore,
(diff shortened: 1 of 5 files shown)


mateuszradomski:
@codex, review

packages/backend/src/modules/interop/plugins/relay/relay.indexer.ts
```diff
@@ -111,35 +116,26 @@ export class RelayIndexer extends ManagedChildIndexer {
       return to
     }
```
adamiak, line 126: **Silent data loss at the tip.** In steady state the parent tick is `UnixTime.now()`, so `syncedTo` here equals the *current, still-filling* wall-clock second: we fetch `[now, now+1)` while most of that second hasn't happened yet, commit `syncedTo = now` permanently, and the next window starts at `now + 1`. Any request whose `updatedAt` lands later in that second — plus anything delayed by the API's own write/replication lag — is never fetched again, with no error and no log.

The old `last.updatedAt - 1` watermark accidentally guarded this by always trailing observed data. And unlike `BlockActivityIndexer`/`BlobIndexer`, whose parents are settled block heights, this parent is a wall clock, so the cited precedent doesn't carry over at the head.

Suggestion: add a safety lag — have `RelayRootIndexer.tick()` return `UnixTime.now() - N` (like `BlockNumberIndexer`'s `delayFromTipInSeconds`), or clamp `syncedTo` strictly below `now`.

packages/backend/src/modules/interop/plugins/relay/relay.indexer.ts
```diff
@@ -111,35 +116,26 @@ export class RelayIndexer extends ManagedChildIndexer {
...
 
+    const syncedTo = from + BATCH_SIZE < to ? from + BATCH_SIZE : to
+
     const res = await this.relayApiClient.getAllRequests({
-      limit: 500,
       startTimestamp: from,
+      endTimestamp: syncedTo + 1,
+      limit: MAX_REQUESTS_PER_UPDATE,
     })
```
adamiak, line 135: **This check throws on windows that were in fact fully fetched, and with the fixed window + infinite retry each false positive is a permanent stall.**

1. `getAllRequests` exits on truthiness (`while (continuation && limit > 0)`) and then assigns the last cursor unconditionally — an empty-string continuation on a final page is "complete" for the client but `'' !== undefined` throws here, on every retry, forever.
2. When the row budget runs out on a page that happens to be the last one (window holds exactly the limit, or the API returns a cursor on any full final page), everything was fetched but `result.continuation` is still set — `RelayApiClient.test.ts` ("reports the cursor when the limit stops the walk") pins exactly this conflation.

Also worth knowing: the client's partial-failure path (transient 5xx mid-pagination) breaks and returns a defined continuation, so this error message points operators at window density when the actual cause was transport.

Suggestion: at minimum `if (res.continuation)`; better, treat it as overflow only together with `res.requests.length >= MAX_REQUESTS_PER_UPDATE`, and surface the client's failure reason so density and transport errors are distinguishable.

packages/backend/src/modules/interop/plugins/relay/relay.indexer.ts
```diff
@@ -49,6 +51,9 @@ export const TokenReceived = createInteropEventType<TokenReceivedArgs>(
   { direction: 'incoming' },
 )
```
adamiak, line 63: **A dense window is still a permanent stall — the threshold moved from 500 rows to 10k/60s, but escaping it requires a redeploy.** The incident this PR fixes had 619 requests in a single second, so ~165 req/s sustained over the 61s window trips the limit (non-`success` rows also count against the budget). When it happens: the client pages at 50 rows, so each attempt burns up to 200 sequential HTTP calls, throws, and the infinite retry re-runs the identical fixed window forever — same failure class as the 4-day stall, and nothing can shrink the window or resume from the returned cursor.

Both indexers this PR mirrors take `batchSize` as an injected dependency, and `InteropModule` already threads `config.interop.financials.batchSize`. Making the window width (and row budget) injectable turns an incident redeploy into a config change.

adamiak (commented):
High-severity findings from review (details inline).

adamiak (approved):

mateuszradomski:
@codex, review

PR #12705, opened 2026-08-31
Configure privacy anonymity sets

sdlyy:
## Summary

Defines the configuration contract for privacy anonymity-set tracking:

- makes bucket minimum amounts the positive tracking source of truth
- restricts tracked buckets to supported EVM deposit extractors
- configures Tornado Cash, Privacy Pools, and Railgun
- marks Umbra and Zama as not applicable
- adds configuration invariants
- adds the shared stable ingestion-configuration identity

The configuration is inert in this PR: no indexer consumes it yet.

## Validation

- shared focused tests, typecheck, lint, and build
- config privacy-project tests, typecheck, and lint

## Stack

1. #12704 - storage
2. #12705 - configuration
3. #12706 - indexer
4. #12707 - frontend

PR #12704, opened 2026-08-31
Add privacy anonymity set storage

sdlyy:
## Summary

Adds the dedicated database foundation for privacy anonymity-set tracking:

- creates the PrivacyAnonymitySetEvent table and migration
- registers the repository and database exports
- supports sender-day aggregation, idempotent writes, trim, and wipe
- adds repository coverage

This PR adds storage only. It does not start a writer or expose frontend data.

## Validation

- database typecheck
- database lint
- Prisma schema validation

The database-backed repository suite was not run locally because no test database was authorized; CI should run it.

## Deployment

Merge and apply this migration before enabling the indexer.

## Stack

1. #12704 - storage
2. #12705 - configuration
3. #12706 - indexer
4. #12707 - frontend

PR #12706, opened 2026-08-31
Index privacy anonymity set deposits

sdlyy:
## Summary

Adds the dedicated per-chain anonymity-set ingestion pipeline:

- derives indexer configurations from tracked buckets
- filters logs to the exact timestamp range before resolving senders
- stores Privacy Pools depositors from the event
- enriches Tornado Cash and Railgun deposits with `transaction.from`
- deduplicates transaction hashes and bounds lookups in batches of 25
- shares log-filtering and protocol extraction helpers with the privacy flow indexer
- preserves existing configuration IDs with a golden test
- fails and retries on missing sender data instead of silently undercounting

This PR starts the writer and historical backfill. It does not expose frontend values.

## Validation

- 402 shared tests
- 68 backend privacy tests
- backend typecheck
- backend lint and formatting

## Deployment

The database migration from #12704 must be applied first. With `NEW_CLIENTS_ENABLED`, the compatibility client encodes multiple event signatures as topic-zero alternatives. Block boundary resolution uses the configured block-indexer client when available and can fall back to RPC queries. Monitor indexer progress, update duration, RPC rate limiting, and transaction lookup failures during backfill.

## Stack

1. #12704 - storage
2. #12705 - configuration
3. #12706 - indexer
4. #12707 - frontend


sdlyy:
@codex review

packages/backend/src/config/features/privacy.ts
```diff
@@ -140,8 +151,28 @@ export async function getPrivacyConfig(
...
+  bucket: ProjectPrivacyBucket,
+  source: PrivacyAnonymitySetDepositSource,
+): PrivacyAnonymitySetIndexerConfig {
+  const privacyAddress = getPrivacyBucketAddress(bucket.address)
+  const config: PrivacyAnonymitySetIndexerConfigProperties = {
+    projectId,
+    bucketId: bucket.id,
+    chain: privacyAddress.chain,
+    address: EthereumAddress(privacyAddress.address),
+    sinceTimestamp: bucket.sinceTimestamp,
```
vilcht: This does not honor PRIVACY_MIN_TIMESTAMP, i know it is stated in the docs that this is for independence but we don't have a way to bound the backfill which feels wrong.

docs/L2B-14626-anonymity-set-indexing-review.md
```diff
@@ -0,0 +1,807 @@
...
+function.
+
+## PR stack
+
+The implementation is split by deployment boundary:
+
+1. [#12704](https://github.com/l2beat/l2beat/pull/12704) adds storage.
+2. [#12705](https://github.com/l2beat/l2beat/pull/12705) adds configuration.
+3. [#12706](https://github.com/l2beat/l2beat/pull/12706) adds indexing.
+4. [#12707](https://github.com/l2beat/l2beat/pull/12707) adds aggregation and UI.
```
vilcht: if the doc is to stay, it should just describe the feature and not reference individual PRs i think

packages/backend/src/modules/privacy/indexers/PrivacyAnonymitySetIndexer.ts
```diff
@@ -0,0 +1,296 @@
...
+    )
+
+    const configMap = buildPrivacyLogConfigurationMap(configurations)
+    const rawRecords = extractRawRecords(logs, configMap)
+    if (rawRecords.length === 0) return []
+
+    const [blockTimestamps, transactionSenders] = await Promise.all([
+      this.$.blockProvider.getBlockTimestamps(
+        unique(rawRecords.map((record) => record.log.blockNumber)),
+      ),
```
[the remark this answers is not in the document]
sdlyy: Fixed in `93bd5ebaad`. `getBlockTimestamps` is now a required `IRpcClient` capability and is implemented by `RpcClientCompat` using bounded batches of lightweight block-header requests, with returned block-number validation. `RpcReplay` was updated to preserve the mirrored interface, and focused mapping/batching/mismatch tests were added.

packages/backend/src/config/features/privacy.ts
sdlyy: Addressed in `6489c0774c`. Anonymity-set configurations now clamp `sinceTimestamp` with `PRIVACY_MIN_TIMESTAMP`, matching the existing privacy flow and price behavior and providing an operational backfill bound. The configuration identity remains stable when the timestamp changes, so an increased floor uses the existing `ManagedMultiIndexer` trimming path. Added a focused `getPrivacyConfig` test covering the configured anonymity-set buckets.

packages/shared/src/clients2/compatibility/RpcClientCompat.ts
```diff
@@ -155,6 +158,38 @@ export class RpcClientCompat implements IRpcClient {
     return toEVMBlock(blockNumber, block)
   }
 
+  async getBlockTimestamps(
```
torztomasz, line 166: i'm not sure it should be in client. client should be pure and basically do the calls, this looks like a function for some service or smth

packages/backend/src/modules/privacy/PrivacyModule.ts
```diff
@@ -93,6 +108,36 @@ export function createPrivacyModule({
...
+            blockProvider: providers.block.getBlockProvider(
+              blockTimestampConfig.chain,
+            ),
+            logsProvider: providers.logs.getLogsProvider(
+              blockTimestampConfig.chain,
+            ),
+            rpcClient: providers.clients.getRpcClient(
+              blockTimestampConfig.chain,
+            ),
+            configurations: anonymitySetConfigs.map((anonymitySetConfig) => ({
```
torztomasz, line 150: shouldnt configurations come from getPrivacyConfig already?

packages/backend/src/modules/privacy/indexers/PrivacyAnonymitySetIndexer.ts
```diff
@@ -0,0 +1,296 @@
...
+        timestamp !== undefined,
+        `Missing block timestamp for block ${record.log.blockNumber}`,
+      )
+
+      if (timestamp < from || timestamp > to) return []
+
+      const sender =
+        record.origin.type === 'event'
+          ? record.origin.sender.toString()
+          : transactionSenders.get(record.log.transactionHash.toLowerCase())
```
torztomasz: I would unify that to some getSenders function instead of having getTransactionSenders and then this ternary

packages/backend/src/config/features/privacy.ts
```diff
@@ -140,8 +156,29 @@ export async function getPrivacyConfig(
...
+    projectId,
+    bucketId: bucket.id,
+    chain: privacyAddress.chain,
+    address: EthereumAddress(privacyAddress.address),
+    sinceTimestamp: Math.max(bucket.sinceTimestamp, minTimestamp),
+    ...source,
+  }
+
+  return {
+    id: createPrivacyAnonymitySetConfigurationId({
```
torztomasz: **Standards (hard):** every sibling indexer exposes `static idToConfigurationId(config)` and `privacy.ts` calls it (`PrivacyPriceIndexer` L123, `PrivacyBlockTimestampIndexer` L142, `StarknetPrivacyFlowIndexer` L209, `PrivacyFlowIndexer` L219). The new indexer alone reaches for `createPrivacyAnonymitySetConfigurationId` from `@l2beat/shared`. Please add the static on `PrivacyAnonymitySetIndexer` and route through it here.

Related: this PR adds a "keeps the existing configuration id" pin test to the four sibling test files, but there is no backend golden for the anonymity-set ID (it relies on the shared test from #12705). Add one alongside the others.

packages/backend/src/modules/privacy/indexers/PrivacyAnonymitySetIndexer.test.ts
```diff
@@ -0,0 +1,253 @@
...
+const TRANSACTION_SENDER = EthereumAddress(
+  '0x3333333333333333333333333333333333333333',
+)
+const FIXED_TOPIC = `0x${'aa'.repeat(32)}`
+const TRANSACTION_HASH = `0x${'bb'.repeat(32)}`
+const privacyPoolsInterface = new utils.Interface([
+  'event Deposited(address indexed depositor, uint256 commitment, uint256 label, uint256 value, uint256 precommitmentHash)',
+])
+
+describe(PrivacyAnonymitySetIndexer.name, () => {
```
torztomasz, line 31: **Standards:** only `multiUpdate` is covered. `wipeData` / `trimData` are part of the ManagedMultiIndexer contract and `PrivacyFlowIndexer.test.ts` L782-868 covers both; mirror that here.

Also, `getTransactionSenders` batching (25 per batch, sequential batches) has no indexer-level test; only the `RpcClientCompat.getBlockTimestamps` batching is tested.

packages/backend/src/modules/privacy/indexers/PrivacyAnonymitySetIndexer.test.ts
```diff
@@ -0,0 +1,253 @@
...
+    maxHeight: null,
+    properties: {
+      id: 'config-1',
+      projectId: 'project-1',
+      bucketId: 'bucket-1',
+      chain: 'ethereum',
+      address: POOL,
+      sinceTimestamp: UnixTime(0),
+      ...extractor,
+    } as PrivacyAnonymitySetIndexerConfig,
```
torztomasz: Minor: the `as PrivacyAnonymitySetIndexerConfig` cast masks the discriminated union, so a wrong `extractor` shape would still compile. Build the properties per variant instead.

packages/backend/src/modules/privacy/indexers/PrivacyAnonymitySetIndexer.ts
```diff
@@ -0,0 +1,296 @@
...
+          tag: $.chain,
+          chain: $.chain,
+        },
+        updateRetryStrategy: Indexer.getInfiniteRetryStrategy(),
+      },
+      logger,
+    )
+  }
+
+  override async multiUpdate(
```
torztomasz, line 74: **Possible Duplicated Code (judgement call, experiment only):** `multiUpdate` / `wipeData` / `trimData` here (L70-144) are near-verbatim copies of `PrivacyFlowIndexer.ts` L59-135, differing only in table and log strings. The doc argues for keeping separate indexers, which is fine; the boilerplate could in principle share a base.

To be clear: this is something to experiment with only if cheap. The human reviewer already noted this is probably redundant given the deliberate separation, so no action required for this PR.

packages/backend/src/modules/privacy/indexers/PrivacyAnonymitySetIndexer.ts
```diff
@@ -0,0 +1,296 @@
...
+          configurationId: configuration.id,
+          from,
+          to,
+          deletedRecords,
+        })
+      }
+    }
+  }
+
+  private async fetchRecordsForGroup(
```
torztomasz: **Possible Mysterious Name (judgement call):** "group" in `fetchRecordsForGroup` is unexplained (inherited from `PrivacyFlowIndexer`). Something like `fetchDepositsForConfigurations` says what it does.

packages/backend/src/modules/privacy/indexers/PrivacyAnonymitySetIndexer.ts
```diff
@@ -0,0 +1,296 @@
...
+    configurations: Configuration<PrivacyAnonymitySetIndexerConfig>[],
+    from: number,
+    to: number,
+  ): Promise<PrivacyAnonymitySetEventRecord[]> {
+    if (configurations.length === 0) return []
+
+    const [blockFrom, blockTo] = await Promise.all([
+      this.$.blockTimestampProvider.getBlockNumberAtOrBefore(
```
torztomasz, line 158: Two notes on block resolution:

1. `blockFrom = atOrBefore(from)` can precede `from`, and its logs are then discarded by the `timestamp < from` filter below. `blockTo = atOrBefore(to)` can miss the tail of the last second, but the next run starts at `adjustedTo` and takes `atOrBefore(adjustedTo)`, so ranges overlap and the `(configurationId, txHash, logIndex)` upsert dedupes. Works, but the correctness relies on that overlap implicitly; a short comment would help the next reader.

2. This uses the live `blockTimestampProvider.getBlockNumberAtOrBefore` (two binary searches per update) rather than the `PrivacyBlockTimestamp` DB mapping the flow indexer uses. Intentional per the doc, but it adds RPC load during backfill and boundaries can differ from the flow indexer's hourly ones. Worth calling out in the PR body under deployment monitoring.

torztomasz (commented):
Review along two axes: repo standards (indexer conventions + code smells) and spec (PR body / L2B-14626 / L2B-14462). Inline comments below. Not repeated inline: tx-hash dedup, inclusive from/to boundary and the config-ID golden tests for the four existing indexers all look correct. Cutoff amounts, 30d aggregation and FE display are correctly deferred to #12705 / #12707.

packages/backend/src/modules/privacy/indexers/PrivacyAnonymitySetIndexer.ts
```diff
@@ -0,0 +1,296 @@
...
+          txHash: record.log.transactionHash,
+          logIndex: record.log.logIndex,
+          sender,
+          amount: record.amount,
+        },
+      ]
+    })
+  }
+
+  private async getTransactionSenders(
```
torztomasz, line 242: **Possible Duplicated Code (judgement call):** this batching loop + "response matches request" assert has the same shape and the same constant (25) as `RpcClientCompat.getBlockTimestamps` (L161-191). A small `batched(items, size, fn)` helper in shared would remove both copies.

packages/backend/src/modules/privacy/PrivacyModule.ts
```diff
@@ -53,6 +55,17 @@ export function createPrivacyModule({
     ])
   }
 
+  const anonymitySetConfigsByChain = new Map<
```
torztomasz, line 65: **Possible Duplicated Code (judgement call):** third copy of the "group configs by chain" loop in this file (`flowConfigsByChain`, `starknetFlowConfigsByChain`, now `anonymitySetConfigsByChain`). A `groupBy(configs, c => c.chain)` would replace all three.

packages/backend/src/modules/privacy/types.ts
```diff
@@ -80,4 +95,5 @@ export interface PrivacyRpcLog {
 export interface PrivacyFlowExtractResult {
   count: number
   amount: bigint
+  sender?: EthereumAddress
```
torztomasz: **Possible Divergent Change / Speculative Generality (judgement call):** `sender?` on `PrivacyFlowExtractResult` (and the conditional spread in `extractPrivacyFlow.ts` L67-75) bolts an anonymity-set concern onto the flow extractor solely so `extractPrivacyAnonymitySetDeposit` can read it back. Consider parsing `depositor` directly in the anonymity extractor and keeping the flow result flow-only.

packages/shared/src/clients2/compatibility/RpcClientCompat.ts
torztomasz, line 166: Not scope creep, but worth stating in the PR body: `BlockProvider.getBlockTimestamps` asserts the client supports this, and `Clients.ts` L113-137 can place `RpcClientCompat` into `blockClients`, so the existing `PrivacyFlowIndexer` would already fail on compat chains without it.

docs/L2B-14626-anonymity-set-indexing-review.md
```diff
@@ -0,0 +1,807 @@
+# L2B-14626 anonymity-set indexing: reviewer guide
```
torztomasz: **Scope / repo hygiene:** this 807-line file is a first-person reviewer guide with PR links, and roughly 60% of it (L491-755) is a unification design proposal (Options A/B/C, migration path) for work this PR does not do. `docs/` otherwise holds durable feature docs (`activity.md`, `tvs.md`, `da-tracking.md`). Suggest keeping the reviewer guide in the PR description (or a Linear/Notion doc) and, if anything lands in `docs/`, a short durable description of what actually ships.

docs/L2B-14626-anonymity-set-indexing-review.md
```diff
@@ -0,0 +1,807 @@
...
+  chain: string
+  transactionHash: string
+  sender: string
+}
+```
+
+It must not cache provider failures as missing or zero-value data. Failed
+lookups remain retryable and do not advance the anonymity consumer.
+
+### Reorg and invalidation behavior
```
torztomasz: This section describes reorg/invalidation behaviour of the *proposed* raw-log design, not what ships. The shipped indexer has no reorg handling: no block hash, no finality lag, and it indexes to the current hour via the `hourlyIndexer` parent (`PrivacyModule.ts` L116). `trimData` / `wipeData` only cover configuration changes. That is consistent with `PrivacyFlowIndexer`, but the doc should say so explicitly to avoid misleading a reader.

docs/L2B-14626-anonymity-set-indexing-review.md
```diff
@@ -0,0 +1,807 @@
...
+### Why not fetch the transaction sender inside `PrivacyFlowIndexer` only for
+tracked deposits?
+
+It makes one indexer conditionally write two business tables and introduces an
+RPC dependency that can stall unrelated withdrawal and pricing metrics.
+
+### Why does anonymity ingestion not depend on `PrivacyBlockTimestampIndexer`?
+
+Flow indexing needs hourly mappings for pricing. Anonymity ingestion needs only
+the update boundaries and may start earlier than the flow timestamp clamp.
```
torztomasz: Contradicts the code: commit 6489c0774c ("Bound anonymity set backfills") clamps `sinceTimestamp` to the same `PRIVACY_MIN_TIMESTAMP` as the flow indexer (`privacy.ts` L162-171), so anonymity ingestion does not start earlier than the flow clamp. Update or drop this paragraph.

torztomasz (changes requested):

docs/L2B-14626-anonymity-set-indexing-review.md
sdlyy: Removed the reviewer guide. PR-specific design and history no longer live under durable `docs/`.

packages/shared/src/clients2/compatibility/RpcClientCompat.ts
sdlyy: I kept this in `RpcClientCompat`. `IRpcClient` exposes the capability, the legacy RPC client already implements it, and `BlockProvider` also needs it for the existing privacy-flow path. Moving it into privacy would leave the compatibility gap unresolved.

packages/backend/src/modules/privacy/PrivacyModule.ts
sdlyy: `getPrivacyConfig` remains the source of these configs. This code only adapts them into the `{ id, minHeight, maxHeight, properties }` shape required by `ManagedMultiIndexer`, matching the sibling indexers.

packages/backend/src/modules/privacy/indexers/PrivacyAnonymitySetIndexer.ts
sdlyy: Done. Sender selection is centralized in `resolveSender`; transaction batching remains separate so event-backed deposits never trigger transaction lookups.

packages/backend/src/modules/privacy/indexers/PrivacyAnonymitySetIndexer.ts
```diff
@@ -0,0 +1,296 @@
...
+      this.$.blockProvider.getBlockTimestamps(
+        unique(rawRecords.map((record) => record.log.blockNumber)),
+      ),
+      this.getTransactionSenders(
+        unique(
+          rawRecords
+            .filter((record) => record.origin.type === 'transaction')
+            .map((record) => record.log.transactionHash.toLowerCase()),
+        ),
+      ),
```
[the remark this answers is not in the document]
sdlyy: Done. Timestamps are fetched and records outside `[from, to]` are discarded before transaction senders are resolved. A regression test verifies that a stale boundary log causes no transaction lookup.

packages/backend/src/config/features/privacy.ts
sdlyy: Done. Added `PrivacyAnonymitySetIndexer.idToConfigurationId()`, routed backend config through it, and added a backend golden-ID test.

packages/backend/src/modules/privacy/indexers/PrivacyAnonymitySetIndexer.test.ts
sdlyy: Done. Added `wipeData`, `trimData`, and a behavioral test proving transaction lookups run in sequential batches of 25.

packages/backend/src/modules/privacy/indexers/PrivacyAnonymitySetIndexer.test.ts
sdlyy: Done. The test builder now accepts `PrivacyAnonymitySetDepositSource` and uses `satisfies`, preserving discriminated-union checking.

sdlyy: Done. The test builder now accepts `PrivacyAnonymitySetDepositSource` and uses `satisfies`, preserving discriminated-union checking.

packages/backend/src/modules/privacy/indexers/PrivacyAnonymitySetIndexer.ts
sdlyy: Leaving the indexers separate. Sharing this small lifecycle boilerplate would couple otherwise distinct persistence and failure domains.

packages/backend/src/modules/privacy/indexers/PrivacyAnonymitySetIndexer.ts
sdlyy: Renamed to `fetchDepositsForConfigurations`.

packages/backend/src/modules/privacy/indexers/PrivacyAnonymitySetIndexer.ts
sdlyy: Added a comment documenting the overlapping-block and upsert behavior, plus a deployment note. The timestamp provider prefers the block-indexer client and can fall back to RPC, so the note does not assume two RPC searches on every update.

packages/backend/src/modules/privacy/indexers/PrivacyAnonymitySetIndexer.ts
sdlyy: Leaving the loops explicit for now. They validate different response types, and two occurrences do not yet justify a shared generic batching abstraction.

packages/backend/src/modules/privacy/PrivacyModule.ts
sdlyy: Leaving this out of the PR because it is unrelated mechanical cleanup with no behavior change.

packages/backend/src/modules/privacy/types.ts
sdlyy: Done. Privacy Pools parsing now uses a shared protocol-specific helper, and `PrivacyFlowExtractResult` is flow-only again.

packages/shared/src/clients2/compatibility/RpcClientCompat.ts
sdlyy: Added this dependency to the PR deployment notes, together with the multi-topic compatibility fix required when `NEW_CLIENTS_ENABLED` is active.

docs/L2B-14626-anonymity-set-indexing-review.md
sdlyy: Removed the reviewer guide rather than keeping PR-specific material in durable `docs/`; the PR description now contains only the relevant implementation and deployment notes.

docs/L2B-14626-anonymity-set-indexing-review.md
sdlyy: Removed the guide and its proposed reorg behavior. The PR description now describes only what this indexer actually ships.

docs/L2B-14626-anonymity-set-indexing-review.md
sdlyy: Removed the contradictory guide. The implementation and PR description retain the `PRIVACY_MIN_TIMESTAMP` backfill clamp.

torztomasz (approved):

PR #12707, opened 2026-08-31
Show privacy anonymity sets

sdlyy:
## Summary

Exposes production anonymity-set metrics on the privacy summary and project pages:

- calculates inclusive-threshold distinct-sender windows using bigint amounts
- gates each series on its exact managed configuration reaching the latest complete UTC day
- shows the headline and charts from synced series only, and names the series still being indexed
- adds available, syncing, unavailable, and not-applicable summary states
- adds the 30-day historical and holding-duration charts
- adds dedicated 30D, 90D, 180D, 1Y, and MAX controls
- keeps raw sender addresses server/database-only
- caches one full-history snapshot per project (10 min TTL, 15 min stale-while-revalidate) and loads history in year-sized pages, so a cache miss never holds a project's full history in memory
- adds shared DAI/USDC/USDT thresholds for Railgun and Privacy Pools; these create 6 new indexer configurations that backfill from each bucket's start

## Validation

- focused aggregation, synchronization, series, summary, paging, and range tests
- frontend and config typecheck
- frontend and config lint and formatting

## Deployment

This UI can be deployed while backfill is running. A project shows Syncing until its first configuration reaches the latest complete UTC day. After that, the headline and charts use the synced series and list the series still being indexed, so the headline token can change while remaining configurations catch up.

## Stack

1. #12704 - storage
2. #12705 - configuration
3. #12706 - indexer
4. #12707 - frontend


sdlyy:
@codex review

packages/frontend/src/server/features/privacy/anonymity-set/getPrivacyAnonymitySetSeries.ts
```diff
@@ -0,0 +1,74 @@
...
+          sinceTimestamp: bucket.sinceTimestamp,
+        }
+      })
+    }),
+  )
+}
+
+function formatTokenAmount(amount: string, decimals: number): string {
+  const padded = amount.padStart(decimals + 1, '0')
+  const whole = padded.slice(0, -decimals) || '0'
```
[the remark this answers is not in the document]
sdlyy: none of the currently configured anonymity-set series use zero-decimal tokens, so the P2 label overstates present impact

sdlyy:
@codex review

packages/frontend/src/server/features/privacy/getPrivacyAnonymitySetChart.ts
```diff
@@ -0,0 +1,177 @@
...
+      firstSeriesDay,
+      Math.min(
+        historyEndpoints.length > 0
+          ? historyStart - 30 * UnixTime.DAY
+          : holdingEndpoint,
+        holdingEndpoint - Math.max(...HOLDING_DURATIONS) * UnixTime.DAY,
+      ),
+    ),
+  )
+  const rows = await db.privacyAnonymitySetEvent.getSenderDaysByProjectIds(
```
vilcht, line 111: for max range i think this will be super expensive and this isnt cached. Trpc endpoint is public so its really not good

sdlyy: Addressed in `b60de6e177`. The endpoint now caches one canonical full-history snapshot per validated project using `FrontendInMemoryCache` with a 10-minute TTL and 15-minute stale-while-revalidate period. Requested ranges are normalized and sliced after the cache lookup, so arbitrary public range inputs cannot increase cache cardinality. The range-independent holding-duration calculation is reused as part of the same snapshot. Added focused range-selection tests.

packages/config/src/projects/railgun/railgun.ts
```diff
@@ -40,9 +40,21 @@ const TRACKED_TOKENS: TrackedToken[] = [
...
+  },
+  {
+    address: '0xA0b86991c6218b36c1d19D4a2e9Eb0cE3606eB48',
+    symbol: 'USDC',
+    minimumAmounts: ['200000000', '20000000000'],
+  },
+  {
+    address: '0x6B175474E89094C44Da98b954EedeAC495271d0F',
+    symbol: 'DAI',
+    minimumAmounts: ['200000000000000000000', '20000000000000000000000'],
```
torztomasz: why strings?

torztomasz:
image
Make the second margin smaller

packages/frontend/src/server/features/privacy/getPrivacyAnonymitySetChart.ts
```diff
@@ -0,0 +1,208 @@
...
+function toResponseSeries(series: PrivacyAnonymitySetSeries[]) {
+  return series.map(({ id, label, token, minimumAmount }) => ({
+    id,
+    label,
+    token,
+    minimumAmount,
+  }))
+}
+
+function emptyResponse(): PrivacyAnonymitySetChartResponse {
```
torztomasz, line 174: why function?

torztomasz (changes requested):

packages/frontend/src/server/features/privacy/anonymity-set/getPrivacyAnonymitySetSummaries.ts
```diff
@@ -0,0 +1,142 @@
...
+      }
+
+      const { syncedSeries, syncingTokens } = getPrivacyAnonymitySetSyncStatus(
+        series,
+        configurations,
+        currentDay,
+      )
+      const firstSeries = syncedSeries[0]
+      if (firstSeries === undefined) {
+        return [project.id, { status: 'syncing' }]
```
torztomasz: This contradicts the description ("gates current summaries on every exact managed configuration") and the deployment note ("remains Syncing until every required configuration reaches the latest complete UTC day"). The summary is `syncing` only when *zero* series are synced; with one synced series it becomes `available` with a number computed from a subset of tokens, and the headline can then flip between tokens as more series catch up. Commit e78a5f1 changed this after the description was written. Either gate on `syncingTokens.length === 0` here, or update the description and deployment note to say the headline is shown from partially synced data.

packages/frontend/src/server/features/privacy/anonymity-set/getPrivacyAnonymitySetSummaries.ts
```diff
@@ -0,0 +1,142 @@
...
+      let bestSeries = firstSeries
+      let bestValue = values[0] ?? 0
+      for (let i = 1; i < values.length; i++) {
+        const value = values[i] ?? 0
+        const candidate = syncedSeries[i]
+        if (candidate !== undefined && value > bestValue) {
+          bestSeries = candidate
+          bestValue = value
+        }
+      }
```
torztomasz: Unnamed "pick the largest synced series" loop. Extract it (e.g. `pickLargestSeries(syncedSeries, values)`) so the headline selection rule is readable and testable; `getPrivacyAnonymitySetSummaries` currently has no tests at all while the choice of headline token is a visible product rule.

packages/frontend/src/server/features/privacy/anonymity-set/getPrivacyAnonymitySetSummaries.ts
```diff
@@ -0,0 +1,142 @@
...
+  currentDay: UnixTime,
+): Promise<Map<string, PrivacyAnonymitySetSummary>> {
+  if (env.MOCK) return getMockSummaries(projects)
+
+  const db = getDb()
+  const trackedProjects = projects.filter(
+    (project) => getPrivacyAnonymitySetSeries(project).length > 0,
+  )
+  const trackedProjectIds = trackedProjects.map((project) => project.id)
+  const cutoff = currentDay - 30 * UnixTime.DAY
```
torztomasz: The 30-day window lives in at least six places: this `30 * UnixTime.DAY`, the `windowDays = 30` default in `calculateAnonymitySets.ts:19` (never overridden), and prose "30" in `PrivacyAnonymitySetSection.tsx`, `anonymitySetTooltip.ts` and the `PrivacySummaryTable.tsx` header/tooltip. Export a single `ANONYMITY_SET_WINDOW_DAYS`, drop the default param, and interpolate it into the copy.

packages/frontend/src/server/features/privacy/anonymity-set/getPrivacyAnonymitySetSummaries.ts
```diff
@@ -0,0 +1,142 @@
...
+
+export async function getPrivacyAnonymitySetSummaries(
+  projects: PrivacyProject[],
+  currentDay: UnixTime,
+): Promise<Map<string, PrivacyAnonymitySetSummary>> {
+  if (env.MOCK) return getMockSummaries(projects)
+
+  const db = getDb()
+  const trackedProjects = projects.filter(
+    (project) => getPrivacyAnonymitySetSeries(project).length > 0,
```
torztomasz: `getPrivacyAnonymitySetSeries(project)` is called three times per project on this path (here, line 64, and inside `getPrivacyAnonymitySetConfigurations`), and the "has an anonymity set" predicate is spelled a third way in `getPrivacyProjectDetails.ts:268`. Compute the series once per project and share one predicate.

packages/frontend/src/server/features/privacy/getPrivacyAnonymitySetChart.ts
```diff
@@ -0,0 +1,208 @@
...
+      ...emptyResponse(),
+      syncingTokens,
+    }
+  }
+
+  const firstSeriesDay = UnixTime.toStartOf(
+    Math.min(...syncedSeries.map((item) => item.sinceTimestamp)),
+    'day',
+  )
+  const holdingEndpoint = currentDay
```
torztomasz, line 104: Description says charts show history "through the common proven cursor", but no common cursor exists: syncing series are dropped and synced ones are rendered through `currentDay`. Fine as behaviour, but the description is stale relative to 1183728 and should say what actually happens.

torztomasz (commented):
Review of the frontend layer against the PR description and L2B-14626. Inline comments below cover everything that looks like it needs a fix; the rest of the description's bullets (inclusive threshold, config id matching, senders staying server-side, both charts, five range controls, four summary states) check out against the code.

One question rather than a fix: the railgun / privacy-pools threshold additions (DAI/USDC/USDT) create new indexer configurations that need backfill. Is that intended to ship in the UI PR rather than in #12705?

packages/frontend/src/server/features/privacy/getPrivacyAnonymitySetChart.ts
```diff
@@ -0,0 +1,208 @@
...
+): PrivacyAnonymitySetChartResponse {
+  const from = range[0]
+  const fromDay = from === null ? null : UnixTime.toStartOf(from, 'day')
+  const toDay = UnixTime.toStartOf(range[1], 'day')
+
+  return {
+    ...snapshot,
+    history: snapshot.history.filter(
+      ([timestamp]) =>
+        (fromDay === null || timestamp >= fromDay) && timestamp <= toDay,
```
torztomasz: `range[1]` comes from `optionToRange`, which is `now - 75min` rounded to the hour. Between 00:00 and 01:15 UTC `toDay` is yesterday, so this filter drops the newest snapshot point (today's `currentDay`) for that window every day. Clamp `toDay` to at least the snapshot's `syncedUntil`, or ignore the upper bound for this daily series.

packages/frontend/src/server/features/privacy/getPrivacyAnonymitySetChart.ts
```diff
@@ -0,0 +1,208 @@
...
+
+  const series = getPrivacyAnonymitySetSeries(project)
+  if (series.length === 0) return emptyResponse()
+
+  const currentDay = UnixTime.toStartOf(UnixTime.now(), 'day')
+  const snapshot = env.MOCK
+    ? getMockResponse(series, currentDay)
+    : await cache.get(
+        {
+          key: ['privacy-anonymity-set-chart', project.id],
```
torztomasz: `currentDay` is computed outside the cache but not part of the key. Across midnight a snapshot built for the previous day is served for up to 25 min (ttl + staleWhileRevalidate) while `syncedUntil` and the summary already report the new day. Add `currentDay` to the key.

packages/frontend/src/server/features/privacy/getPrivacyAnonymitySetChart.ts
```diff
@@ -0,0 +1,208 @@
...
+  }
+
+  const historyEndpoints = generateTimestamps(
+    [UnixTime(firstSeriesDay), UnixTime(holdingEndpoint)],
+    'day',
+  )
+  const rows = await db.privacyAnonymitySetEvent.getSenderDaysByProjectIds(
+    [project.id],
+    UnixTime(firstSeriesDay),
+    holdingEndpoint,
```
torztomasz: This loads every sender-day row since the earliest series `sinceTimestamp` into memory on each cache miss. For Tornado Cash that is all rows since 2019. Since the history is a rolling 30-day distinct count, consider aggregating per day in SQL, or at least paginating by year and folding the result, before this reaches prod traffic.

packages/frontend/src/server/features/privacy/getPrivacyAnonymitySetChart.ts
```diff
@@ -0,0 +1,208 @@
...
+    PrivacyAnonymitySetSeries,
+    'id' | 'label' | 'token' | 'minimumAmount'
+  >[]
+  history: PrivacyAnonymitySetHistoryPoint[]
+  holdingDuration: PrivacyAnonymitySetHoldingDurationPoint[]
+  syncingTokens: string[]
+  syncedUntil: number | undefined
+}
+
+const HOLDING_DURATIONS = Array.from({ length: 359 }, (_, index) => index + 7)
```
torztomasz: `359` is `MAX_HOLDING_DAYS - MIN_HOLDING_DAYS + 1` (365 - 7 + 1). Derive it from named constants so the intent is readable.

packages/frontend/src/server/features/privacy/anonymity-set/getPrivacyAnonymitySetSync.ts
```diff
@@ -0,0 +1,49 @@
...
+      configuration.currentHeight >= target
+        ? syncedSeries
+        : syncingSeries
+    destination.push(item)
+  }
+
+  return {
+    syncedSeries,
+    syncingSeries,
+    syncingTokens: unique(syncingSeries.map((item) => item.token)),
```
torztomasz: `syncingTokens` is keyed by token symbol, but a project can have several series per token (Tornado Cash ETH denominations, and now the two-threshold stablecoin buckets). One lagging bucket labels the whole token as excluded in the tooltip while its other series are still shown in the number. Key on series label (or series id) instead.

packages/frontend/src/server/features/privacy/anonymity-set/getPrivacyAnonymitySetSync.ts
```diff
@@ -0,0 +1,49 @@
...
+  const syncedSeries: PrivacyAnonymitySetSeries[] = []
+  const syncingSeries: PrivacyAnonymitySetSeries[] = []
+
+  for (const item of series) {
+    const configuration = configurationsById.get(item.configurationId)
+    const destination =
+      configuration !== undefined &&
+      configuration.maxHeight === null &&
+      configuration.currentHeight !== null &&
+      configuration.currentHeight >= target
```
torztomasz, line 30: This re-implements the `maxHeight === null && currentHeight !== null` check from `features/utils/getConfigurationsSyncedUntil.ts`. Per-series partitioning is a real difference, but a shared `isConfigurationSyncedTo(config, target)` would keep the two from drifting.

packages/config/src/processing/getProjects.test.ts
```diff
@@ -432,6 +432,39 @@ describe('getProjects', () => {
         }
       })
     }
+
+    it('configures comparable ETH and stablecoin anonymity-set thresholds', () => {
```
torztomasz: The test name says "comparable" but the body asserts the literal thresholds, so it only re-states `railgun.ts` and `privacy-pools.ts` and will need editing on every threshold change. Either assert the actual invariant (same USD-equivalent tiers across tokens/projects, via a shared constant both configs import) or drop the test.

packages/config/src/projects/railgun/railgun.ts
torztomasz: These stablecoin thresholds are duplicated verbatim in `privacy-pools.ts` (and a third time in the test). Move them to a shared common value so the two projects cannot diverge.

sdlyy: Base-unit amounts go up to 2e22, above `Number.MAX_SAFE_INTEGER`, and the config is serialized to JSON so `bigint` is not an option. `minimumAmounts: string[]` came in with #12705 and Tornado denominations use the same convention; the frontend parses them with `BigInt`.

packages/frontend/src/server/features/privacy/getPrivacyAnonymitySetChart.ts
sdlyy: It is spread into cached snapshots and returned straight to tRPC, so each call gets its own object rather than a shared mutable one. Kept as is.

packages/frontend/src/server/features/privacy/anonymity-set/getPrivacyAnonymitySetSync.ts
sdlyy: Left separate for now. `getConfigurationsSyncedUntil` falls back to finished configurations and takes the min across configs, whereas this needs the exact active configuration per series. Sharing would need a mode flag on one of them.

packages/frontend/src/server/features/privacy/anonymity-set/getPrivacyAnonymitySetSummaries.ts
sdlyy: Kept the partial display and fixed the description and deployment note instead. They now say each series is gated on its own configuration, that the headline and charts use synced series while naming the rest, and that the headline token can change while the remaining configurations catch up.

packages/frontend/src/server/features/privacy/anonymity-set/getPrivacyAnonymitySetSummaries.ts
sdlyy: Done in 48fa99540b. Per-project logic is now the pure `getPrivacyAnonymitySetSummary` with a named `pickLargestSeries` (largest count wins, ties keep configuration order), covered by 5 tests: not-applicable, unavailable, syncing, largest wins with syncing labels, and ties.

packages/frontend/src/server/features/privacy/anonymity-set/getPrivacyAnonymitySetSummaries.ts
sdlyy: Done in 48fa99540b. Single `ANONYMITY_SET_WINDOW_DAYS` exported from `calculateAnonymitySets.ts`, default param removed, and interpolated into the section, tooltip, and summary table copy.

packages/frontend/src/server/features/privacy/anonymity-set/getPrivacyAnonymitySetSummaries.ts
sdlyy: Done in 48fa99540b. Series are computed once per project in the summaries path, `getPrivacyAnonymitySetConfigurations` takes series instead of projects, and `hasPrivacyAnonymitySet` is the one predicate, used by project details too.

packages/frontend/src/server/features/privacy/getPrivacyAnonymitySetChart.ts
sdlyy: Description updated to describe the actual behavior: syncing series are dropped and synced ones are rendered through the latest complete UTC day.

packages/frontend/src/server/features/privacy/getPrivacyAnonymitySetChart.ts
sdlyy: Done in 48fa99540b. The upper bound is no longer applied, since the snapshot already ends at the latest complete UTC day; only the range start filters history. Added a test for the post-midnight case.

packages/frontend/src/server/features/privacy/getPrivacyAnonymitySetChart.ts
sdlyy: Done in 48fa99540b. `currentDay` is part of the cache key.

packages/frontend/src/server/features/privacy/getPrivacyAnonymitySetChart.ts
sdlyy: Done in 48fa99540b, the "paginate and fold" variant. `loadAnonymitySetCharts` fetches 365-day pages aligned to the current day, each with a 30-day overlap, and runs the existing pure functions per page; the last page doubles as the holding-duration input. Pages load sequentially, so peak memory is about 13 months of sender days instead of the full history. Total DB scan is unchanged; a backend per-day aggregate is the follow-up for that. Test asserts the paged result equals the unpaged calculation over a 1,000-day fixture.

packages/frontend/src/server/features/privacy/getPrivacyAnonymitySetChart.ts
sdlyy: Done in 48fa99540b. Derived from `MIN_HOLDING_DAYS` and `MAX_HOLDING_DAYS` with lodash `range`.

packages/frontend/src/server/features/privacy/anonymity-set/getPrivacyAnonymitySetSync.ts
sdlyy: Done in 48fa99540b. Renamed to `syncingLabels` and keyed by series label, so a lagging Tornado denomination no longer marks every ETH series as excluded. Response shape, UI copy, and tests updated.

packages/config/src/processing/getProjects.test.ts
sdlyy: Dropped in 48fa99540b. Both configs now import the shared constant, so the invariant holds by construction.

packages/config/src/projects/railgun/railgun.ts
sdlyy: Done in 48fa99540b. Thresholds live in `packages/config/src/common/privacyAnonymitySets.ts` as `PRIVACY_ANONYMITY_SET_MINIMUM_AMOUNTS`, looked up by symbol in both projects. Generated config output is unchanged.

sdlyy:
Re the backfill question: yes, adding `anonymitySet` to the DAI/USDC/USDT buckets creates 6 new indexer configurations (3 Railgun, 3 Privacy Pools) that backfill from each bucket's start. #12705 is already merged, so they stay in this PR; the description now says so. Until they catch up those projects show the ETH/WETH headline with the stablecoin series listed as still indexing.

Inline comments addressed in 48fa99540b, replies on each thread.

sdlyy:
Done in 09f3e5f227. The second heading's top margin went from `mt-8` to `mt-4`, so the gap between the first chart's footnote and "Estimated anonymity set by holding duration" is now about 21px, in line with the 17px above the first chart.

torztomasz (approved):

PR #12709, opened 2026-08-31
Repair invalid Coingecko circulating supply values, unskip Venice and Marlin

adamiak:
Resolves L2B-14768

## What happened

Coingecko's hourly market cap history for `venice-token` contains a single poisoned point: `market_cap = -1` at 2026-08-16 17:00 UTC (verified still served by their API today, 15 days later — the only such point in the token's entire history). `approximateCirculatingSupply` derives supply as `marketCap / price`, so the value became `NaN` via `Math.log10(negative)`, and `BigInt(NaN)` in `CirculatingSupplyAmountIndexer.multiUpdate` threw `RangeError`. Since that indexer is shared by all circulating-supply tokens and retries infinitely, the whole TVS pipeline stalled for ~12 hours (2026-08-16 19:12 → 08-17 06:13). The mitigation was to filter `venice-token` out of the TVS config entirely (#12587), like `marlin` before it (#12453/#12455).

## The fix

- **`CoingeckoQueryService.getCirculatingSupplies`** now validates every computed supply (`Number.isFinite && >= 0`) and repairs corrupt gaps of up to `MAX_REPAIRED_SUPPLY_GAP_HOURS = 24` consecutive hours from the neighboring hour (previous hour preferred, next for a leading gap). Every repair is logged at **error** level with the coingeckoId, count, gap length and timestamps, so a long stretch of bad data can't go unnoticed. Stretches longer than 24h (or a fully-invalid window) are not fabricated: an error is logged and the existing `Insufficient data in response` error is thrown, which skips only that configuration's window while the rest of TVS keeps syncing.
- **`CirculatingSupplyAmountIndexer`** additionally drops any non-finite/negative value right before the `BigInt` conversion (defense in depth), logging the `apiId` — the missing piece that made the original incident hard to diagnose.
- **Removed the temporary `venice-token` / `marlin` filter** from `config/features/tvs.ts`. Both tokens' data was wiped when they were filtered, so they resync from their `sinceTimestamp`s automatically. Base gets ~$822M of VVV back; Arbitrum and Polygon PoS get POND back.

## Why unskipping is safe

- Venice: scanned the full history (Jan 2025 → today) via the pro API at hourly granularity — exactly one invalid point, which the repair absorbs. Verified end-to-end by running the new code against the real poisoned API response: one error log, valid BigInt conversion for every hour.
- Marlin: its amount is a sum of Ethereum escrow balances — the only Coingecko dependency is the price, so its Aug 4 failure was a different (price-path) response issue that Coingecko has since fixed: full history (Dec 2020 → today) scanned type-strict, zero anomalies in both series.

## Expected after deploy

One "Invalid circulating supply values, repaired from neighboring hours" **error log for venice-token** when its resync reaches Aug 16 2026 — that's the fix working, not an incident.

## Tests

- `getCirculatingSupplies`: poisoned mid-point repairs from previous hour, leading gap (incl. zero-price NaN) repairs from next hour, zero market cap keeps supply 0, exactly-24h gap repairs, >24h gap throws insufficient-data, all-invalid throws.
- `approximateCirculatingSupply`: NaN for negative cap, NaN for zero price, 0 for zero cap.
- Indexer: non-finite value dropped and the rest of the batch saved.
- Full suites green (shared 406, backend 1369), typecheck and biome clean.

### DIFF
--- a/packages/backend/src/modules/tvs/indexers/CirculatingSupplyAmountIndexer.ts
+++ b/packages/backend/src/modules/tvs/indexers/CirculatingSupplyAmountIndexer.ts
@@ -76,7 +76,8 @@ export class CirculatingSupplyAmountIndexer extends ManagedMultiIndexer<Circulat
               (p) => Number.isFinite(p.value) && p.value >= 0,
             )
             if (validSupplies.length !== supplies.length) {
-              this.logger.error(
+              // Use critical level to trigger maintenance alert
+              this.logger.critical(
                 `Dropped invalid circulating supply values for ${configuration.properties.apiId}`,
                 {
                   priceId: configuration.properties.apiId,
--- a/packages/backend/src/modules/tvs/indexers/CirculatingSupplyAmountIndexer.ts
+++ b/packages/backend/src/modules/tvs/indexers/CirculatingSupplyAmountIndexer.ts
@@ -69,7 +69,23 @@ export class CirculatingSupplyAmountIndexer extends ManagedMultiIndexer<Circulat
                 CoingeckoId(configuration.properties.apiId),
                 { from: from, to: adjustedTo },
               )
-            const supplyRecords: TvsAmountRecord[] = supplies.map((p) => ({
+
+            // defense in depth: a non-finite value would crash the BigInt
+            // conversion below and halt the whole indexer
+            const validSupplies = supplies.filter(
+              (p) => Number.isFinite(p.value) && p.value >= 0,
+            )
+            if (validSupplies.length !== supplies.length) {
+              this.logger.error(
+                `Dropped invalid circulating supply values for ${configuration.properties.apiId}`,
+                {
+                  priceId: configuration.properties.apiId,
+                  dropped: supplies.length - validSupplies.length,
+                },
+              )
+            }
+
+            const supplyRecords: TvsAmountRecord[] = validSupplies.map((p) => ({
               configurationId: configuration.id,
               timestamp: p.timestamp,
               amount: BigInt(p.value * 10 ** configuration.properties.decimals),
--- a/packages/shared/src/services/coingecko/CoingeckoQueryService.ts
+++ b/packages/shared/src/services/coingecko/CoingeckoQueryService.ts
@@ -14,6 +14,7 @@ import type {
 
 export const MAX_DAYS_FOR_HOURLY_PRECISION = 80
 export const COINGECKO_INTERPOLATION_WINDOW_DAYS = 14
+export const MAX_REPAIRED_SUPPLY_GAP_HOURS = 24
 
 export interface QueryResultPoint {
   value: number
@@ -52,7 +53,7 @@ export class CoingeckoQueryService {
       coingeckoId,
       range,
     )
-    return zip(queryResult.prices, queryResult.marketCaps).map(
+    const supplies = zip(queryResult.prices, queryResult.marketCaps).map(
       ([price, marketCap]) => {
         assert(price && marketCap && price.timestamp === marketCap.timestamp)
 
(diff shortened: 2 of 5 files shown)


packages/shared/src/services/coingecko/CoingeckoQueryService.ts
```diff
@@ -63,6 +64,86 @@ export class CoingeckoQueryService {
...
+      coingeckoId: coingeckoId.toString(),
+      invalidCount,
+      longestGapHours,
+      from: firstInvalid,
+      to: lastInvalid,
+    }
+
+    if (
+      invalidCount === points.length ||
+      longestGapHours > MAX_REPAIRED_SUPPLY_GAP_HOURS
```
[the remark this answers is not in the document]
adamiak: Acknowledged, accepting this edge case deliberately. Context: (1) in steady-state sync the windows are ~1h, so an ongoing corruption makes the whole window invalid, which throws and skips instead of filling — the boundary case only exists during large historical backfills; (2) it requires a >24h persistent corrupt stretch in historical data (never observed — worst seen is a single poisoned hour) landing exactly across a ~52-day batch boundary; (3) both halves still emit the error log, so it cannot happen silently — worst case is ≤48h of one token neighbor-filled, announced by two error logs. If a >24h stretch is ever observed in the wild, the follow-up is to measure runs on a padded series using the ±14d interpolation context the raw fetch already includes.

packages/backend/src/modules/tvs/indexers/CirculatingSupplyAmountIndexer.ts
```diff
@@ -69,7 +69,23 @@ export class CirculatingSupplyAmountIndexer extends ManagedMultiIndexer<Circulat
...
+              (p) => Number.isFinite(p.value) && p.value >= 0,
+            )
+            if (validSupplies.length !== supplies.length) {
+              this.logger.error(
+                `Dropped invalid circulating supply values for ${configuration.properties.apiId}`,
+                {
+                  priceId: configuration.properties.apiId,
+                  dropped: supplies.length - validSupplies.length,
+                },
+              )
```
torztomasz: I guess we should have some alert on that?

packages/backend/src/modules/tvs/indexers/CirculatingSupplyAmountIndexer.ts
```diff
@@ -69,7 +69,23 @@ export class CirculatingSupplyAmountIndexer extends ManagedMultiIndexer<Circulat
                 CoingeckoId(configuration.properties.apiId),
                 { from: from, to: adjustedTo },
               )
-            const supplyRecords: TvsAmountRecord[] = supplies.map((p) => ({
+
+            // defense in depth: a non-finite value would crash the BigInt
+            // conversion below and halt the whole indexer
+            const validSupplies = supplies.filter(
+              (p) => Number.isFinite(p.value) && p.value >= 0,
+            )
```
torztomasz, line 77: isn't this already validated in `getCirculatingSupplies`?

torztomasz (changes requested):

packages/backend/src/modules/tvs/indexers/CirculatingSupplyAmountIndexer.ts
adamiak: Correct — it's intentionally redundant, as the last line of defense before the conversion that took TVS down for 12h. The upstream guarantee lives two packages away behind the provider indirection, and nothing type-level enforces it, so a refactor of the query service (an area under active churn) could silently lose the repair without touching this file. The post-incident recommendation also explicitly asked for validation in both getCirculatingSupplies and multiUpdate. If the upstream repair ever regresses, this guard turns a full pipeline halt into an error-logged skip of the affected hours.

packages/shared/src/services/coingecko/CoingeckoQueryService.ts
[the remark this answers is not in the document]
adamiak: acknowledged, accepted deliberately — see the thread above

packages/backend/src/modules/tvs/indexers/CirculatingSupplyAmountIndexer.ts
adamiak: Switch to critical to trigger alert if happens

torztomasz (approved):

PR #12710, opened 2026-08-31
test(interop): correct regression comment wording

elonfliter:
Update the regression explanation in `InMemoryEventDb.test.ts` so the event timing is stated precisely.


radinat:
Thanks for contributing, but we currently do not accept purely cosmetic or typo fixes.

PR #12711, opened 2026-08-31
Show per-plugin transfer evidence in token ingestion preview and log

adamiak:
Resolves L2B-14727

## Problem

Researchers reviewing the token ingestion queue can't tell where an address actually appears — which interop plugins bridge it and via which transfers ([L2B-14727](https://linear.app/l2beat/issue/L2B-14727/add-more-context-to-the-token-ingestion-queue)). The issue proposed storing that context on the queue entry, but the queue deliberately carries only `(chain, address)` + status: enqueueing an already-queued address is a no-op (stored provenance would be first-writer-wins), many enqueues come from propagation (a useless "reason"), and the transfer table is a ~7-day sliding window (stored references would dangle). See the discussion on the Linear issue.

## Solution

Derive the evidence fresh from the current transfer window instead, through the existing trace pipeline:

- `getTokenRoutes()` now groups by `plugin` and joins the aggregate back to `InteropTransfer` on the sample transfer's primary key to carry that row's `srcTxHash`/`dstTxHash` (per-column `max()` would stitch hashes from different transfers). Still one query per index refresh, index size still proportional to distinct routes (× plugins, almost always 1–2).
- The `transfer-evidence` trace step summarizes matches per distinct plugin: total transfer count + one sample transfer's tx hashes (a hashed sample preferred over a busier hash-less one).
- `formatIngestionTrace` renders the summary, so it shows up in the queue **Preview** dialog and in the ingestion log persisted to `TokenDbHistory` — no UI changes needed:

```
2. Found 128 transfers (127 non-swapping). Other sides resolve to: USDC01:USDC. Transfers by plugin:
   - cctp: 125 transfers (sample: src tx 0xabc…, dst tx 0xdef…)
   - oft: 3 transfers (sample: dst tx 0x123…)
```

The design doc (`automatic_token_ingestion.md`) is updated accordingly.

## Testing

- `packages/database`: full `InteropTransferRepository` suite (112 tests) passes against the test DB, including the per-plugin `getTokenRoutes` aggregation.
- `packages/token-backend`: all 203 tests pass, including new unit tests for the per-plugin summarizer and index enrichment.
- Typechecks clean across `database`, `token-backend`, `token-ui`.

Query-cost note: the aggregation was already a full scan + hash aggregate (no index covered the old GROUP BY either); the added join does PK lookups over the small aggregated set, and the query runs once per ingestion tick, not per entry or page row.


packages/database/src/repositories/InteropTransferRepository.ts
```diff
@@ -366,14 +373,33 @@ export class InteropTransferRepository extends BaseRepository {
...
-        ...groupColumns,
-        eb.fn.countAll().as('transferCount'),
-        eb.fn.max('transferId').as('sampleTransferId'),
+      .with('route', (qb) =>
+        qb
+          .selectFrom('InteropTransfer')
+          .select((eb) => [
+            ...groupColumns,
+            eb.fn.countAll().as('transferCount'),
+            eb.fn.max('transferId').as('sampleTransferId'),
```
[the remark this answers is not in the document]
adamiak: Addressed in 251eb0c3e9 — the sample id is now picked with FILTER aggregates (both hashes → any hash → any row), so a fully hashed transfer wins within the group regardless of transfer-id ordering. Covered by two new repository tests. No extra scan or join: two conditional aggregates over rows the seq scan already reads.

packages/token-backend/src/ingestion/tokenIngestionUtils.ts
```diff
@@ -45,6 +61,58 @@ export function buildInteropTransferIndex(
...
+  candidate: InteropTransferMatch,
+  current: InteropTransferMatch,
+): boolean {
+  const candidateHasHash =
+    candidate.sampleSrcTxHash !== undefined ||
+    candidate.sampleDstTxHash !== undefined
+  const currentHasHash =
+    current.sampleSrcTxHash !== undefined ||
+    current.sampleDstTxHash !== undefined
+  if (candidateHasHash !== currentHasHash) return candidateHasHash
```
sdlyy: This can discard a fully hashed sample. Once both candidates have at least one hash, the busier route wins, so a partial sample can replace one containing both hashes. Does that make sense? If so, can we rank by hash completeness first, then transfer count, and test that case?

packages/token-backend/src/ingestion/tokenIngestionUtils.ts
```diff
@@ -19,6 +22,16 @@ export interface InteropTransferIndex {
...
 }
 
+/** Per-plugin summary of the transfers involving an address: the total count
+ * and one sample transfer's tx hashes, so a researcher can jump from the
+ * ingestion trace to an explorer. */
+export interface TransferPluginEvidence {
+  plugin: string
+  transferCount: number
+  sampleSrcTxHash: string | undefined
+  sampleDstTxHash: string | undefined
```
sdlyy: I’m not sure how useful these hashes are without chain context. When a plugin spans multiple chains, we cannot tell which explorer each source or destination hash belongs to. Should we include the source/destination chains and render chain-qualified links?

packages/token-backend/src/ingestion/tokenIngestionUtils.ts
```diff
@@ -19,6 +22,16 @@ export interface InteropTransferIndex {
   findInvolving(address: TokenAddress): InteropTransferMatch[]
 }
 
+/** Per-plugin summary of the transfers involving an address: the total count
+ * and one sample transfer's tx hashes, so a researcher can jump from the
+ * ingestion trace to an explorer. */
+export interface TransferPluginEvidence {
```
sdlyy: nit: I feel like this interface belongs closer to `IngestionStep` rather than in a generic utility file.

packages/token-backend/src/ingestion/formatIngestionTrace.ts
```diff
@@ -46,7 +47,12 @@ export function describeIngestionStep(step: IngestionStep): string {
         step.abstractTokens.length === 0
           ? 'no abstract tokens'
           : step.abstractTokens.map(formatRef).join(', ')
-      return `Found ${step.total} transfers (${step.nonSwapping} non-swapping). Other sides resolve to: ${refs}.`
+      const base = `Found ${step.total} transfers (${step.nonSwapping} non-swapping). Other sides resolve to: ${refs}.`
+      if (step.plugins.length === 0) return base
+      const plugins = step.plugins
+        .map((evidence) => `\n   - ${formatPluginEvidence(evidence)}`)
```
sdlyy: We now embed multiline list layout inside `describeIngestionStep()`. Would it make more sense for `formatIngestionTrace()` to handle the newlines and indentation?

sdlyy (approved):

packages/token-backend/src/ingestion/tokenIngestionUtils.ts
adamiak: Good catch — this was the same bug Codex found in the SQL, one level up. Fixed in 0996af589e: samples are now ranked by number of tx hashes (2 → 1 → 0) before transfer count, with a test covering both merge orders.

packages/token-backend/src/ingestion/tokenIngestionUtils.ts
adamiak: Agreed on the chain context — fixed in 0996af589e. The route's src/dst chains are now carried through to the evidence and the log reads `src tx 0x… on ethereum, dst tx 0x… on arbitrum`. Clickable explorer links would need the UI to render structured step data instead of the persisted text blob, so I left that out of this PR — happy to file it as a follow-up.

packages/token-backend/src/ingestion/tokenIngestionUtils.ts
adamiak: Moved `TransferPluginEvidence` into IngestionTrace.ts next to `IngestionStep` in 0996af589e. Note this makes the type imports between the two modules mutual (IngestionTrace already takes `TokenAddress` from the utils file) — type-only, so erased at runtime.

packages/token-backend/src/ingestion/formatIngestionTrace.ts
adamiak: Agreed — done in 0996af589e. `describeIngestionStep` returns a single line again; a new `describeIngestionStepDetails` returns the per-plugin lines and `formatIngestionTrace` owns the newlines and indentation. Output is unchanged.

PR #12714, opened 2026-08-31
Fix stale cache requests overwriting newer results

ginavalent:
## Summary

- Prevent superseded cache fallbacks from overwriting results produced by newer requests.
- Prevent superseded fallbacks from deleting the current `inFlight` entry.
- Apply the same generation check to background revalidation.
- Add regression coverage for both stale writes and stale cleanup.

## Root cause

After an in-flight fallback exceeds `promiseTimeout`, `InMemoryCache` allows a new fallback to start for the same key.

The superseded promise previously had no way to determine that it was no longer the current request. When it settled, it could therefore:

- overwrite a newer cached result
- delete the newer request's `inFlight` entry

This affected both regular cache misses and background revalidation.

## Fix

Use the stored Promise identity as a generation token.

A fallback now updates the cache or clears `inFlight` only when its Promise is still the current Promise for the key. Superseded callers still receive their own fallback result, but their completion can no longer mutate newer shared cache state.

## Validation

- `pnpm --filter @l2beat/shared-pure test` — 446 passing on Node.js 22
- `pnpm --filter @l2beat/shared-pure typecheck`
- `pnpm --filter @l2beat/shared-pure lint`
- `pnpm --filter @l2beat/shared-pure format`


torztomasz:
Thank you for tracking this down and for the clear write-up of the root cause! The stale-overwrite race is real and the regression tests made it easy to reason about.

I pushed two commits on top of your branch:

**1. Merge `main`.** #12743 landed in the meantime and already covers the `inFlight` cleanup half (guarded delete on promise identity), and it changed the internals (`store()`, array-keyed `_get`/`_set`, no `initialCache`). I kept your structure for `get` / `revalidateInBackground` and ported the tests to the new API and the existing controlled-clock helper.

**2. Switch the write guard from promise identity to a timestamp check.** The identity check skips the write whenever the request is no longer the current one in `inFlight`, which also covers the case where the newer request has *failed*:

- A starts at 00:00 and is slow.
- At 00:31 B starts and replaces A in `inFlight`.
- B rejects at 00:32 and cleans itself up.
- A resolves at 00:40 with valid data, sees it is not current, and discards it. The cache stays empty and the next visitor pays for a third query.

Under the overload conditions that trigger the 30s timeout, a fast failure of the retry is the likely case, so throwing away a completed expensive query seemed worse than the stale write we're fixing.

The write is now skipped only when the cache already holds an entry written after this request started (`existing.timestamp >= startedAt` in `store()`). That still prevents the stale overwrite your tests cover, and it keeps A's result when nothing newer made it into the cache. Added two tests for both sides of that rule.

Happy to adjust if you see a hole in the timestamp rule.

torztomasz (approved):

PR #12717, opened 2026-09-01
Switch Avail da-beat client from WebSocket to HTTP

torztomasz:
## Summary

Replaces the polkadot-js `WsProvider` transport in the Avail da-beat client with `HttpProvider`. The staking queries are plain request/response storage reads; the WebSocket was only ever the transport and is the reason the backend went down on 2026-08-31.

## Root cause this fixes

When the avail WSS endpoint started failing, `WsProvider`'s background reconnect timers rejected promises with raw WebSocket `ErrorEvent`s outside any awaited chain — fatal unhandled rejections that killed the process. The old code couldn't catch it: `api.isReady` never rejects on connection failure, so `connect()` hung while the process died around it, and since `connect()` sat outside the `try/finally`, `disconnect()` never ran and the reconnect loop kept firing every ~2.5s. Both prod and staging died on the same hourly tick, 28s apart, and crash-looped all night.

## Changes

- `AvailWsClient` → `AvailClient` (`packages/shared/src/clients/avail/`), now on `HttpProvider`: no sockets, no background timers, every failure rejects the awaited promise (`isReadyOrError` + `throwOnConnect`)
- Connect/disconnect lifecycle removed from the public API — it was per-call runtime-metadata initialization, not a connection; it now lives inside the single stateless `getStakingEraOverview()` method with teardown in `finally`
- `DaBeatStatsProvider.getAvailStats()` collapses to a plain call
- Config: `availWsUrl` → `availRpcUrl`, env `AVAIL_RPC_URL[_FOR_DA_BEAT]`, default switched to the official `https://mainnet-rpc.avail.so/rpc` (publicnode's avail endpoint currently returns 503)
- `@polkadot/api` stays a lazy `await import()` — loading it costs ~700ms and ~82MB RSS, so it's paid only in the process that actually queries avail

## Deployment note

`AVAIL_WS_URL` is no longer read. Set `AVAIL_RPC_URL` if the default endpoint shouldn't be used.

## Testing

- Live-verified against `mainnet-rpc.avail.so`: returns the current era's validator overview; also verified a failing endpoint (publicnode, 503) surfaces as a clean rejected promise instead of a crash
- 20 `DaBeatStatsProvider` tests passing; typecheck, lint, format clean on both packages


mateuszradomski:
@codex, review

mateuszradomski (approved):

PR #12718, opened 2026-09-01
Store permissions in a separate discovered.json key

mateuszradomski:
Part of L2B-10994


packages/discovery/src/discovery/modelling/combinePermissionsIntoDiscovery.ts
```diff
@@ -1,73 +1,97 @@
...
+    permissionsOutput.eoasWithUpgradePermissions?.includes(entry.address) &&
+    !isZeroAddress(entry.address) &&
+    !isAlias(entry.address, allAddresses) &&
+    upgradesCriticalContract(receivedPermissions, discovery)
+
+  return {
+    receivedPermissions,
+    directlyReceivedPermissions: pick(false),
+    eoaWithUpgradePermissions: isEoaWithUpgradePermissions ? true : undefined,
+  }
```
adamiak: 🤖 **Comment by Claude Fable (AI)** — posted from Adrian's account, non-blocking

`buildPermissionEntry` always emits all three keys, with explicit `undefined` values for the missing ones. But `attachPermissions` deliberately uses `Object.assign` so that "a permission the project does not hold stays an absent key, not a key holding undefined - the diff treats those two as different" — and `Object.assign` copies an `undefined`-valued key onto the entry as an own key, violating exactly that invariant.

No live path hits this today, but only by accident of serialization: `DiscoveryRunner` and `updateDiffHistory` pass the output through `withoutUndefinedKeys` (a deep JSON round-trip) and disk writes go through `JSON.stringify`, all of which strip the `undefined`-valued keys before the map is ever read. A future in-memory combine → attach → diff path would produce phantom diffs.

Cheap hardening: only set the fields that are actually defined (then `isEmpty` collapses to `Object.keys(entry).length === 0`).

packages/backend/src/modules/update-monitor/UpdateDiffer.ts
```diff
@@ -100,11 +101,14 @@ export class UpdateDiffer {
...
-    const diff = diffDiscovery(discovery.entries, latestDiscovery.entries)
+    const diff = diffDiscovery(
+      entriesForDiff(discovery),
+      entriesForDiff(latestDiscovery),
+    )
 
     return this.getUpdateDiffs(
       diff,
-      latestDiscovery.entries,
+      entriesForDiff(latestDiscovery),
```
adamiak: 🤖 **Comment by Claude Fable (AI)** — posted from Adrian's account, non-blocking

`entriesForDiff(latestDiscovery)` is computed twice in this function (here and on L111). Unlike the old bare `latestDiscovery.entries`, each call shallow-copies every entry and redoes the permissions join, once per project per UM run. Hoisting it to a local (`const latestEntries = entriesForDiff(latestDiscovery)`) avoids the double work and makes it explicit that the diff and `getUpdateDiffs` must see the same entries.

Same pattern in `updateDiffHistory.ts` L106/L111 with `rerun.prevDiscovery`.

packages/l2b/src/commands/Inspect.ts
```diff
@@ -116,28 +116,33 @@ export const Inspect = command({
       )
     }
 
+    const permissions = discovery.permissions?.[entry.address] ?? {}
```
adamiak: 🤖 **Comment by Claude Fable (AI)** — posted from Adrian's account, non-blocking

Consistency thought, take it or leave it: this hand-rolls the map lookup, and `loadDiscoveryMappings.ts:144` does the same, so the read-path join now lives in three places (here, there, and `attachPermissions`). Calling `attachPermissions([discovery])` right after `readDiscovery`, the way `generateEntrypoints.ts` does, would have left the printing block below untouched and keeps the join semantics in one place if the map's keying ever changes.

adamiak:
🤖 **Comment by Claude Fable (AI)** — posted from Adrian's account, question / non-blocking

One thing worth double-checking outside this repo: the format change is breaking for any **external consumer** of `discovered.json`. An out-of-repo reader of `entry.receivedPermissions` / `entry.directlyReceivedPermissions` / `entry.eoaWithUpgradePermissions` won't error after this lands — it will silently see no permissions at all.

Are there any external tools, repos, scripts, or data pipelines (monitoring, research tooling, exports) reading these files that need a heads-up or a matching join on their side?

adamiak (approved):

mateuszradomski:
@codex, review

PR #12725, opened 2026-09-01
centralized sequencing research, ethereum baseline and OP clock-extension fix

vilcht:
Production split of the research prototype in #12430 (research/config half). Closes the config part of L2B-14607.

- fixes the OP fault-proof cumulative clock extension (10d → 10d 3h) and derives the 1.09493 bond factor from FaultDisputeGame gas literals, with tests
- models sequencing as a tagged union: sequencer-set spec (existing) + new 9-field centralized spec
- adds centralized specs for OP Mainnet, Base (shared Flashblocks helper), Arbitrum and Starknet — exit delay/economics derived from discovery with build-breaking asserts on chain state
- adds Ethereum, Gnosis and Polygon PoS sequencer-set specs and stake-distribution snapshots (generated by the l2b tooling in the stacked PR)
- base/optimism rediscovered on the same block to expose OptimismPortal2 minimum deposit gas limits
- stake-distribution.json is parsed with a validate schema at config load (incl. date-format checks), so malformed regenerations fail the build
- review fixes: bond-cost helper no longer throws on small inputs (display-approximate, documented)

Deliberately not carried from the prototype: the sequencingWindowSeconds template plumbing (no-op today — both users pass the default value).

Frontend is only adapted mechanically to the renamed field; the centralized spec renders nothing yet. The UI lands in the stacked PR.


packages/config/src/types.ts
```diff
@@ -58,15 +60,41 @@ export interface ProjectTechnologyChoice {
...
   isIncomplete?: boolean
   isUnderReview?: boolean
 }
 
+/**
+ * The sequencing section carries fields no other technology choice has.
+ * The union ties the inclusion-delay chart to sequencer sets: the chart
+ * models validator-set censorship, so a centralized spec cannot carry one.
+ */
+export type ProjectSequencingTechnologyChoice = ProjectTechnologyChoice & {
```
torztomasz: This three-way `ProjectTechnologyChoice & {...} & (A | B | C)` union with `?: never` fields is a type-level workaround for a data-model problem. The chart and its description belong *to the sequencer-set spec*, so nest them there and the union disappears:

```ts
export interface ProjectSequencerSetSpec {
  type: 'sequencer-set'
  blockTime?: TableReadyValue
  // ...
  inclusionDelayChart?: ProjectInclusionDelayChart
  inclusionDelayChartDescription?: string
}

export interface ProjectSequencingTechnologyChoice extends ProjectTechnologyChoice {
  sequencingSpec?: ProjectSequencingSpec
  censorshipResistance?: string
}
```

No `never` tricks, no JSDoc explaining them, and the frontend narrows once on `spec.type` and then reads `spec.inclusionDelayChart`.

packages/frontend/src/server/features/layer2s/risks/sequencing/getL2RiskSequencingEntries.ts
```diff
@@ -91,7 +91,7 @@ function getInclusionDelayComparison(
     .map((project) => {
       const sequencing = project.scalingTechnology.sequencing
       const chart = sequencing?.inclusionDelayChart
-      if (!sequencing?.sequencerSetSpec || !chart) {
+      if (sequencing?.sequencingSpec?.type !== 'sequencer-set' || !chart) {
```
torztomasz: This `sequencingSpec?.type !== 'sequencer-set' || !chart` shape now appears three times across the frontend. If the chart moves into `ProjectSequencerSetSpec` (see comment in `types.ts`), this collapses to a single `type` check followed by `spec.inclusionDelayChart`.

packages/config/src/projects/optimism/optimism.ts
```diff
@@ -1,18 +1,89 @@
...
 const discovery = new ProjectDiscovery('optimism')
 const genesisTimestamp = UnixTime(1636665399)
 const chainId = 10
+const l2BlockTimeSeconds = HARDCODED.OPTIMISM.L2_BLOCK_TIME_SECONDS
+const flashblockIntervalMilliseconds =
+  HARDCODED.OPTIMISM.FLASHBLOCK_INTERVAL_MILLISECONDS
+const sequencingWindowSeconds = HARDCODED.OPTIMISM.SEQUENCING_WINDOW_SECONDS
+const sequencingWindowBlocks = HARDCODED.OPTIMISM.SEQUENCING_WINDOW_BLOCKS
+const maxDepositCalldataBytes = HARDCODED.OPTIMISM.MAX_DEPOSIT_CALLDATA_BYTES
+const proofMaturityDelaySeconds = discovery.getContractValue<number>(
```
torztomasz: This ~60-line prelude (and the near-identical one in `base.ts`) re-reads `proofMaturityDelaySeconds`, `disputeGameFinalityDelaySeconds`, `respectedGameType`, `initBondGame{N}`, `maxClockDuration`, `maxGameDepth`, `clockExtension`, `challengePeriod`. The opStack template already reads every one of these in `getFinalizationPeriod`, `getChallengePeriod`, `getPermissionlessGameBond`, `getFaultDisputeGameName` and the `getStateValidation` switch, keyed on `getFraudProofType`.

The cleaner move: let the template derive `exitDelay` / `exitEconomics` from the fraud-proof type in the same switch that already computes the `describeOPFP` inputs (`Permissionless` → FDG math, `AggregateProof` → AggregateVerifier math). Then a project passes only what is genuinely project-specific:

```ts
sequencing: getOpStackCentralizedSequencing(templateVars, {
  hardcoded: HARDCODED.OPTIMISM,
  sequencer: { /* ... */ },
  trustedPreconfirmationDescription: '...',
  description, references, censorshipResistance,
})
```

Each project's 130-line block collapses to prose plus a sequencer cell.

packages/config/src/projects/optimism/optimism.ts
```diff
@@ -1,18 +1,89 @@
...
+const clockExtension = discovery.getContractValue<number>(
+  'FaultDisputeGame',
+  'clockExtension',
+)
+const respectedGameType = discovery.getContractValue<number>(
+  'OptimismPortal2',
+  'respectedGameType',
+)
+assert(
+  respectedGameType === 8,
```
torztomasz: This assert duplicates the `respectedGameType === 8` branch inside `getPermissionlessGameBond` in the template. If the exit-economics derivation lives in the template, the assert goes with it and there's one place that knows which game type is respected.

packages/config/src/projects/base/base.ts
```diff
@@ -1,23 +1,79 @@
...
 const discovery = new ProjectDiscovery('base')
 const genesisTimestamp = UnixTime(1686074603)
 const chainId = 8453
+const l2BlockTimeSeconds = HARDCODED.BASE.L2_BLOCK_TIME_SECONDS
+const flashblockIntervalMilliseconds =
+  HARDCODED.BASE.FLASHBLOCK_INTERVAL_MILLISECONDS
+const sequencingWindowBlocks = HARDCODED.BASE.SEQUENCING_WINDOW_BLOCKS
+const sequencingWindowSeconds = HARDCODED.BASE.SEQUENCING_WINDOW_SECONDS
+const maxDepositCalldataBytes = HARDCODED.BASE.MAX_DEPOSIT_CALLDATA_BYTES
+const proofMaturityDelaySeconds = discovery.getContractValue<number>(
```
torztomasz: Same as the comment on `optimism.ts`: this prelude re-derives values the opStack template already owns. The `AggregateProof` fraud-proof type in `getStateValidation` is the natural home for the AggregateVerifier exit-delay math.

packages/config/src/templates/opStack.ts
```diff
@@ -287,6 +289,80 @@ export interface OpStackConfigL3 extends OpStackConfigCommon {
   hostChain: string
 }
 
+export function getOpStackCentralizedSequencingCommon({
```
torztomasz: Two things here even if the bigger template-owned refactor is deferred:

1. This should take the `HARDCODED.<CHAIN>` object rather than five loose numbers that both callers copy out of it one by one.
2. `opStack.ts` goes from 2829 to 2955 lines in this PR and is already ~3x the healthy size. The `templates/opStack/` directory exists; put this builder in `templates/opStack/sequencing.ts` and the bond/clock math in `templates/opStack/faultDisputeGame.ts`, and have `opStack.ts` import them. The new test file should then import from those modules instead of the monolith.

packages/config/src/templates/opStack.ts
```diff
@@ -287,6 +289,80 @@ export interface OpStackConfigL3 extends OpStackConfigCommon {
...
+}: {
+  discovery: ProjectDiscovery
+  l2BlockTimeSeconds: number
+  flashblockIntervalMilliseconds: number
+  sequencingWindowSeconds: number
+  sequencingWindowBlocks: number
+  maxDepositCalldataBytes: number
+  trustedPreconfirmationDescription: string
+  sequencer: TableReadyValue
+}): Omit<
```
torztomasz: Returning `Omit<Spec, 'type' | 'exitDelay' | 'exitEconomics'>` and having every caller spread it and re-tag with `type: 'centralized'` is an awkward partial-object contract. Take `exitDelay` / `exitEconomics` as inputs and return the complete, typed `ProjectCentralizedSequencingSpec`.

packages/config/src/templates/opStack.ts
```diff
@@ -1332,6 +1408,46 @@ function getStateValidation(
...
 
+export function getOpStackBondScalingFactor(gameMaxDepth: number): number {
+  return (
+    (HARDCODED.OPTIMISM.FAULT_PROOF_HIGH_GAS_CHARGED /
+      HARDCODED.OPTIMISM.FAULT_PROOF_BASE_GAS_CHARGED) **
+    (1 / gameMaxDepth)
+  )
+}
+
+export function getOpStackFullDisputeGameBondCost(
```
torztomasz: This keeps machinery its only caller throws away: scale by 100,000, loop, round, multiply back into a `bigint`, accept `number | string`, and document it as "display-approximate". The one production caller immediately does `Number.parseFloat(formatEther(...))`.

Return ether as a `number` from a closed-form geometric sum:

```ts
export function getOpStackFullDisputeGameBondCostEther(initialBondWei: number, gameMaxDepth: number): number {
  const f = getOpStackBondScalingFactor(gameMaxDepth)
  return Number(formatEther(initialBondWei)) * (f ** (gameMaxDepth + 1) - 1) / (f - 1)
}
```

The `scaleFactor`, the `bigint`, the union param, the caveat, and the small-input test all go away.

packages/config/src/discovery/values/hardcoded.ts
```diff
@@ -14,23 +14,101 @@ UNDERSTAND WHAT YOU ARE DOING BEFORE YOU UPDATE THIS FILE
 
 // This is the place to insert hardcoded values
 // which should be used inside project's hardcoded tests
+const ETHEREUM_BLOCK_TIME_SECONDS = 12
+const OP_MAINNET_SEQUENCING_WINDOW_BLOCKS = 3_600
+const BASE_SEQUENCING_WINDOW_BLOCKS = 3_600
+
 export const HARDCODED = {
```
torztomasz, line 18: The file header says this is for data that discovery cannot fetch, used in hardcoded tests. This PR turns it into a general protocol-constants bag (Ethereum consensus params, Starknet Tendermint sizing, Flashblock intervals) and rewrites `assumedBlockTime = 12` in five files to point here. Meanwhile `arbitrum.ts` still inlines `delayBufferReplenishIntervalSeconds = 1_200` and `basisPointsDivisor = 10_000`.

Pick one and be consistent: either update the header to say this is the home for offchain protocol parameters and move the Arbitrum numbers here too, or keep display constants local to the project.

packages/config/src/discovery/values/hardcoded.ts
```diff
@@ -14,23 +14,101 @@ UNDERSTAND WHAT YOU ARE DOING BEFORE YOU UPDATE THIS FILE
...
+    // block_time and seq_window_size can be independently verified by calling
+    // optimism_rollupConfig on an OP Mainnet op-node RPC. They deliberately
+    // remain hardcoded because this RPC describes node configuration rather
+    // than historical L1 state and is not supported by every RPC provider.
     // https://docs.optimism.io/chain-operators/guides/configuration/rollup#sequencerwindowsize
     // https://github.com/ethereum-optimism/optimism/blob/develop/op-deployer/pkg/deployer/state/deploy_config.go#L93
     // https://github.com/ethereum-optimism/optimism/blob/51eeb76efeb32b3df3e978f311188aa29f5e3e94/packages/contracts-bedrock/deploy-config/mainnet.json#LL10C26-L10C30
-    SEQUENCING_WINDOW_SECONDS: 3600 * 12, // blocks * blocktime
+    L2_BLOCK_TIME_SECONDS: 2,
+    SEQUENCING_WINDOW_BLOCKS: OP_MAINNET_SEQUENCING_WINDOW_BLOCKS,
```
torztomasz: `SEQUENCING_WINDOW_BLOCKS` and `SEQUENCING_WINDOW_SECONDS` stored side by side (here and for `BASE`) is a derived value stored next to its inputs, with module-level temporaries to make it work. Keep blocks and derive seconds at the call site.

packages/config/src/projects/arbitrum/arbitrum.ts
```diff
@@ -132,8 +138,63 @@ const maxTimeVariation = discovery.getContractValue<{
...
+const worstCaseExitDelaySeconds =
+  currentForceInclusionDelayBlocks * assumedBlockTime +
+  worstCaseStateFinalizationDelaySeconds
+const worstCaseExitDelayDisplaySeconds =
+  Math.ceil(worstCaseExitDelaySeconds / 3_600) * 3_600
+const worstCaseStateFinalizationDisplaySeconds =
+  Math.ceil(worstCaseStateFinalizationDelaySeconds / 3_600) * 3_600
 
 const selfSequencingDelay = maxTimeVariation.delaySeconds
+const delayBufferReplenishIntervalSeconds = 1_200
```
torztomasz: These two magic numbers contradict the PR's own convention of moving offchain parameters into `HARDCODED`. Either they go to `HARDCODED.ARBITRUM` with a source link like the others, or the convention isn't one.

packages/config/src/projects/arbitrum/arbitrum.ts
```diff
@@ -132,8 +138,63 @@ const maxTimeVariation = discovery.getContractValue<{
...
+)
+const edgeChallengePeriodSeconds = edgeChallengePeriodBlocks * assumedBlockTime
+const worstCaseStateFinalizationDelaySeconds =
+  minimumAssertionPeriodSeconds +
+  edgeChallengePeriodSeconds * 2 +
+  challengeGracePeriodSeconds
+const worstCaseExitDelaySeconds =
+  currentForceInclusionDelayBlocks * assumedBlockTime +
+  worstCaseStateFinalizationDelaySeconds
+const worstCaseExitDelayDisplaySeconds =
```
torztomasz: `Math.ceil(x / 3_600) * 3_600` is done twice here for display rounding. `formatSeconds` already truncates to two units. Either drop the rounding or extract a single helper; inline arithmetic duplicated in a project file is the kind of thing that drifts.

packages/config/src/projects/ethereum/ethereum.ts
```diff
@@ -1,10 +1,34 @@
...
+  SEQUENCING_SPEC,
+} from '../../common'
 import { linkByDA } from '../../common/linkByDA'
-import type { BaseProject } from '../../types'
+import { HARDCODED } from '../../discovery/values/hardcoded'
+import { type BaseProject, ProjectStakeDistributionSchema } from '../../types'
 import { readProjectMarkdown } from '../../utils/readMarkdown'
+import stakeDistributionJson from './stake-distribution.json'
+
+const stakeDistribution = ProjectStakeDistributionSchema.parse(
```
torztomasz: `ProjectStakeDistributionSchema.parse(json)` is copy-pasted into four project files, and ethereum/gnosis then each `assert(validatorCount !== undefined)`. Add a `readStakeDistribution(json, { requireValidatorCount })` helper next to `readProjectMarkdown` in `utils/`, or split the schema so the ethereum-like chart input requires `validatorCount` and the asserts disappear.

packages/config/src/projects/ethereum/ethereum.ts
```diff
@@ -69,40 +176,41 @@ export const ethereum: BaseProject = {
...
         // BPO2: Blob Parameter Only fork 2
         size: 2_752_512, // 2.625 MiB (max 21 blobs × 128 KiB)
         target: 1_835_008, // 1.75 MiB (target 14 blobs × 128 KiB)
-        frequency: 12, // unchanged: 12 s slot time
+        frequency: ethereumBlockTimeSeconds,
         sinceTimestamp: 1767747671, // 2026-01-07 01:01:11 UTC – epoch 419072
       },
     ],
-    finality: 768, // seconds
+    finality: ethereumBlockTimeSeconds * 2 * HARDCODED.ETHEREUM.SLOTS_PER_EPOCH,
```
torztomasz: `ethereumBlockTimeSeconds * 2 * SLOTS_PER_EPOCH` is computed twice (here and for `consensusFinality`). One `const ethereumFinalitySeconds`.

packages/config/src/projects/stakeDistribution.test.ts
```diff
@@ -0,0 +1,56 @@
+import { expect } from 'earl'
+import { ProjectStakeDistributionSchema } from '../types'
+
+describe('stake distribution schema', () => {
```
torztomasz: This tests a schema that lives in `types.ts` but sits in `projects/`. Move it beside what it tests (or beside the `readStakeDistribution` helper if you add one).

packages/shared-pure/src/utils/dateStrings.ts
```diff
@@ -0,0 +1,17 @@
...
+  const parsed = Date.parse(`${value}T00:00:00Z`)
+  // The round-trip rejects rolled-over dates like 2026-02-31, which
+  // Date.parse silently accepts as March 3rd.
+  return (
+    !Number.isNaN(parsed) && new Date(parsed).toISOString().startsWith(value)
+  )
+}
+
+/** True for any string Date.parse understands, e.g. an RFC 3339 timestamp. */
+export function isTimestamp(value: string): boolean {
```
torztomasz: `isTimestamp` is a one-line wrapper over `Date.parse` with a name broader than its behavior (it accepts `"2026"`, `"March 5"`, etc.). Inline `!Number.isNaN(Date.parse(d.date))` in the schema check. `isDateOnly` is worth keeping.

packages/config/src/common/sequencingSpec.ts
```diff
@@ -0,0 +1,35 @@
...
+      description ??
+      'The centralized sequencer can censor transactions submitted through the normal L2 path.',
+  }
+}
+
+function OpenBlockProduction(description?: string): TableReadyValue {
+  return {
+    value: 'Open',
+    sentiment: 'good',
+    ...(description !== undefined ? { description } : {}),
```
torztomasz: `exactOptionalPropertyTypes` isn't enabled in this repo, so `...(description !== undefined ? { description } : {})` can just be `description`. Same at line 25.

packages/config/src/projects/starknet/starknet.ts
```diff
@@ -398,6 +423,133 @@ export const starknet: ScalingProject = {
         OPERATOR.CENTRALIZED_OPERATOR.description +
         ' Typically, the Operator is the hot wallet of the Starknet service submitting state updates for which proofs have been already submitted and verified.',
     },
+    sequencing: {
```
torztomasz, line 426: Softer version of the `opStack.ts` size comment: this file goes 2206 → 2358 lines. The 150-line sequencing block is self-contained and could live in `starknet/sequencing.ts`.

torztomasz (changes requested):
The research content is strong and the clock-extension fix is right, but the config-side structure keeps a lot of incidental complexity that has a visible simpler shape. Inline comments below; the ones I'd hold the merge on are the three-way sequencing union in `types.ts`, the duplicated OP/Base prelude that re-derives what the opStack template already owns, and the bond-cost helper. Each of those deletes a layer rather than rearranges it.

Not run locally: typecheck and tests on this branch.

torztomasz (approved):

packages/config/src/projects/optimism/optimism.ts
vilcht: Done, with one change to the shape. `templateVars` isn't reachable from inside the project's own `opStackL2({...})` literal, so instead of `getOpStackCentralizedSequencing(templateVars, {...})` the project passes a `centralizedSequencing: { hardcoded, description, trustedPreconfirmationDescription, sequencer, censorshipResistance, references }` option and the template builds `technology.sequencing` itself.

The exit derivation is keyed on `getFraudProofType` (Permissionless → FDG clocks + bond scaling, AggregateProof → AggregateVerifier delays + checkpoint bond, anything else throws so a respected-game-type change fails loudly). It lives in `templates/opStack/sequencing.ts` rather than inside the `getStateValidation` switch, to keep `opStack.ts` from growing again. Both preludes are gone; Optimism and Base each pass ~40 lines of prose plus the sequencer cell.

packages/config/src/discovery/values/hardcoded.ts
vilcht: Done for Base, where the seconds key was new. Kept `HARDCODED.OPTIMISM.SEQUENCING_WINDOW_SECONDS` because it predates this PR and has about a dozen readers that assume the OP Mainnet default (`opStack.ts`, `agglayer.ts`, Katana, Zircuit, Lightlink, Bugbuster). Dropped the module-level temporaries and left a comment on the remaining key. Happy to sweep those readers in a separate PR if you want the key gone.

packages/config/src/projects/arbitrum/arbitrum.ts
vilcht: Extracted a `ceilToFullHours` helper so the arithmetic isn't duplicated, but kept rounding up rather than relying on `formatSeconds` truncation: these are worst-case figures, and truncating would show 16h59m as 16 hours, under-reporting the delay. Left that reasoning in a comment next to the helper.

packages/config/src/projects/starknet/starknet.ts
vilcht: Left this one as is. The block is self-contained and project-specific, and unlike `opStack.ts` nothing else would import it, so a split only moves lines around. Will pick it up the next time this file is touched if you feel strongly about it.

PR #12726, opened 2026-09-01
centralized sequencing table, ethereum baseline row and stake distribution tooling

vilcht:
Production split of the research prototype in #12430 (tooling/frontend half). Stacked on the research PR;

- moves DuneQueryService from backend to @l2beat/shared (pure relocation; backend changes are import churn)
- fixes getstakedistrib default output path to resolve from .discovery.json instead of cwd
- l2b getstakedistrib: ethereum source via Dune (validated row schema, Lido split by node operator) and gnosis via the analytics API (snapshot-alignment check, aggregate-only); emits the dateType/date shape; first tests for this module
- sequencing risk page: centralized-sequencers table (9 research columns), Ethereum pinned as the decentralized baseline row (links to the DA Ethereum page), comparison chart uses Ethereum's real config chart and marks largest-entity cumulative stakes per series, stake-freshness tooltip on set size
- project pages: centralized sequencing spec sheet

Changes vs the prototype: no view-layer sentiment override (RT CR renders config red), hand-rolled Dune row validation replaced by the parser the service natively accepts, aztec pagination hardening dropped, frontend date-union adapter unnecessary (config carries the discriminant).


torztomasz:
image

packages/frontend/src/components/projects/sections/sequencing/CentralizedSequencingSpecSheet.tsx
```diff
@@ -0,0 +1,65 @@
...
+              </td>
+            </tr>
+          ))}
+        </tbody>
+      </table>
+    </div>
+  )
+}
+
+function SpecValue({ value }: { value: TableReadyValue }) {
```
torztomasz: This whole component is a copy of `SequencerSetSpecSheet`: the wrapper div, table, header row, row markup and this `SpecValue` helper are all duplicated. And the copy already drifts: this `SpecValue` always wraps in `SentimentText` with a neutral fallback, while the original branches on sentiment.

Can we have one `SpecSheet({ title, rows: { label, value }[] })` component that owns the markup and `SpecValue`? Both sheets then collapse to a field list plus a one-line render. Please unify the sentiment behavior while you're there.

packages/frontend/src/pages/layer2s/risk/sequencing/components/table/columns.tsx
```diff
@@ -78,7 +95,24 @@ export const l2SequencingColumns = [
     columnHelper.accessor((entry) => adjustTableValue(entry[key]), {
       id: key,
       header,
-      cell: (ctx) => <TableValueCell value={ctx.row.original[key]} />,
+      cell: (ctx) => {
+        if (key !== 'sequencerCount') {
```
torztomasz: This adds a special-case branch inside the loop that builds every table-value column. Either lift `sequencerCount` out into its own explicit column definition, or give the `tableValueColumns` descriptor an optional `cell` override so the loop stays uniform.

packages/frontend/src/server/features/layer2s/risks/sequencing/getL2RiskSequencingEntries.ts
```diff
@@ -137,22 +182,94 @@ function getInclusionDelayComparison(
     { key: ETHEREUM_SERIES_KEY, label: 'Ethereum', type: 'ethereum' as const },
   ]
 
-  return { data, series, maxCensorFraction }
+  const entityMarkers: InclusionDelayComparisonEntityMarker[] = [
```
torztomasz: Ethereum is threaded through as a special case three times here: appended separately to the merged data (line 173), the series list (line 182), the entity markers (this block) and `maxCensorFraction`.

I think there's a code-judo move: build a single list of `{ key, label, type, points, entityMarkers, maxCensorFraction }` with Ethereum as the first element, then every output becomes one map over that list. `ETHEREUM_SERIES_KEY` and the duplicated marker-prefixing spread disappear.

packages/frontend/src/server/features/layer2s/risks/sequencing/getL2RiskSequencingEntries.ts
```diff
@@ -94,38 +141,36 @@ function getInclusionDelayComparison(
...
       }
     })
     .filter(notUndefined)
     .sort((a, b) => a.name.localeCompare(b.name))
 
-  const [first] = projectDelays
-  if (!first) {
+  const ethereumChart =
+    ethereum.scalingTechnology.sequencing?.inclusionDelayChart
+  if (projectDelays.length === 0 || !ethereumChart) {
```
torztomasz: Inconsistent boundary: the entry path asserts that Ethereum sequencing config exists, but this path silently returns `undefined` when the chart is missing. Both come from the same config object that is always populated. Assert once at the top and pass the non-optional `sequencing` down so this fallback goes away.

packages/frontend/src/server/features/layer2s/risks/sequencing/getL2RiskSequencingEntries.ts
```diff
@@ -137,22 +182,94 @@ function getInclusionDelayComparison(
...
+      ...marker,
+      id: `${ETHEREUM_SERIES_KEY}-${marker.id}`,
+      seriesKey: ETHEREUM_SERIES_KEY,
+    })),
+  ]
+
+  return { data, series, entityMarkers, maxCensorFraction }
+}
+
+function hasDelay(
```
torztomasz: This is identical to `hasFiniteDelay` in `ProjectInclusionDelayChart.tsx`. Put it in `calculateInclusionDelay.ts`, or have `getInclusionDelayData` return the filtered `entityMarkers` alongside `entityLegendEntries` so neither caller needs the guard.

packages/frontend/src/server/features/layer2s/risks/sequencing/getL2RiskSequencingEntries.ts
```diff
@@ -137,22 +182,94 @@ function getInclusionDelayComparison(
...
+    name: project.name,
+    shortName: project.shortName,
+    backgroundColor: 'blue',
+    statuses: undefined,
+    description: project.display.description,
+    ...values,
+  }
+}
+
+type SequencingValues = Pick<
```
torztomasz: `Omit<L2RiskSequencingEntry, keyof CommonProjectEntry>` says the same thing and can't drift when a column is added.

packages/frontend/src/server/features/layer2s/risks/sequencing/getL2RiskSequencingEntries.ts
```diff
@@ -32,25 +45,36 @@ export interface L2RiskSequencingEntry extends CommonL2Entry {
   additionalCrGadgets: TableReadyValue | undefined
 }
 
+export interface L2RiskCentralizedSequencingEntry
+  extends CommonL2Entry,
```
torztomasz: `L2RiskSequencingEntry` now extends `CommonProjectEntry` but this one still extends `CommonL2Entry`. Neither table uses `tab` or `filterable`. Use `CommonProjectEntry` for both.

packages/frontend/src/server/features/layer2s/risks/sequencing/getL2RiskSequencingEntries.ts
```diff
@@ -137,22 +182,94 @@ function getInclusionDelayComparison(
...
+    return undefined
+  }
+
+  return {
+    ...getCommonL2Entry({ project, changes }),
+    ...values,
+  }
+}
+
+function getEthereumSequencingEntry(
```
torztomasz: This is the second hand-rolled Ethereum `CommonProjectEntry` (the activity feature has `getEthereumEntry`). A shared `getEthereumCommonEntry()` under `server/features/utils` would be the canonical home for both.

packages/l2b/src/implementations/getStakeDistrib.ts
```diff
@@ -77,6 +179,16 @@ export class StakeDistributionFetcher {
 
   async fetchAndDisplay(): Promise<void> {
     const projects = this.getProjectsToFetch()
+    // Resolve prerequisites (Dune credentials, output location) upfront so a
+    // misconfigured environment fails before any network work is done.
+    if (projects.includes('ethereum')) {
+      assert(process.env[DUNE_API_KEY_ENV_NAME], MISSING_DUNE_API_KEY_MESSAGE)
+    }
+    const defaultOutputRoot =
```
torztomasz: `defaultOutputRoot` is computed as `string | undefined` here, threaded into `writeJsonOutput`, then re-asserted non-undefined there. Separately, `extractProjectData` merges `project` into the distribution and `getStakeDistributionOutput` strips it back out.

Can we keep `{ project, distribution }` pairs and resolve an output plan of `{ path, data }` up front? Then writing is `Promise.all(plan.map(writeJsonFile))` and `ExtractedStakingProjectData`, `getStakeDistributionOutput`, `getJsonOutput`, and the assert all go away. The Dune key assert here is also repeated in `executeDuneSql`.

packages/frontend/src/pages/layer2s/risk/sequencing/components/centralized-table/CentralizedSequencingTable.tsx
```diff
@@ -0,0 +1,51 @@
...
+    onSortingChange: setSorting,
+    initialState: {
+      columnPinning: {
+        left: ['#', 'logo'],
+      },
+    },
+  })
+
+  return (
+    <PrimaryCard className="mt-6">
```
torztomasz: Card ownership is inconsistent between the two tables: the page now wraps `L2RiskSequencingTable` in a `PrimaryCard`, while this component renders its own card, heading and description. Pick one owner for the card and heading across both.

packages/frontend/src/components/projects/sections/sequencing/centralizedSequencingFields.ts
```diff
@@ -0,0 +1,75 @@
...
+  },
+  exitEconomics: {
+    label: 'Exit economics',
+    header: 'Exit\neconomics',
+    tooltip:
+      'Capital and proof or dispute work required to self-propose and defend one state update needed for an exit under blanket censorship or operator walkaway.',
+  },
+}
+
+export const CENTRALIZED_SEQUENCING_FIELD_KEYS = Object.keys(
```
torztomasz: The record plus `Object.keys(...) as` cast exists only to recover ordering. The sibling patterns (`SPEC_ROWS`, `tableValueColumns`) use an ordered array of `{ key, label, ... }`, which needs no cast. Match them.

torztomasz (commented):
Code quality review. Behavior looks fine and everything passes locally (biome clean, frontend + l2b typecheck, l2b + shared tests green), but there are a few structural issues I'd like addressed before merging. Moving `DuneQueryService` into `@l2beat/shared` is the right call, no changes needed there.

**Structural (please address):**
1. `CentralizedSequencingSpecSheet` is a copy of `SequencerSetSpecSheet` (see inline)
2. Special-case branch inside the generic column map (see inline)
3. Ethereum is special-cased three times in the comparison builder (see inline)
4. Inconsistent Ethereum boundary: assert on one path, silent `undefined` on the other (see inline)
5. `getStakeDistrib.ts` output orchestration carries optionality churn (see inline)

**Smaller cleanups:** duplicate `hasDelay` / `hasFiniteDelay` type guard, hand-listed `SequencingValues` Pick, mixed `CommonL2Entry` / `CommonProjectEntry` bases, third hand-rolled Ethereum entry, inconsistent `PrimaryCard` ownership between the two tables, `Object.keys(...) as` cast for field ordering. All inline.

**Decomposition:** `getStakeDistrib.ts` is at 625 lines and now mixes CLI orchestration, four source fetchers, a 60-line Dune query, response schemas and console formatting. Not past the 1k line, but it's already four concerns in one file. Splitting per-source fetchers into `stake-distribution/sources/*.ts` with the fetcher class on its own would make the next source addition trivial.

torztomasz (approved):

packages/frontend/src/components/projects/sections/sequencing/CentralizedSequencingSpecSheet.tsx
vilcht: Unified into a single `SpecSheet({ title, rows })`. On the sentiment behavior I went with always rendering `SentimentText` with `sentiment ?? 'neutral'`: the branching version dropped the description tooltip for values without a sentiment, which is most of the centralized cells (and Ethereum's rate limit on the sequencer-set sheet). Neutral renders in the default text color, so plain values look the same as before.

packages/frontend/src/server/features/layer2s/risks/sequencing/getL2RiskSequencingEntries.ts
vilcht: Added `getEthereumCommonEntry()` under `server/features/utils` and pointed both the sequencing and activity entries at it. Left the DA summary's Ethereum entry alone since it overrides most of the common fields (icon by layer slug, `statuses: {}`, `nameSecondLine`, href).

PR #12729, opened 2026-09-01
Model permissions across referenced projects

mateuszradomski:
Part of L2B-10994


mateuszradomski:
@codex, review

mateuszradomski:
@codex, review

packages/discovery/src/discovery/modelling/modelPermissions.ts
```diff
@@ -54,6 +59,60 @@ export class DiscoveryRegistry {
...
+// is newer than what is on disk.
+export function addReferencedDiscoveries(
+  discoveries: DiscoveryRegistry,
+  project: string,
+  configReader: ConfigReader,
+  logger: Logger = Logger.SILENT,
+): void {
+  let referenced: DiscoveryRegistry
+  try {
+    referenced = loadDiscoveriesForModelling(project, configReader)
```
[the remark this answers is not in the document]
mateuszradomski: We are fine with this, we accept this case

packages/backend/src/modules/update-monitor/DiscoveryRunner.ts
```diff
@@ -54,27 +54,17 @@ export class DiscoveryRunner {
...
-      toDiscover,
+      [projectName],
       discoveryTimestamp,
       configReader,
       logger,
     )
+    // Projects reached through an entrypoint are deliberately not
+    // rediscovered: modelling runs against their committed discovery, and
+    // keeping the two sides in sync is handled through Update Monitor.
+    addReferencedDiscoveries(discoveries, projectName, configReader, logger)
```
adamiak, line 68: _Comment left by Claude Fable 5.1 on Adrian Adamiak's behalf._

`combinePermissionsIntoDiscovery` below (line 78) is called without the cluster entries that both other write paths pass: `modelPermissionsCommand.ts` and the historical rerun in `updateDiffHistory.ts` hand it `clusterEntries(discoveries)` as the third argument. Without them `isAlias` and `upgradesCriticalContract` only see the project's own entries, so an EOA whose upgrade target lives in the referenced module gets `eoaWithUpgradePermissions: true` from the CLI but nothing from Update Monitor, which would then report a permanent phantom diff.

No committed project hits this today (I checked every `discovered.json` for a flagged holder with an upgrade target outside its project), so it is latent, but the write paths should agree. Passing `clusterEntries(discoveries)` here fixes it.

packages/discovery/src/discovery/modelling/combinePermissionsIntoDiscovery.ts
```diff
@@ -17,25 +17,35 @@ import type {
...
 
   const allAddresses = new Set(
-    discovery.entries.map((e) => e.address.toLowerCase()),
+    clusterEntries.map((e) => e.address.toLowerCase()),
+  )
+
+  // Keyed by receiver rather than by entry: modelling now spans the whole
+  // cluster, so an address that holds a permission is often owned by a
+  // referenced project and has no entry here beyond a Reference stub.
+  const receivers = new Set(
```
adamiak, line 36: _Comment left by Claude Fable 5.1 on Adrian Adamiak's behalf._

This now stores an entry for every receiver in the cluster, but the diff path does not follow: `entriesForDiff` in `diffDiscovery.ts` still joins the map onto the project's own entries only. For zkcandy, 18 of the 23 stored holders have no entry in the project (the Security Council multisig, for example), so a change to their permissions is stored but never appears in Update Monitor, the dashboard, or `diffHistory.md` for the consumer. Holders that happen to have a Reference stub (ProxyAdmin, ProtocolUpgradeHandler) instead get the module's whole graph dumped onto the stub, as the new zkcandy `diffHistory.md` entry shows.

Not a regression, since the module's own run still catches its own changes, but the stored snapshot and the diffed snapshot now disagree. Options: diff the map directly, attach onto the cluster's entries in `entriesForDiff`, or store only what the diff can see.

packages/discovery/src/discovery/modelling/modelPermissions.ts
```diff
@@ -146,7 +208,11 @@ export async function modelPermissionFactsUsingClingo(
 
   const result = facts.map(parseClingoFact)
 
-  const permissionsConfigHash = generatePermissionConfigHash(clingoForProject)
+  // Scoped to this project: a change inside a shared module must not turn
```
adamiak: _Comment left by Claude Fable 5.1 on Adrian Adamiak's behalf._

This is less insulated than the comment (and the matching one in `config.test.ts`) says. The project's own clingo is generated with the cluster-wide `addressToNameMap`: `tryCastingToName` emits an id when the cluster knows the address and a quoted raw string when it does not, and `buildPermissionsModel` drops the permission fact entirely in the latter case. So when a shared module starts or stops discovering any address a consumer's values mention, that consumer's own clingo, and therefore this hash, changes, and `model-permissions is up to date` fails for every such consumer until each is re-modelled.

The fix is mechanical (`l2b model-permissions <consumer>`), so this may be acceptable, but then the two comments should say that module drift can still turn consumers red, or the hash should be computed from inputs that are truly project-local.

mateuszradomski:
@codex, review

mateuszradomski:
@codex, review

packages/discovery/src/discovery/modelling/modelPermissions.ts
```diff
@@ -146,7 +208,15 @@ export async function modelPermissionFactsUsingClingo(
...
-  const permissionsConfigHash = generatePermissionConfigHash(clingoForProject)
+  // Scoped to the clingo generated for this project, which keeps most shared
+  // module churn out of it. Not fully insulated though: that clingo is written
+  // against the cluster's address map, so a module that starts or stops
+  // discovering an address this project's values mention does change the hash
+  // and does turn this project red. The fix is `l2b model-permissions`, which
+  // is offline, so the trade is accepted.
+  const ownClingo = clingoByProject[project]
+  assert(ownClingo !== undefined, `No clingo generated for ${project}.`)
+  const permissionsConfigHash = generatePermissionConfigHash(ownClingo)
```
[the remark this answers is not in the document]
mateuszradomski: We accept this for now

mateuszradomski:
@codex, review

mateuszradomski:
@codex, review

mateuszradomski:
@codex, review

mateuszradomski:
@codex, review

adamiak (approved):

PR #12730, opened 2026-09-01
Support for Relay V3

sdlyy:
Resolves L2B-14850

@codex review

Smoke test is failing since I haven't added envs yet

## Summary

Migrates Relay interoperability ingestion from the unauthenticated `/requests/v2` endpoint to the authenticated `/requests/v3` API. The change adapts Relay’s new response schema, adds rate limiting and bounded retries, and prevents incomplete paginated windows from advancing the indexer checkpoint.

## Key implementation details

- `RelayApiClient` now calls `/requests/v3` with `INTEROP_RELAY_API_KEY` in the `x-api-key` header.
- Raw V3 responses are validated and normalized into a small internal `RelayRequest` model consumed by the indexer:
  - `txHash` is normalized to `hash`.
  - Source and destination transactions come from the first `inTxs` and `outTxs` entries.
  - Actual route amounts/currencies are preferred, with quoted route data used as fallback.
  - Nullable or missing route sections are supported.
- The indexer continues processing successful requests into `TokenSent` and `TokenReceived` events, but now depends only on normalized fields rather than Relay’s raw API schema.
- Pagination remains ordered by `updatedAt` ascending and walks continuation cursors with a maximum page size of 50.
- A repeated continuation cursor is rejected to prevent an infinite pagination loop.
- Reaching `INTEROP_RELAY_MAX_REQUESTS_PER_UPDATE` with a continuation remaining fails the window instead of committing partial data.
- Later-page failures are no longer returned as partial success. The entire window retries, preventing the UIF checkpoint from skipping requests.
- Network errors, HTTP 408/429, and 5xx responses are retried with bounded exponential backoff. Permanent 4xx errors fail immediately.
- Requests are rate-limited through `INTEROP_RELAY_CALLS_PER_MINUTE`, defaulting to 200.

## Configuration and schema

`InteropFeatureConfig.relay` is now independently feature-gated and can be `false`. When enabled, it requires:

- `INTEROP_RELAY_API_KEY`
- `INTEROP_RELAY_CALLS_PER_MINUTE` — defaults to `200`
- Existing batch-size, request-limit, and safe-time-offset settings

Relay indexers are only constructed and started when `interop.relay` is enabled. Relay ingestion is no longer implicitly tied to whether matching is enabled.

The API validation schema was replaced with the consumed subset of the V3 response and an internal normalized request contract. There are no database schema changes or migrations.

## Tests

Coverage includes:

- V3 endpoint and authentication header.
- Raw V3 transaction and route normalization.
- Actual-to-quoted route fallback.
- Multi-page continuation handling.
- Rate-limit and transient-error retries.
- Permanent and exhausted page failures.
- Repeated cursor and invalid limit guards.
- Rejection of incomplete dense windows.
- `TokenSent` and `TokenReceived` generation from normalized V3 requests.

The standalone Relay inspection script was also updated to require the API key and consume the normalized fields.


packages/backend/src/modules/interop/plugins/relay/RelayApiClient.ts
```diff
@@ -29,214 +19,265 @@ type GetAllRequestsOptions = Omit<
...
+
+  return {
+    id: request.id,
+    status: request.status,
+    sourceTx: normalizeTransaction(request.data.inTxs?.[0]),
+    destinationTx: normalizeTransaction(request.data.outTxs?.[0]),
+    sourceCurrency:
+      actual?.origin?.inputCurrency ??
+      quoted?.origin?.inputCurrency ??
+      undefined,
```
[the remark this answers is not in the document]
sdlyy: The current behavior - preserving incomplete actual data - is safer than silently fabricating actual values from a quote. If we want defensive hardening, the correct change is observability for incomplete successful actual data, not merging it.

packages/backend/src/config/features/interop.ts
```diff
@@ -98,14 +98,18 @@ export async function getInteropFeatureConfig(
...
     },
-    relay: {
-      batchSize: env.integer('INTEROP_RELAY_BATCH_SIZE', 60),
-      maxRequestsPerUpdate: env.integer(
-        'INTEROP_RELAY_MAX_REQUESTS_PER_UPDATE',
-        10_000,
-      ),
-      safeTimeOffset: env.integer('INTEROP_RELAY_SAFE_TIME_OFFSET', 10),
-    },
+    relay: flags.isEnabled('interop', 'relay')
```
adamiak, line 101: _Written by Claude on behalf of Adrian Adamiak._

Gating on `interop.relay` widens where Relay ingestion runs. `FeatureFlags.resolve` treats a child flag as enabled whenever its parent is, so every deployment with `interop` or `*` in `FEATURES` now resolves this to true unless it explicitly sets `!interop.relay`. Before this PR the Relay indexers only started under `interop.matching`.

Two consequences:
- Every such environment must define `INTEROP_RELAY_API_KEY` or `getConfig()` throws at startup. That is the current smoke-test failure for staging and production, and local `FEATURES=*` setups from the README are affected as well.
- If any second backend instance shares the DB with interop on but matching off, it will now also run the `relay_indexer` UIF indexer against the same indexer state and the same API key.

Before merge: add the key to both Coolify envs and confirm only one instance resolves this flag to true, or keep the start coupled to `interop.matching` as before.

packages/backend/src/modules/interop/plugins/relay/RelayApiClient.ts
```diff
@@ -29,214 +19,265 @@ type GetAllRequestsOptions = Omit<
...
+    chainId: transaction.chainId,
+  }
+}
+
+function isRetryableRelayError(error: unknown): boolean {
+  if (!(error instanceof Error)) {
+    return true
+  }
+
+  const match = /^HTTP error: (\d{3})/.exec(error.message)
```
adamiak: _Written by Claude on behalf of Adrian Adamiak._

`fetchWithRetry` re-implements what `ClientCore` in `@l2beat/shared` already provides (`RateLimiter` + `RetryHandler` + metrics), and retryability here is decided by regex-matching the `HTTP error: NNN` text thrown by `HttpClient.fetch` in another package. If that message wording ever changes, every 4xx silently becomes retryable (or nothing does), and no test would catch it.

Suggestion: extend `ClientCore` like the other shared API clients, or at minimum use `RetryHandler` together with `httpClient.fetchRaw` so this branch reads `res.status` instead of parsing a message.

One caveat if moving to `ClientCore`: `RetryHandler.retry` logs `url` and `init` on each retry, so a header-borne `x-api-key` would land in warn logs. That hazard already exists for the Coingecko query-param key and is best fixed in `RetryHandler` itself (omit/sanitize) rather than being a reason to keep a parallel retry loop here.

packages/backend/src/modules/interop/plugins/relay/relay.indexer.ts
```diff
@@ -136,7 +133,7 @@ export class RelayIndexer extends ManagedChildIndexer {
 
     if (res.continuation) {
       throw new Error(
-        `Window ${from}-${syncedTo} incomplete after ${res.requests.length} requests. Check the client warning for the reason and lower INTEROP_RELAY_BATCH_SIZE if the window is too dense`,
+        `Window ${from}-${syncedTo} exceeds INTEROP_RELAY_MAX_REQUESTS_PER_UPDATE=${this.relayConfig.maxRequestsPerUpdate}. Fetched ${res.requests.length} requests but a continuation remains`,
```
adamiak, line 136: _Written by Claude on behalf of Adrian Adamiak._

Efficiency: line 140 below filters `status === 'success'` client-side after `getAllRequests` has already paged through everything. The v3 endpoint documents a `status` query filter (`success`, `failure`, `refund`, `pending`, `depositing`). Since the window filters on `updatedAt`, a request currently appears in every window in which it changed state (pending, depositing, success, ...), and each of those pages costs rate-limit budget and counts toward `INTEROP_RELAY_MAX_REQUESTS_PER_UPDATE`.

Passing `status: 'success'` from the indexer should cut pages substantially and make this cap count only relevant items. Semantics should hold: the success transition bumps `updatedAt`, and `sentIds`/`receivedIds` dedupe anyway. Worth verifying against the live API, since the docs don't spell out the interaction with `sortBy=updatedAt`.

adamiak (commented):
_Review comments written by Claude on behalf of Adrian Adamiak._

Three inline comments below (feature gating, retry/rate-limit implementation, server-side status filter). A fourth note is posted as a reply in the existing Codex thread on `RelayApiClient.ts`.

packages/backend/src/modules/interop/plugins/relay/RelayApiClient.ts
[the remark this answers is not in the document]
adamiak: _Written by Claude on behalf of Adrian Adamiak._

Adding a data point to this thread: the v2 code fell back at the amount level (`routeInput?.amount !== undefined ? routeInput : metadata?.currencyIn`), while `normalizeRequest` now falls back only when the whole `inputCurrency`/`outputCurrency` object is null or missing. So a `success` request whose actual currency object lacks `amount` now yields an event with `amount: undefined`, where v2 would have used the quote. That is a silent behavior change relative to v2.

Not fabricating actual values is a reasonable position, but the mitigation proposed above (observability for incomplete actual data) isn't in the PR yet. Either add a warn log when a success request has an actual currency without `amount`, or restore the amount-level fallback. Not a blocker.

packages/backend/src/config/features/interop.ts
sdlyy: The inheritance mechanics are right, but this isn't a new class of config risk — `interop.capture.<chain>` already works exactly this way. Every chain in `INTEROP_CHAINS` resolves true under a bare `interop`, and we manage that with explicit `!` entries today. `interop.relay` is no less explicit than what already ships.

On "widens": it doesn't, with one exception. `interop.relay` and `interop.matching` resolve through the same inheritance, so the set of instances running Relay is identical for every FEATURES string **except** `interop` combined with `!interop.matching`. That single shape is the only new-writer case, and it's a grep.

One correction on the failure mode, because it changes what to look for. A second writer would not collide loudly — `eventId` is `generateId('evt')` (random), `InteropEventRepository.insertMany` is a plain insert with no conflict clause, and the `sentIds`/`receivedIds` dedup is per-process in memory. Two writers would silently duplicate `InteropEvent` rows and double-count transfers. Conversely the startup throw is the *safe* outcome, and it is the guard: an instance can only become a second writer if `INTEROP_RELAY_API_KEY` is also deployed there.

The risk I'd actually weight higher is the inverse one. If a FEATURES string enumerates explicit leaves rather than a bare `interop`, `interop.relay` resolves false and Relay ingestion silently stops after deploy — no crash, no alert, just a data gap. The same grep answers both questions.

So: keeping the namespace consistent with `interop.capture.*` rather than adding a separate non-inherited switch. Before merge I'll add the key to both Coolify envs and check both FEATURES strings for the two shapes above. The local `FEATURES=*` case is a fair hit — documented in the README in dd80cb636b (`FEATURES=*,!interop.relay`).

packages/backend/src/modules/interop/plugins/relay/RelayApiClient.ts
sdlyy: Agreed on the regex, fixed in dd80cb636b. Not migrating to `ClientCore` though, for two concrete reasons:

- `RetryHandler.retry` retries on *any* error, including permanent 4xx — a bad API key would burn three retries per page before failing.
- It logs `init` verbatim in the `Scheduling retry` warning (`packages/shared/src/tools/RetryHandler.ts`), and our `init` carries `headers: { 'x-api-key': ... }`. That would put the key in the logs.

Instead the client now uses `httpClient.fetchRaw` and throws a `RelayHttpError` carrying `status`, so `isRetryableRelayError` switches on the status directly instead of parsing another package's message text. Non-HTTP failures (network, timeout, malformed body) stay retryable. The response body is included truncated for diagnosis — the key travels as a header, so nothing sensitive lands in the URL or the message.

Tests now drive real status codes: 429 retried at the same cursor, 401 not retried, rather than asserting on `HTTP error: NNN` strings.

packages/backend/src/modules/interop/plugins/relay/relay.indexer.ts
sdlyy: Measured this against the live API, and the premise doesn't hold. A request appears in the feed **once**, at its current `updatedAt` — it does not reappear in every window in which it changed state. So there are no extra pages to save from state churn.

The numbers (measured on v2, since I don't have a v3 key locally — @codex independently checked v3 and got the same shape):

- 500 consecutive requests span 16.0s ≈ **31 req/s**. Non-success is ~1% of the feed (480 success / 16 pending / 4 refund in a 500-record sample).
- A 60s window is therefore ~1,900 records ≈ **38 pages, with or without the filter**. Codex's live v3 check: 1,398 → 1,384 records, 28 pages either way.
- The cap: 10,000 records ≈ 320s of traffic, while `INTEROP_RELAY_BATCH_SIZE=60` caps each window at 60s. ~5x headroom, so `INTEROP_RELAY_MAX_REQUESTS_PER_UPDATE` isn't reachable at current volume.
- Rate-limit budget: ~38 calls/min steady state against `INTEROP_RELAY_CALLS_PER_MINUTE=200`, leaving ~5x catch-up headroom (a 1h backlog clears in ~12 min).

No page savings and no cap relief, against making ingestion depend on the server-side filter's semantics matching ours. Not taking it — the client-side filter stays.

packages/backend/src/modules/interop/plugins/relay/RelayApiClient.ts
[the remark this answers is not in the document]
sdlyy: @adamiak the data point is correct and I confirmed it: v2 fell back at the amount level, `normalizeRequest` falls back only when the whole currency object is null or missing. The two differ in exactly one case — `actual` present but `amount` absent.

Restoring the amount-level fallback would still be wrong, because quoted is not a proxy for settled. On 480 successful live requests, quoted differs from actual on the destination side in **425 of them (89%)**:

| symbol | actual | quoted | error |
|---|---|---|---|
| PAPER | 505,324.79 | 729,108.04 | **+44.3%** |
| CUCK | 41,535.25 | 59,067.61 | **+42.2%** |
| USDC | 17.218781 | 14.541210 | −15.6% |
| — | | | median 0.27% |

Source side is tamer but non-zero: 8/480, worst `ETH actual=0.002722315960942499 vs quoted=0.003032669194909321` (+11.4%).

The trade is: a missing amount yields a Transfer with `srcAmount`/`dstAmount` undefined that `RelayPlugin` still matches by `id`; a quoted amount yields a wrong number that silently corrupts USD volume. We take the gap.

How often does the gap actually occur? **0/500** in my sample — `actual` currency objects and their amounts were present on every request, including `pending` and `refund`. Theoretical today, but the tail isn't bounded, which is why the observability is now implemented rather than deferred (dd80cb636b): `warnOnIncompleteActualData` emits one aggregated warning per page with `missingAmount`/`missingAddress` counts and sample ids, and a test asserts we warn *and* still return `amount: undefined` rather than merging.

(Measured on v2, mapping `metadata.route` → v3 `actual` and `metadata.currencyIn/Out` → v3 `quoted`, since I don't have a v3 key locally. @codex's independent v3 check found 601 amount and 6 token-address differences — same conclusion.)

adamiak (approved):

PR #12732, opened 2026-09-02
Index DeFi TVL from DeFiLlama

sdlyy:
## Summary

- Adds an explicit DeFi TVL source contract to project configuration.
- Adds hourly DeFiLlama ingestion for protocols whose TVL would require high-maintenance custom escrow discovery.
- Configures Ethereum TVL for Uniswap V3 and V4 while marking Liquity V2 and Polymarket as native L2BEAT sources.

## Why

Native TVS remains the right path when L2BEAT tracks meaningful escrows. Pool-heavy protocols need different discovery logic and frequent maintenance, so external TVL is kept in a separate provider-neutral pipeline instead of being squeezed into TVS.

## Configuration contract

Projects explicitly choose either `source: l2beat` or `source: defillama`. External projects also define:

- the exact DeFiLlama protocol slug
- the first historical timestamp
- an explicit mapping between L2BEAT and provider chain names

The source, project, slug, and canonical chain mapping form the stable configuration identity. `sinceTimestamp` is a range boundary rather than identity, so moving it trims or backfills the existing configuration instead of creating a new one.

## Indexing behaviour

- A shared hourly parent triggers one independently managed indexer per project, making one protocol request per configured project per normal hourly run.
- Each project fetches the public DeFiLlama `/protocol/{slug}` payload.
- Backfills import midnight UTC daily points in the requested range. Incremental runs reconcile the previous 30 days and add the current hourly observation.
- The provider's daily point is canonical when it collides with the current observation at midnight.
- Snapshot rows are replaced atomically, so a chain omitted from a corrected response cannot survive as stale data.
- Writes happen only after the complete payload is mapped successfully.
- Missing chains/history, invalid or negative values, source data older than 36 hours, and timestamps more than 3 hours in the future fail closed. The project keeps its previous durable checkpoint and retries.

## Lifecycle behaviour

- Changing a slug or chain mapping changes the configuration ID; initialization of that still-active project indexer wipes rows for the obsolete ID.
- Changing `sinceTimestamp` keeps the same ID and uses UIF range trimming/backfilling.
- If the whole project tag disappears because its source/project/feature flag is removed, persisted state is intentionally retained. This matches other tagged indexers and makes temporary feature disables non-destructive; the frontend no longer consumes the data. Permanent purging remains an explicit operation.
- The module is controlled by the `defi-tvl` feature flag. Normal wildcard feature configuration enables it unless explicitly excluded.

## DeFiLlama constraints and deployment decision

- The implementation uses the unauthenticated public endpoint. At the current two-project scope, the normal schedule is approximately 48 requests per day; backfills do not require one request per historical point.
- DeFiLlama's public documentation does not state a numeric free-endpoint quota. The request volume is technically small, but this is not a contractual rate-limit guarantee.
- The endpoint supplies latest USD values, per-chain data, and historical TVL. The configured chain allowlist limits ingestion to researched chains.
- Token breakdown is deferred. Live Uniswap V3/V4 payloads checked during implementation exposed empty `tokensInUsd` arrays, so storing them now would add schema surface without usable data.
- Production enablement requires product/legal approval of DeFiLlama's terms for automated commercial use and republication. The code does not require a paid API key, but a paid plan or separate licensing agreement may still be the required commercial outcome. See [API docs](https://defillama.com/docs/api), [subscription details](https://defillama.com/subscription), and [terms](https://defillama.com/terms).

## Stack

1. Database foundation: https://github.com/l2beat/l2beat/pull/12731
2. **Configuration and ingestion — this PR**
3. Frontend and API integration: https://github.com/l2beat/l2beat/pull/12733

## Validation

- DeFi TVL backend tests: 16 passing
- DeFi TVL configuration invariants: 4 passing
- Config typecheck
- Backend/config Biome
- Full backend typecheck reaches only unrelated locally missing Railgun dependencies (`@waku/sdk` and `@scure/base`); no DeFi TVL diagnostics
- `git diff --check`


sdlyy:
@codex review

packages/backend/src/modules/defi-tvl/DefiTvlModule.ts
```diff
@@ -0,0 +1,61 @@
...
+    logger.info('DefiTvlModule disabled')
+    return
+  }
+
+  logger = logger.tag({ feature: 'defi-tvl', module: 'defi-tvl' })
+  const hourlyIndexer = new HourlyIndexer(logger, clock)
+  const indexerService = new IndexerService(db)
+  const client = new DefiLlamaClient(new HttpClient(), config.defiTvl.apiUrl)
+  const indexers = config.defiTvl.projects.map(
+    (project) =>
```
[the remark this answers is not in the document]
sdlyy: This retention is intentional. A missing tagged indexer means that indexing is disabled or no longer consumed; it does not mean that its persisted state and history should be destroyed. TVS, DA, privacy, and DA Beat use the same tagged-indexer lifecycle and retain state when a whole tag is absent, which also allows feature-flag rollbacks to resume from the saved checkpoint. For DeFi TVL, changing the slug, chains, or sinceTimestamp keeps the same project tag but changes its configuration, so ManagedMultiIndexer initializes that active indexer and wipeData removes the obsolete rows. Switching away from DeFiLlama or removing the project makes the frontend stop reading this data. A startup sweep would instead make a temporary per-project feature-flag disable destructive, while a global feature disable remains non-destructive. Permanent purging should be an explicit operation or migration, not inferred from config absence, so no cleanup is added here.

packages/backend/src/modules/defi-tvl/DefiLlamaClient.ts
```diff
@@ -0,0 +1,41 @@
...
+export type DefiLlamaProtocol = v.infer<typeof DefiLlamaProtocol>
+
+export class DefiLlamaClient {
+  private readonly apiUrl: string
+
+  constructor(
+    private readonly httpClient: HttpClient,
+    apiUrl: string,
+  ) {
+    this.apiUrl = apiUrl.replace(/\/$/, '')
```
torztomasz, line 27: kindaa weird

packages/backend/src/modules/defi-tvl/DefiLlamaClient.ts
```diff
@@ -0,0 +1,41 @@
...
+})
+
+export type DefiLlamaProtocol = v.infer<typeof DefiLlamaProtocol>
+
+export class DefiLlamaClient {
+  private readonly apiUrl: string
+
+  constructor(
+    private readonly httpClient: HttpClient,
+    apiUrl: string,
```
torztomasz, line 25: shouldnt apiUrl be a part of DefiLlamaClient instead of being passed down through constructor?

packages/backend/src/modules/defi-tvl/getDefiTvlConfigurationId.ts
```diff
@@ -0,0 +1,28 @@
...
+    source: 'defillama',
+    metric: 'tvl',
+    projectId: config.projectId,
+    protocolSlug: config.protocolSlug,
+    sinceTimestamp: config.sinceTimestamp,
+    chains: [...config.chains].sort((a, b) => a.chain.localeCompare(b.chain)),
+  }
+
+  return createHash('sha1')
+    .update(JSON.stringify(identity))
```
torztomasz, line 29: **Standards (hard)** – `JSON.stringify(identity)` where `chains` holds `{chain, providerChain}` objects taken verbatim from config. Only the array is sorted; object key order is not normalised. Repo convention: don't hash `JSON.stringify` of params, key order isn't guaranteed; if params must participate, sort keys first. A project written as `{ providerChain, chain }` yields a different id and wipes data. Siblings hash a joined string list (`config/features/privacy.ts`, `ecosystemToken.ts`).

packages/backend/src/modules/defi-tvl/getDefiTvlConfigurationId.ts
```diff
@@ -0,0 +1,28 @@
...
+
+export function getDefiTvlConfigurationId(
+  config: DefiTvlConfigurationIdentity,
+): string {
+  const identity = {
+    source: 'defillama',
+    metric: 'tvl',
+    projectId: config.projectId,
+    protocolSlug: config.protocolSlug,
+    sinceTimestamp: config.sinceTimestamp,
```
torztomasz: **Standards (hard)** – `sinceTimestamp` is part of the identity. Repo convention ("identity vs range"): `since*`/`until*` are not hashed so ranges update in place without a wipe. Here moving `sinceTimestamp` forces a full wipe + re-import. Relatedly there's no `trimData` override on `DefiTvlIndexer`; it's only masked because the range is baked into the id.

packages/backend/src/modules/defi-tvl/getDefiTvlConfigurationId.ts
```diff
@@ -0,0 +1,28 @@
+import type { ProjectDefiTvlChain } from '@l2beat/config'
+import type { ProjectId, UnixTime } from '@l2beat/shared-pure'
+import { createHash } from 'crypto'
+
+interface DefiTvlConfigurationIdentity {
```
torztomasz, line 5: **Smell (judgement, Data Clump)** – `DefiTvlConfigurationIdentity` is `Omit<DefiTvlProjectConfig, 'configurationId'>` restated. `PrivacyPriceIndexer` uses the `Omit` directly.

packages/backend/src/modules/defi-tvl/DefiTvlIndexer.ts
```diff
@@ -0,0 +1,70 @@
...
+      logger,
+    )
+  }
+
+  override async multiUpdate(
+    from: number,
+    to: number,
+    configurations: Configuration<DefiTvlProjectConfig>[],
+  ): Promise<() => Promise<number>> {
+    assert(configurations.length === 1, 'Expected one DeFi TVL configuration')
```
torztomasz, line 43: **Standards (judgement, Refused Bequest)** – extends `ManagedMultiIndexer` then asserts exactly one configuration. `EcosystemsModule` uses one multi-indexer holding all configurations. Either go single-indexer with N configs, or don't inherit the multi-config base.

Also: no `static idToConfigurationId` on the class. The documented contract and every privacy indexer put id derivation on the indexer; here it lives in a standalone file.

packages/backend/src/modules/defi-tvl/DefiTvlIndexer.ts
```diff
@@ -0,0 +1,70 @@
...
+export class DefiTvlIndexer extends ManagedMultiIndexer<DefiTvlProjectConfig> {
+  constructor(
+    private readonly $: DefiTvlIndexerDeps,
+    logger: Logger,
+  ) {
+    super(
+      {
+        ...$,
+        name: INDEXER_NAMES.DEFI_TVL,
+        updateRetryStrategy: Indexer.getInfiniteRetryStrategy(),
```
torztomasz, line 32: **Spec (questionable)** – with `getInfiniteRetryStrategy()` any assertion in `mapDefiLlamaTvl` (stale, negative, missing chain) blocks this project's indexer forever rather than failing closed once. Combined with the 6h staleness check, a protocol/chain that DeFiLlama only updates daily can never sync.

packages/backend/src/modules/defi-tvl/DefiTvlIndexer.ts
```diff
@@ -0,0 +1,70 @@
...
+        project: config.projectId,
+        records: records.length,
+        from,
+        to,
+      })
+      return to
+    }
+  }
+
+  override async wipeData(
```
torztomasz, line 60: **Spec (wrong)** – PR says *"Removing or changing a configuration deletes data belonging to the obsolete configuration ID"*. Removing does not. Indexers are only built for projects still in config (`DefiTvlModule.ts:23-43`), so a removed or flag-disabled project has no indexer, `mergeConfigurations` never runs for it and `wipeData` is never called. Its rows plus `indexer_configurations` entries stay orphaned. Only *changing* the id within a still-present project wipes.

packages/backend/src/modules/defi-tvl/DefiTvlModule.ts
```diff
@@ -0,0 +1,61 @@
...
+}: ModuleDependencies): ApplicationModule | undefined {
+  if (!config.defiTvl) {
+    logger.info('DefiTvlModule disabled')
+    return
+  }
+
+  logger = logger.tag({ feature: 'defi-tvl', module: 'defi-tvl' })
+  const hourlyIndexer = new HourlyIndexer(logger, clock)
+  const indexerService = new IndexerService(db)
+  const client = new DefiLlamaClient(new HttpClient(), config.defiTvl.apiUrl)
```
torztomasz, line 22: **Standards (judgement)** – client constructed here with a bare `HttpClient`. Siblings get third-party clients from `providers.clients` (`packages/shared/src/clients/*`, extending `ClientCore` for retries/metrics).

packages/backend/src/modules/defi-tvl/DefiLlamaClient.ts
torztomasz, line 27: **Smell (judgement, Speculative Generality)** – trailing-slash stripping (and its test) has no precedent in sibling clients and the env default has no trailing slash.

packages/backend/src/modules/defi-tvl/DefiLlamaClient.ts
```diff
@@ -0,0 +1,41 @@
...
+  date: v.number(),
+  totalLiquidityUSD: v.number(),
+})
+
+const DefiLlamaChainTvl = v.object({
+  tvl: v.array(DefiLlamaTvlPoint).optional(),
+})
+
+export const DefiLlamaProtocol = v.object({
+  currentChainTvls: v.record(v.string(), v.union([v.number(), v.null()])),
```
torztomasz, line 14: **Spec (missing)** – Linear asks for *"tokens breakdown"* as a nice-to-have. `tokensInUsd` is dropped by the schema and nothing stores it. If intentionally deferred, say so in the PR.

packages/backend/src/modules/defi-tvl/mapDefiLlamaTvl.ts
```diff
@@ -0,0 +1,107 @@
...
+    )
+    assert(
+      history !== undefined && history.length > 0,
+      `${config.protocolSlug}: Missing TVL history for ${chainConfig.providerChain}`,
+    )
+
+    const latest = history.reduce((a, b) => (a.date > b.date ? a : b))
+    const sourceTimestamp = UnixTime(latest.date)
+    assert(
+      sourceTimestamp >= to - MAX_STALENESS,
```
torztomasz: **Spec (questionable)** – 6h staleness against DeFiLlama's per-chain last point. Ethereum currently updates hourly so this passes, but any protocol/chain updated only daily fails permanently, and with infinite retry the indexer is stuck rather than failing closed once.

packages/backend/src/modules/defi-tvl/mapDefiLlamaTvl.ts
```diff
@@ -0,0 +1,107 @@
...
+      sourceTimestamp <= to + MAX_FUTURE_SKEW,
+      `${config.protocolSlug}: Future TVL timestamp for ${chainConfig.providerChain}`,
+    )
+
+    for (const point of history) {
+      const timestamp = UnixTime(point.date)
+      if (
+        timestamp < historyFrom ||
+        timestamp > to ||
+        !UnixTime.isFull(timestamp, 'day')
```
torztomasz, line 58: **Spec (unspecified)** – non-midnight history points are silently dropped. Not in either spec; worth a comment or a note in the PR.

packages/backend/src/modules/defi-tvl/mapDefiLlamaTvl.ts
```diff
@@ -0,0 +1,107 @@
...
+    }
+
+    const currentRecord = createRecord(
+      config,
+      chainConfig.chain,
+      to,
+      sourceTimestamp,
+      currentValue,
+    )
+    records.set(recordKey(currentRecord), currentRecord)
```
torztomasz: **Spec (questionable)** – PK is `(timestamp, configurationId, chain)`. When `to` is midnight this snapshot shares a key with the daily point. In-batch the snapshot wins (last `set`), but the next hour's reconciliation overwrites it with `totalLiquidityUSD`, so precedence flips and `sourceTimestamp` changes. Pick one source for midnight rows.

**Smell (judgement, Mysterious Name)** – `mapDefiLlamaTvl` also validates staleness/skew and merges history with current; "map" undersells it.

packages/backend/src/config/features/defiTvl.ts
```diff
@@ -0,0 +1,46 @@
...
+  flags: FeatureFlags,
+  apiUrl: string,
+): Promise<DefiTvlConfig | false> {
+  const projects = (await ps.getProjects({ select: ['defiInfo'] }))
+    .filter((project) => flags.isEnabled('defi-tvl', project.id))
+    .flatMap((project) => {
+      const tvl = project.defiInfo.tvl
+      if (tvl?.source !== 'defillama') return []
+
+      assert(
```
torztomasz: **Smell (judgement, Duplicated Code)** – these four assertions repeat `packages/config/src/processing/getProjects.test.ts:405-416` (slug non-empty, chains non-empty, unique `chain`, unique `providerChain`). Keep one.

packages/backend/src/modules/defi-tvl/DefiTvlIndexer.test.ts
```diff
@@ -0,0 +1,73 @@
...
+        configurations: [
+          {
+            id: config.configurationId,
+            minHeight: config.sinceTimestamp,
+            maxHeight: null,
+            properties: config,
+          },
+        ],
+        tags: { tag: 'defi-tvl-indexer-test' },
+        db: mockObject<Database>({ defiTvl }),
```
torztomasz: **Standards (judgement)** – siblings use `mockDatabase({ defiTvl })` from `packages/backend/src/test/database.ts` rather than `mockObject<Database>`.

packages/config/src/processing/getProjects.test.ts
```diff
@@ -392,6 +392,32 @@ describe('getProjects', () => {
...
+      const tvl = project.defiInfo?.tvl
+      if (!tvl) continue
+
+      if (tvl.source === 'l2beat') {
+        it(`${project.id} has TVS config for its L2BEAT TVL source`, () => {
+          expect(project.tvsConfig).not.toEqual(undefined)
+        })
+      } else {
+        it(`${project.id} has a complete external TVL scope`, () => {
+          expect(project.tvsConfig).toEqual(undefined)
```
torztomasz, line 406: **Spec (scope creep)** – enforcing `tvsConfig === undefined` for defillama projects (and required for `l2beat` ones) is a coupling rule in neither the Linear issue nor the PR description. If it's a deliberate invariant, document it in the `ProjectDefiTvlConfig` JSDoc.

torztomasz:
Two-axis review (Standards vs Spec). Line-anchored findings are inline; the rest are here.

## Standards (not anchorable)

- **Third-party client location** – `DefiLlamaClient` is module-local and built with a bare `HttpClient`. Third-party clients in this repo live in `packages/shared/src/clients/*` and extend `ClientCore` (retries, metrics), then get wired via `providers.clients` (see `EcosystemsModule.ts`). This one loses that shared behaviour.

## Spec (Linear L2B-14554 + PR description)

**Missing / partial**
- Linear: *"plans and rate limits and if it fits our indexing schedule, can we use it for free"* – not addressed. The client hits public `api.llama.fi` with no key, no 429 handling and no documented limits; only the generic infinite backoff. The PR description doesn't defer this either.
- Linear: *"nice to have – historical chart, tokens breakdown"* – tokens breakdown is neither implemented nor deferred. The schema parses only `currentChainTvls`/`chainTvls`; `tokensInUsd` is dropped and the DB has no token column.
- Linear: *"After this, have a conclusion if we can go with it"* – no written conclusion in the PR.
- PR: *"Incremental runs reconcile the previous 30 days"* – only daily midnight points are reconciled; previously written hourly snapshots are never revisited.
- Tests vs the *"fail closed"* list: no test for negative values, future skew, missing history, or `wipeData`.

**Questionable**
- PR: *"The whole module is opt-in through the `defi-tvl` feature flag"* – gating matches the `tvs` pattern, but the non-local default is `FEATURES=*` (`makeConfig.ts:33-34`), so it is on unless explicitly excluded. "Opt-in" overstates it.
- Initial vs incremental detection works (uif passes `from = height+1`, unsynced config starts at `minHeight`), but it's a heuristic: any outage >30d re-imports everything. Harmless, worth a comment.
- Test counts (9 backend, 4 config) match.

**Clean**: `INDEXER_NAMES` entry, `$` deps naming, thunk pattern, `wipeData` logging, `logger.tag`, `Config.ts`/`makeConfig.ts` wiring, `describe(X.name)`, config-package `defiInfo.tvl` literals and JSDoc all match siblings.

_Inline findings (15) are in a separate review._

torztomasz (changes requested):

packages/backend/src/modules/defi-tvl/DefiTvlModule.ts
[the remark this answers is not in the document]
sdlyy: Correction after the later identity review: `sinceTimestamp` no longer changes the configuration ID. It is now a range boundary, so UIF trims or backfills the same active configuration. Slug and chain-mapping changes still create a new ID and wipe the obsolete active configuration. The absent-tag retention behavior described here remains intentional.

packages/backend/src/config/features/defiTvl.ts
```diff
@@ -0,0 +1,46 @@
...
+  ps: ProjectService,
+  flags: FeatureFlags,
+  apiUrl: string,
+): Promise<DefiTvlConfig | false> {
+  const projects = (await ps.getProjects({ select: ['defiInfo'] }))
+    .filter((project) => flags.isEnabled('defi-tvl', project.id))
+    .flatMap((project) => {
+      const tvl = project.defiInfo.tvl
+      if (tvl?.source !== 'defillama') return []
+
```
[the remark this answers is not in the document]
sdlyy: The PR description was inaccurate and is corrected. An absent project tag means inactive or unconsumed, not permanently deleted. Other tagged indexers retain state in this case so a temporary feature disable can resume from its checkpoint. For a still-active project, slug or chain changes create a new ID and invoke `wipeData`; changing `sinceTimestamp` now trims or backfills the same ID. Switching away from DeFiLlama also stops the frontend from reading these rows. Permanent purging should remain an explicit operation rather than a startup inference.

packages/backend/src/modules/defi-tvl/DefiLlamaClient.ts
sdlyy: This normalizes the injected base URL before appending `/protocol/{slug}`, so an environment override with one trailing slash does not produce a double slash. It stays at the client boundary and is covered by the client test.

packages/backend/src/modules/defi-tvl/DefiLlamaClient.ts
sdlyy: It is part of `DefiLlamaClient`: the constructor stores it as the private `apiUrl` property. Constructor injection is intentional so runtime configuration and tests can provide the endpoint without hardcoding deployment configuration in the client.

packages/backend/src/modules/defi-tvl/getDefiTvlConfigurationId.ts
sdlyy: Fixed in `b0eb96c8db`. The identity now maps each chain to a primitive `[chain, providerChain]` tuple and sorts the tuples before serialization, so object property insertion order cannot affect the hash. The test covers reversed array order and reversed property order.

packages/backend/src/modules/defi-tvl/getDefiTvlConfigurationId.ts
sdlyy: Fixed in `b0eb96c8db`. `sinceTimestamp` is excluded from identity, and `DefiTvlIndexer.trimData` deletes the inclusive range supplied by UIF. Tests verify that changing the import boundary preserves the ID and that trim delegates the exact range to the repository.

packages/backend/src/modules/defi-tvl/getDefiTvlConfigurationId.ts
sdlyy: After removing `sinceTimestamp` from identity, this is intentionally no longer equivalent to `Omit<DefiTvlProjectConfig, 'configurationId'>`. The explicit type now documents the smaller set of fields that are allowed to affect identity and prevents a range field from being added back accidentally.

packages/backend/src/modules/defi-tvl/DefiTvlIndexer.ts
sdlyy: The per-project split is intentional: one provider failure must not block every DeFi project, and each tagged indexer gets an independent checkpoint and retry lifecycle. `ManagedMultiIndexer` is used for configuration reconciliation, wipe, trim, and checkpoint management; it does not require batching unrelated configurations in one request. The ID helper is needed by config assembly before an indexer exists and no base-class contract requires it to be static, so it remains a standalone pure function.

packages/backend/src/modules/defi-tvl/DefiTvlIndexer.ts
sdlyy: The 6-hour assumption was too strict and is fixed: source age now allows one daily interval plus 12 hours of publishing jitter. Infinite retry remains intentional. Here fail-closed means the save callback is not executed and the checkpoint does not advance, so the last complete snapshot stays durable until the provider recovers or configuration is corrected. Because indexers are split per project, one bad protocol does not block the others.

packages/backend/src/modules/defi-tvl/mapDefiLlamaTvl.ts
sdlyy: Fixed in `b0eb96c8db`. The source-age limit is now 36 hours: one expected daily interval plus 12 hours of publication jitter. A focused test accepts 24-hour-old data and another rejects data older than 36 hours.

packages/backend/src/modules/defi-tvl/DefiTvlIndexer.ts
sdlyy: The 6-hour assumption was too strict and is fixed: source age now allows one daily interval plus 12 hours of publishing jitter. Infinite retry remains intentional. Here fail-closed means the save callback is not executed and the checkpoint does not advance, so the last complete snapshot stays durable until the provider recovers or configuration is corrected. Because indexers are split per project, one bad protocol does not block the others.

packages/backend/src/modules/defi-tvl/mapDefiLlamaTvl.ts
sdlyy: Documented in the PR. Historical import deliberately stores midnight UTC daily points because the free protocol history is primarily daily and can contain irregular intraday observations. The current value is stored separately at each hourly indexing timestamp.

packages/backend/src/modules/defi-tvl/mapDefiLlamaTvl.ts
sdlyy: Fixed the precedence issue in `b0eb96c8db`: when both sources have the same midnight key, the canonical daily historical point wins on every run. The regression test covers this collision. I kept the function name because this is still one provider-boundary transformation: it validates and maps one payload into durable records; splitting or renaming it would not clarify a separate reusable responsibility.

packages/backend/src/modules/defi-tvl/DefiTvlIndexer.ts
sdlyy: Agreed that the original PR wording was wrong; it now describes the distinction explicitly. Changed IDs within a still-active tag are wiped, `sinceTimestamp` changes trim or backfill the same ID, and a wholly absent tag retains its persisted checkpoint and history intentionally. The latter matches the existing tagged-indexer lifecycle and keeps feature-flag rollback non-destructive.

packages/backend/src/config/features/defiTvl.ts
sdlyy: Fixed in `b0eb96c8db`. The duplicate runtime assertions were removed; the config-wide `getProjects` invariant test remains the single validation point for these static project definitions.

packages/backend/src/modules/defi-tvl/DefiTvlModule.ts
sdlyy: Keeping this module-local for now. It has one endpoint and one backend consumer, and the backend-local `RelayApiClient` is an existing precedent. `HttpClient` supplies HTTP status handling and timeouts, while retry ownership is at the indexer layer. Moving it to shared and adding `ClientCore` becomes justified if it gains shared consumers or needs per-request provider metrics.

packages/backend/src/modules/defi-tvl/DefiTvlIndexer.test.ts
sdlyy: Fixed in `b0eb96c8db`. The test now uses `mockDatabase({ defiTvl })`, with the repository mock typed as `Database['defiTvl']`.

packages/backend/src/modules/defi-tvl/DefiLlamaClient.ts
sdlyy: The default has no trailing slash, but `apiUrl` is an environment override. Normalizing that boundary prevents malformed joining without changing the client contract, and the existing client test covers the override case. Keeping this small guard.

packages/config/src/processing/getProjects.test.ts
sdlyy: Documented on `ProjectDefiTvlConfig`. The native source requires `tvsConfig`; the external DeFiLlama source must not be combined with it. The invariant prevents two competing TVL authorities for one project.

packages/backend/src/modules/defi-tvl/DefiLlamaClient.ts
sdlyy: Documented as intentionally deferred in the PR. Live Uniswap V3 and V4 protocol payloads checked during implementation exposed empty `tokensInUsd` arrays, so persisting the field now would add schema and API surface without usable data. The current scope provides latest USD TVL, researched-chain breakdown, and daily history.

torztomasz (approved):

PR #12739, opened 2026-09-02
feat(frontend): set Postgres statement_timeout on frontend connections

torztomasz:
## Summary
- Abandoned or timed-out HTTP requests (e.g. spammed tRPC chart routes) kept their queries running on Postgres, holding pool connections and DB CPU long after the 25s `connect-timeout` gave up on the response.
- Adds `DATABASE_STATEMENT_TIMEOUT_MS` (default `20000`, validated as a positive integer) and applies it as a server-side `statement_timeout` to both the main and token database pools. Postgres cancels the statement itself, unlike `query_timeout`, which only rejects the client promise.
- `pg` sends `statement_timeout` in the connection startup packet, so every pooled connection gets it with no changes to `@l2beat/database`.

## Before merging
Check p99 of the heaviest chart queries in production (e.g. `layer2` filter on the `max` range). Anything legitimately running over 20s will now fail with `canceling statement due to statement timeout`. The env var can be raised on Coolify without a redeploy.

## Test plan
- [x] `pnpm typecheck` in `packages/frontend`
- [x] `pnpm test` in `packages/frontend` (464 passing)
- [x] `pnpm lint:fix` and `pnpm format:fix`
- [ ] Deploy to staging and confirm `SHOW statement_timeout` on an `FE-staging` connection returns `20s`

### DIFF
--- a/packages/frontend/README.md
+++ b/packages/frontend/README.md
@@ -40,6 +40,9 @@ If you are running `pnpm dev` or `pnpm build` you need to set the following envi
 - create file `.env` in frontend root folder
 - `DATABASE_URL` - database connection url (read-only access is sufficient)
 
+Optional database tuning:
+- `DATABASE_STATEMENT_TIMEOUT_MS` - Postgres `statement_timeout` applied to every frontend connection, in milliseconds (default `20000`). Keep it below the HTTP request timeout so the database gives up before the response does.
+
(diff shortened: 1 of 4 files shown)


mateuszradomski:
@codex, review

mateuszradomski (approved):

PR #12759, opened 2026-09-04
add token graph on tokens detailed pages

vilcht:
Adds the relations graph to the token detail pages (L2B-14707): deployments grouped into burn-and-mint clusters, lock-and-mint backing edges, per-deployment and per-cluster 24h stats computed from raw transfers, and a full-screen view.

**Conventions**

- Stats value each transfer at the larger of its two sides (`getInteropTransferValue`, `GREATEST` in SQL), the same definition the protocols table and flows use, and count it once per deployment and per cluster even when both sides belong to it. The aggregated sections count per matching project, so totals can differ for tokens bridged by overlapping projects.
- A deployment pair whose relations claim both lock directions is shown without an edge. That is contradictory evidence about direction, not a route.


packages/database/src/repositories/InteropTransferRepository.ts
```diff
@@ -968,6 +978,61 @@ export class InteropTransferRepository extends BaseRepository {
...
+    }>`
+      SELECT
+        CASE WHEN "srcAbstractTokenId" = ${abstractTokenId} THEN "srcChain" END AS "srcChain",
+        CASE WHEN "srcAbstractTokenId" = ${abstractTokenId} THEN "srcTokenAddress" END AS "srcTokenAddress",
+        CASE WHEN "dstAbstractTokenId" = ${abstractTokenId} THEN "dstChain" END AS "dstChain",
+        CASE WHEN "dstAbstractTokenId" = ${abstractTokenId} THEN "dstTokenAddress" END AS "dstTokenAddress",
+        COUNT(*) AS "transferCount",
+        COUNT("duration") AS "transfersWithDurationCount",
+        COALESCE(SUM("duration"), 0) AS "totalDurationSum",
+        COALESCE(SUM(GREATEST("srcValueUsd", "dstValueUsd")), 0) AS "volume"
```
torztomasz: Valuation change vs. main: the deleted aggregate credited src with `srcValueUsd` and dst with `dstValueUsd` (`aggregation.ts:~430`), while this uses `GREATEST(src, dst)` for both sides and applies no plugin/config filter (the old path went through `aggregationService.aggregate(transfers, configs)`). Per-deployment volumes in the table will shift vs. main and vs. the other "Last 24h" numbers on the same page. If that's intended, a note in the PR description would help.

packages/frontend/src/server/features/layer2s/interop/token/getInteropTokenRelations.ts
```diff
@@ -0,0 +1,96 @@
...
+  return { routes, pairStats }
+}
+
+async function getPairStats(tokenId: string) {
+  const snapshotTimestamp = await getAggregatedInteropSnapshotTimestamp()
+  if (!snapshotTimestamp) return undefined
+  const from = snapshotTimestamp - UnixTime.DAY
+  // Aggregates outlive raw transfers, so an aggregates timestamp override can
+  // point at a day the cleaner has already emptied.
+  if (from < UnixTime.now() - 7 * UnixTime.DAY) return undefined
```
torztomasz: This 7-day literal duplicates `InteropCleanerLoop.ts:30,33`. The commit is titled "share transfer retention" but no shared constant exists, so if the cleaner window changes this silently returns `undefined` (all stats null) for a valid snapshot. Consider exporting the retention from one place.

packages/frontend/src/server/features/layer2s/interop/token/buildTokenRelationsGraph.ts
```diff
@@ -0,0 +1,151 @@
...
+ * directions, are omitted.
+ */
+export function buildTokenRelationsGraph<T extends Endpoint>(
+  deployments: T[],
+  routes: TokenRelationRoute[],
+): TokenRelationsGraph<T> {
+  const byKey = new Map(deployments.map((d) => [deploymentKey(d), d]))
+  const relevant = routes.filter(
+    (route): route is GraphRoute =>
+      route.plugin !== MANUAL_RELATION_PLUGIN &&
```
torztomasz: Spec: "the way the relations are formed is the same as on the graph above". The existing graph (`token-backend/.../deployedTokens/index.ts` `isGraphRelation`) keeps every burnAndMint/lockAndMint relation regardless of plugin; excluding `MANUAL_RELATION_PLUGIN` here means manually-added relations visible on tokens.l2beat.com will be missing on the token page. Intentional?

packages/frontend/src/server/features/layer2s/interop/token/buildTokenRelationsGraph.ts
```diff
@@ -0,0 +1,151 @@
...
+    const to = groupOf.get(
+      endpointKey(route, route.lockedToken === 'A' ? 'B' : 'A'),
+    ) as string
+    if (from === to) continue
+    const edge = edges.get(edgeId(from, to)) ?? { from, to, sources: [] }
+    addSource(edge.sources, route)
+    edges.set(edgeId(from, to), edge)
+  }
+  for (const edge of [...edges.values()]) {
+    if (edges.has(edgeId(edge.to, edge.from))) {
```
torztomasz, line 77: Dropping both directions when a pair conflicts hides relations that the source graph shows as conflicts. Matches prototype #12564, but diverges from the tokens graph the spec points to.

packages/frontend/src/server/features/layer2s/interop/token/buildTokenRelationsGraph.ts
```diff
@@ -0,0 +1,151 @@
...
+  to: string
+  sources: TokenRelationsGraphSource[]
+}
+
+export interface TokenRelationsGraph<T> {
+  nodes: TokenRelationsGraphNode<T>[]
+  edges: TokenRelationsGraphEdge[]
+}
+
+interface Endpoint {
```
torztomasz: Possible Data Clump: this `Endpoint` shape is declared again in `utils/deploymentKey.ts:3`, and `DeployedTokenPrimaryKey` in `@l2beat/database` already models `{ chain, address }`. The `${chain.id}|${address}` key is also rebuilt in `RelationsDetails.tsx`, `RelationsNode.tsx`, and `RelationsPrimitives.tsx`.

packages/frontend/src/server/features/layer2s/interop/token/buildTokenRelationsGraph.ts
```diff
@@ -0,0 +1,151 @@
...
+      }))
+      .toSorted((a, b) => a.id.localeCompare(b.id)),
+    edges: [...edges.values()].toSorted(
+      (a, b) => a.from.localeCompare(b.from) || a.to.localeCompare(b.to),
+    ),
+  }
+}
+
+function edgeId(from: string, to: string): string {
+  return `${from}->${to}`
```
torztomasz, line 99: Possible Duplicated Code: same shape as `edgeKey` in `relations-graph/graphSelectors.ts:9`. Fine if the server/client split is deliberate, otherwise share one.

packages/frontend/src/server/features/layer2s/interop/token/getInteropTokenRelationsGraph.ts
```diff
@@ -0,0 +1,254 @@
...
+  }
+  return new Map(
+    [...sums].map(([group, sum]) => [
+      group,
+      {
+        volume: sum.volume,
+        transferCount: sum.transferCount,
+        avgDuration:
+          sum.transfersWithDurationCount > 0
+            ? Math.floor(sum.totalDurationSum / sum.transfersWithDurationCount)
```
torztomasz, line 249: Behaviour change: `Math.floor` here, the deleted code used `Math.round`. Harmless, just flagging.

packages/frontend/src/components/projects/sections/interop/onchain-deployments/relations-graph/RelationsEdges.tsx
```diff
@@ -0,0 +1,116 @@
...
+  )
+}
+
+export function RelationsEdgeBadge({
+  edge,
+  at,
+  isDimmed,
+}: {
+  edge: InteropTokenRelationsEdge
+  at: EdgePath
```
torztomasz: Nit, possible Mysterious Name: `at: EdgePath` doesn't say what it holds. `path` or `position` would read better.

torztomasz (commented):
Two-axis review (standards vs. spec L2B-14707) at `a575c25c`. Inline comments below; summary here.

**Standards**
- No hard violations of documented standards. `biome check` clean on all changed files, 24/24 new frontend tests pass. Mock convention satisfied (`getInteropTokenRelations.ts` early-returns under `env.MOCK`).
- Judgement calls (Fowler smells): duplicated `edgeId`/`edgeKey`; `{ chain; address }` `Endpoint` declared twice while `DeployedTokenPrimaryKey` exists in `@l2beat/database`, and `${chain.id}|${address}` keys rebuilt in three client components; 7-day retention literal duplicated from `InteropCleanerLoop`; `buildTokenRelationsGraph<T>` only ever instantiated with one type; `RelationsDetails.tsx` (423 lines, six subcomponents) and `RelationsPrimitives.tsx` (11 exports) are fairly divergent.

**Spec**
- Cluster double counting is correctly avoided (`aggregateStats` dedupes per pair row, covered by tests). Graph renders on every abstract token page with deployments, and in MOCK mode.
- Partial: spec says "the way the relations are formed is the same as on the graph above", but this builder drops `MANUAL_RELATION_PLUGIN` routes and conflicting bidirectional pairs, which the tokens.l2beat.com graph keeps. Worth documenting in the PR description if intentional.
- Questionable: per-deployment volume valuation changed from per-side (`srcValueUsd` for src, `dstValueUsd` for dst) to `GREATEST(src, dst)` for both sides, with no plugin/config filter. Table numbers will disagree with other "Last 24h" widgets on the same page.
- Mild creep: tooltip wording changes in `columns.tsx`, and ~1.7k lines of interaction UI (drag/two-finger pan, depth-10 path explorer). Prototype-consistent, just beyond "visualise this info".
- Behaviour change: avg duration now `Math.floor` where the deleted code used `Math.round`.

packages/frontend/src/server/features/layer2s/interop/token/getInteropTokenRelationsGraph.ts
vilcht: floor is correct, every other similar usecase has floor

packages/frontend/src/server/features/layer2s/interop/token/buildTokenRelationsGraph.ts
vilcht: split is intentional, code is cleaner that way

packages/database/src/repositories/InteropTransferRepository.ts
vilcht: The valuation change is intentional: the graph and deployment table now share pair-based statistics, using `max(srcValueUsd, dstValueUsd)` consistently with `getInteropTransferValue`. This also lets us count transfers within a cluster only once. The missing filtering was an oversight. Fixed by applying the existing classifier’s plugin/config rules and the transfer list’s active-chain and crosschain restrictions. Transfers matching multiple configs still count once. The token summary retains its previous valuation, so that difference remains; this change aligns the graph and deployment table.

packages/frontend/src/server/features/layer2s/interop/token/buildTokenRelationsGraph.ts
vilcht: Not sure about it, for now i'll live it as it mirrors the prototype

vilcht: The double back should never be the case and isn't right now, I think hiding  these on users end is correct

torztomasz:
image
Make it cover full screen really

torztomasz:
## Data flow review (queries checked on production, read-only)

Both new queries perform fine, but the server-side flow around them can be tightened.

**Measured on `l2beat_production`** (24h window before the latest aggregated snapshot, full plugin/chain lists):

| Query | Case | Time |
|---|---|---|
| `getDeployedTokenPairStats` | busiest token (~266k rows in window, 4.5k groups out) | 0.6–0.9 s |
| `getDeployedTokenPairStats` | mid token (~42k rows) | 0.2 s |
| `getRoutesBetween` | 18 endpoints | ~1 ms |

Pair-stats plan is a BitmapOr on the existing `srcAbstractTokenId,timestamp` / `dstAbstractTokenId,timestamp` indexes, no seq scan. ~300 ms of the busiest case is JIT overhead; not worth changing here, but `SET LOCAL jit = off` is a cheap win if it ever matters.

Replacing the `AggregatedInteropDeployedToken` read with raw transfers is justified: the aggregate is per project config and per single deployment, so it double counts transfers matching two projects and transfers inside one burn-and-mint cluster.

### Current flow

```
getCachedData
  Promise.all
    getInteropTokenData
    getDeploymentsAndRelations                          serial
      getInteropTokenOnchainDeployments
        Q1 deployedToken.getByAbstractTokenId
        Q2 tokenRelation.getMintingPluginsForMany
      getInteropTokenRelations
        Q3 tokenRelation.getRoutesBetween               new, ~1 ms
        getPairStats
          getAggregatedInteropSnapshotTimestamp         already resolved inside getInteropTokenData
          ps.getProjects(interopConfig)                 already loaded by getCachedData
          Q4 interopTransfer.getDeployedTokenPairStats  new, 0.4–0.9 s
```

Q4 is the slow step and it starts last, behind two serial tokenDb round trips, even though it only needs the token id.

### Suggested changes

```diff
 getCachedData
   getInteropAbstractTokens
+  getAggregatedInteropSnapshotTimestamp        once
   Promise.all
     getInteropTokenData
-    getDeploymentsAndRelations
-      getInteropTokenOnchainDeployments
-        Q1 deployedToken.getByAbstractTokenId
-        Q2 tokenRelation.getMintingPluginsForMany
-      getInteropTokenRelations
-        Q3 tokenRelation.getRoutesBetween
-        getPairStats
-          getAggregatedInteropSnapshotTimestamp
-          ps.getProjects(interopConfig)
-          Q4 interopTransfer.getDeployedTokenPairStats
+    getInteropTokenDeployments
+      Q1 deployedToken.getByAbstractTokenId
+      Q2' tokenRelation.getRelationsTouching(deployments)
+    Q4 interopTransfer.getDeployedTokenPairStats(tokenId, snapshotTs, projects)
     ps.getProjects(chainConfig)
     ps.getProjects(interopConfig)
   getInteropTokenRelationsGraph
+    splitRelations(relations, deploymentSet)
+      routes  = both endpoints in set
+      minters = burnAndMint either end | lockAndMint opposite lockedToken
     buildTokenRelationsGraph(deployments, routes)
```

1. **Merge Q2 and Q3 into one TokenRelation query.** Both read the same table for the same deployments. Fetch every relation touching any deployment once, then derive routes ("both endpoints in the set") and minters (the `mintedAtEndpoint` rule) in TS. This removes `getRoutesBetween`, `ROUTE_COLUMNS`, `toRoute`, and the minting-plugins map building. Note that Q2 is not a subset of Q3: for `9HN5PN` 3 of 215 touching relations and for `C0Hmkq` 11 of 73 have an endpoint outside the token's deployment set, so keep the "touching" query rather than deriving minters from routes.
2. **Start Q4 immediately.** Move `getPairStats` into the top-level `Promise.all`; it does not depend on deployments.
3. **Pass inputs down instead of refetching.** `getPairStats` re-resolves the snapshot timestamp (2 queries) and re-loads interop projects; both are already available to the caller.
4. **Q4 nit.** `min("srcEventId")` / `min("dstEventId")` only feed `isOneSided`, and the group already keys on `"srcEventId" IS NULL`. Select those booleans instead of the two `min` aggregates.
5. `deploymentsCount` in `getInteropTokenEntry` reduces over graph nodes; it equals `deployments.length`.

### One thing to decide

The graph counts each transfer once regardless of how many projects match it. The "Volume and flows" section on the same page sums `AggregatedInteropTransfer` per project id, so a transfer matching two projects counts twice there. Totals on the two sections can disagree for tokens bridged by overlapping projects. Worth picking one convention for the page.

vilcht:
Thanks for the measurements. Landed:

- **1. One relation query** — 8373c41603. `getRelationsTouching` replaces `getRoutesBetween` and `getMintingPluginsForMany`; routes and minters are derived in TS, and the minted-endpoint rule is exported as `isMintedAtEndpoint` next to the SQL predicate it mirrors. `ROUTE_COLUMNS`/`toRoute` stay, `getAllRoutes` still uses them.
- **2 + 3. Start pair stats first, resolve inputs once** — 8fc84f4f78. The page loader resolves the snapshot timestamp and interop projects once and passes them down; `getInteropTokenData` and `getLatestAggregatedInteropTransferWithTokens` take them as optional inputs so the tRPC path is unchanged. Pair stats run in the top-level `Promise.all`. `getInteropAbstractTokens` keeps its own lookup since it sits behind a 10 min cache.
- **5. Deployments count** — 6e2a6087b8.
- **4. Skipped.** `isOneSided` takes the ids as `string | undefined`, so booleans would need a sentinel threaded through the classifier for no measurable gain.

On the counting convention: keeping the graph's once-per-transfer count. Documented in the PR description together with the valuation.

packages/database/src/repositories/InteropTransferRepository.ts
vilcht: Intentional, keeping it. The plugin/config filter half was addressed in 31b0bd27b2. On the valuation:

- `GREATEST(src, dst)` with Postgres' null-skipping is the SQL form of `getInteropTransferValue`, so the graph uses the same definition of a transfer's value as the protocols table and the flows on the same page. Matching main's per-side crediting would instead disagree with those two sections.
- One figure per transfer is what makes the once-per-cluster count well defined. With per-side values, a burn-and-mint node holding both sides of a transfer has no principled side to take; the per-token aggregate dodges that by always picking `src`.
- The numeric difference is the bridge fee on the receiving side, and where a side is missing main's aggregate added 0 while this falls back to the other side.

Documented in the PR description.

packages/frontend/src/server/features/layer2s/interop/token/buildTokenRelationsGraph.ts
vilcht: Intentional. The tokens site's conflict is two endpoints with different abstract token ids, which cannot occur inside one abstract token's graph. What this drops is two `lockAndMint` relations claiming opposite locked sides for the same deployment pair: contradictory evidence about direction, not a route we can draw. Hiding the pair beats picking one of the two arrows. Noted in the PR description.

packages/frontend/src/components/projects/sections/interop/onchain-deployments/relations-graph/relationsCamera.ts
```diff
@@ -0,0 +1,132 @@
...
+
+const MIN_SCALE = 0.2
+const MAX_SCALE = 4
+const FIT_PADDING = 32
+/** Below this the node text stops being readable. */
+const READABLE_SCALE = 0.55
+const WHEEL_ZOOM_STEP = 1.15
+
+/** Fits the width and anchors the top: a tall graph is panned, not shrunk into a smudge. */
+function fitCamera(content: Size, viewport: Size, focusX?: number): Camera {
```
torztomasz: nit: `fitCamera`/`zoomCamera` are private helpers but come before the exported `useRelationsCamera` (L69). Could we put the hook first and the helpers below it (like `clamp` already is), so the file reads top-down?

packages/frontend/src/components/projects/sections/interop/onchain-deployments/relations-graph/RelationsNode.tsx
```diff
@@ -0,0 +1,215 @@
...
+const BORDER_Y = 2
+const TITLE_HEIGHT = 20
+const META_GAP = 4
+const META_HEIGHT = 16
+const LIST_GAP = 12
+const CLUSTER_ROW_HEIGHT = 26
+const CLUSTER_FOOTER_HEIGHT = 20
+const HEADER_HEIGHT = TITLE_HEIGHT + META_GAP + META_HEIGHT
+
+export function getNodeSize(node: InteropTokenRelationsNode): {
```
torztomasz: nit: `getNodeSize`, `getClusterColumns` and `getShownMembers` sit above the `RelationsNode` component (L61). Could the component go first, with the helpers below next to `Meta`/`Bridges`?

packages/frontend/src/components/projects/sections/interop/onchain-deployments/relations-graph/graphSelectors.ts
```diff
@@ -0,0 +1,214 @@
...
+  }
+  for (const edge of edges) {
+    if (edge.from !== activeId) continue
+    edgeKeys.add(edgeKey(edge))
+    nodeIds.add(edge.to)
+  }
+  return { nodeIds, edgeKeys }
+}
+
+export interface RelationsPath {
```
torztomasz: nit: `RelationsPath` and the `MAX_PATH_*` constants (L76-77) are declared halfway down the file. Could they move to the top with the other declarations?

packages/frontend/src/server/features/layer2s/interop/utils/createInteropProjectResolver.ts
```diff
@@ -41,9 +41,13 @@ export function createMatchingProjectsResolver<TTarget>(
  * Unlike `createTransferBridgeResolver`, several matches are legitimate:
  * more than one bridge can mint the same deployment.
  */
+export type InteropProjectResolver = (
```
torztomasz, line 40: The new `InteropProjectResolver` type ended up between `createInteropProjectResolver`'s doc comment and the function, so the comment now documents the type. Could the type go above the comment?

packages/frontend/src/server/features/layer2s/interop/token/getInteropTokenOnchainDeployments.ts
```diff
@@ -1,160 +1,165 @@
...
+    [...result].map(([key, plugins]) => [key, [...plugins.values()]]),
   )
 }
 
-function deploymentKey(chain: string, address: string): string {
-  return `${chain}|${address.toLowerCase()}`
+function endpoint(
+  relation: TokenRelationRoute,
+  slot: 'A' | 'B',
+): { chain: string; address: string } {
```
torztomasz: nit: this could return `Endpoint` from `../utils/deploymentKey` instead of spelling out `{ chain: string; address: string }` again.

packages/frontend/src/server/features/layer2s/interop/token/getInteropTokenRelationsGraph.ts
```diff
@@ -0,0 +1,263 @@
...
+      iconUrl: manifest.getUrl(`/icons/${project.slug}.png`),
+      explorerUrl: project.chainConfig.explorerUrl,
+    }
+  )
+}
+
+/** Count a transfer once per group, even when both endpoints belong to it. */
+function aggregateStats(
+  rows: InteropTransferDeployedTokenPairStats[],
+  groupOf: (side: { chain: string; address: string }) => string | undefined,
```
torztomasz: nit: `side: Endpoint` here too.

packages/frontend/src/server/features/layer2s/interop/token/buildTokenRelationsGraph.test.ts
```diff
@@ -0,0 +1,167 @@
...
+
+    expect(graph.nodes.map((node) => node.members)).toEqual([
+      [arbitrum],
+      [ethereum],
+    ])
+  })
+})
+
+function route(
+  a: { chain: string; address: string },
```
torztomasz: nit: `a`, `b` and `locked` could be typed as `Endpoint`.

packages/frontend/src/server/features/layer2s/interop/token/buildTokenRelationsGraph.ts
```diff
@@ -0,0 +1,144 @@
...
+}
+
+export interface TokenRelationsGraphNode<T> {
+  id: string
+  members: T[]
+  sources: TokenRelationsGraphSource[]
+}
+
+export interface TokenRelationsGraphEdge {
+  /** `from` backs `to`. */
```
torztomasz: This `/** from backs to */` comment is repeated on three edge types (here, `getInteropTokenRelationsGraph.ts:51` and `layoutRelationsGraph.ts:9`). What if we renamed the fields to `backer`/`backed`? Then the comment wouldn't be needed, and every place that reads an edge would say what it means.

packages/backend/src/modules/interop/engine/cleaner/InteropCleanerLoop.ts
```diff
@@ -27,7 +27,7 @@ export class InteropCleanerLoop extends TimeLoop {
       now - 1 * UnixTime.DAY,
     )
     const expiredTransfers = await this.db.interopTransfer.deleteBefore(
-      now - 7 * UnixTime.DAY,
+      now - INTEROP_TRANSFER_RETENTION,
     )
     const expiredPrices = await this.db.interopRecentPrices.deleteBefore(
       now - 7 * UnixTime.DAY,
```
torztomasz: Transfers now use `INTEROP_TRANSFER_RETENTION`, but prices right below still hardcode `7 * UnixTime.DAY`. If prices are meant to share the transfer retention, could we use the constant here too? If not, a separately named constant would make it clear the two are independent.

packages/shared-pure/src/types/InteropTransferRetention.ts
```diff
@@ -0,0 +1,4 @@
+import { UnixTime } from './UnixTime.js'
+
+/** How long raw `InteropTransfer` rows are kept before the cleaner drops them. */
+export const INTEROP_TRANSFER_RETENTION = 7 * UnixTime.DAY
```
torztomasz: hmm maybe this should be in database? not sure

packages/frontend/src/server/features/layer2s/interop/utils/getLatestAggregatedInteropTransferWithTokens.ts
```diff
@@ -35,7 +38,8 @@ export async function getLatestAggregatedInteropTransferWithTokens({
     return { records: [], snapshotTimestamp: undefined }
   }
 
-  const snapshotTimestamp = await getAggregatedInteropSnapshotTimestamp()
+  const snapshotTimestamp =
+    resolvedSnapshotTimestamp ?? (await getAggregatedInteropSnapshotTimestamp())
```
torztomasz: shouldnt it be
```suggestion
  const resolvedSnapshotTimestamp =
    snapshotTimestamp ?? (await getAggregatedInteropSnapshotTimestamp())
```

packages/frontend/src/server/features/layer2s/interop/token/getInteropTokenPairStats.ts
```diff
@@ -0,0 +1,69 @@
...
+/** Past 24h transfer stats per deployment pair, or undefined without a snapshot. */
+export async function getInteropTokenPairStats(
+  tokenId: string,
+  snapshotTimestamp: UnixTime | undefined,
+  projects: Project<'interopConfig'>[],
+): Promise<InteropTransferDeployedTokenPairStats[] | undefined> {
+  if (env.MOCK) {
+    return MOCK_INTEROP_TOKEN_PAIR_STATS
+  }
+  if (!snapshotTimestamp) return undefined
```
torztomasz: then I guess do not accept snapshotTimestamp being undefined in arguments and move this check in call site of this function

packages/frontend/src/server/features/layer2s/interop/token/getInteropTokenEntry.ts
```diff
@@ -18,9 +10,9 @@ export interface InteropTokenEntry {
 export function getInteropTokenEntry(
   tokenId: string,
   interopChains: InteropChainWithIcon[],
-  projectsWithChains: Project<'chainConfig'>[],
-  interopProjects: Project<'interopConfig'>[],
-  deployments: InteropTokenOnchainDeployment[],
+  deploymentsCount: number,
+  /** Undefined when the token has no deployments. */
```
torztomasz: remove

packages/frontend/src/pages/interop/token/getInteropTokenPageData.ts
```diff
@@ -102,33 +106,48 @@ async function getCachedData({
...
+  // Everything below reads the same snapshot, and the slow per-pair transfer
+  // query needs nothing but the token id, so it starts alongside the rest.
+  const [
+    abstractTokens,
+    snapshotTimestamp,
+    projectsWithChains,
+    interopProjects,
+  ] = await Promise.all([
+    getInteropAbstractTokens(activeInteropChainIds),
+    env.MOCK ? undefined : getAggregatedInteropSnapshotTimestamp(),
```
torztomasz: I guess move that check inside function

packages/frontend/src/components/projects/sections/interop/onchain-deployments/relations-graph/TokenRelationsGraphView.tsx
```diff
@@ -0,0 +1,167 @@
...
+
+  const details = selectedNode && (
+    <RelationsDetails
+      graph={graph}
+      node={selectedNode}
+      onSelectNode={selectNode}
+      onClose={() => setSelectedNodeId(undefined)}
+    />
+  )
+  const diagram = (
```
torztomasz: Why some lowercase function, it should be simply a component

packages/frontend/src/components/projects/sections/interop/onchain-deployments/relations-graph/RelationsNode.tsx
```diff
@@ -0,0 +1,215 @@
...
+        width: box.width,
+        height: box.height,
+        padding: `${PADDING_Y}px ${PADDING_X}px`,
+      }}
+      className={cn(
+        'absolute flex flex-col overflow-hidden rounded-xl border bg-surface-primary text-left',
+        'focus-visible:outline-2 focus-visible:outline-brand/50',
+        isSelected ? 'border-brand ring-1 ring-brand' : 'border-divider',
+        isUnconnected && 'border-dashed',
+        isSource && 'outline outline-4 outline-brand/20 outline-offset-2',
```
torztomasz: 'outline-4' applies the same CSS properties as 'outline'.tailwindcss(cssConflict)

torztomasz:
On mobile we need to lock scrolling when using two fingers to move the diagram
image

torztomasz:
image
make the header of details sticky if possible, when scrolled I cant close it

torztomasz:
image

something wrong with sorting

torztomasz (approved):

PR #12763, opened 2026-09-07
Attribute RateLimiter waits and Http metrics per consumer

sdlyy:
RateLimiter.call() accepts an optional label and records per-label enqueue/dispatch counts, enqueue-to-dispatch wait (total/max), queue depth at enqueue, and limiter-wide in-flight counts, exposed via takeStats(). Dispatch order and spacing are unchanged.

ClientCore resolves the label once per fetch() from the RPC metrics context coreFeature (falling back to 'uncategorized'), reuses it across retries, and passes it to both the limiter and the metrics aggregator. ClientMetricsAggregator now emits one 'Http metrics' line per (client, label) with the limiter statistics merged in, so limiter queueing can be told apart from RPC latency per consumer. The clients2 Http path gets the same treatment and now includes its label in the emitted line.

### DIFF
--- a/packages/backend-tools/src/metrics/MetricsAggregator.ts
+++ b/packages/backend-tools/src/metrics/MetricsAggregator.ts
@@ -34,19 +34,24 @@ export abstract class MetricsAggregator<T> {
   }
 
   private flush() {
-    if (!this.buffer.length) {
-      return
-    }
-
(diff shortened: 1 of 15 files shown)


sdlyy:
@codex review

adamiak (approved):

PR #12769, opened 2026-09-07
add tokens page

vilcht:
Resolves L2B-14708. Stacked on #12759 (L2B-14707); merge that one first.

## What

A `/tokens` page listing every token that has at least one observed relation, busiest first, with a compact preview of how its deployments back one another. Opening a card shows the full interactive graph from the token page in a dialog; on phones a tap goes straight to the token page.

Everything is behind `CLIENT_SIDE_TOKENS_PAGE` (router, nav entry, search bar entry, sitemap path).

## How

- Database: two lean whole-table reads (`abstractToken.getAllSummaries`, `deployedToken.getAllAssignments`) so the tiles never load the heavy JSON columns.
- Server: a cached tiles feature built on the existing `buildTokenRelationsGraph`, in-memory paging, and a `tokens` tRPC router. The per-token graph lookup was extracted into `getInteropTokenRelationsGraphById` and the token page now uses it too.
- Page: infinite-scroll card grid, SVG preview per card, dialog reusing `TokenRelationsGraphView` (two new optional props).
- Wiring: nav, search bar, `pagePaths`, OG image.

Verified against the production read-only database: 1188 tokens with relations, first page server-rendered, dialog graph loads in under a second.


vilcht:
Opened by mistake, closing.

PR #12772, opened 2026-09-07
Cut interop syncer DB round trips and prepare capture items once per block

sdlyy:
## Why

Interop event syncers run one instance per resyncable cluster per chain (17 clusters today). In following mode every one of them did, for every block:

- `setLastError(null)` three times (before the state ran, inside the save transaction, and once more un-awaited afterwards),
- a SELECT of the resync request and a SELECT of the last synced range,
- a `BEGIN` / range upsert / `COMMIT` even when the block produced no events,
- its own `getItemsToCapture` pass over the block, with a per-transaction filter over all logs.

That is 8 statements per block per cluster (136 per Robinhood block, 7 of them awaited in sequence) and 18 conversions of the same block. On the two 340-row bookkeeping tables this showed up as ~3,000 row updates/s and heavy bloat, and it is the main reason Robinhood only catches up at ~1.4× chain speed.

## What

- **Clear the sync error only on the error→ok transition.** A private flag tracks whether a stored error may exist; `clearChainSyncError` writes only then. The floating post-save clear is removed (it was also an unhandled-rejection hazard).
- **Cache the synced range in `FollowingState`.** The syncer is the only writer of its range while following. Catch-up and wipe both go through `CatchingUpState`, which creates a new `FollowingState`, so the cache is dropped with the state. The resync check runs at most every 10 s from the block path; the `checkStatus` tick alone is not enough because it is skipped while a block is being processed.
- **Skip the transaction when only the range changes.** A single upsert is atomic on its own; the transaction stays whenever events, derived updates or an error clear are written.
- **Prepare capture items once per block.** `BlockIndexer` passes the same block and logs objects to every processor, and the manager to every syncer, so `getItemsToCapture` memoizes per block object (WeakMap, checked against chain and logs reference). Logs are grouped by transaction in one pass instead of a filter per transaction.

## Effect

| per block per cluster, steady state | before | after |
|---|---|---|
| SQL statements | 8 (7 awaited in sequence) | 1 (+1 SELECT per 10 s) |

| capture CPU per block, 17 clusters + legacy processor | before | after |
|---|---|---|
| 40 txs / 40 logs (Robinhood median) | 1.25 ms | 0.27 ms |
| 200 txs / 1,000 logs | 33.7 ms | 2.6 ms |

## Behaviour changes to be aware of

- A resync or wipe request from the backoffice is noticed up to 10 s later than before (the catch-up path already had that latency).
- The stored error now stays visible until a retry actually succeeds, instead of flickering to `null` when the retry starts.
- All clusters now share the prepared capture items. No plugin mutates its input (grepped for in-place mutation of `txLogs`, `topics`, `transactions`, `Object.assign`, `delete`); the `block` object and `txLogs` arrays were already shared within a cluster before.

## After deploy

- `BlockProcessingStats` wall time should converge on CPU time; the row-update rate on `InteropPluginSyncState` / `InteropPluginSyncedRange` should drop to roughly the block rate × 17.
- One-off `VACUUM FULL` + reindex on those two tables; autovacuum will not reclaim the accumulated bloat.

## Tests

- 6 new/updated cases in `InteropEventSyncer.test.ts` and `FollowingState.test.ts`, 2 new in `getItemsToCapture.test.ts`.
- All 1,476 interop tests and backend typecheck pass.

Discovery notes: `docs/interop-performance-discovery-claude.md` (S1.4, S1.5, S1.6, S1.10).

### DIFF
--- a/packages/backend/src/modules/interop/engine/sync/InteropEventSyncer.test.ts
+++ b/packages/backend/src/modules/interop/engine/sync/InteropEventSyncer.test.ts
@@ -2,6 +2,7 @@ import { Logger } from '@l2beat/backend-tools'
 import type { InteropPluginName } from '@l2beat/config'
 import type {
   BlockRangeWithTimestamps,
+  Database,
   InteropEventRecord,
   InteropPluginSyncedRangeRecord,
 } from '@l2beat/database'
@@ -76,9 +77,9 @@ describe(InteropEventSyncer.name, () => {
     it('does not clear errors when checking blockProcessor status', async () => {
       const setLastError = mockFn().resolvesTo(undefined)
       const syncer = createSyncer({
-        db: mockObject<InteropEventSyncer['db']>({
+        db: mockObject<Database>({
           interopPluginSyncState: mockObject<
-            InteropEventSyncer['db']['interopPluginSyncState']
+            Database['interopPluginSyncState']
           >({
             setLastError,
             findByPluginNameAndChain: mockFn().resolvesTo(undefined),
@@ -226,9 +227,9 @@ describe(InteropEventSyncer.name, () => {
     it('stores last error when state throws', async () => {
       const setLastError = mockFn().resolvesTo(undefined)
       const syncer = createSyncer({
-        db: mockObject<InteropEventSyncer['db']>({
+        db: mockObject<Database>({
           interopPluginSyncState: mockObject<
-            InteropEventSyncer['db']['interopPluginSyncState']
+            Database['interopPluginSyncState']
           >({
             setLastError,
           }),
@@ -252,9 +253,9 @@ describe(InteropEventSyncer.name, () => {
     it('clears a possibly stale stored error once on the first success', async () => {
       const setLastError = mockFn().resolvesTo(undefined)
       const syncer = createSyncer({
-        db: mockObject<InteropEventSyncer['db']>({
+        db: mockObject<Database>({
           interopPluginSyncState: mockObject<
-            InteropEventSyncer['db']['interopPluginSyncState']
+            Database['interopPluginSyncState']
           >({
             setLastError,
           }),
@@ -274,9 +275,9 @@ describe(InteropEventSyncer.name, () => {
     it('clears the stored error once after recovering from a failure', async () => {
       const setLastError = mockFn().resolvesTo(undefined)
       const syncer = createSyncer({
-        db: mockObject<InteropEventSyncer['db']>({
+        db: mockObject<Database>({
           interopPluginSyncState: mockObject<
-            InteropEventSyncer['db']['interopPluginSyncState']
+            Database['interopPluginSyncState']
           >({
             setLastError,
           }),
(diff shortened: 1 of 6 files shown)


sdlyy:
@codex review

sdlyy:
@codex review

sdlyy:
Re the AI review finding "Keep stored sync errors until a retry actually completes": confirmed and fixed in 80bdafd576.

`triggerState` now clears the stored error only when a state run ends in a block-processor state (the syncer is following again). A `Following → CatchingUp` transition after a gap, or a catch-up tick that ends waiting for a wipe or a smaller-range retry, no longer clears it. Data that is actually persisted still clears it inside `saveProducedInteropEvents`, so a successful catch-up range clears at the right moment. Two tests cover the transition directions.

For the record, `main` had the same behaviour (the pre-clear ran before the state did), so this fix goes one step beyond the previous semantics rather than restoring them.

_Reply written with Claude Code._

packages/backend/src/modules/interop/engine/sync/InteropEventSyncer.test.ts
```diff
@@ -245,9 +245,82 @@ describe(InteropEventSyncer.name, () => {
...
+      expect(setLastError.calls[0]?.args[2]).toInclude('boom')
+    })
+
+    it('clears a possibly stale stored error once on the first success', async () => {
+      const setLastError = mockFn().resolvesTo(undefined)
+      const syncer = createSyncer({
+        db: mockObject<InteropEventSyncer['db']>({
+          interopPluginSyncState: mockObject<
+            InteropEventSyncer['db']['interopPluginSyncState']
+          >({
```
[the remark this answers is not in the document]
sdlyy: Done in e2368420f2: every repository mock in this test file is now typed as `Database['<repo>']` / `Database`, per AGENTS.md. Other sync test files still use the old pattern but are outside this PR.

packages/backend/src/modules/interop/engine/sync/InteropEventSyncer.ts
```diff
@@ -333,15 +357,21 @@ export class InteropEventSyncer extends TimeLoop {
...
   async clearChainSyncError() {
+    if (!this.storedErrorMayExist) {
+      return
+    }
     await this.db.interopPluginSyncState.setLastError(
       this.cluster.name,
       this.chain,
       null,
     )
+    this.storedErrorMayExist = false
```
adamiak: Posted by Claude on behalf of Adrian, from a review of this branch.

`storedErrorMayExist` is flipped to `false` here, but `clearChainSyncError()` also runs inside the transaction in `saveProducedInteropEvents` (line 348). If that transaction rolls back after this statement, the in-memory flag says "no error stored" while the DB still holds the old `lastError`.

Today this is safe only because the clear is the last statement before `COMMIT`, and a failed commit throws into the `triggerState` catch, which calls `saveChainSyncError` and resets the flag. Anyone reordering the statements in that transaction would silently break it and leave a stale error visible in the backoffice until the next failure.

Suggestion: keep the `setLastError(null)` write inside the transaction, but set `this.storedErrorMayExist = false` only after the transaction has resolved (e.g. after `runInTransaction` returns in `saveProducedInteropEvents`, and directly in the non-transactional call from `triggerState`).

adamiak (approved):

packages/backend/src/modules/interop/engine/sync/InteropEventSyncer.ts
sdlyy: Agreed, fixed in 12de77337f. `clearChainSyncError` now only writes; `storedErrorMayExist` is reset after `runInTransaction` resolves in `saveProducedInteropEvents`, and right after the direct call in `triggerState`. Added a test that fails the commit after the clear and checks the next save writes the clear again.

_Reply written with Claude Code._

PR #12773, opened 2026-09-07
Probe block timestamps without tx bodies and demote per-block logs

sdlyy:
## Why

Two small block-sync costs identified in the interop performance discovery (`docs/interop-performance-discovery-claude.md`, S1.8 and S0.3):

- `BlockNumberIndexer` ticks every 10 s and calls `BlockProvider.getBlockNumberAtOrBefore`, which interpolates/binary-searches with ~5–8 block probes. Each probe fetched the block **with transaction bodies** even though only the timestamp is used. On Robinhood that is ~1 MB of payload per tick and about 5% of the chain's RPC budget; on Base-sized blocks it is several MB.
- `BlockIndexer` logged `Processor finished` and `Processed block` at info for every block and processor, ~2.6M documents a day for Robinhood alone, each one stringified and shipped to Elasticsearch.

## What

- **Header-only timestamp probes.** `BlockClient` gets an optional `getBlockTimestamp(blockNumber)` capability. `RpcClient` implements it with `eth_getBlockByNumber(n, false)`. `BlockProvider.getBlockNumberAtOrBefore` prefers it and falls back to `getBlockWithTransactions` for clients that do not implement it (Fuel, Starknet, Aztec). Same RPC method, same call count, same limiter accounting; only the `includeTxs` flag changes.
- **Per-block log lines to debug.** The two per-block messages move to debug. The per-batch `Processed blocks` and `Fetched blocks and logs` lines stay at info, and the `Metrics` line that the "[Interop] Out of sync" alert and the RPC budget dashboard rely on is untouched.

## Checked before changing the log level

Searched Kibana saved objects (visualization `visState` and `searchSourceJSON`, lens `state`, saved searches) and all 23 alerting rules for `Processed block` / `Processor finished`: no references.

## Blast radius

`getBlockNumberAtOrBefore` is also used by activity's and DA's `BlockTargetIndexer`, TVS's `BlockTimestampIndexer` and privacy's timestamp indexer. They all get the same timestamps from smaller responses; there is no semantic change.

## Tests

- `RpcClient.getBlockTimestamp` sends `[n, false]` and returns the timestamp.
- `BlockProvider` prefers the header probe and never calls `getBlockWithTransactions` when the capability exists; existing fallback tests keep passing with the capability absent.
- Shared and backend typecheck pass.

### DIFF
--- a/packages/backend/src/modules/block-sync/BlockIndexer.ts
+++ b/packages/backend/src/modules/block-sync/BlockIndexer.ts
@@ -132,7 +132,7 @@ export class BlockIndexer extends ManagedChildIndexer {
             () => processor.processBlock(block, logs),
           )
           const duration = Date.now() - start
-          this.logger.info('Processor finished', {
+          this.logger.debug('Processor finished', {
             processor: processor.constructor.name,
             durationMs: Number.parseFloat(duration.toFixed(2)),
           })
@@ -144,7 +144,7 @@ export class BlockIndexer extends ManagedChildIndexer {
           })
         }
       }
-      this.logger.info('Processed block', {
+      this.logger.debug('Processed block', {
         blockNumber: block.number,
         logs: logs.length,
       })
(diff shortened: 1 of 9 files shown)


sdlyy:
@codex review

sdlyy:
@codex review

adamiak (approved):

PR #12780, opened 2026-09-08
Add Tokens page

vilcht:
Stacked on #12759. Resolves L2B-14708.

Adds `/tokens`, a grid of every token that has relations between its deployments, each card with a compact preview of its relations graph. Clicking a card opens the full graph in a dialog, reusing `TokenRelationsGraphView` from the token detail page.

The whole page is gated behind `CLIENT_SIDE_TOKENS_PAGE` (router, nav, search bar, sitemap, page paths).

## Commits

- **repositories methods** – lightweight `getAllSummaries` / `getAllAssignments` reads for the tiles.
- **server functions** – `buildTokenGraphTiles` (pure), cached whole-table loader with paging, shared `getInteropTokenRelationsGraphById`, `tokens` tRPC router, mock branch.
- **ui** – page, router, grid with infinite scroll, tile, preview diagram, dialog with error/retry; graph view gains `embedded` for inline details in a dialog.
- **fix layout** – preview rows come from the real layout at real card sizes, so wrapping matches the dialog; card size formula extracted to `nodeSize.ts`.
- **nav, svg, ogimage** – flag-gated nav entry, search entry, `pagePaths`, icon, OG image.
- **unify orders** – per-deployment volume on tile nodes via one all-tokens pair-stats query, so sibling order matches the dialog exactly.

## Notes for review

- Tiles link only to tokens the detail page can resolve; others render without a link.
- The preview is a second renderer over the same layout and routing rows as the dialog. A shared-router version was tried and produced visibly worse thumbnails.
- The all-tokens pair-stats query is a second consumer of raw 24h transfers with the same semantics as the dialog. When per-deployment volume moves into the aggregation snapshot, both should switch together.

## Verified

- Typecheck, lint, frontend suite green.
- Page, both tRPC procedures, and mock mode render without a database.
- Row order identical to the dialog for the six busiest tokens against production data.
- Phone viewport: dialog is full-screen, node details render inline, no stacked modals.


packages/database/src/repositories/AbstractTokenRepository.ts
```diff
@@ -182,6 +187,14 @@ export class AbstractTokenRepository extends BaseRepository {
     return result.map(toRecord)
   }
 
+  /** Skips the coingecko/comment JSON that `getAll` carries. */
```
torztomasz: remove

packages/database/src/repositories/DeployedTokenRepository.ts
```diff
@@ -229,6 +234,14 @@ export class DeployedTokenRepository extends BaseRepository {
     return rows.map(toRecord)
   }
 
+  /** Skips the metadata/assignment-proof JSON that `getAll` carries. */
```
torztomasz: remove

packages/database/src/repositories/InteropTransferRepository.ts
```diff
@@ -983,19 +994,49 @@ export class InteropTransferRepository extends BaseRepository {
...
+  ): Promise<InteropTransferDeployedTokenPairStatsByToken[]> {
+    return this.queryDeployedTokenPairStats(timeRange, selection, (eb) =>
+      eb.or([
+        eb('srcAbstractTokenId', 'is not', null),
+        eb('dstAbstractTokenId', 'is not', null),
+      ]),
+    )
+  }
+
+  /** `getAllDeployedTokenPairStats` narrowed to one abstract token. */
```
torztomasz: Remove

packages/frontend/src/pages/tokens/TokensRouter.ts
```diff
@@ -0,0 +1,30 @@
...
+  render: RenderFunction,
+  cache: InMemoryCache,
+) {
+  if (!env.CLIENT_SIDE_TOKENS_PAGE) return null
+  const router = express.Router()
+
+  router.get('/tokens', async (req, res) => {
+    const data = await cache.get(
+      {
+        key: ['tokens', req.originalUrl],
```
torztomasz: do not cache on URL, query params misses cache

packages/frontend/src/pages/tokens/components/TokenGraphTileCard.tsx
```diff
@@ -0,0 +1,64 @@
...
+          </span>
+          {tile.issuer && (
+            <span className="block truncate text-label-value-12 text-secondary">
+              Issued by <span className="capitalize">{tile.issuer}</span>
+            </span>
+          )}
+        </span>
+        {tile.volume !== null && (
+          <span className="ml-auto shrink-0 font-medium text-label-value-13 text-secondary">
+            {formatCurrency(tile.volume, 'usd')}
```
torztomasz: I would add some label `VOLUME` above with `text-secondary` probably and smaller font size

packages/frontend/src/server/features/tokens/getTokenGraphTilesPage.ts
```diff
@@ -0,0 +1,31 @@
...
+export async function getTokenGraphTilesPage({
+  cursor = 0,
+}: TokenGraphTilesParams): Promise<TokenGraphTilesPage> {
+  const tiles = await getTokenGraphTiles()
+  const end = cursor + PAGE_SIZE
+  return {
+    items: tiles.slice(cursor, end),
+    nextCursor: end < tiles.length ? end : undefined,
+    total: tiles.length,
+  }
```
torztomasz, line 30: wouldn't the page be faster if we didnt fake the pagination?

torztomasz:
https://github.com/user-attachments/assets/f1a8fba6-75e1-4df8-8743-dd95ddfa9879

I guess we can make the layout not shift on load

packages/frontend/src/server/features/layer2s/interop/token/getInteropTokenRelationsGraphById.ts
```diff
@@ -0,0 +1,52 @@
...
+    () => getInteropTokenRelationsGraphById(tokenId),
+  )
+}
+
+export async function getInteropTokenRelationsGraphById(
+  tokenId: string,
+): Promise<InteropTokenRelationsGraph | undefined> {
+  const [snapshotTimestamp, [projectsWithChains, interopProjects]] =
+    await Promise.all([
+      env.MOCK ? undefined : getAggregatedInteropSnapshotTimestamp(),
```
torztomasz: move env.MOCK inside function

packages/frontend/src/server/features/layer2s/interop/token/getInteropTokenRelations.ts
```diff
@@ -0,0 +1,24 @@
+import { INTEROP_TRANSFER_RETENTION, UnixTime } from '@l2beat/shared-pure'
+import { ps } from '~/server/projects'
+import { getAggregatedInteropSnapshotTimestamp } from '../utils/getAggregatedInteropTimestamp'
+import { getActiveInteropChainIds } from '../utils/getInteropChains'
+
+/** Shared time window and eligibility for the tiles and detailed token graphs. */
+export async function getPairStatsParams() {
```
torztomasz: this function is named completely different than file

torztomasz (changes requested):

torztomasz:
idk if that's just me but I can only see the "Backs" on the diagram, rest of the legend is something I do not recognize on the graph 
image

torztomasz:
Why are some of the tokens having two circles around? Why some of them have highlighted circles? Why some circles are more highlighted than the others?
image

packages/frontend/src/server/features/layer2s/interop/token/getInteropTokenRelations.ts
```diff
@@ -0,0 +1,24 @@
+import { INTEROP_TRANSFER_RETENTION, UnixTime } from '@l2beat/shared-pure'
+import { ps } from '~/server/projects'
+import { getAggregatedInteropSnapshotTimestamp } from '../utils/getAggregatedInteropTimestamp'
+import { getActiveInteropChainIds } from '../utils/getInteropChains'
+
+/** Shared time window and eligibility for the tiles and detailed token graphs. */
```
torztomasz: Comment restates what a good name would say. Once the function/file naming is sorted (see comment below), something like `getTokenGraphPairStatsParams` makes this doc line redundant.

packages/frontend/src/server/features/layer2s/interop/token/getInteropTokenRelations.ts
```diff
@@ -0,0 +1,24 @@
...
+  if (from < UnixTime.now() - INTEROP_TRANSFER_RETENTION) return undefined
+  const projects = await ps.getProjects({ select: ['interopConfig'] })
+  const chains = getActiveInteropChainIds()
+  return {
+    timeRange: { from, to: snapshotTimestamp },
+    selection: {
+      plugins: projects.flatMap((project) => project.interopConfig.plugins),
+      sourceChains: chains,
+      destinationChains: chains,
+    },
```
torztomasz: **Duplicated code.** This is a copy of `getInteropTokenPairStats.ts:21-36`: same 24h window, same retention guard (the why-comment is copied verbatim too), same active-chains + plugins selection. The PR description says both consumers must keep the same semantics, but nothing enforces it. Make `getInteropTokenPairStats` call this helper so there is one definition. That file also still inlines `.filter((chain) => !chain.isUpcoming).map(...)` instead of the new `getActiveInteropChainIds()`.

packages/frontend/src/server/features/layer2s/interop/utils/getRelationsGraphProjects.ts
```diff
@@ -0,0 +1,9 @@
+import { ps } from '~/server/projects'
+
+/** Projects the relations graph needs for chain display and bridge names. */
```
torztomasz: Comment restates the two `select` calls below; the function name already says it.

**Duplicated code / data clump.** `getInteropTokenPageData.ts:118-119` still does the same two `ps.getProjects` calls inline, so this helper has not replaced the original. Also the result is a positional tuple `[projectsWithChains, interopProjects]` that travels through `getInteropTokenRelationsGraphById`, `getTokenGraphTiles` and into `getInteropTokenRelationsGraph`. Return a named object `{ projectsWithChains, interopProjects }` so callers cannot swap them.

packages/frontend/src/server/features/layer2s/interop/token/getInteropTokenRelationsGraphById.ts
```diff
@@ -0,0 +1,52 @@
...
+): Promise<InteropTokenRelationsGraph | undefined> {
+  return relationsGraphCache.get(
+    {
+      key: ['interop-token-relations-graph', tokenId],
+      ttl: 5 * 60,
+      staleWhileRevalidate: 25 * 60,
+    },
+    () => getInteropTokenRelationsGraphById(tokenId),
+  )
+}
```
torztomasz: **Middle man.** `getCachedInteropTokenRelationsGraphById` only delegates to the function below, and the uncached one has no other caller (the token page assembles the graph itself). Keep a single exported function with the cache inside.

packages/frontend/src/server/features/layer2s/interop/token/getInteropTokenRelationsGraphById.ts
```diff
@@ -0,0 +1,52 @@
...
+  ])
+  if (deployments.length === 0) return undefined
+
+  return getInteropTokenRelationsGraph(
+    tokenId,
+    deployments,
+    { routes, pairStats },
+    projectsWithChains,
+    interopProjects,
+  )
```
torztomasz, line 62: **Duplicated code.** Same assembly as `getInteropTokenPageData.ts:126-144`: deployments + pair stats in parallel, `deployments.length === 0` guard, then `getInteropTokenRelationsGraph(...)` with identical args. The page should call this function (passing the snapshot/projects it already has) instead of keeping a second copy.

packages/frontend/src/server/features/tokens/buildTokenGraphTiles.test.ts
```diff
@@ -0,0 +1,195 @@
...
+    })
+
+    expect(tile).toEqual({
+      id: usdc.id,
+      symbol: 'USDC',
+      issuer: 'circle',
+      iconUrl: TOKEN_PLACEHOLDER_ICON_URL,
+      href: '/interop/tokens/usdc01/circle/usdc',
+      volume: null,
+      // Optimism has no relation and is left out.
```
torztomasz: Move this into the test: either the `it(...)` title or a separate case like `it('leaves out deployments without a relation')`. A comment inside the expected object is easy to miss and will rot when the fixture changes.

packages/frontend/src/server/features/tokens/buildTokenGraphTiles.test.ts
```diff
@@ -0,0 +1,195 @@
...
+        nodes: [
+          {
+            id: 'base|0xb1',
+            volume: null,
+            chains: [{ id: 'base', iconUrl: undefined }],
+          },
+          {
+            id: 'ethereum|0xe1',
+            volume: null,
+            // Display-name order, not id order.
```
torztomasz: Same here. Chain ordering inside a burn-and-mint node is its own behaviour, so give it its own `it('orders cluster chains by display name')` instead of a comment.

packages/frontend/src/pages/tokens/components/TokenGraphGrid.tsx
```diff
@@ -0,0 +1,61 @@
+import { useInfiniteQuery } from '@tanstack/react-query'
+import { useMemo } from 'react'
+import { Skeleton } from '~/components/core/Skeleton'
+import { NoDataBanner } from '~/components/NoDataBanner'
+import {
+  InfiniteScrollTrigger,
+  useInfiniteScrollTrigger,
+} from '~/pages/interop/components/tokens/infiniteScroll'
```
torztomasz: **Feature envy.** A top-level page reaching into another page's private components folder. `infiniteScroll` now has three consumers across two pages, so move it to `~/components/`.

packages/frontend/src/pages/tokens/components/tilePreviewLayout.ts
```diff
@@ -0,0 +1,341 @@
...
+  TokenGraphTileNode,
+} from '~/server/features/tokens/buildTokenGraphTiles'
+
+export const VIEW_WIDTH = 320
+export const VIEW_HEIGHT = 132
+export const X_PADDING = 12
+export const Y_PADDING = 14
+export const BASE_RADIUS = 6
+export const SOURCE_RING_GAP = 3
+export const LINE_GAP = 0.5
```
torztomasz: **Speculative generality.** `X_PADDING`, `Y_PADDING` and `LINE_GAP` are exported but nothing outside this file uses them. Drop the `export`.

packages/frontend/src/pages/tokens/components/tilePreviewLayout.ts
```diff
@@ -0,0 +1,341 @@
...
+  radius: number,
+  sourceIds: ReadonlySet<string>,
+): number {
+  const ringGap = (SOURCE_RING_GAP / BASE_RADIUS) * radius
+  return radius + (row.some((node) => sourceIds.has(node.id)) ? ringGap : 0)
+}
+
+function getVerticalSpan(nodeCount: number, rowCount: number): number {
+  if (rowCount <= 1) return 0
+  if (rowCount === 2 && nodeCount <= 8) return 58
```
torztomasz: **Primitive obsession / magic number.** `58` and the `nodeCount <= 8` threshold sit outside `SIZE_BUCKETS`, so there are two places that decide vertical span. Either fold this into the buckets or name it (`TWO_ROW_SPAN`).

packages/frontend/src/pages/tokens/components/tilePreviewLayout.ts
```diff
@@ -0,0 +1,341 @@
...
+}
+
+interface RowGroup {
+  row: number
+  busY: number
+  targets: Mark[]
+}
+
+/** Source drops to a bus above each backed row; deeper rows go via a side lane. */
+function buildPaths(
```
torztomasz: **Duplicated code.** `buildPaths` + `getSideLane` re-implement the idea of `relations-graph/routeRelationsEdges.ts`: adjacent row via a bus, deeper rows via a side lane, departure Y, lane index per source. Rows are shared through `layoutRelationsGraph` but routing is not, so the thumbnail and the dialog can drift apart. The PR description says a shared router looked worse. If so, write that *why* here (one bus per source row instead of per edge) and extract at least the adjacent-vs-deep split so both routers agree on it.

packages/frontend/src/pages/tokens/components/tilePreviewLayout.ts
```diff
@@ -0,0 +1,341 @@
...
+
+    const targets = (outgoing.get(sourceId) ?? []).map((edge) => edge.target)
+    const groups: RowGroup[] = [...Map.groupBy(targets, (target) => target.row)]
+      .toSorted(([a], [b]) => a - b)
+      .map(([row, targets]) => {
+        const targetY = Math.min(
+          ...targets.map((target) => target.y - target.radius - LINE_GAP),
+        )
+        const available = Math.max(0, targetY - startY)
+        const clearance = Math.max(3, Math.min(6, available * 0.3))
```
torztomasz: Magic numbers: `3`, `6`, `0.3`, and `5` on line 235. Name them (`MIN_BUS_CLEARANCE`, `MAX_BUS_CLEARANCE`, ...) next to `MIN_GAP` at the top.

packages/frontend/src/pages/tokens/components/tilePreviewLayout.ts
```diff
@@ -0,0 +1,341 @@
...
+    from.x - from.halfWidth,
+    ...passed.map((mark) => mark.x - mark.halfWidth),
+  )
+  const rightEdge = Math.max(
+    from.x + from.halfWidth,
+    ...passed.map((mark) => mark.x + mark.halfWidth),
+  )
+  const laneGap = 7 + lane * 1.5
+  const left = Math.max(3, leftEdge - laneGap)
+  const right = Math.min(VIEW_WIDTH - 3, rightEdge + laneGap)
```
torztomasz: Magic numbers again: `7 + lane * 1.5` and the `3` edge inset. `LANE_GAP`, `LANE_STEP`, `VIEW_EDGE_INSET`.

torztomasz (commented):
Second pass focused on comments that restate code and on code smells (all judgement calls, none blocking).

One that has no single line to attach to: **shotgun surgery** around pair stats. The 24h window + eligibility rules now live in `getInteropTokenPairStats.ts`, `getInteropTokenRelations.ts` and two repository methods; changing the semantics means touching all of them. Inline comments below cover the concrete spots.

packages/frontend/src/pages/tokens/components/tilePreviewLayout.ts
```diff
@@ -0,0 +1,341 @@
...
+  const shown = Math.min(MAX_CLUSTER_ICONS, node.chains.length)
+  const remaining = node.chains.length - shown
+  const iconDiameter = 8.5 * scale
+  const iconStep = 6.5 * scale
+  const iconsWidth = iconDiameter + (shown - 1) * iconStep
+  const countWidth =
+    remaining > 0 ? (String(remaining).length * 4 + 5) * scale : 0
+  const contentWidth = iconsWidth + (remaining > 0 ? 3 * scale : 0) + countWidth
+  return {
+    width: contentWidth + 10 * scale,
```
torztomasz: **Primitive obsession.** `8.5`, `6.5`, `4`, `5`, `3`, `10` are all cluster geometry in base-radius units. Group them into one `CLUSTER` constant object (`iconDiameter`, `iconStep`, `charWidth`, `countPadding`, `countGap`, `horizontalPadding`) so the pill shape is defined in one readable place.

packages/frontend/src/server/features/tokens/getTokenGraphTilesPage.ts
vilcht: Measured on prod: 1283 tokens get a tile, so the client needs pages either way. Server side the tile list is built once per 10 min and cached; each page is an array slice, and the volume sort needs all tokens before slicing, so real pagination would not skip any work. Happy to raise the page size if the round trips bother you.

PR #12781, opened 2026-09-08
Preparse diff history into the project database

mateuszradomski:
Part of L2B-14828

Preparse diff history into the project database

The frontend used to read every project's diffHistory.md from the config checkout at request time, parse it with DiffHistoryParser, filter it down to the public sections, and cache the result in memory keyed by file mtime. That was the only reason the parser lived in @l2beat/shared: the frontend cannot depend on @l2beat/discovery, and shared was the one package both could import.

This PR moves the work to build time. @l2beat/config parses each project's diff history once in getProjects and stores the public updates as a new discoveryUpdates column in the SQLite database, next to everything else the frontend already reads. The frontend selects that column like any other project field. It no longer imports the parser, touches the filesystem under config/src/projects, or keeps a parse cache.

With the frontend dependency gone, DiffHistoryParser and the diff markdown helpers move from @l2beat/shared to @l2beat/discovery, the package that produces the markdown they read. Shared has no diff history code left.

What stays the same:

- The filtering rules are unchanged. Watched changes and initial discovery are public, config related changes are reduced to contracts being created or verified, and source code changes are never public.
- Update ids are unchanged. The stored id is the same content fingerprint the update card's copy link has always used, so existing ?update= links keep resolving. Verified on arbitrum, taiko, ccip, railgun, and relay: the old and new code produce identical output on every field.
- The hasDiscoUi gate and the seven day recentUpdatesCount stay in the frontend, since they depend on request time.

The recent changes overview reads the column once per process and keeps the first 20 entries per project, then filters by the seven day window per request. That matches what the old mtime cache gave it without decoding the full history on every open of the dialog.

Cost: the database grows from 14 MB to about 29 MB, and the config build spends under two seconds parsing. Both are build artifacts, and no page sends more bytes than before.


mateuszradomski:
@codex, review

torztomasz (approved):

torztomasz (commented):
Validation: 22 focused tests passed. Old and new parsing produced identical output across 240 histories containing 3,798 public updates. Reviewed against the PR description; the linked Linear issue was unavailable.

Standards: no findings. The duplicated-scanner suggestion was withdrawn because the small local scanner respects the intended package boundary: diff-history helpers leave shared, and the frontend cannot depend on discovery.

Spec: two P2 findings.

packages/frontend/src/server/features/layer2s/project/getL2ProjectEntry.ts
```diff
@@ -226,7 +224,7 @@ export async function getL2ProjectEntry(
 
   const projectLiveness = liveness[project.id]
   const discoveryUpdates = project.discoveryInfo?.hasDiscoUi
-    ? getDiscoveryUpdates(project.id)
+    ? (project.discoveryUpdates ?? [])
```
torztomasz, line 227: **Spec — P2: Restore the 50-update limit on project pages.** The previous `getDiscoveryUpdates(project.id)` defaulted to 50 results; this now forwards the entire history into the updates section and serialized SSR props, contradicting “no page sends more bytes than before.” Privacy and interop make the same change. Taiko now receives 105 updates instead of 50, increasing serialized update JSON from roughly 168 KB to 390 KB. Restore the previous limit in all three callers.

packages/frontend/src/server/features/projects/recent-changes/getRecentChangesOverview.ts
```diff
@@ -25,6 +21,7 @@ export interface RecentChangesOverview {
 
 export async function getRecentChangesOverview(): Promise<RecentChangesOverview> {
   const projects = await ps.getProjects({
+    select: ['discoveryUpdates'],
```
torztomasz, line 24: **Spec — P2: Cache the overview’s truncated histories.** This does not implement “reads the column once per process and keeps the first 20 entries per project.” Every invocation queries the applicable projects and `ProjectDatabase` JSON-decodes each complete history before `.slice(0, 20)` runs. Neither `ProjectService` nor `ProjectDatabase` caches this operation, so repeatedly opening the dialog repeats full-history reads, parsing, and allocations. Cache the initial query’s truncated results per process, keeping the seven-day filter request-dependent.

PR #12786, opened 2026-09-09
Adversaries (privacy)

sekuba:


packages/config/src/projects/tornado-cash/adversaries.ts
```diff
@@ -0,0 +1,89 @@
...
+        },
+      ],
+    },
+    chainAnalyst: {
+      condition: 'anonymity set per pool, no delay',
+      sentiment: 'good',
+      exposure:
+        'Every withdrawal spends one deposit of the same fixed amount, so the candidates are the unspent notes in that pool, the tracked anonymity set, which differs sharply between pools. No delay is enforced; timing and address reuse narrow the set further.',
+      advice:
+        'Use a pool with a large anonymity set, wait days or weeks before withdrawing, withdraw to a fresh address that never touches your other wallets, and deposit a large sum as one note.',
```
lucadonnoh, line 30: this is not good enough. we should warn the users on other types of more subtle fingerprinting: for example, smart wallets impl (e.g. ambire 7702 impl) shared by sender and recipient. another one: some wallets can be fingerprinted by how they set the priority fee. there might be others we should think about and maybe try to do the analysis ourselves on one of the pools. another one could be how wallet sets the max gas limit.

packages/config/src/projects/tornado-cash/adversaries.ts
```diff
@@ -0,0 +1,89 @@
...
+          title: 'Blockchain is Watching You (arXiv:2005.14051)',
+          url: 'https://arxiv.org/abs/2005.14051',
+        },
+      ],
+    },
+    networkObserver: {
+      condition: 'note never leaves the device',
+      sentiment: 'good',
+      exposure:
+        'The app downloads all deposits and searches locally, so a node learns only which pool you looked at. The relayer only submits the finished withdrawal.',
```
lucadonnoh, line 45: the user should connect to the relayer via tor! the real ip is a big fingerprint. or at least a vpn. but now the vpn is trusted. many vpns can definitely NOT be trusted

lucadonnoh (changes requested):

packages/config/src/projects/strk20/adversaries.ts
```diff
@@ -0,0 +1,192 @@
...
+    protects: 'linkage',
+    text: 'Hides amounts and senders inside the pool, and the link between deposit and withdrawal. First contacts are public.',
+  },
+  cells: {
+    publicObserver: {
+      sentiment: 'warning',
+      exposure:
+        'Transfers inside are encrypted and the client proof is not published, but the public part of every transaction lists the recipient address whenever you open a channel to someone new, the token and amount of the fee you pay, and the target contract and calldata of any DeFi action. Deposits and withdrawals show address, token and amount. Transactions reach the chain through a shared relayer, so the address that submits them does not name you.',
+      advice:
+        'Pay fees in the most common token and treat the first payment to a new recipient, and any DeFi action, as public.',
```
sergeyshemyakov: AFAIU when Alice opens a channel to Bob, the only thing that could publicly attribute this action to Alice is if she actually sends this transaction herself. Pls check, and if this is true, add an advice to open channels via relayed transactions, not signed by you.

packages/config/src/projects/privacy-pools/adversaries.ts
```diff
@@ -0,0 +1,138 @@
...
+        {
+          title: 'Relayer quote carries amount and recipient',
+          url: 'https://github.com/0xbow-io/privacy-pools-website/blob/62a962f68def3cbeb801c4259e743fd3d4762928/src/utils/relayerClient.ts#L45-L58',
+        },
+      ],
+    },
+    privilegedInsider: {
+      sentiment: 'warning',
+      exposure:
+        'The ASP postman new lists and the pool accepts only the latest one, so they can deny you a private exit or publish a list with just your deposit in it. The website shows the anonymity set for your amount but does not block a tiny one. It requests only pool-wide data from the ASP, so the operator and rpc provider can learn which deposits are yours only from the RPC traffic described under network observer.',
```
sergeyshemyakov: I don't understand this paragraph. Do you mean "The ASK posts new lists"? WDYM with "The website shows the anonymity set for your amount but does not block a tiny one"? If you mean that nobody stops you from depositing a small and specific amount, I suggest to drop it because it's also the case for other projects.
In the end, I am not sure why sentiment is yellow here, because of ASP that could sieve out a private withdrawal for a particular deposit but leaving only it in the whitelist? IDK if this is a solid enough deanon technique.

sergeyshemyakov:
I don't see how, but CROPS page is merged into this PR. Is this intended? I thought we were not releasing CROPS yes.

packages/config/src/projects/tornado-cash/adversaries.ts
sergeyshemyakov: I would say it's still unresolved. Maybe recommend depositing and withdrawing via different wallets? Operating at different times of day?

packages/config/src/projects/railgun/adversaries.ts
```diff
@@ -0,0 +1,143 @@
...
+          url: 'https://github.com/Railgun-Community/engine/blob/main/src/note/transact-note.ts',
+        },
+      ],
+    },
+    chainAnalyst: {
+      sentiment: 'good',
+      exposure:
+        'The candidates for an unshield are the shields of the same token, although the DeFi bundles create shields with new tokens that likely have a clean sender. Transfers inside the pool break the one-to-one match of a mixer. Timing and exact or round amounts narrow the set.',
+      advice:
+        'Keep funds shielded for a while, unshield uneven amounts that differ from any single shield, and use a fresh exit address every time.',
```
sergeyshemyakov: PP advice says: "Withdraw common amounts rather than everything at once", I think it reads nicer. I suggest to align the phrasing here.

packages/config/src/projects/umbra/adversaries.ts
```diff
@@ -0,0 +1,165 @@
...
+    protects: 'recipient',
+    text: 'Hides which recipient controls a stealth payment from the public. The payer knows the intended recipient; sender, asset, amount, stealth address and withdrawal destination remain public.',
+  },
+  cells: {
+    publicObserver: {
+      sentiment: 'good',
+      exposure:
+        'Announcements identify each stealth payment but do not identify the recipient whose public keys were used; the registry exposes candidate recipients and their key history, not a payment-to-recipient mapping.',
+      advice:
+        'Withdraw to a fresh address without an ENS name or prior identity-linked activity, and use a token relayer to avoid funding the stealth address for gas from your known wallet.',
```
sergeyshemyakov: I would also suggest to add: "Don't merge funds from different stealth addresses."

sergeyshemyakov (changes requested):
I have also made some edits cleaning up some extra verbosity from ai, feel free to review them.
One more soft suggestion to reduce the number of links, I think there are too many.

packages/config/src/projects/privacy-pools/adversaries.ts
```diff
@@ -9,7 +9,7 @@ export const privacyPoolsAdversaries = definePrivacyAdversaries({
     publicObserver: {
       sentiment: 'good',
       exposure:
-        'Which approved deposit a withdrawal spends is hidden. Partial withdrawals with change notes are supported and indistinguishable from full withdrawals, but cannot consume more than a single deposit.',
```
sekuba: this is def needed, i thought you can combine them for example and it is super important for privacy

packages/config/src/projects/privacy-pools/adversaries.ts
sekuba: is it clearer now? yellow because ASP can deanon you and frontend does not protect you, also because of conditional privacy in general

packages/frontend/src/pages/privacy/adversaries/privacyAdversaryUi.ts
```diff
@@ -7,7 +7,7 @@ import type {
 import type { PrivacyAdversarySummaryCell } from '~/server/features/privacy/types'
 
 export const PRIVACY_ADVERSARIES_TOOLTIP =
-  'On public blockchains like Ethereum, every address registers its past and future actions publicly. A privacy protocol can at best cut the link between addresses or offer privacy while deposited. One coloured dot per adversary, in order of reach. The colour says whether a careful user can keep the link, amount or recipient private against that adversary using the privacy protocol: green yes, yellow only outside supported options or by accepting another leak, red no.'
```
sekuba: disagree with the cut, they need to know that we are going out of our way doing this 'perspective from ethereum' thing. zama already criticized this and we had to explain

packages/config/src/projects/fluidkey/adversaries.ts
```diff
@@ -20,7 +20,7 @@ export const fluidkeyAdversaries = definePrivacyAdversaries({
     publicObserver: {
       sentiment: 'good',
       exposure:
-        'A payment can reach a predicted Safe before it is deployed, without an onchain announcement linking it to the recipient account. Deployment later exposes the individual stealth owner and any enabled modules, not the parent account. Auto-earn deposits remain attributable to that Safe, and combining balances or sending to an identified personal wallet publicly connects the activity.',
```
sekuba: 👍

packages/config/src/projects/umbra/adversaries.ts
sekuba: hmm, the difference is that umbra actually never merges your funds and the others do by default. so you would have to 'withdraw' to the same address two times, like in a priv pool, to loose anonymity. added more details to say that

packages/config/src/projects/strk20/adversaries.ts
sekuba: tbh i do not really get the details here but i agree and changed it, hope its better now

packages/config/src/projects/tornado-cash/adversaries.ts
sekuba: added wallet exposure and advice snippets

lucadonnoh (approved):

sergeyshemyakov (approved):

PR #12796, opened 2026-09-10
Export the CROPS modules from the config root

torztomasz:
## Summary
- `@l2beat/config` now exports a single `CROPS` object from its root, grouping `CROPS.eas`, `CROPS.attestations` and `CROPS.canonicalCrops`, with their types (`CropKey`, `CropSentiment`, `ResolvedCrops`, `ResolvedCropEvaluation`, `CropAttestation`, `CropAttestationLedger`, `RevokedCropAttestation`, `AttestationNetwork`, `AttestationNetworkConfig`) exported by name.
- The frontend imports these from `@l2beat/config` instead of `@l2beat/config/build/crops/...`. The nine client files only import types, which are erased at build, so the browser bundle is unchanged. The four server files that use values take them from `CROPS`.
- The l2b CLI keeps its deep imports for now.

The public API migration originally in this PR was dropped; the garden endpoints stay in the frontend.

## Test plan
- [x] `pnpm typecheck` in config, frontend, l2b, public-api
- [x] `pnpm test` in frontend
- [x] `pnpm lint` in config and frontend

### DIFF
--- a/packages/config/src/crops/index.ts
+++ b/packages/config/src/crops/index.ts
@@ -1,0 +1,19 @@
+import * as attestations from './attestations'
+import * as canonicalCrops from './canonicalCrops'
+import * as eas from './eas'
+
+export type {
+  CropAttestation,
+  CropAttestationLedger,
+  RevokedCropAttestation,
+} from './attestations'
+export type {
+  CropKey,
+  CropSentiment,
+  ResolvedCropEvaluation,
+  ResolvedCrops,
+} from './canonicalCrops'
+export type { AttestationNetwork, AttestationNetworkConfig } from './eas'
+
+/** The CROPS modules, grouped so the root index gains a single export. */
+export const CROPS = { attestations, canonicalCrops, eas }
--- a/packages/config/src/index.ts
+++ b/packages/config/src/index.ts
@@ -2,6 +2,7 @@
 // | DO NOT ADD MORE EXPORTS! |
 // +--------------------------+
 
+export * from './crops'
 export { PROJECT_COUNTDOWNS } from './global/countdowns'
 export {
   INTEROP_CHAINS,
--- a/packages/frontend/src/components/projects/sections/GardenCropsSection.tsx
+++ b/packages/frontend/src/components/projects/sections/GardenCropsSection.tsx
@@ -1,4 +1,4 @@
-import type { ResolvedCrops } from '@l2beat/config/build/crops/canonicalCrops'
+import type { ResolvedCrops } from '@l2beat/config'
 import { CustomLinkIcon } from '~/icons/Outlink'
 import { GARDEN_PATH } from '~/pages/garden/submit/links'
 import { CropsBed } from './crops/CropsBed'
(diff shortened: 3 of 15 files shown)


packages/public-api/src/OpenApi.ts
```diff
@@ -320,3 +330,18 @@ export class OpenApi {
...
+    .map((segment) =>
+      segment.startsWith(':') ? '[^/]+' : escapeRegExp(segment),
+    )
+    .join('/')
+  return new RegExp(`^${pattern}/?$`).test(path)
+}
+
+function escapeRegExp(value: string): string {
+  return value.replace(/[.*+?^${}()|[\]\\]/g, '\\$&')
+}
```
torztomasz: uh, i do not like this matching, cant this be simplified?

packages/public-api/src/routes/garden/routes.ts
```diff
@@ -0,0 +1,114 @@
...
+  type GardenCropsProjectResult,
+  GardenCropsProjectResultSchema,
+  GardenCropsResultSchema,
+  GardenLookupResultSchema,
+} from './types'
+
+const CACHE_TTL = 5 * UnixTime.MINUTE
+
+// Keyless on purpose, so wallets need no onboarding.
+export function addGardenRoutes(
```
torztomasz: is this a pattern to expose functions starting with `add...` as route entries?

packages/public-api/src/middleware/authMiddleware.ts
```diff
@@ -4,9 +4,17 @@ import type { AuthConfig } from '../config/Config'
...
+  config: AuthConfig,
+  logger: Logger,
+  isPublicPath: (path: string) => boolean,
+) {
   logger = logger.for('Auth')
   return (req: Request, res: Response, next: NextFunction) => {
+    if (isPublicPath(req.path)) {
+      return next()
+    }
+
```
torztomasz: can't we just register paths before authMiddleware?

packages/public-api/src/OpenApi.ts
torztomasz: Removed in 6cee4643fc. No matching anymore: the keyless routes are registered before the auth middleware, and `public` now only shapes the docs (no security requirement, no 401).

packages/public-api/src/middleware/authMiddleware.ts
torztomasz: Done in 6cee4643fc. The logger middleware now creates the request context and runs first, auth fills in the user afterwards, and the garden routes sit between the two. The finished-request log still carries the user for keyed requests.

packages/public-api/src/routes/garden/routes.ts
torztomasz: Yes, it follows the existing convention in this package: `addProjectsRoutes`, `addTvsRoutes`, `addActivityRoutes`, `addInteropRoutes` are all called from `index.ts` the same way.

PR #12807, opened 2026-09-10
Add 404 page

vilcht:
Closes L2B-14471

## Summary

- Adds a `NotFoundPage` rendered inside the normal app layout with the side nav, a large 404, "This page doesn't exist", and a "Go to L2BEAT" button linking to the scaling summary (or `/` when the home page flag is on). The page is marked noindex. Content fills the remaining height and is vertically centered; on mobile the card background is dropped so the screen reads as one surface.
- Adds `renderNotFoundPage`, which responds with the page and status 404. Page routes that previously answered `res.status(404).send('Not found')` when their params matched nothing now call it, so the outcome is explicit at the decision site. The interop token OG image endpoint keeps its plain 404 since it serves an image.
- Adds a catch-all `NotFoundHandler` at the end of the server page router for GET/HEAD page paths no route matched. Other methods and `/api/*` fall through to the default Express 404, as before.
- Registers the API router and `/health` before the page router so the catch-all does not swallow them.

404 responses are not edge-cached, consistent with the existing comment in `ServerPageRouter`.

## Test plan

- [x] `tsc --noEmit`, Biome, and mocha pass
- [x] Unknown path and unknown project, ZK catalog, publication, and interop token slugs return 404 with the HTML page and no `Cache-Control`
- [x] Existing pages, `/health`, and unknown `/api/*` paths behave as before
- [x] Checked desktop and phone viewports in the browser

### DIFF
--- a/packages/frontend/src/pages/not-found/renderNotFoundPage.ts
+++ b/packages/frontend/src/pages/not-found/renderNotFoundPage.ts
@@ -1,0 +1,16 @@
+import type { Response } from 'express'
+import type { RenderFunction } from '~/ssr/types'
+import type { Manifest } from '~/utils/Manifest'
+import { getNotFoundData } from './getNotFoundData'
+
+/** Responds with the 404 page. Call it where a route decides the resource does not exist. */
+export async function renderNotFoundPage(
+  manifest: Manifest,
+  render: RenderFunction,
+  url: string,
+  res: Response,
+) {
+  const data = await getNotFoundData(manifest, url)
+  const html = await render(data, url)
+  res.status(404).send(html)
+}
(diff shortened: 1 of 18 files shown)


torztomasz (changes requested):
Replacing `res.status(404).send('Not found')` with a bare `next()` in the page routes hurts readability and is fragile.

At the call site, `next()` only says "I'm not handling this". It does not say that the outcome is a 404 page. To learn that, a reader has to leave the router, open `ServerPageRouter`, and notice that the chain of ~20 sub-routers happens to end with `NotFoundHandler`. The intent ("this slug does not exist, show the 404 page") is no longer visible where the decision is made.

It is also order-dependent. Any router registered later that adds a broader pattern will capture these requests instead of the 404 page, and nothing at the `next()` site will hint that this can happen. Fallthrough is the right tool for "this route does not apply to the URL", not for "the route matched and decided the resource does not exist".

Please make the not-found outcome explicit at the place where the route decides it.

packages/frontend/src/pages/not-found/renderNotFoundPage.ts
```diff
@@ -0,0 +1,16 @@
+import type { Response } from 'express'
+import type { RenderFunction } from '~/ssr/types'
+import type { Manifest } from '~/utils/Manifest'
+import { getNotFoundData } from './getNotFoundData'
+
+/** Responds with the 404 page. Call it where a route decides the resource does not exist. */
+export async function renderNotFoundPage(
```
torztomasz: `sendNotFoundPage` will be better here

torztomasz (approved):

PR #12830, opened 2026-09-14
Trim project page SSR payload from 10 MB to 1 MB

torztomasz:
## Why

Project pages inlined a huge `window.__SSR_DATA__`. Arbitrum's HTML was 10.65 MB with a 9.9 MB JSON payload. The browser downloads, parses and hydrates all of it before the page becomes interactive.

## What

**Chart sections get only what they render.** The TVS, activity, costs, data-posted and liveness sections take a `ChartProject` (id, name, shortName, iconUrl), but the server handed each the whole project with an icon spread on top. Every section serialized its own copy, including `discoveryUpdates` and `tvsConfig`. Six copies at 1.5 MB each on Arbitrum. New `toChartProject` helper, applied to the L2, DA and ZK catalog builders.

**Discovery update diff bodies load on demand.** The updates section shows five collapsed cards per page, yet all 82 Arbitrum updates shipped with full diff bodies (1.2 MB). The server now strips bodies past the first page; the section fetches missing bodies for the visible page through a new `projects.discoveryUpdateSections` procedure and shows a skeleton until they arrive. Deep links to an update on a later page still open that page.

## Numbers

Measured locally with the production bundle against the real database, Arbitrum page:

| | before | after |
|---|---|---|
| `__SSR_DATA__` | 9.92 MB | 1.01 MB |
| HTML | 10.65 MB | 1.74 MB |
| warm TTFB | 135 ms | 75 ms |
| page-2 update fetch | n/a | 5 KB, 12 ms |

Remaining payload is rendered content: contracts 228 KB, state validation 122 KB, chart query state 300 KB.

## Verified

- typecheck, biome, 465 frontend tests
- browser: paging to page 2 renders fetched diffs; `?update=<id>` for a page-2 update opens page 2 with the body rendered

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #12832, opened 2026-09-14
Mount charts on approach to the viewport instead of during hydration

torztomasz:
## Why

Follow-up to #12830 and #12831. After hydration the project page froze the main thread for seconds. A CPU profile of the Arbitrum page showed the cause was not React hydration itself but charts: the server renders no chart markup because recharts draws nothing until it has measured its container, so right after hydration all six charts measured themselves, dispatched their size and re-rendered synchronously, measuring every axis tick label with a forced layout on the way. That was 310 forced style recalculations in one task.

## What

**Charts mount when they approach the viewport.** `ChartContainer` renders its recharts children only once the container is within 300 px of the viewport, flipped inside a transition so React can time-slice the mount. Until then it renders the same fixed-height box and loader the server sends, so hydration still matches. Charts below the fold mount as the user scrolls to them.

**Scroll gradient no longer forces layout mid-commit.** `ScrollWithGradient` read scroll metrics inside its ref callback, between React's DOM writes. The `ResizeObserver` attached right after already delivers an initial notification after layout, so the synchronous read was redundant.

## Numbers

Headless Chrome, 4x CPU throttling, fresh context per run, Arbitrum page, three runs each:

| | before | after |
|---|---|---|
| longest main-thread task | 3.7 to 4.0 s | 1.4 to 1.5 s |
| total blocking time | 4.2 to 4.5 s | 2.7 to 3.1 s |
| first React DOM commit | 5.1 to 5.2 s | 2.6 to 2.8 s |

Charts mounted after load: 1 at the top, 5 after scrolling to costs, all 6 at the bottom.

## Tried and dropped

A `Suspense` boundary per section, so React would hydrate each in its own task, made the numbers worse in three consistent runs: longest task 2.1 to 2.2 s, first commit 3.3 to 3.8 s. Not included.

## What remains

The remaining 1.4 s task is the hydration render itself plus the Radix collapsible content measuring on mount for 46 collapsibles. Markdown is also re-parsed on the client during hydration.

## Verified

- typecheck, biome, 465 frontend tests
- browser: charts appear progressively while scrolling; the scrollable sidebar still gets its fade class


packages/frontend/src/components/core/chart/Chart.tsx
```diff
@@ -117,7 +117,11 @@ function ChartContainer<T extends { timestamp: number }>({
   noDataSourceMessage?: string
 }) {
   const ref = React.useRef<HTMLDivElement>(null)
-  const isClient = useIsClient()
+  // Recharts renders nothing until it has measured its container, and every
+  // chart measuring and re-rendering right after hydration was the longest
+  // main-thread task on project pages. Mount each chart only when it is
+  // about to be seen.
+  const isClient = useIsNearViewport(ref)
```
torztomasz:
```suggestion
  const isNear = useIsNearViewport(ref)
```

packages/frontend/src/components/core/chart/Chart.tsx
```diff
@@ -141,7 +145,7 @@ function ChartContainer<T extends { timestamp: number }>({
             (isLoading || !hasData) && 'pointer-events-none',
           )}
         >
-          {children}
+          {isClient ? children : <div />}
```
torztomasz:
```suggestion
          {isNear ? children : <div />}
```

packages/frontend/src/hooks/useIsNearViewport.ts
vilcht, line 1: I don't like the effect here, i think we can do a callback and it would automatically reattatch. 
Here's some code that fable wrote, looks ok for me, but didnt test it

vilcht:
```js
import { type RefCallback, startTransition, useCallback, useState } from 'react'

/**
 * False on the server and during hydration, true once the element comes
 * within `rootMargin` of the viewport. The flip happens in a transition so
 * React can time-slice whatever the caller mounts in response.
 */
export function useIsNearViewport(
  rootMargin = '300px',
): [RefCallback<Element>, boolean] {
  const [isNear, setIsNear] = useState(false)

  const ref: RefCallback<Element> = useCallback(
    (element) => {
      if (!element || isNear) return
      if (typeof IntersectionObserver === 'undefined') {
        startTransition(() => setIsNear(true))
        return
      }
      const observer = new IntersectionObserver(
        (entries) => {
          if (entries.some((entry) => entry.isIntersecting)) {
            startTransition(() => setIsNear(true))
          }
        },
        { rootMargin },
      )
      observer.observe(element)
      return () => observer.disconnect()
    },
    [isNear, rootMargin],
  )

  return [ref, isNear]
}
```

torztomasz: Done in a907f9671a: the hook now returns a callback ref. Chart composes it with the object ref ChartMilestones reads.

vilcht (approved):

PR #12836, opened 2026-09-14
Add Privacy Boost privacy flow stats

vilcht:
Resolves L2B-14763

## Problem

Privacy Boost's pool emits no event we can build flow stats from:

- Epoch withdrawals settle in batches. No pool event carries a per-withdrawal amount.
- `DepositRequested` does carry an amount, but the pool upgrade on 2026-09-03 added an indexed `origin` arg, so its topic0 changed. A bucket config holds one event hash, so the prototype in #12701 would either miss all deposits since the upgrade or the five months before it.

Every flow does show up as a standard ERC-20 `Transfer` on the token contract. This PR tracks both directions that way: a deposit is a `Transfer` to the pool, a withdrawal a `Transfer` from it.

## Changes

**`@l2beat/shared`: positional topic filters**
`getLogs` on `LogsClient`, `LogsProvider`, `RpcClient` and `RpcClientCompat` takes a `LogsTopicFilter` with standard eth_getLogs semantics (positional from topic0, array = alternatives, null = wildcard). Previously the list was wrapped into position 0, so callers could never filter on indexed args. The privacy helpers wrap their event lists as `[events]`, so their queries are unchanged on the wire.

**Backend: `erc20Transfer` extractor**
- New extractor variant `{ extractor: 'erc20Transfer', params: { from?, to? } }`, amount = `value`, count = 1. Zero-value and self transfers are skipped.
- `toFlowConfig` points the flow at the token contract and derives topic1/topic2 from `from`/`to`. A source with neither is rejected so nothing can fetch every transfer of a token.
- `fetchPrivacyLogMatches` groups configurations by topic filter, queries once per group and matches only within the group. eth_getLogs ANDs positions within one filter, so deposit and withdrawal filters cannot share a query. Configurations without filters keep forming a single group, so the other 108 flow configs still make one call per update.
- Configuration ids are unchanged: topics are derived from `params`, which already feed the hash.

**Config: Privacy Boost**
One pool bucket per registered token (11, from the discovered `TokenRegistry`), deposit = `Transfer(to = pool)`, withdrawal = `Transfer(from = pool)`.

Each bucket starts at the later of the pool deployment and the token's CoinGecko listing, following umbra and railgun. The price indexer covers a bucket from its start, and there are no prices before listing, so flows before listing are dropped rather than stored without USD value. No registered token was listed after the pool deployed, so nothing is dropped today.

## Metric

Gross flows across the pool boundary. Deposit requests, portal deposits and withdrawals are counted, but so are cancellation refunds, treasury fee legs and gateway DeFi round trips. A failed gateway operation that never moves tokens is correctly recorded as nothing. This is documented in the project config.

## Verification

- Reconciled against the full deployment history (block 150,294,733 to 156,767,322): all 18,127 deposit requests match exactly one inbound transfer by transaction, token and amount, and net transfers equal the pool's on-chain balance for all 11 tokens.
- New tests: extractor (3), config mapper derived topics and rejection (3), indexer per-group queries and routing (1), positional passthrough in `RpcClientCompat`.
- Config 24,740 passing, backend 1,490 passing, typecheck clean.
- Dry run of `getPrivacyConfig`: 22 Privacy Boost flow configs on OP Mainnet, other projects byte-identical.


packages/backend/src/modules/interop/examples/snapshot/snapshot.ts
```diff
@@ -161,10 +162,10 @@ export class RpcSnapshotClient extends RpcClientCompat {
     from: number,
     to: number,
     addresses?: string[],
-    topics?: string[],
+    topics?: LogsTopicFilter,
   ): Promise<EVMLog[]> {
     const addressKey = addresses?.join(',') ?? 'all'
-    const topicsKey = topics?.join(',') ?? 'all'
+    const topicsKey = topics ? JSON.stringify(topics) : 'all'
```
torztomasz: **Spec / latent bug.** This now writes `JSON.stringify(topics)`, but `replay.ts` still builds the key with `topics?.join(',')` and keeps the old `topics?: string[]` signature. Any newly recorded interop example that passes topics will throw `ReplayError` on replay. No checked-in snapshot has `logs|` keys today, so nothing breaks yet, but replay should use the same key format (and `LogsTopicFilter`).

packages/backend/src/modules/privacy/utils/extractPrivacyFlow.ts
```diff
@@ -58,6 +64,21 @@ export function extractPrivacyFlow<T extends PrivacyFlowSource>(
...
 }
 
+// Zero-value and self transfers move nothing across the pool boundary, so
+// they do not count as flows.
+function extractErc20Transfer(
+  log: PrivacyRpcLog,
+): PrivacyFlowExtractResult | undefined {
+  const parsedLog = erc20Interface.parseLog(log)
+  const amount = BigInt(parsedLog.args.value.toString())
+
```
torztomasz, line 73: **Spec gap.** #12701 re-checked `from`/`to` against the pool here as belt-and-braces, and the indexer re-checked per-config topics client-side. Neither survives, so correctness now rests entirely on every `LogsClient` honouring positional topics. If that is deliberate, say so in the PR body; otherwise a cheap `parsedLog.args.to === params.to` style check here would restore the safety net.

packages/backend/src/config/features/privacy.ts
```diff
@@ -306,6 +318,31 @@ function toFlowConfig(
...
+function getErc20TransferTopics(params: {
+  from?: EthereumAddress
+  to?: EthereumAddress
+}): (string | null)[] {
+  assert(
+    params.from !== undefined || params.to !== undefined,
+    'erc20Transfer source needs a from or to filter',
+  )
+  const topics = [addressTopic(params.from), addressTopic(params.to)]
+  while (topics.length > 0 && topics[topics.length - 1] === null) {
```
torztomasz, line 338: **Standards: say why.** Trailing `null`s are harmless to `eth_getLogs`, so the reader cannot tell whether this trim is cosmetic or works around a client quirk. Either extract as `withoutTrailingWildcards(topics)` or add a one-line why (e.g. keeps the `[pool]` shape the tests and snapshot keys expect).

torztomasz (commented):
Two-axis review (repo standards / spec = L2B-14763 → prototype #12701). Inline comments below; items without an in-diff anchor are here.

**Standards, not anchorable**
- Commits `7600199a6e cleanup` and `604a772c77 cleanup` are undescriptive and touch the same files with near-identical stats (looks like a rebase duplicate); `a49ab50a84` is a self-merge of the branch. Please squash into something like "Group flow configurations by topic filter before querying logs".

**Spec: verified claims (no action)**
- Config-id stability holds: `idToConfigurationId` does not hash `topics`, and erc20Transfer configs use the token address, so no existing id changes.
- Topic positions are correct: `{to}` → `[null, pool]`, `{from}` → `[pool]`, indexer prepends `[events, ...]`.
- Grouping does not cross: config map is rebuilt per group, so a withdrawal log cannot match a deposit config on the same token.
- Dropping `privacyBoostDeposit` is justified: the prototype's `DepositRequested` hash is the pre-`origin` signature; the discovered ABI now hashes differently.
- All other `getLogs` callers were migrated or pass a single topic0 (equivalent under both semantics). Only `replay.ts` was missed (see inline).

**Spec: accepted deviations from #12701**
- No per-direction `address`/`topics` fields on `PrivacyFlowSource`; `toFlowConfig` hard-wires the token address instead. Simpler, fine for this use case.
- 11 buckets from the discovered TokenRegistry instead of the hard-coded 4. Reasonable.
- Listing-date floor on `sinceTimestamp`, zero/self transfers skipped. Both explained.

Summary: Standards 3 hard + 5 judgement calls, worst = undescriptive commits. Spec 2 gaps, worst = snapshot/replay key mismatch, then the dropped client-side `from`/`to` re-check.

packages/backend/src/modules/privacy/utils/privacyLogIndexerUtils.ts
```diff
@@ -42,38 +44,62 @@ export async function fetchPrivacyLogMatches<T extends PrivacyLogIndexerConfig>(
...
 
-  const { addresses, events } = buildPrivacyLogFilter(configurations)
-  const logs = await deps.logsProvider.getLogs(
-    blockFrom,
-    blockTo,
-    addresses,
-    events,
+  // eth_getLogs ANDs the topic positions of a single filter, so
+  // configurations with different indexed-arg filters cannot share a query.
+  // Each group is fetched and matched on its own.
```
torztomasz: **Standards: comment restates code.** The first two lines are a good why. "Each group is fetched and matched on its own." restates the code below; drop it.

packages/backend/src/modules/privacy/utils/privacyLogIndexerUtils.ts
```diff
@@ -42,38 +44,62 @@ export async function fetchPrivacyLogMatches<T extends PrivacyLogIndexerConfig>(
...
+  const groups = await Promise.all(
+    groupByTopics(configurations).map(async (group) => {
+      const { addresses, events } = buildPrivacyLogFilter(group)
+      const topics = group[0].properties.topics ?? []
```
torztomasz: **Data clump (judgement call).** `groupByTopics` computed this key and then we reach back into `group[0].properties.topics` to recover it. Returning `{ topics, configurations }[]` from `groupByTopics` removes the round trip and the `?? []` fallback.

packages/backend/src/modules/privacy/utils/privacyLogIndexerUtils.ts
```diff
@@ -13,6 +13,8 @@ import type { Configuration } from '../../../tools/uif/multi/types'
 interface PrivacyLogIndexerConfig {
   address: EthereumAddress
   event: string
+  /** Filters on indexed event args starting at topic1, null = wildcard. */
+  topics?: (string | null)[]
```
torztomasz: **Duplicated type (judgement call).** `topics?: (string | null)[]` with a paraphrased doc comment also lives in `modules/privacy/types.ts`. Export one named type (or reuse shared's `LogsTopicFilter`) and use it in both places.

packages/config/src/projects/privacy-boost/privacy-boost.ts
```diff
@@ -50,15 +50,71 @@ function formatBasisPoints(value: number): string {
   return `${Number((value / 100).toFixed(4))}%`
 }
 
+// topic0 of the standard ERC-20 Transfer(address,address,uint256)
+const TRANSFER_EVENT =
```
torztomasz: **Duplicated literal (judgement call).** The Transfer topic0 is hard-coded here and in `privacy.test.ts`, but derived via `getEventTopic` in `PrivacyFlowIndexer.test.ts`; `erc20Interface` is also declared three times and the 32-byte address padding is re-implemented in two tests. One exported `ERC20_TRANSFER_TOPIC` / interface would remove all of it.

packages/config/src/projects/privacy-boost/privacy-boost.ts
```diff
@@ -50,15 +50,71 @@ function formatBasisPoints(value: number): string {
...
       token.tokenAddress as ChainSpecificAddress,
     )
-    return getTokenByAddress(address.toString(), OP_MAINNET_CHAIN_ID).symbol
+    return {
+      address,
+      info: getTokenByAddress(address.toString(), OP_MAINNET_CHAIN_ID),
+    }
   })
 
+// The pool's own events carry no usable amounts: epoch withdrawals settle in
```
torztomasz, line 71: **Comment placement (judgement call).** Good rationale, but the accepted-errors list (refunds, fee legs, gateway round trips) is user-relevant and will rot here. Consider moving it into the project's markdown/description so it is visible on the site.

packages/config/src/projects/privacy-boost/privacy-boost.ts
```diff
@@ -50,15 +50,71 @@ function formatBasisPoints(value: number): string {
   return `${Number((value / 100).toFixed(4))}%`
 }
 
+// topic0 of the standard ERC-20 Transfer(address,address,uint256)
+const TRANSFER_EVENT =
+  '0xddf252ad1be2c89b69c2b068fc378daa952ba7f163c4a11628f55a4df523b3ef'
+
+const poolAddress = ChainSpecificAddress.address(pool.address)
+
 const registeredTokens = discovery
```
torztomasz, line 59: **Naming (minor).** `registeredTokens` now holds `{ address, info }` pairs and `info` is a whole token record. `tokenInfo` or destructuring to `token` reads clearer.

torztomasz (approved):

torztomasz (approved):

PR #12838, opened 2026-09-14
Fix frontend dev server: keep UpdatesSection out of the server bundle

vilcht:
## Problem

Since #12830, `getUpdatesSectionProps.ts` (server code) imports the runtime constant `UPDATES_PAGE_SIZE` from the `UpdatesSection.tsx` component. The dev server's esbuild `tsx-boundary-check` plugin rejects `.tsx` files in the server bundle, so `pnpm dev` in `packages/frontend` fails immediately:

```
✘ [ERROR] "src/components/projects/sections/UpdatesSection.tsx" is a React component (.tsx) and must not be bundled into the server.
```

## Fix

Move the constant into `src/components/projects/sections/updatesPageSize.ts` and import it from both the component and the server loader. The server loader now imports `UpdatesSectionProps` as a type-only import, so nothing from the `.tsx` reaches the server bundle.

## Verification

- `pnpm dev` starts and serves pages again (verified locally on a page with the updates section).
- Frontend typecheck clean.


vilcht:
Superseded by #12839, which landed the same fix on main.

PR #12894, opened 2026-09-17
Remove @l2beat/test-utils and use only what vitest ships

torztomasz:
No shared test helpers and no custom matchers. Test counts are identical in every package.

- `mockObject<T>({...})` becomes a plain object literal of `vi.fn()` members cast to the target type; members driven as mocks go through `vi.mocked(...)`.
- `toEqualUnsorted` becomes `toHaveLength` plus `expect.arrayContaining`, or a sort on both sides for arrays of literals.
- `toThrowWithMessage` had no call sites outside the package's own tests.
- The Postgres harness moves to `packages/database/src/test/postgresHarness.ts`; backend deep-imports it. A subpath export was tried and backed out because declaring `exports` breaks declaration emit for repository types in backend and token-backend.
- `packages/test-utils` is deleted along with every `setupFiles` and tsconfig `types` reference.

Part of the [Migrate to vitest](https://linear.app/l2beat/project/migrate-to-vitest-b5617c435a83) stack. Closes L2B-14947.

<sub>Stack created with <a href="https://github.com/github/gh-stack">GitHub Stacks CLI</a> • <a href="https://gh.io/stacks-feedback">Give Feedback 💬</a></sub>


torztomasz:
Superseded by a fresh PR from the same branch; GitHub never computed mergeability for this one so no CI ran.

PR #12899, opened 2026-09-17
Use toBe and toEqual by default, toStrictEqual only where undefined keys matter

torztomasz:
The migration mapped every earl `toEqual` to `toStrictEqual` to preserve semantics. A review of all ~6,200 sites found the strictness load-bearing in 17 of them, and the rest read unlike ordinary vitest.

Verified on vitest 5: `toStrictEqual` differs from `toEqual` only on undefined-valued keys, sparse holes, and user-defined class identity. `toEqual` still distinguishes Map, Set and Date. This repo's branded types are primitives at runtime and no test expects a user-defined class instance.

Two mechanical commits, each checkable with one grep:
- `toStrictEqual` → `toEqual` everywhere except 17 sites that are about an optional key being present with value `undefined` (validate, cloneWithComments, two liveness repository mappings, OpenApi). Each cluster carries a comment saying why.
- Primitive literal expectations → `toBe`. 2,963 sites.

| matcher | before | after |
|---|---|---|
| toStrictEqual | 6,251 | 17 |
| toEqual | 444 | 3,715 |
| toBe | 31 | 2,994 |

Per-package test counts identical to the parent commit, database and backend included, verified against a real Postgres.

Part of the [Migrate to vitest](https://linear.app/l2beat/project/migrate-to-vitest-b5617c435a83) stack. Closes L2B-14950.

<sub>Stack created with <a href="https://github.com/github/gh-stack">GitHub Stacks CLI</a> • <a href="https://gh.io/stacks-feedback">Give Feedback 💬</a></sub>


torztomasz:
Superseded by a fresh PR from the same branch; GitHub never computed mergeability for this one so no CI ran.

PR #12907, opened 2026-09-17
Keep top tokens tab chips at their natural width

vilcht:
Closes L2B-14943

## Problem
On https://l2beat.com/interop/intent-bridges the Top Tokens by Volume tab list wraps. The shared `TabsTrigger` applies `size-full flex-1`, so the chips stretch to fill each row, and a chip that lands alone on the last row (Stargate) spans the whole widget width and looks more important than the others.

## Fix
Override the trigger sizing in `InteropTopTokensWidget` with `size-auto flex-none`, so every chip only takes the width of its content, even when it is the only one on its row.

Verified in the mock dev server: intent bridges chips are now natural width on all three wrapped rows, and the Token frameworks page (non-wrapping list) is unchanged.


packages/frontend/src/pages/interop/components/InteropTopTokensWidget.tsx
```diff
@@ -13,6 +13,11 @@ import { InteropTokenRow, type InteropTokenRowData } from './InteropTokenRow'
 import type { InteropTransferDefaults } from './InteropTransferTrigger'
 import { Last24HoursBadge } from './Last24HoursBadge'
 
+const TAB_LIST_CLASS_NAME =
+  '-mr-1 -mb-1 block h-auto rounded-none bg-transparent p-0 text-balance leading-none'
+const TAB_TRIGGER_CLASS_NAME =
+  'mr-1 mb-1 size-auto gap-1 rounded-full bg-surface-secondary px-2.5 py-[3px] align-top font-bold text-sm leading-[1.15]'
```
torztomasz: imo just inline it

torztomasz (approved):

